PR #803, opened 2016-08-16
add livepatch redirect

grahambancroft:
## Done

Add livepatch redirect

## QA

Can’t be QA’d yet as the redirect destination doesn’t exist


anthonydillon:
LGTM :+1:

PR #832, opened 2016-09-07
Remove YUI dep from partners.js

anthonydillon:
## Done
- Removed the YUI and replaced with ES5
- Moved and functioned some logic

## QA
- Go to `http://localhost:8001/server`
- You will get a XOR error
- Change line 17 of partners.js to:
`var feedURI = 'http://www.stmgrts.org.uk/taleo/partners.json';` and you should see the feed come through successfully

### DIFF
--- a/static/js/partners.js
+++ b/static/js/partners.js
@@ -1,35 +1,44 @@
-/**
- * Generate a warnin only if warnings are available.
- */
-function warn(message) {
-    if (typeof(console) == 'object' && typeof(console.warn) == 'function') {
-        console.warn(message);
-    }
-}
 
 /**
- * Call the partners API and load
+ * Partners logo cloud
+ *
+ * @project        Ubuntu Core Front-End Framework
+ * @author         Web Team at Canonical Ltd
+ * @copyright      Canonical Ltd
  */
-YUI().use('jsonp', 'node', function(Y) {
-    dependencies = true;
 
-    if (typeof(partnerLogoClouds) != 'object') {
-        warn('Load partner logos: Expected array "partnerLogoClouds" not found. Stopping.');
-        dependencies = false;
-    }
+if (!core) { var core = {}; }
 
+// Loops through the partner clouds and loads the feeds for each.
+core.loadPartnerClouds = function() {
     for (var cloudIter in partnerLogoClouds) {
-      var cloud = partnerLogoClouds[cloudIter];
-      var partnersAPI = "http://partners.ubuntu.com/partners.json";
-      var url = partnersAPI + cloud['params'] + "&callback={callback}";
-      Y.jsonp(
-        url,
-        (function(elementId) {
-          return function(response) {
+        var cloud = partnerLogoClouds[cloudIter];
+        var partnersAPI = 'http://partners.ubuntu.com/partners.json';
+        var feedURI = partnersAPI + cloud['params'] + '&callback={callback}';
+        core.loadJSON(feedURI, cloud['elementId', function(response) {
+            var response = JSON.parse(response);
             return core.renderJSON(response, elementId);
-          }
-        })(cloud['elementId'])
-      );
+        });
     }
     return true;
-});
+}
+
+// Loads a given URI and returns to response to the callback function.
+core.loadJSON = function(uri, elementId, callback) {
+    var xobj = new XMLHttpRequest();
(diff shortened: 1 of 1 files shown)


static/js/partners.js
```diff
@@ -1,35 +1,44 @@
...
-    }
-}
 
 /**
- * Call the partners API and load
+ * Partners logo cloud
+ *
+ * @project        Ubuntu Core Front-End Framework
+ * @author         Web Team at Canonical Ltd
+ * @copyright      Canonical Ltd
```
nottrobin: What am I gonna say... :)

static/js/partners.js
```diff
@@ -1,35 +1,44 @@
...
  */
-YUI().use('jsonp', 'node', function(Y) {
-    dependencies = true;
 
-    if (typeof(partnerLogoClouds) != 'object') {
-        warn('Load partner logos: Expected array "partnerLogoClouds" not found. Stopping.');
-        dependencies = false;
-    }
+if (!core) { var core = {}; }
```
nottrobin: This may not be how it was done in the past, but I would rather use namespaces to actually mean something useful rather than using one single `core` namespace. Thus, `core.js` simply states `var core = {svgFallback: function() {...}}` - completely owning the `core` namespace, and this file would similarly declare:

``` javascript
var partners = {
    loadPatnerClouds = function() {...}
}
```

static/js/partners.js
```diff
@@ -1,35 +1,44 @@
-/**
- * Generate a warnin only if warnings are available.
- */
-function warn(message) {
```
nottrobin: Are you no longer worried about `console.warn` not existing? Or is this taken care of by moderniser?

nottrobin:
I'm afraid this doesn't seem to work.

There only seem to be two places on the site that actually use this code - both on `/server/` - "A SELECTION OF OUR PUBLIC CLOUD PARTNERS" and "A SELECTION OF OUR BIG DATA PARTNERS".

To test this out yourself, run the site on port 80:

``` bash
PORT=80 make run
```

Then edit your `/etc/hosts` file to add:

``` bash
127.0.0.1 www.ubuntu.com
```

Now visit <http://www.ubuntu.com/server>. You'll see that those sections are empty. They shouldn't be.

[Here](http://pastebin.ubuntu.com/23147105/) is the equivalent curl command for the one the works, and [here's](http://pastebin.ubuntu.com/23147107/) the one for the one that failed.

You'll see that in your one it says `callback=\{callback\}'`. But I think your problem probably goes deeper than that - I don't think you're implementing JSONP properly. You might want to see if there's a nice JSONP plugin that can help.

nottrobin:
I've proposed a pull request into your branch to implement JSONP. Turns out [it's not so hard](http://stackoverflow.com/questions/6132796/how-to-make-a-jsonp-request-from-javascript-without-jquery):

https://github.com/anthonydillon/www.ubuntu.com/pull/1

nottrobin:
It looks to me like the [cloud partners page](http://www.ubuntu.com/cloud/partners) should be using the same method as `/server`, but isn't. And it is broken. On live. We should fix that sometime.

static/js/partners.js
anthonydillon: Na, if we are moving stuff to ES5 then browser support is shifting anyways.

nottrobin: We still need to worry about the same set of browsers don't we? To me moving to ES5 just means adding a bunch of polyfills. This could easily be one of them.

nottrobin: It [looks like](http://caniuse.com/#search=console.warn) any calls to `console.*` in IE8 & IE9 will cause an error. Is that really acceptable? But maybe we should move it to `polyfills.js`.

nottrobin:
I noticed the breadcrumbs look broken - is this something to do with this JavaScript work we've been doing?

broken breadcrumbs

deadlight:
@nottrobin: can you replicate the bug you were getting on http://www.ubuntu.com/cloud/partners? I can't but, if you still can, can you file an issue with some more detail?

nottrobin:
@deadlight ah it's just my ublock origin plugin messing things up. Never mind.

nottrobin:
I've filed an issue for the breadcrumbs styling issue:
https://github.com/ubuntudesign/www.ubuntu.com/issues/834

PR #839, opened 2016-09-09
Unbreak breadcrumbs

nottrobin:
The `.after` element being added by JS was breaking the layout of the breadcrumbs menu on the expanded version of the menu on wider screens.

This PR hides that element in those displays.

Fixes #834.

QA
--

```
make run
```

Visit e.g. <http://127.0.0.1:8001/cloud/openstack>.

Check breadcrumbs looks okay on wider screens, and the expand/collapse functionality still looks okay and works on smaller screens.

PR #840, opened 2016-09-09
remove breadcrumb padding on large screen

### LINKED ISSUE #834: Breadcrumbs styling is broken
I noticed the breadcrumbs look broken in master (not in live). We should fix this before it gets released.

broken breadcrumbs

pmahnke:
## Done

* remove the <a class="after" styling that is added for the mobile nav chevron

## QA

1. go to /phone and /cloud
2. see that the nav looks ok in large and small screens

## Issue / Card

Fixes #834

## Screenshots

image

### DIFF
--- a/static/css/section/_local.scss
+++ b/static/css/section/_local.scss
@@ -179,6 +179,13 @@ body {
   color: $warm-grey;
 }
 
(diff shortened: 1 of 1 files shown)


pmahnke:
@nottrobin fixed this with PR #839

PR #842, opened 2016-09-09
use full size image on desktop/features

### LINKED ISSUE #569: Desktop > Features: the image in "Smarter searching" should not be physically cropped, only visibly

pmahnke:
## Done

* uploaded full size version of smarter searching image to the asset server
* replaced on desktop/features with get string cropping

## QA

1. go to /desktop/features
2. see that the smarter searching image is nearly identical

## Issue / Card

Fixes #569

## Screenshots

image


anthonydillon:
Thanks, does indeed look the same but not sure it solves the issue. I believe @yaili wants the image to load as a full image and be masked by the row below. I think they way you have done it is good so will +1 but can you quickly check with @yaili.

yaili:
I think having the image cropped in the asset server is better than what is currently there. Thanks.

anthonydillon:
Thanks @yaili

PR #845, opened 2016-09-10
Tidy up scratch

anthonydillon:
# Done
- Stepped through the scratch.js removing the requirement of YUI and converting to ES5
- Removed any functions not used
- Add some CSS that is required and fixed a few issues

## QA
Sorry this is a long one and will require a thorough QA with the console open checking for errors.

- [x] Go to /phone/features and scroll down to the full width video (`.row--video`). Click the row and check the video starts to play.
- [x] Go to /phone and click the Watch the video button in the hero row. Check the video plays. Then click the Close this video link and the hero should go back to how it was. Check it matches live.
- [x] Go to the homepage. Scroll down to the section titled 'Multiple devices, one experience'. Check that the device morph kicks off as you scroll to it. Once the animation is finished click on the device and check the animation plays again.
- [x] Go to /phone/developers and scroll down to the 'Build more than just an app' section. Check the scopes are scrolling. I made it loop as I think its nicer and how the phone works.
- [x] Pop open your inspector and head over to your Applications tab. Delete the `_cookie_accepted` cookie. Refresh the page and see that cookie policy appears. Refresh the page and see the cookie policy remains. Click the close and refresh and check its gone and the `_cookie_accepted` cookie is back.
- [x] Go on down to small screen and head to the footer. You should see the main sections stacked up. Click on one of the section rows (not the word as it will load that page). The section should open to reveals the wonders contained inside. Click again and the row should conceal itself.
- [x] Stay on small screen in the footer. Click on the Ubuntu websites title and check the links are shown and can be toggled open and closed. Resize your browser up to large screen and check the global menu jumps to the top of the page. Resize it back to small screen and check it jumps back down without any issues. And continues to open and close.
- [x] Go to /download/cloud and check the code snippets are selected when you click on the text. Now click the copy to clipboard button and it should change the border to green and save the snippet to your clipboard.

### DIFF
--- a/templates/cloud/partners/index.html
+++ b/templates/cloud/partners/index.html
@@ -154,11 +154,12 @@
 {% endblock content %}
 
 {% block footer_extra %}
-<script>
-    YUI().use('event-base', function (Y) {
-        Y.on('domready', function () {
-            core.loadPartners('?programme__name=Certified%20Public%20Cloud&programme__name=Charm%20Partner%20Programme&programme__name=OpenStack&featured=true', '#cloud-partners');
-        });
-    });
-</script>
+  <script>
+      partnerLogoClouds = [{
+          'params': '?programme__name=Certified%20Public%20Cloud&programme__name=Charm%20Partner%20Programme&programme__name=OpenStack&featured=true',
+          'elementId':'#cloud-partners'
+      }];
+  </script>
+  {% load versioned_static  %}
+  <script src="{% versioned_static 'js/partners.js' %}"></script>
 {% endblock %}
--- a/static/css/section/_footer.scss
+++ b/static/css/section/_footer.scss
@@ -57,19 +57,19 @@ footer.global {
       margin-bottom: 0;
       padding: .833333333em 10px;
     }
-    
+
     h2 {
       /// open/close arrow closed
       background-image: url("#{$asset-server}7bd1bd7b-arrow_down_9fa097.svg");
       background-position: 96% 50%;
       background-repeat: no-repeat;
       background-size: 13px 13px;
-      
+
       a:link,
       a:visited {
         color: #888;
       }
-      
+
       a::after {
         content: ""; /* hide the default > */
       }
@@ -79,11 +79,15 @@ footer.global {
         display: none;
       }
 
-      &.active {
-        background-image: url("#{$asset-server}43e2b367-arrow_up_9fa097.png");
-        
-        + ul.second-level-nav {
-          display: block;
+      @media only screen and (max-width: $breakpoint-medium) {
+        cursor: pointer;
+
+        &.active {
+          background-image: url("#{$asset-server}43e2b367-arrow_up_9fa097.png");
+
+          + ul.second-level-nav {
+            display: block;
+          }
         }
       }
     }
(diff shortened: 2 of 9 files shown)


nottrobin:
Code LGTM. The only tiny issue I found is for the expandable mobile footer. When I click on the title of one of the footer rows, yes it redirects me to that page, but before that happens that footer section expands/collapses, which is a slightly clunky experience. This doesn't happen on live. Apart from that, everything works as specified.

templates/cloud/partners/index.html
```diff
@@ -155,10 +155,9 @@ <h2 class="muted-heading">A selection of our cloud partners</h2>
 
 {% block footer_extra %}
 <script>
-    YUI().use('event-base', function (Y) {
-        Y.on('domready', function () {
-            core.loadPartners('?programme__name=Certified%20Public%20Cloud&programme__name=Charm%20Partner%20Programme&programme__name=OpenStack&featured=true', '#cloud-partners');
-        });
-    });
+    partnerLogoClouds = [{
```
nottrobin: This isn't working. I assume because `partnerLogoClouds` is only interpreted in `partners.js`, which isn't currently included in this page. `partners.js` will have to load *after* `partnerLogoClouds` is defined to work.

It's lucky I thought to look for this specifically, otherwise I'd have never caught it, as it didn't throw any kind of error.

Which actually brings me to a slightly different topic. I'd rather we followed the pattern that was previously used in this page. In the old code, `core.loadPartners` was called from here. This would mean that if `core.loadPartners` wasn't available, you would see an error - as you should.

This method of more passively declaring a `partnerLogoClouds` variable is much more prone to failing silently.

So I would prefer that we took the `partners.loadPartnerClouds();` function call out of `partners.js` and put it in the bottom of every page that needs it. This would also allow us to pass through the partner information directly. Something like:

``` html
{% block footer_extra %}
<script src="static/js/partners.js"></script>
<script>
partners.loadPartnerCloud(
  '#cloud-partners',
  '?programme__name=Certified%20Public%20Cloud&programme__name=Charm%20Partner%20Programme&programme__name=OpenStack&featured=true'
);
// And if you had more than one cloud, call it again:
partners.loadPartnerCloud(
  '#another-cloud',
  '?programme__name=Certified%20Public%20Cloud'
);
</script>
{% endblock %}
```

anthonydillon: I agree, thanks. Think this is best tackled in a follow up once this one and the local branch land. As they both touch this code.

anthonydillon:
Fixed the expandable mobile footer @nottrobin. Good spot.

templates/cloud/partners/index.html
nottrobin: Could you at least add `<script src="{% versioned_static 'js/partners.js' %}"></script>` below the `partnerLogoClouds` before we merge this branch? I'm not comfortable merging work into `master` which knowingly breaks something that's currently working - since `master` could be released at any time.

nottrobin:
@anthonydillon Yep that works.

Could you please just do the quick fix for the partner logo cloud on `/cloud/partners`, then we can merge this.

anthonydillon:
@nottrobin of course, done.

nottrobin:
@anthonydillon Great, works a charm. :+1:

PR #850, opened 2016-09-12
Add polyfills for basic ECMAScript 5 features

nottrobin:
Port over polyfills used on snapcraft.io/docs for:

- Element.prototype.classList ([browser support](http://caniuse.com/#search=classlist))
- Element.prototype.addEventListener ([browser support](http://caniuse.com/#search=addeventlistener))
- String.prototype.startsWith ([browser support](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith#Browser_compatibility))

QA
--

Just check it doesn't throw any JavaScript errors.


static/js/polyfills.js
```diff
@@ -35,3 +35,15 @@ Object.getOwnPropertyNames(Array.prototype).forEach(
...
+
+/* String.prototype.startsWith polyfill, by Mathius Bynens, https://mathiasbynens.be/
+ * https://github.com/mathiasbynens/String.prototype.startsWith */
+String.prototype.startsWith||!function(){"use strict";var t=function(){try{var t={},r=Object.defineProperty,e=r(t,t,t)&&r}catch(n){}return e}(),r={}.toString,e=function(t){if(null==this)throw TypeError();var e=String(this);if(t&&"[object RegExp]"==r.call(t))throw TypeError();var n=e.length,i=String(t),a=i.length,o=arguments.length>1?arguments[1]:void 0,h=o?Number(o):0;h!=h&&(h=0);var u=Math.min(Math.max(h,0),n);if(a+u>n)return!1;for(var g=-1;++g<a;)if(e.charCodeAt(u+g)!=i.charCodeAt(g))return!1;return!0};t?t(String.prototype,"startsWith",{value:e,configurable:!0,writable:!0}):String.prototype.startsWith=e}();
+>>>>>>> Add polyfills for basic ECMAScript 5 features
```
anthonydillon: Broken conflict?

anthonydillon:
Is it worth getting unminified versions and we will get a minifier on build?

static/js/polyfills.js
nottrobin: Yep. Can't believe you +1ed this. You are too easy.

nottrobin:
@anthonydillon I think I'd rather leave them minified - even if we do minify again. The reason is because I think this form is fundamentally more readable to us. The comment to tell you what the polyfill is for and where it comes from gives you the important information. The expanded polyfills are very long and not pretty code, and they would make the `polyfills.js` file very difficult to parse.

PR #857, opened 2016-09-13
Remove YUI from global.js

nottrobin:
But keep the same interface - global.js should still work everywhere, although it requires polyfills for the following to work cross-browser (arriving in #850):

- NodeList.forEach
- Element.addEventListener
- Element.classList

I also uploaded the new global.js to the assets server at: https://assets.ubuntu.com/v1/ab9f5f4b-global.js

QA
--

Check global nav behaves as expected - as it does on live.


deadlight:
Works well and the code looks good.

PR #860, opened 2016-09-15
update gsa server to butlerov.internal

pmahnke:


pmahnke:
Self approve for QA on staging,

PR #880, opened 2016-09-21
Fixes secondary nav issue

grahambancroft:
## Done

Add the new full width section classes to the 'extra rules for full width sections' SCSS

## QA

Pop on over to http://www.ubuntu.com-1610-dev.demo.haus/cloud and have a look at the ginormous  gap between the top of the secondary nav and the bottom of the man nav and then have a look locally. This fixes those problems


anthonydillon:
It works. thanks. :+1:

PR #905, opened 2016-09-30
Ensure resources are loaded through HTTPS

nottrobin:
- Make some ambiguous resources explicitly load over HTTPS
- Fix styling for sliders
- For domains that we *know* support HTTPS (like insights.ubuntu.com), make *all links* use HTTPS

QA
--

Click around the site, especially on touched pages, and make sure everything works and looks right.


static/js/global.js
```diff
@@ -1,7 +1,7 @@
 var ai = [
   { url: "http://www.ubuntu.com", title:"Ubuntu" },
```
pmahnke: what about this link? will we be https everywhere? I assume we have to change it AFTER we move to https?  Perhaps we should file and issue for that.

also, we need up update this file in the [new repo](https://github.com/ubuntudesign/global-nav) as well.

nottrobin: Yep. We should merge this first, but I've made an issue for replacing www.ubuntu.com links:
https://github.com/ubuntudesign/www.ubuntu.com/issues/907

I'll add a comment now about updating global nav.

nottrobin: I also filed this PR: https://github.com/ubuntudesign/global-nav/pull/6

nottrobin:
Okay I'm gonna merge this.

PR #908, opened 2016-10-01
Fix errors in period calculations

nottrobin:
Quick before it goes live.


nottrobin:
Self-approving urgent hotfix

PR #911, opened 2016-10-03
Fix clicks on "Multiple devices, one experience" row

### LINKED ISSUE #910: The Phone link on the main page doesn't work
## Summary

If you navigate to ubuntu.com and scroll down to Multiple Devices one experience and then click on it it does nothing if you right click on it and select open in another tab it then opens

## Process

1. open a browser (Tested in ubuntu-browser, firefox and chromium)
2. goto ubutnu.com
3. click on Multiple Devices one experience
4. right click on it and select open in another tab

I expect the page to open in the current tab
Actual result nothing happens unless I right click on select open in another tab

nottrobin:
No need to preventDefault for this one. It's breaking links!

This fixes #910

QA
---

Visit homepage, check clicking "Multiple devices one experience" goes to phone page. Check the row otherwise continues to behave as expected - clicking restarts the animation.

### DIFF
--- a/static/js/scratch.js
+++ b/static/js/scratch.js
@@ -68,7 +68,6 @@ core.setupAnimations = function() {
             });
 
             sliderAnimation.addEventListener('click', function(e) {
-                e.preventDefault();
                 sliderAnimation.classList.add('run');
                 setTimeout(function(e) {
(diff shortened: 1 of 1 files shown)


pmahnke:
Good catch!

👍

PR #912, opened 2016-10-03
Move Insights feed to backend

WillMoggridge:
## Done

Move javascript feed parsing to backend template logic.

## QA

Check feeds on:
- `/`
- `/cloud`
- `/management`

They should not look any different, and should not pop in! Yay!
In between refreshes, they should be cached.


nottrobin:
As part of this PR, could we make correct the spelling of `FEED_TIMEOUT` and maybe set the default to be higher than one second [here](https://github.com/ubuntudesign/www.ubuntu.com/blob/master/webapp/lib/feeds.py#L17)?

WillMoggridge:
Note to self, Django can provide pretty dates with:
`django.contrib.humanize.naturaltime`
```
{% load humanize %}
{{ date|naturaltime }}
```

Issue #399, opened 2016-04-12
Review template files


WillMoggridge:
## Context

It looks like we have some extraneous template files that may not be needed. We should review them and remove any unneeded files.

For example, is `/templates/mu-b760b7ee-790eb464-8b1bbe62-de00513f.html` used anywhere?

pmahnke:
that was to get access to google webmaster toolkit

anthonydillon:
@WillMoggridge as your PR landed I'll close this issue but please reopen if not fully complete.

PR #916, opened 2016-10-04
Remove unmentioned template includes

nottrobin:
The number of old and unused template files polluting our repository has been bugging me for a while.

I ran [this script](https://gist.github.com/nottrobin/dc4a147260d9089405b1e2f920a06d96) in the `templates` directory to delete them. Don't worry, they'll still be in version history.

QA
--

```
make run
```

Browse around the whole site and check pages still look fine and as they do on live. Especially footers, contextual navs, sub navs, breadcrumbs etc.


WillMoggridge:
Fixes #399

WillMoggridge:
We talked before about having some default rotating takeovers.
I imagine we want to confirm which these are and keep them in the folder, rather than keep deleting and restoring them.

nottrobin:
That's a good point @WillMoggridge. Maybe @deadlight can comment?

pmahnke:
@nottrobin I think we really need to exempt the takeover folder and perhaps create an issue to clean that up?

nottrobin:
@pmahnke I've added back all files in the `templates/takeovers` folder, and [created an issue](https://github.com/ubuntudesign/www.ubuntu.com/issues/924) to remove then instead. But under protest.

deadlight:
We should create a companion repo called "out back" with all takeovers as a reference ;)

nottrobin:
Or you could just use a command like this:

``` bash
$ git log --pretty=format: --name-status | cut -f2- | grep 'templates/takeovers'
templates/takeovers/_1310_takeover.html
templates/takeovers/_14.04_takeover.html
templates/takeovers/_14.10_takeover.html
...
```

deadlight:
That's not practical at all for most of our use-cases. 90% of the time a takeover is based on or very similar to another. We often re-use them but there's no pattern in when we might do that. The ability to quickly slap them out and check them visually is invaluable.

That said we can probably cull a lot of them and I agree that a better way of handling repeating/generic takeovers would make things tidier.

IMO this is library code, not unused code. My preference is for keeping them but organising the file structure, culling entirely useless code and naming the files better. Also stand-alone CSS files for bespoke elements would avoid the problem we have in the homepage sass file of variously-kept/removed CSS.

nottrobin:
@deadlight All good point. Let's continue the discussion re: takeovers specifically over on [this issue](https://github.com/ubuntudesign/www.ubuntu.com/issues/924).

PR #917, opened 2016-10-04
Minimise redirects by removing slashes

nottrobin:
Any URLs ending in "/" will redirect to without the "/".

With HTTPS this will be exacerbated, because the TemplateFinder will redirect from https://xxxx/ to http://xxxx, which will then be redirected back to https://xxxx.

So removing any internal links with "/" on the end will reduce unnecessary redirects.

I used [a simple ack command](https://gist.github.com/nottrobin/28f8570add88222727b85a55e7f02c90) to find these links, then I removed them.

### DIFF
--- a/templates/desktop/developers.html
+++ b/templates/desktop/developers.html
@@ -12,7 +12,7 @@
     <div class="six-col">
         <h1>Ubuntu Desktop for&nbsp;developers</h1>
         <p>Whether you&rsquo;re a mobile app developer, an engineering manager, a music or video editor or a financial analyst with large-scale models to run &mdash; in fact, anyone in need of a powerful machine for your work &mdash; Ubuntu is the ideal platform.</p>
-        <p><a href="/download/desktop/" class="button--primary">Get Ubuntu now</a></p>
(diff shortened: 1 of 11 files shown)


grahambancroft:
[Server section full width](https://trello.com/c/6aU6HEyP/1788-server-section-full-width)

WillMoggridge:
We intentionally changed the format of our URLs recently but left these links unchanged, it isn't an upstream bug if you mean that.

Potentially, this may be avoided by using the django `{% url %}` template tag rather than writing direct links but it isn't much of a problem.

nottrobin:
@WillMoggridge I thought `{% url %}` was for named URLs set in `urls.py`? Which none of ours are.

nottrobin:
@pmahnke this does partly exposes a bug in django-template-finder-view. It shouldn't redirect HTTPS to HTTP. I've filed an issue - https://github.com/ubuntudesign/django-template-finder-view/issues/1.

WillMoggridge:
@nottrobin It can use a name, or you can type the module/app namespacing.
I don't mean to say it works right now or if it is even a net gain, just a musing.

PR #934, opened 2016-10-10
1610 fw desktop

grahambancroft:
## Done

Design review amends on full width desktop section

## QA

Check the desktop section in s/m/l

Awaiting final design sign off and there are still a few more amends to make

## Issue / Card

Card: https://trello.com/c/wcj7FtDU


grahambancroft:
Hi @barrymcgee, I’ve spoken with @yaili and these are issues with live also, they do need to be addressed but shouldn’t be blockers on the full width/16.10 updates. She is, however, going to look at some of the critical issues, such as desktop/overview to see if they need a white box.

PR #958, opened 2016-10-12
Fix up offset image in desktop/developers

barrymcgee:
## Done
Fix up offset image in desktop/developers

## QA
- Go to /desktop/developers
- Scroll to What's New 16.10
- Confirm that laptop image is offset


WillMoggridge:
At a smaller width the laptop overlaps some of the text. Otherwise looks good.
This is at 1210px wide:
selection_010

PR #969, opened 2016-10-13
Only remove row-videos border on biggish screens

nottrobin:
QA
---

On `/desktop/features`, check there is still no border between "Videos" and "Gaming" on a large screen, but on smaller screens, when there's no background image for "Gaming", the border should appear.

### DIFF
--- a/static/css/section/_desktop.scss
+++ b/static/css/section/_desktop.scss
@@ -199,6 +199,12 @@
     }
   }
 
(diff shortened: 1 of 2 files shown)


nottrobin:
[/desktop/features fixes](https://trello.com/c/8VLpy3Qr/1966-desktop-features-fixes)

PR #1019, opened 2016-10-14
Redirect all certification URLs

nottrobin:
QA
---

Run site, visit `/certification`, `/certification/desktop` etc. and check you are redirected to the right place.


nottrobin:
@barrymcgee well, ideally `/certification/foo` should 404 in place, as it was never a page that existed, but for that to happen I would have to write each certification URL in manually, which is more trouble than it's worth. But yes, `/certification/foo` leading to a 404 is definitely correct.

PR #1032, opened 2016-10-19
Change `/livepatch` redirect to point to https

WillMoggridge:
Link `/livepatch` directly to https://auth.livepatch.canonical.com and avoid the extra redirect from http to https.


deadlight:
:+1:

PR #1033, opened 2016-10-19
live patch takeover

deadlight:
## Done

Created a homepage takeover for live patching.

## QA

Check that the homepage takeover renders in all viewports and that matches the [copy](https://docs.google.com/document/d/1c0Nse3mevYyMdJ9a7Bnd2XjBPRYiNA2Pm7rfcmJjDJ8/edit?ts=57b451b8)

### DIFF
--- a/templates/takeovers/_livepatch.html
+++ b/templates/takeovers/_livepatch.html
@@ -1,0 +1,8 @@
+<section class="row row-hero no-border-bottom">
+    <div class="strip-inner-wrapper">
+        <h1 class="seven-col">Apply critical kernel patches&nbsp;without rebooting</h1>
+        <p class="six-col">The Canonical Livepatch Service for Ubuntu 16.04 LTS reduces downtime while maintaining compliance and security.</p>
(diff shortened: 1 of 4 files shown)


templates/takeovers/_livepatch.html
```diff
@@ -0,0 +1,8 @@
+<section class="row row-hero no-border-bottom">
+    <div class="strip-inner-wrapper">
+        <h1 class="seven-col">Apply critical kernel patches&nbsp;without rebooting</h1>
+        <p class="six-col">The Canonical Livepatch Service for Ubuntu 16.04 LTS reduces downtime while maintaining compliance and security.</p>
+        <img src="{{ ASSET_SERVER_URL }}7a3a71fc-dell-xps-laptop-front.png?w=768" alt="laptop with live patch work" class="for-small six-col row-hero__image" />
```
WillMoggridge: https://assets.ubuntu.com/v1/7a3a71fc-dell-xps-laptop-front.png?w=768 looks pretty pixelated.
Can we try resizing the asset with a better tool and see if it looks better? If so, we can use and upload that.

WillMoggridge: OK, we will fix this shortly but this is OK for the moment

PR #1035, opened 2016-10-20
update /cloud events

pmahnke:
## Done

* updated the /cloud overview events

## QA

1. go to /cloud
2. see that all the events are current

## Issue / Card

[trello card](https://trello.com/c/NR8J87s9)

### DIFF
--- a/templates/cloud/shared/_events_overview.html
+++ b/templates/cloud/shared/_events_overview.html
@@ -4,42 +4,42 @@
             <div class="newsevent clearfix" itemscope itemtype="http://data-vocabulary.org/Event">
                 <div class="six-col equal-height--vertical-divider__item no-margin-bottom last-col">
                     <div class="newsevent__map">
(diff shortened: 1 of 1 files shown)


pmahnke:
[update /cloud events](https://trello.com/c/NR8J87s9/1837-update-cloud-events)

anthonydillon:
It doesnt look like this has the latest events in it. Seems to be missing:
- Huawei Connect Europe
- Enterprise Open Source Roadshow – Espoo
- Enterprise Open Source Roadshow – Zurich event

Is there a reason for this?
Seen at the bottom of https://insights.ubuntu.com/cloud-and-server

pmahnke:
@anthonydillon I get the events from https://insights.ubuntu.com/category/events/feed/?group=cloud-and-server&2016-10-20d0000=true

I think you have uncovered an insights bug that is showing non cloud and server events on the cloud and server page...

I think my list is more correct.

anthonydillon:
I think the feed you are using is wrong. The first event [1] has Cloud and server as a tag and is not shown on your list.

[1] https://insights.ubuntu.com/event/enterprise-open-source-roadshow-10/

pmahnke:
@anthonydillon well... I would still say that is an insights bug that it isn't in that feed... but on a practical level, I think promoting ODS over a sales event in Finland is the right thing now.

PR #1072, opened 2016-11-03
Design review tweaks

grahambancroft:
## Done

make ubuntu core logo + os image size graph smaller
on the hero, keep the illustration (between copy and the cta)
make sure we use the same padding between title+ copy and copy + cta?

## QA

check in s/m/l


deadlight:
Look good.

PR #1093, opened 2016-11-10
Updated Spotlight

pmahnke:
## Done

 - Added spotlight article support to the homepage news box
 - Added a way to exclude results from a get_rss_feed_content call (to allow removal of spotlight articles from the general feed, avoiding duplication)
- Added GTM tracking onclicks

## QA

Go to the homepage and see that three general articles and the one article tagged "spotlight" are pulled in from insights.

Check the page in all viewports.

Extra-credit: try using different feeds in index.html (lines 13-14) and check that they behave as expected (removing duplicates, etc.)

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -12,28 +12,33 @@
 
 
 {% get_rss_feed "https://insights.ubuntu.com/tag/spotlight/feed" limit=1 as spotlight_feed %}
-{% get_rss_feed "https://insights.ubuntu.com/feed" limit=3 exclude_items_in=spotlight_feed as insights_feed %}
-
+{% if spotlight_feed %}
+    {% get_rss_feed "https://insights.ubuntu.com/feed" limit=3 exclude_items_in=spotlight_feed as insights_feed %}
+{% else %}
+    {% get_rss_feed "https://insights.ubuntu.com/feed" limit=4 as insights_feed %}
+{% endif %}
 <section class="row row--ubuntu-news strip no-border">
     <div class="strip-inner-wrapper equal-height--vertical-divider">
-        <div class="nine-col equal-height--vertical-divider__item">
-            <h2 class="external row--ubuntu-news__title"><span><a href="https://insights.ubuntu.com/"onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'insights', 'eventAction' : 'clicks insights feed link', 'eventLabel' : 'ubuntu.com homepage', 'eventValue' : undefined });">Latest news from Insights</a></span></h2>
-            {% for item in insights_feed %}
+        <div class="{% if spotlight_feed %}nine-col{% else %}twelve-col{% endif %} equal-height--vertical-divider__item">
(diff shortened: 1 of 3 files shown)


templates/index.html
```diff
@@ -10,21 +10,29 @@
...
+            {% for item in insights_feed %}
+            <div class="three-col {% if forloop.last %}last-col no-margin-bottom{% endif %}">
+                <h3><a href="{{ item.link }}" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'insights', 'eventAction' : 'clicks news link', 'eventLabel' : 'ubuntu.com homepage', 'eventValue' : undefined });">{{ item.title }}</a></h3>
+                <p><time pubdate datetime="{{ item.updated }}">{{ item.updated_datetime|date:"j F Y" }}</time></p>
+            </div>
+            {% endfor %}
+        </div>
+        <div class="three-col no-margin-bottom last-col equal-height--vertical-divider__item">
+            <h2 class="external row--ubuntu-news__title"><span><a href="https://insights.ubuntu.com/tag/spotlight/" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'insights', 'eventAction' : 'clicks spotlight feed link', 'eventLabel' : 'ubuntu.com homepage', 'eventValue' : undefined });">Spotlight</a></span></h2>
+            {% for item in spotlight_feed %}
```
nottrobin: Do we need to consider the case of if there is no spotlight feed (`{% if spotlight_feed %} ... {% endif %}`)? Which as there's only 1 right now seems like it could easily happen.

webapp/lib/feeds.py
```diff
@@ -59,6 +59,10 @@ def get_rss_feed_content(url, offset=0, limit=None):
         )
         content = []  # Empty response
 
+    if exclude_items_in:
+        exclude_ids = [x['guid'] for x in exclude_items_in]
+        content = filter(lambda x: x['guid'] not in exclude_ids, content)
```
pmahnke: I am going to leave this as I haven't written it... but I agree.... @deadlight  to fix please later

templates/index.html
```diff
@@ -10,21 +10,29 @@
...
-                    {% for item in cloud_feed %}
-                    <li class="three-col {% if forloop.last %}last-col{% endif %}">
-                        <h3><a href="{{ item.link }}">{{ item.title }}</a></h3>
-                        <p><time pubdate datetime="{{ item.updated }}">{{ item.updated_datetime|date:"j F Y" }}</time></p>
-                    </li>
-                    {% endfor %}
-                </ul>
+    <div class="strip-inner-wrapper equal-height--vertical-divider">
+        <div class="nine-col equal-height--vertical-divider__item">
+            <h2 class="external row--ubuntu-news__title"><span><a href="https://insights.ubuntu.com/"onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'insights', 'eventAction' : 'clicks insights feed link', 'eventLabel' : 'ubuntu.com homepage', 'eventValue' : undefined });">Latest news from Insights</a></span></h2>
```
pmahnke: Done

templates/index.html
nottrobin: I don't think there would be an error as it stands, you'd just see the "spotlight" heading, with nothing underneath it. We could put a paragraph saying "There is no spotlight" or similar?

pmahnke:
@nottrobin can you re-review... I added in logic if there is no spotlight to show 4 news items

nottrobin:
@pmahnke nice solution! works a charm.

PR #1166, opened 2016-12-02
add mobile partners page

deadlight:
## Done

Added a new mobile partners page

## QA

Check the `/mobile/partners` page in browser at all viewports and check it against the [copydoc](https://docs.google.com/document/d/1kf_dm2alpBcDAxX3t8TeV1SndCbtCB4lzaQzcUPHSug/edit#heading=h.6sx4vafy85fs)

### DIFF
--- a/static/css/section/_mobile.scss
+++ b/static/css/section/_mobile.scss
@@ -39,6 +39,48 @@
   }
 }
 
+.mobile-partners {
+  .row-hero {
+    overflow: hidden;
+
+    &__image {
+      margin: 10px 0;
+    }
+
+    &:before {
+      @media only screen and (min-width : $breakpoint-medium) {
+        background: url('#{$asset-server}969bb3c1-9a.-scope-task-switcher-tablet-portrait-landscape.png?w=700') center center no-repeat;
+        background-size: contain;
+        content: '';
+        display: block;
+        height: 80%;
+        overflow: visible;
+        position: absolute;
+        right: -24vw;
+        top: 10%;
+        width: 70vw;
+      }
+
+      @media only screen and (min-width : 769px) {
+        right: 2vw;
+        width: 47vw;
+      }
+
+      @media only screen and (min-width : 1400px) {
+        right: 2vw;
+        width: 50vw;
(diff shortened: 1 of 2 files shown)


deadlight:
[Build mobile/partners page](https://trello.com/c/vVqbZybZ/2176-build-mobile-partners-page)

static/css/section/_mobile.scss
```diff
@@ -1,40 +1,84 @@
-.mobile-features {
-  .row-hero {
-    overflow: hidden;
-
-    @media only screen and (min-width : $breakpoint-medium) {
+.mobile- {
```
nottrobin: Here you're creating a class called `.mobile-`. I don't know if that's what you intended. If not, I think you should instead have top-level classes with the full names you want to use, e.g. `.mobile-features {}`, `.mobile-partners {}`. Or if you want, you could have a top-level class called `.mobile {}` and then put `&__features {}` inside it, but I'd only suggest doing that if you were actually going to use the `.mobile` class.

deadlight: It's to avoid the repetition. There's no content in the style so the class ".mobile-" never exists in the CSS.

It's a matter of style but I can change it.

deadlight:
Thank's @nottrobin I think I reused some sections from the old site without noticing the updates.

deadlight:
@nottrobin all fixed other than the quote order. The order doesn't matter but the smaller one being in the middle looks better.

static/css/section/_mobile.scss
nottrobin: I discussed it with @anthonydillon, he agreed it's better not to have partial class names to reduce a tiny amount of repetition in this way.

nottrobin:
@deadlight replied to the comment in code. Also, could you change the copydoc to reflect the order?

static/css/section/_mobile.scss
deadlight: @nottrobin fixed - fascist :P

deadlight:
@nottrobin all fixed

PR #1181, opened 2016-12-12
Update the mobile overview page content

anthonydillon:
## Done
- Updated content to match copy doc
- Reorder two sections to match copy doc content order

## QA
- Pull down this branch and run `make run`
- Go to http://127.0.0.1:8001/mobile
- Check that the content is the same as the copy doc
- Check the order is also correct


pmahnke:
LGTM 👍

PR #1186, opened 2016-12-13
Add meta tag to define Content-Secure-Policy

barrymcgee:
## Done

Added meta tag to define Content Security Policy so that local development can more accurately reflect the production environment.

## QA

Check code

## Issue / Card

Fixes: https://github.com/ubuntudesign/www.ubuntu.com/issues/1141


nottrobin:
Well done for taking this on @barrymcgee but I'm afraid it is not quite as simple as this:

errors

PR #1284, opened 2017-01-27
Add link checking to automated (travis) tests

nottrobin:
This will run linkchecker across a local run of the whole site, and will error if it finds any 404s or 500s.

There are a couple of downsides to doing this:

- This will probably mean builds take an extra minute
- Builds could break without any code having changed - if a URL we're linking to goes down.
  This will unfortunately mean someone may be tasked with updating a 404 link that was not their fault to get their PR to pass.

I think on balance the value we get from automatically checking for 404s and 500s is worth these downsides. The reviewer can decide if they agree with me.

QA
---

Check the travis build passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,24 @@
 language: python
 python:
   - "2.7"
+
+before_install:
+  - nvm install 7.4.0
+  - nvm use 7.4.0
+
 install:
-  - pip install flake8==3.0.4
+  - npm install --legacy-bundling
+  - npm install node-sass
+  - node_modules/node-sass/bin/node-sass --include-path node_modules static/css --output static/css
+  - pip install --upgrade pip
+  - pip install -r requirements.txt
+  - pip install flake8 requests[security]
+  - pip install https://github.com/nottrobin/linkchecker/archive/fix-install.zip
+
+before_script:
+  - python manage.py runserver &
+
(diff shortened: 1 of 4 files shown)


deadlight:
@nottrobin - is there any way of testing both the previous build and the new build and, if the 404 occurs in both, the build passes (but the 404 is flagged in another way)?

nottrobin:
@deadlight I don't know if it would be possible, but I don't really see the point. There are currently no 404s and shouldn't be any for any long period of time.

I know the travis build is failing. I'm having some trouble with getting it to work properly.

deadlight:
@nottrobin my point is that V1 has no 404s, V2 introduces a 404 and it should be fixed. BUT if, at the time of checking, both V1 and V2 have the 404 then it's clearly an external link that has started 404ing. It's not the fault of V2 and V2 should be mergable

nottrobin:
@deadlight oh I see. My thinking was that it's not exactly an arduous task to fix a 404, and we definitely want to fix any that appear ASAP. And so if one pops up between the last PR and yours, it is not unreasonable to simply ask that you fix it.

In any case, I don't know how to implement that right now. Would it be okay if we try it out like this for now, and if it turns out to be a burden then we roll it back until we work out how to implement your suggestion?

PR #1290, opened 2017-02-02
Replace GPL with LGPL in LICENSE.txt

nottrobin:
The `README.md` states:

> ... and the underlying code used to format and display that content is licensed under the LGPLv3

But the contents of `LICENSE.md` shows GPL. This makes them consistent.

I've opted to change `LICENSE.md` to LGPL, rather than updating `README.md` to state GPL,
because LGPL is the more permissive license of the two, and as such if we decide later
to upgrade to GPL then there's no problem. However, downgrading to LGPL is more problematic.


deadlight:
@pmahnke any input on this one?

PR #1357, opened 2017-02-20
fix for stepped lists until it's fixed upstream

deadlight:
## Done

Fixed stepped lists at exactly 768px. We need a better solution in Vanilla but this is a quick fix for now.

## QA

Go to any stepped-list page (e.g., `/download/cloud/conjure-up`) and check it at all viewports but specifically exactly 768px wide.


grahambancroft:
LGTM 👍

PR #1361, opened 2017-02-20
snags for voerview page

deadlight:
## Done

Design tweaks from @joanasa89 

## QA

Check that the page renders and the the code is not bad - `/containers`


grahambancroft:
LGTM 👍

PR #1362, opened 2017-02-21
add context params to the contact URLs of the containers section

deadlight:
## Done

Added context parameters to the contact us links

## QA

See that the parameter is added to the contact us link URLs


grahambancroft:
LGTM 👍

PR #1365, opened 2017-02-21
add webinar view to k8s row

deadlight:
## Done

Add webinar row to k8s page

## QA

See that the row is now present


grahambancroft:
lgtm 👍

PR #1372, opened 2017-02-22
update whitepaper URL

deadlight:
## Done

Updated the whitepaper include with the correct URL.

## QA

Go to `/containters` and see that the whitepaper has the correct URL and no longer has the note that the URL is missing.


grahambancroft:
Lgtm 💯

PR #1374, opened 2017-02-22
update URL for k8s webinar to the on-demand one

deadlight:
## Done

Updated webinar URL
NOTE: the URL isn't live yet so will 404

## QA

See that the URL is updated


grahambancroft:
LGTM 👍 and the url is live

PR #1382, opened 2017-02-23
turns out that 'whitepaper' isn't a word; corrected two instances

deadlight:
## Done

Corrected two misspellings of "white paper" as "whitepaper"


grahambancroft:
LGTM 👍

PR #1388, opened 2017-02-26
fixed angle brackets on provisioning download page

pmahnke:
## Done

1. added html entities for local host

## QA

1. go to /download/server/provisioning
2. see that Login to the MAAS UI at http://&lt;your.maas.ip&gt;:5240/MAAS/ is visible


pmahnke:
[fixed angle brackets on provisioning download page](https://trello.com/c/1WGcEqFH/2683-fixed-angle-brackets-on-provisioning-download-page)

PR #1417, opened 2017-03-08
Fix contributions for empty fields

### LINKED ISSUE #1288: Contribution error
## Summary
Removing the content of a contribution value input results in the total displaying an error.

## Process
- Go to /download/desktop/contribute
- Delete the number from a selection
- See that the total shows NaN and the icon image is broken

### Current
Shows `NaN` in total and setting the image to undefined.

### Expected
When the user removes the numbers it's set to `0`. There should also be more error checking here.

## Screenshot
image

nottrobin:
If a field has its value deleted:

- Always equate it to zero for all calculations
- Set it back to zero when the field loses focus

Fixes #1288.

QA
--

Run the site and go to <http://127.0.0.1:8001/download/desktop/contribute?version=16.04.2&architecture=amd64>.

Delete the value from a field. See that the form page still looks sane, acting as if the empty field had 0 in it.

Now click off the field. See that it fills in with "0".

### DIFF
--- a/static/js/forms.js
+++ b/static/js/forms.js
@@ -21,7 +21,7 @@ forms.addSliders = function (numberElementsSelector, slidersClass) {
     var updateValueClosure = function (sisterNode) {
       return function(changeEvent) {
         targetNode = changeEvent.target;
-        sisterNode.value = targetNode.value;
+        sisterNode.value = targetNode.value || 0;
       }
     }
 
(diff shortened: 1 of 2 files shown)


nottrobin:
[Contribution error](https://trello.com/c/DTDNuG9x/2751-contribution-error)

PR #1419, opened 2017-03-08
Lxd support image

deadlight:
## Done

Fixed the support image on `/containers/lxd` to 200px wide.

## QA

See that the support image size is updated on `/containers/lxd`


deadlight:
I don't know why github thinks that this will update managed cloud too. That change is landed already and you can see it on master.

templates/cloud/managed-cloud.html
```diff
@@ -36,8 +36,8 @@
     <ul class="no-bullets equal-height--vertical-divider">
       <li class="four-col equal-height--vertical-divider__item">
         <img src="{{ ASSET_SERVER_URL }}c8eb3c73-cloud-econonmics-webinar.png?w=300" alt="">
-        <h3>Learn why Private Open Cloud Makes Financial Sense</h3>
```
nottrobin: yuck

PR #1447, opened 2017-03-21
Remove reference to "run node" command

nottrobin:
This command never existed. If we wanted to provide it we
could do so with `docker run --entrypoint node ubuntudesign/yarn`, but
I don't think we should complicate the run command until we have a need
for it.

QA
--

Confirm the `node` command doesn't exist:

``` bash
$ ./run node
Command not recognised.

Usage
===
etc.
```

Confirm that the usage instructions no longer mention node:

``` bash
$ ./run --help
...
Commands
---

- start [-w|--watch]: Run the development server (optionally watching for Sass changes) \
- watch: Run `yarn run watch`
- build: Run `yarn run build`
- test: Run `yarn run test` and `./manage.py test`
- yarn <args>: Run yarn
- django <args>: Run ./manage.py <args>
- clean: Remove all images and containers, any installed dependencies and the .docker-project file
```

### DIFF
--- a/run
+++ b/run
@@ -25,7 +25,6 @@ Commands
 - watch: Run \`yarn run watch\`
 - build: Run \`yarn run build\`
 - test: Run \`yarn run test\` and \`./manage.py test\`
-- node <args>: Run node
 - yarn <args>: Run yarn
 - django <args>: Run ./manage.py <args>
 - clean: Remove all images and containers, any installed dependencies and the .docker-project file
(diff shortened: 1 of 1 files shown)


barrymcgee:
No mention of `run node` anywhere.

Only issue is when I try to run `./run test`

screen shot 2017-03-21 at 17 44 38

nottrobin:
Yes, `yarn run test` isn't implemented, so that's the expected behaviour.

PR #1455, opened 2017-03-22
Use new docker image

nottrobin:
This will do a better job of detecting whether a database is needed.

I was having a problem with the old method - for some reason my django
container thought there was a service available called `db`, so it was
trying to connect to the database, and it was breaking Django. This
solves that.

QA
--

Check `./run` still works.


nottrobin:
[Improve database detection in ./run script](https://trello.com/c/NtbOwDbZ/2881-improve-database-detection-in-run-script)

PR #1459, opened 2017-03-24
Move ./run to use host yarn cache folder

WillMoggridge:
This mounts the host yarn cache into the docker image, so it can share the cache and keep everything wonderfully fast!

## QA
 - Install yarn on your computer - `npm install -g yarn`
 - `yarn install`
 - Delete node_modules
 - `./run yarn install` - should be fast using cache, with no errors
 - Delete node_modules again
 - `yarn install` - no errors

This needs reviews from someone on Linux and on OSX

### DIFF
--- a/run
+++ b/run
@@ -28,6 +28,7 @@ Commands
 - yarn <args>: Run yarn
 - django <args>: Run ./manage.py <args>
 - clean: Remove all images and containers, any installed dependencies and the .docker-project file
+- clean-cache: Empty cache files, which are saved between projects (eg, yarn)
 
 Persistent options
 ---
@@ -61,6 +62,9 @@ if grep -q '^docker:' /etc/group && ! groups | grep -q '\bdocker\b'; then
     exit 1
 fi
 
+# Check if yarn is locally installed
+command -v yarn >/dev/null 2>&1 && LOCAL_YARN_INSTALLED=1 || LOCAL_YARN_INSTALLED=0
+
 # Generate the project name
 if [[ -f ".docker-project" ]]; then
   project=$(cat .docker-project)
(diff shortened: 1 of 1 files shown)


WillMoggridge:
[Make ./run use central cache](https://trello.com/c/lT45bKW7/2899-make-run-use-central-cache)

WillMoggridge:
I had a quick test with @grahambancroft but can you have a look @pmahnke?

And then I would like someone on Linux to test this.

anthonydillon:
Linux :+1:

WillMoggridge:
Going to add some Docker volume functionality after talking with Robin and he has helped with permissions via: https://github.com/canonical-webteam/docker-yarn/pull/1

run
```diff
@@ -141,20 +168,33 @@ docker_django () {
...
+clean_cache () {
+  # Clean yarn cache volume
+  echo "Deleting yarn_cache Docker volume..."
+  docker volume rm yarn_cache 2>/dev/null || echo "No volume found"
+  if [ $LOCAL_YARN_INSTALLED = true ]; then
+    if [ $OPT_CACHE_CLEAN_FORCE = true ]; then
+      yarn cache clean
+    else
+      echo "#"
+      echo "# Yarn is locally installed, use --force to clean your local cache"
```
nottrobin: This is making me think - maybe we should never use the local yarn cache? I think it might just confuse things. For all intents and purposes I think it makes sense to consider `docker_yarn` to be a separate application from `yarn` - it may well be a different version etc. - so maybe we should simplify things considerably by only ever storing the `docker_yarn` cache in a docker volume.

nottrobin: @WillMoggridge if you're getting sick of the amount of discussion going on here, I'm also happy for you to merge this and then I can propose my ideas as a future PR.

PR #1466, opened 2017-03-29
Remove interaction with system yarn

nottrobin:
To make the ./run script more self-encapsulated, don't use the local
system yarn cache even if it's available.

This was discussed in https://github.com/canonical-websites/www.ubuntu.com/pull/1459#pullrequestreview-29230811.

The default behaviour should be to always keep anything that the ./run tooling relies on completely within docker.

If you want to you can override using the `YARN_CACHE_VOLUME` env var, but I'm not documenting this at the moment as I don't think it should be encouraged.

However, if someone else would like to document it at the bottom of HACKING.md, I'd have no problem with that.

QA
---

1. `./run clean && ./run clean-cache` - clean everything
2. `./run yarn` - note that step `[2/4] Fetching packages...` takes a couple of seconds
3. `./run clean` - delete `node_modules`
4. `./run yarn` - note that step 2 is almost instant
5. `./run clean && ./run clean-cache` - delete `node_modules` *and* the cache volume
6. `./run yarn` - note that step 2 takes a couple of seconds as it builds the cache
7. `./run clean && ./run clean-cache` - clean again
8. `yarn && rm -r node_modules` - build your local yarn cache
9. ``CANONICAL_WEBTEAM_YARN_CACHE_VOLUME=`yarn cache dir` ./run yarn`` - run using local cache - step 2 should be almost instant
10. `./run clean-cache` - you should see "No cache volume found" - to prove we weren't using the docker cache

### DIFF
--- a/run
+++ b/run
@@ -83,15 +83,9 @@ if [ -f .env ]; then
     export $(cat .env | grep -v ^\# | xargs)
 fi
 
-# Bind yarn cache to named volume or local yarn cache if available
-yarn_cache_host="yarn_cache"
-yarn_cache_docker="/home/shared/.cache/yarn/"
-# If yarn is installed, use its local cache directory
-if [ "$LOCAL_YARN_INSTALLED" = true ]; then
-  yarn_cache_host=`yarn cache dir`
-  mkdir -p ${yarn_cache_host}
-fi
-yarn_cache_arg="--volume ${yarn_cache_host}:${yarn_cache_docker}"
+# Bind yarn cache to "yarn-cache" volume
+# Or override with $YARN_CACHE_VOLUME
+yarn_cache_arg="--volume ${YARN_CACHE_VOLUME:-yarn-cache}:/home/shared/.cache/yarn/"
 
 # Optional arguments
 # ==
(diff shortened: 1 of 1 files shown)


run
```diff
@@ -181,16 +166,11 @@ yarn_install () {
...
-    if [ $OPT_CACHE_CLEAN_FORCE = true ]; then
-      yarn cache clean
-    else
-      echo "#"
-      echo "# Yarn is locally installed, use --force to clean your local cache"
-      echo "#"
-    fi
+  if docker volume ls | grep -q yarn-cache; then
+    echo "Deleting cache volume:"
+    docker volume rm --force yarn-cache
```
WillMoggridge: I am getting an error from --force but it looks like I am also getting it on the clean command. May be outside the scope of this.

run
```diff
@@ -114,15 +108,6 @@ module_volumes=()
 
 while [[ -n "${1:-}" ]] && [[ "${1:0:1}" == "-" ]]; do
     key="$1"
-    case $run_command in
-      "clean-cache")
-          case $key in
-              -f|--force)
-                  OPT_CACHE_CLEAN_FORCE=true
```
WillMoggridge: The default value for this is still set in the file.

run
nottrobin: Yep let's ignore that for now :)

WillMoggridge:
Awesome!

PR #1478, opened 2017-04-03
Add ./run detach option, with bonus namespacing and cleaning

WillMoggridge:
## Done

- Added `./run --detach` to run in background
- Namespaced all of the containers, for easy management
- `./run clean` now cleans all said namespaced containers
- Also, added yarn cache volume to yarn_install

## QA

- `./run` should work fine
- `./run --detach` should run in background. Can test with `docker ps`
- `./run` should kill background process and run as normal
- Running `./run --detach` should be killed and cleaned when running `./run clean`

### DIFF
--- a/run
+++ b/run
@@ -184,6 +184,7 @@ yarn_install () {
   # Install yarn dependencies, without module overrides
   docker_run \
     --name ${container_name} \
+    --volume ${CANONICAL_WEBTEAM_YARN_CACHE_VOLUME:-canonical-webteam-yarn-cache}:/home/shared/.cache/yarn/  `# Bind yarn cache to volume` \
     ${yarn_image} install
 }
 
--- a/run
+++ b/run
@@ -120,6 +120,15 @@ while [[ -n "${1:-}" ]] && [[ "${1:0:1}" == "-" ]]; do
     shift
 done
 
+kill_docker_container() {
+  container_name="${1}"
+  # Kill any previously running containers
+  if docker ps | grep -q ${container_name}; then
+    echo "Killing previous container:"
+    docker kill ${container_name} || true
+  fi
+}
+
 docker_run() {
   # The standard set of "docker run" arguments
 
@@ -138,12 +147,7 @@ docker_run() {
 
 docker_django () {
   container_name="${project}-django"
-
-  # Kill any previously running containers
-  if docker ps | grep -q ${container_name}; then
-    echo "Killing previous container:"
-    docker kill ${container_name} || true
-  fi
+  kill_docker_container "${container_name}"
 
   # Run Django using the docker image
   docker_run  \
@@ -157,12 +161,7 @@ docker_django () {
 
 docker_yarn () {
   container_name="${project}-yarn"
-
-  # Kill any previously running containers
-  if docker ps | grep -q ${container_name}; then
-    echo "Killing previous container:"
-    docker kill ${container_name} || true
-  fi
+  kill_docker_container "${container_name}"
 
   # Run "yarn" from the "node" image
   docker_run  \
@@ -174,12 +173,7 @@ docker_yarn () {
 
 yarn_install () {
   container_name="${project}-yarn-install"
-
-  # Kill any previously running containers
-  if docker ps | grep -q ${container_name}; then
-    echo "Killing previous container:"
-    docker kill ${container_name} || true
-  fi
+  kill_docker_container "${container_name}"
 
   # Install yarn dependencies, without module overrides
   docker_run \
@@ -211,7 +205,7 @@ case $run_command in
     # Run watch command in the background
     if [[ ${1:-} == "--watch" ]] || [[ ${1:-} == "-w" ]]; then
       container_name="${project}-yarn-$(date +'%s')"
-      trap 'docker kill ${container_name}' EXIT
+      kill_docker_container "${container_name}"
       docker_run  \
         --name ${container_name}  `# Name the unit so it can be killed` \
         --detach                  `# Run in the background` \
--- a/run
+++ b/run
@@ -13,6 +13,7 @@ USAGE="Usage
   $ ./run \\
     [-p|--port NUM]          # Specify local port for server \\
     [-m|--node-module PATH]  # A path to a local node module to use instead of the installed dependencies \\
+    [-d|--detach]            # Detach container to the backgorund \\
     [--no-debug]             # Turn off Django's DEBUG setting \\
     [COMMAND]                # Optionally provide a command to run
 
@@ -73,6 +74,7 @@ else
 fi
 
 # Defaults
+DETACH=''
 PORT=8000
 DJANGO_DEBUG=true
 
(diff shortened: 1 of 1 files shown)


WillMoggridge:
[Fix demos](https://trello.com/c/4SJYCaAN/2961-fix-demos)

anthonydillon:
Code :+1:

grahambancroft:
[fix new global nav on mobile](https://trello.com/c/plxbRwRZ/2945-fix-new-global-nav-on-mobile)

anthonydillon:
Getting a QA issue when running detached:
```bash
./run --detach
yarn install v0.21.3
warning ubuntu-website@0.0.1: License should be a valid SPDX license expression
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.46s.
yarn run v0.21.3
$ node-sass --include-path node_modules --output static/css static/css 
Rendering Complete, saving .css file...
Wrote CSS to /home/ant/Documents/Canonical/Git/www.ubuntu.com/static/css/core-print.css
Rendering Complete, saving .css file...
Wrote CSS to /home/ant/Documents/Canonical/Git/www.ubuntu.com/static/css/phone.ie9.css
Rendering Complete, saving .css file...
Wrote CSS to /home/ant/Documents/Canonical/Git/www.ubuntu.com/static/css/styles.ie9a.css
Rendering Complete, saving .css file...
Wrote CSS to /home/ant/Documents/Canonical/Git/www.ubuntu.com/static/css/styles.ie9b.css
Rendering Complete, saving .css file...
Wrote CSS to /home/ant/Documents/Canonical/Git/www.ubuntu.com/static/css/styles.css
Wrote 5 CSS files to /home/ant/Documents/Canonical/Git/www.ubuntu.com/static/css
Done in 2.12s.
Conflicting options: --rm and -d
```

run
```diff
@@ -177,8 +204,8 @@ case $run_command in
 
     # Run watch command in the background
     if [[ ${1:-} == "--watch" ]] || [[ ${1:-} == "-w" ]]; then
-      container_name="yarn-$(date +'%s')"
-      trap 'docker kill ${container_name}' EXIT
```
nottrobin: I've tested this without this line, and it does, as I suspected, leave the watcher running:

``` bash
$ docker ps -q
$ ./run start --watch
** run the server**
** ctrl + c **
$ docker ps -q
d56ab6af95e8
```

We need this line.

run
```diff
@@ -177,8 +204,8 @@ case $run_command in
...
+      container_name="${project}-yarn-$(date +'%s')"
```
nottrobin: For consistency with the others, this should probably be `yarn-watch`, rather than `yarn $(date ...)`

run
```diff
@@ -202,6 +229,10 @@ case $run_command in
     docker_django test
   ;;
   "clean")
+    running_containers="$(docker ps -aq --filter name=${project})"
```
nottrobin: This isn't just running containers, the `-a` means it'll catch all stopped containers too (which is a good thing). We should probably call the variable `project_containers` or something.

nottrobin:
I love the idea of explicitly naming each of the containers, and then making sure they're dead before trying to spin them up again. That seems like a great way of keeping everything clean.

It's making me think that maybe it's actually better not to use `--rm` at all. Leaving the stopped containers in Docker until either the next run or `clean` is called would allow the dev to inspect the stopped containers if they wanted, which may be helpful.

If we don't have the `--rm`, am I right in thinking that the old `nohup ./run &` behaviour should work in Jenkins? Because then maybe we don't need the `--detach` option at all. I'd prefer that, because I'd rather the `./run` interface was a little opinionated about how you use it. I don't really want devs to start detaching their containers just so they don't have to open another terminal window, for example. I think it'll make supporting local dev a lot more messy.

So my suggestions (mostly also mentioned in inline comments):

- Keep the explicit naming of each container
- Fixup the `watch` container to
  a) have an explicit name
  b) reinstate the `trap`
- Remove `--detach`

Let me know your thoughts.

WillMoggridge:
@anthonydillon `--rm` was made to work with `--detach` in docker 1.13. If you upgrade Docker it should work nicely.

run
WillMoggridge: Agreed

run
WillMoggridge: Yeah, this was a misunderstanding from my part, though it did seem broken. It wasn't using `${container_name}` in that scope, and was trying to delete the django container.

I have added more signals to trap and echod the variable directly into the trap command.

run
WillMoggridge: Yes very true, that won't work too well with a date!

WillMoggridge:
I have been trying to think about my position on this.
It does feel weird to not use Docker's `--detach` option designed for this use case, and use a nohup solution which is what caused us some problems in the first place.

Other benefits is that it keeps the script blocking. One way this is useful is that the build will actually fail if `./run` fails. A few times people have been confused because a build passes but there was an error.
(Also nohup makes Jenkins complain, which is a slight annoyance :) )


As for `--rm`, I find myself quite pro towards it. I was really liking the auto cleaning up. It is nice on my machine but especially nice on the demo server if you want to look at docker containers.

run
nottrobin: I'd like to see the error. I'm sure I tested it out plenty of times.

run
```diff
@@ -177,12 +204,13 @@ case $run_command in
...
     if [[ ${1:-} == "--watch" ]] || [[ ${1:-} == "-w" ]]; then
-      container_name="yarn-$(date +'%s')"
-      trap 'docker kill ${container_name}' EXIT
+      container_name="${project}-yarn-watch"
+      kill_docker_container "${container_name}"
       docker_run  \
         --name ${container_name}  `# Name the unit so it can be killed` \
         --detach                  `# Run in the background` \
         $yarn_image run watch     `# Use the image for node version 7`
+      trap "kill_docker_container '${container_name}'" EXIT INT TERM
```
nottrobin: Have you confirmed that the extra signals are necessary?

run
```diff
@@ -117,6 +120,15 @@ while [[ -n "${1:-}" ]] && [[ "${1:0:1}" == "-" ]]; do
     shift
 done
 
+kill_docker_container() {
+  container_name="${1}"
+  # Kill any previously running containers
+  if docker ps | grep -q ${container_name}; then
+    echo "Killing previous container:"
+    docker kill ${container_name} || true
```
nottrobin: Although `--rm` shouldn't leave stopped containers behind, if a stopped container does get left behind then the "run" command will fail. We should delete stopped containers as well, by using `docker ps --all` for detection, and `docker rm --force` for killing & removing.

run
nottrobin: Oh I can see how it could have been erroring. Because `${container_name}` gets parsed at the point of `ctrl+c` rather than on instantiation, if its value has changed then it will kill the wrong thing. My testing probably never hit that particular state. Good catch.

nottrobin:
@WillMoggridge Okay, I'm convinced. I'm happy to leave the `--detach` option. If you just fix up the remaining couple of inline points I'm happy to approve this. Good work.

run
WillMoggridge: I was reading that EXIT doesn't always fire from a shell, as it isn't a shell standard. So run could be stopped externally and not fire it off.

But you bringing this up made me look into it again and it is easier/safer not to worry about that unless it comes up.

WillMoggridge:
@nottrobin OK cool, I have pushed changes up.

PR #1488, opened 2017-04-05
Add Landscape row to cloud overview page

anthonydillon:
## Done
Added a landscape section to the cloud overview page.

## QA
- Pull code
- Run `./run`
- Go to http://localhost:8001/cloud
- Check that the landscape section matches to copy doc
- The image was removed by design

## Details
Copy doc: https://drive.google.com/open?id=1mQjVXDcoP8ChOzRdmuI55BdXhbURD53qBNc2bgJhfhg

PR #1490, opened 2017-04-05
Update landscape self service on server overview

anthonydillon:
## Done
Removed OpenStack from the "What’s new in 16.04 LTS" section. 
Added the "Landscape: Server management" section. 
Added the "Ubuntu Advantage: World class support" section.

## QA
- Pull code
- Run `./run`
- Go to http://localhost:8001/server
- Check the new sections with the copy doc

Copy doc https://docs.google.com/document/d/1vxLhtbjgPs9PCNEhr7fbhEoErCPMXYFfU4UTKH7Kbq4/edit#heading=h.2dc8qpnnp7j1

### DIFF
--- a/templates/server/index.html
+++ b/templates/server/index.html
@@ -82,17 +82,17 @@
             <p>
                 Landscape allows you to manage thousands of Ubuntu machines as
                 easily as one, making the administration of Ubuntu desktops,
-                servers and cloud instances more cost-effective. It&rsquo;s
+                servers and cloud instances more cost-effective.<br />It&rsquo;s
                 easy to set up, easy to use and it requires no special
                 hardware. It features:
             </p>
             <div class="combined-list">
-                <ul class="list-ticks four-col">
+                <ul class="list-ticks--compact four-col">
                     <li>Management at scale</li>
                     <li>Deploy or rollback security updates</li>
                     <li>Kernel livepatching </li>
                 </ul>
-                <ul class="list-ticks four-col last-col">
+                <ul class="list-ticks--compact four-col last-col">
                     <li>Compliance reporting</li>
                     <li>Role-based access</li>
                     <li>Informative monitoring</li>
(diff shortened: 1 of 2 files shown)


anthonydillon:
[Update /server for landscape-self-service](https://trello.com/c/9iWd6O05/2982-update-server-for-landscape-self-service)

templates/server/index.html
```diff
@@ -28,34 +28,84 @@
...
+                </ul>
+            </div>
+            <p>
+                <a href="https://landscape.canonical.com/"
+                    class="button--primary">
+                    <span class="external">
+                        Learn more about Landscape
+                    </span>
+                </a>
+            </p>
```
grahambancroft: This seems to be the same copy as the new section in #1488 and https://github.com/canonical-websites/www.ubuntu.com/pull/1487 could it be a shared include? The only difference I can see is one of the list items here is 'Kernel livepatching' whereas on the other two it’s 'Live patching’, guessing this needs to be consistent.

templates/server/index.html
```diff
@@ -28,34 +28,84 @@
...
-                <li>Certified by Microsoft to host Windows Server 2012 and Windows Server 2008 R2 as guests, under its Server Virtualisation Validation Program (SVVP)</li>
-                <li>OpenStack software, Juju, MAAS, LXD on IBM LinuxONE mainframe</li>
-            </ul>
+    </div>
+</section>
+
+<section class="row strip-light">
+    <div class="strip-inner-wrapper equal-height">
+        <div class="four-col equal-height__item equal-height__align-vertically align-center">
+            <img class="priority-0" src="{{ ASSET_SERVER_URL }}e047e34e-picto-landscape.svg"
```
grahambancroft: There’s no image in the copy doc

anthonydillon: I borrowed the image from a similar section. Though design would like to see an image.

templates/server/index.html
anthonydillon: The problem is that they have a different layout. combined list vs a single list. Not sure it would be an easy include. 
@Greggless should they all be "Kernel livepatching"

Greggless: @anthonydillon yes: 'kernel livepatching' is what we'll be using.

PR #1501, opened 2017-04-07
remove mobile section, remove the styles and references in the nav

deadlight:
## Done

Removed the mobile section, its styling and navigation links.

Note: links to the site elsewhere haven't been removed yet.

## QA

Check that the site still works with the section removed.


deadlight:
Yeah, my checks failed. Deal with it! (By which I mean, I'll deal with it in future commits)

PR #1504, opened 2017-04-07
remove TV section and associated styling. Redirected all TV URLs to site homepage

deadlight:
## Done

Removed the TV files, their SCSS files and redirected the TV section to the homepage

## QA

Check that the site still works and that /tv redirects to the homepage.


redirects.yaml
```diff
@@ -639,14 +639,14 @@ devices/tablet/contact-us/?: "/mobile"
...
-devices/tv/industry/?: "/tv/industry"
-devices/tv/thank-you/?: "/tv"
+devices/tv/?: "/"
+devices/tv/commercial-info/?: "/"
+devices/tv/contact-us/?: "/"
+devices/tv/contributors/?: "/"
+devices/tv/experience/?: "/"
+devices/tv/features-and-specs/?: "/"
+devices/tv/industry/?: "/"
+devices/tv/thank-you/?: "/"
```
anthonydillon: I would prefer these to 404

deadlight: sure thing

PR #1508, opened 2017-04-10
Remove mobile from homepage

deadlight:
## Done

 - Removed mobile row from homepage
 - Removed mobile icon and associated styling from signpost

## QA

view the homepage in-browser.

[Demo](http://www.ubuntu.com-remove-mobile-from-homepage.demo.haus/)

### DIFF
--- a/static/css/section/_homepage.scss
+++ b/static/css/section/_homepage.scss
@@ -511,13 +511,13 @@ body.homepage .row--cloud-products {
     @media only screen and (min-width : $breakpoint-medium) {
       position: absolute;
       top: -40px;
-      right: 0;
+      left: -280px;
       width: 700px;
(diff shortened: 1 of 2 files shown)


barrymcgee:
@yaili - is it ok for both of the row images to be on the right? Illustration: https://cl.ly/0343361H1E0e

yaili:
@barrymcgee no, I think Juju should moved to the left.

anthonydillon:
Looks good to me but would need a designer to check it.

joanasa89:
Can we move the 'openstack' copy to the right? Maybe  "class="four-col prepend-**eight**"

anthonydillon:
Graham is following this one with a fix.

PR #1513, opened 2017-04-11
remove mobile redirects

deadlight:
## Done

Removed a lot of old mobile redirects.

## QA

Check that the site still works and none of the redirects are essential.


anthonydillon:
Looks like there are some 404 links on the homepage. Is that why this has the do not merge label?

PR #1547, opened 2017-04-20
remove old homepage takeover/takeunder partials/css

### LINKED ISSUE #1510: Takeover include partial need review
There is a huge number of takeover partials in `templates/takeovers` - I assume many are outdated. 

Prior to Vanilla v1 migrations, these should be reviewed to determine which partials are still relevant.

pmahnke:
## Done

* removed all but current and generic takeovers and takeunders
* removed related css

## QA

1. go to the homepage
2. see that is looks fine
3. go to https://github.com/ubuntudesign/ubuntu.com-takeover-archive and see that it looks complete

## Issue / Card

Fixes #1510


anthonydillon:
The only homepage that is broken is `index2` for a/b testing the position of the insights row. This will need to be  updated to `_1704_takeover` instead `_containers-ebook`.

PR #1566, opened 2017-04-24
Update to latest run script

nottrobin:
Update the `run` script and `package.json` to use latest patterns from [generator-canonical-webteam](https://www.npmjs.com/package/generator-canonical-webteam) version 0.1.6.

This involves updating the standard set of NPM scripts in `package.json`:

- build: Use `node-sass` to build sass from the `static/sass` dir into `static/css`
- watch: Watch for changes to `static/sass/styles.scss`
- test: Use `sass-lint` to check Scss against standard rules
- clean: Remove local files, e.g. `node_modules`

To accommodate these scripts, we are also:

- Moving `node-sass` and `ubuntu-vanilla-theme` from `dependencies` to `devDependencies`
- Using the standard version of `node-sass` from generator-canonical-webteam@0.1.6
- Adding `sass-lint` to `devDependencies`
- Updating `yarn.lock` to match these new dependencies
- Moving all `*.scss` files form `static/css` to `static/sass`
- Adding `.sass-lint.yml` sass linting rules from vanilla-framework

QA
--

Check out each of the functions of the run script:

``` bash
./run clean                   # Tidy up any previous settings
./run                         # Run the server
./run build                   # Build sass
./run watch                   # Watch for changes to sass files
./run serve --watch           # Run server with background watcher
./run serve --detach          # Run the server in the background 
./run serve --detach --watch  # Run the server in the background and watch
./run test                    # Test the sass config
```


barrymcgee:
Looks very interesting indeed.. you have some conflicts to resolve though first?

barrymcgee:
For all commands, I get a permission error without running `sudo`

screen shot 2017-04-25 at 09 17 56

barrymcgee:
Can we have it so the watcher watches all `.scss` files in the /sass folder and runs whenever any of them is changed? Anything `.scss` file without a leading underscore would then be transpiled into the /css folder?

nottrobin:
I don't think it's feasible to avoid the warning about the yarn version. We are deliberately using a specific version of yarn, and it will probably become out of date pretty fast. We'll update it as and when we can or need to.

I agree it would ideally watch for changes in any sass files. The reason it's only watching the `styles.scss` is because `node-sass` ended up in a weird race condition when I tried to recursively watch a directory. I've filed [an issue](https://github.com/canonical-webteam/generator-canonical-webteam/issues/13), but would you mind if we consider solving that out-of-scope of this PR for now? Feel free to experiment yourself by editing the package.json.

I don't understand that permissions error you're seeing. I'll try to get access to someone's mac right now to solve that.

deadlight:
@nottrobin a workaround that would help us continue working in the meantime would be to build styles-v1.scss as well.

deadlight:
@nottrobin no issues running `./run` and `./run watch` for me after running `./run clean`

barrymcgee:
Good to merge then...

PR #1718, opened 2017-05-04
Cookie bar

deadlight:
## Done

Updated the cookie bar to the latest design and sandboxed the changes to the v1 rollout

## QA

If you do not see the cookie bar on `/about` then add `state('open');` to line 114 of `scratch-v1.js`.

Check the cookie bar against [the design](https://cloud.githubusercontent.com/assets/9105999/23611548/cf5fd186-026f-11e7-89e4-693e4f1b6d08.png) and ensure that it works correctly at all viewports.


static/sass/_v1_pattern_notifications.scss
```diff
@@ -0,0 +1,41 @@
+// Notification style patterns
+@mixin ubuntu-p-notification {
+  @include vf-notifications-default;
+  @include vf-notifications-positive;
+  @include vf-notifications-warning;
+  @include vf-notifications-negative;
```
anthonydillon: These should be included by the framework.

static/sass/_v1_pattern_notifications.scss
```diff
@@ -0,0 +1,41 @@
...
+    bottom: 0;
+    position: fixed;
+    text-align: left;
+    width: 100%;
+    z-index: 2;
+
+    @media (min-width: $breakpoint-large) {
+      left: 50%;
+      margin-left: calc(#{$breakpoint-large} / -2);
+      width: $breakpoint-large;
```
anthonydillon: We should use `$grid-max-width` here.

deadlight:
@anthonydillon ready for rereview

PR #1724, opened 2017-05-04
Contextual footer top border should only be 3px

barrymcgee:
## Done

Contextual footer top border should only be 3px

## QA

- Pull code
- Go to /about
- Check that top orange border on contextual footer is only 3px


anthonydillon:
Was this merged without a review?

PR #1728, opened 2017-05-05
Remove small-screen-specific styling for the footer inline list

deadlight:
## Done

Made the link list in the footer ("Contact us" etc.) look the same on all viewports. This required updated the compact variant of button to stop it being 100% at mobile.

## QA

Check the footer of `/about` page in all viewports.

### DIFF
--- a/static/sass/_v1_pattern_footer.scss
+++ b/static/sass/_v1_pattern_footer.scss
@@ -190,14 +190,6 @@
     a:hover {
       color: currentColor;
     }
-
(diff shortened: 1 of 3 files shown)


barrymcgee:
screen shot 2017-05-05 at 16 23 06

deadlight:
looks like there is no simple way to rectify inline lists at small viewports. Closing this for now and I'll propose a simpler solution (copying existing styling)

PR #1781, opened 2017-05-15
Add -v1 partials for each section

barrymcgee:
## Done

Add -v1 partials for each section to make picking up a new section easier

## QA

Check code and make sure everything makes sense.


anthonydillon:
It's time to kill keywords. Let's remove them from the v1 base templates, please.

PR #1789, opened 2017-05-16
1621 download alternative downloads

### LINKED ISSUE #1621: /download/alternative-downloads/ to VBT

pmahnke:
## Done

* updated alternative downloads - http://0.0.0.0:8001/download/alternative-downloads
* fixed contextual footers to use p-link--external, instead of external and added equal height to each section so the dividers look even

## QA

- Check out this feature branch or the [demo](http://www.ubuntu.com-1621-download-alternative-downloads.demo.haus/download/alternative-downloads)
- look at the above page and see if it looks ok
- fixed external link class and made equal heights

## Issue / Card

Fixes #1621

### DIFF
--- a/static/sass/_v1_pattern_link.scss
+++ b/static/sass/_v1_pattern_link.scss
@@ -1,0 +1,12 @@
+@mixin ubuntu-p-link {
+  h1.p-link--external,
+  h2.p-link--external,
+  h3.p-link--external,
+  h4.p-link--external,
+  h5.p-link--external,
+  h6.p-link--external {
+    background: url('#{$assets-path}e1bba201-external-link-cool-grey.svg') left center no-repeat;
+    background-size: 20px;
+    padding-left: 26px;
+  }
+}
--- a/static/sass/_v1_pattern_link.scss
+++ b/static/sass/_v1_pattern_link.scss
@@ -6,7 +6,7 @@
   h5,
   h6 {
     &.p-link--external {
-      background: url('#{$assets-path}e1bba201-external-link-cool-grey.svg') left #{(1rem / 3)} no-repeat;
+      background: url('#{$assets-path}e1bba201-external-link-cool-grey.svg') left #{(1em / 3)} no-repeat;
       background-size: $sp-medium;
       padding-left: $sp-large;
     }
--- a/static/sass/_v1_pattern_link.scss
+++ b/static/sass/_v1_pattern_link.scss
@@ -1,12 +1,14 @@
 @mixin ubuntu-p-link {
-  h1.p-link--external,
-  h2.p-link--external,
-  h3.p-link--external,
-  h4.p-link--external,
-  h5.p-link--external,
-  h6.p-link--external {
-    background: url('#{$assets-path}e1bba201-external-link-cool-grey.svg') left center no-repeat;
-    background-size: 20px;
-    padding-left: 26px;
+  h1,
+  h2,
+  h3,
+  h4,
+  h5,
+  h6 {
+    &.p-link--external {
+      background: url('#{$assets-path}e1bba201-external-link-cool-grey.svg') left #{(1em / 2)} no-repeat;
+      background-size: $sp-medium;
+      padding-left: $sp-large;
+    }
   }
 }
--- a/static/sass/_v1_pattern_link.scss
+++ b/static/sass/_v1_pattern_link.scss
@@ -6,7 +6,7 @@
   h5,
   h6 {
     &.p-link--external {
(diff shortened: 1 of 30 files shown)


static/sass/_v1_pattern_lists.scss
```diff
@@ -119,3 +121,57 @@
...
+          }
+
+          &:nth-last-child(0) {
+            border-bottom: 0;
+          }
+        }
+      }
+    }
+  }
+}
```
barrymcgee: A lot of duplicated code here - can it be abstracted?

static/sass/_v1_pattern_link.scss
```diff
@@ -0,0 +1,12 @@
...
+  h2.p-link--external,
+  h3.p-link--external,
+  h4.p-link--external,
+  h5.p-link--external,
+  h6.p-link--external {
+    background: url('#{$assets-path}e1bba201-external-link-cool-grey.svg') left center no-repeat;
+    background-size: 20px;
+    padding-left: 26px;
+  }
+}
```
barrymcgee: Cleaner like this:

```css
h1,
h2,
h3,
h4,
h5,
h6 {
  &.p-link--external {
     background: url('#{$assets-path}e1bba201-external-link-cool-grey.svg') left center no-repeat;
     background-size: 20px;
     padding-left: 26px;
   }
}

```

templates/download/alternative-downloads.html
```diff
@@ -1,169 +1,181 @@
...
-            <p><a class="external download-pastreleases" href="http://releases.ubuntu.com/">Past releases</a></p>
-        </div><!-- /.eight-col -->
-    </div><!-- /.strip-inner-wrapper -->
+<div class="p-strip">
+  <div class="row">
+    <div class="col-8">
+      <h2>Past releases and other flavours</h2>
+      <p>Looking for an older release of Ubuntu? Whether you need an obsolete release or a previous LTS point release with its original stack, you can find them in past releases.</p>
+      <p><a class="p-link--external" href="http://releases.ubuntu.com/">Past releases</a></p>
+    </div><!-- /.col-8 -->
```
barrymcgee: I've been removing these comments are most modern IDE's have element collapse built in making them redundant - they also get outdated really easily.

barrymcgee:
Another point to be aware of is that the divider pattern already has a change merged upstream that will make sure all items have equal height dividers meaning you don't need to use the util here unless you consider it critical in the meantime.

yaili:
Some notes:
- The 3 columns in "Select the nearest mirror" need the have the same gutter as if they were using the normal columns/grid, it looks narrower right now
- The the viewport becomes too narrow (840px and lower) the icons in the sub headings in BitTorrent are vertically centre aligned with the title, the icon should be top aligned — maybe add this as a bug to Vanilla?

templates/download/alternative-downloads.html
pmahnke: done

static/sass/_v1_pattern_link.scss
pmahnke: done

static/sass/_v1_pattern_lists.scss
pmahnke: it is a copy of the .is-divided pattern... perhaps if this is promoted we do that all at once?

pmahnke:
@barrymcgee @yaili demo updated... please re-review

@barrymcgee I will make a note to remove the equal hight on the _contextual-footer when that lands.

yaili:
The position of the external icon in the headings doesn't look right. I think if we make it 0.3rem instead of 0.5rem it looks better.

pmahnke:
@yaili updated... and updated the demo

yaili:
This is still not right, in large screens the size and position of the icon doesn't work. 

Should this be worked out in Vanilla, or the brochure theme? Should we be using ems here, so that the icon is in proportion to the heading, regardless of which heading size we're using? If I make the background size 0.9em, the size remains balanced across screen sizes, the position seems a bit trickier, 0.1em works in most viewports, 0.2em too, but not in all. And the left padding seems to work well at 1em across screen sizes. Thoughts? @pmahnke @anthonydillon @barrymcgee

pmahnke:
@yaili how about this

image

```
@mixin ubuntu-p-link {
  h1,
  h2,
  h3,
  h4,
  h5,
  h6 {
    &.p-link--external {
      background: url('#{$assets-path}e1bba201-external-link-cool-grey.svg') left #{(1em / 3)} no-repeat;
      background-size: .6em;
      padding-left: .75em;
    }
  }
}
```

barrymcgee:
All reviews passed, thanks @pmahnke 👌🏻

PR #1830, opened 2017-05-24
update /core/contact-us to vbt

### LINKED ISSUE #1599: /core/contact-us/ to VBT

pmahnke:
## Done

* updated /core/contact-us to VBT

## QA

- View the site locally in your web browser at: [page](http://0.0.0.0:8001/core/contact-us)
- [demo](http://www.ubuntu.com-1599-core-contact-us.demo.haus/core/contact-us)

## Issue / Card

Fixes #1599


yaili:
- [ ] There is a double border in the breadcrumbs at S screen because there is only one item

screen shot 2017-05-25 at 08 58 03

pmahnke:
@yaili I had fixed in contact us... but moved it here

yaili:
Seeing the same double border in S screen breadcrumbs bug

pmahnke:
@yaili fixed in both demos

yaili:
The Overview row in the S screen breadcrumbs is taller than the Core row, it should be the same height if both only have one item. There is also a couple of pixels gap between the breadcrumb and the left edge of the viewport but I've seen this fixed in about demos. Has this been fixed somewhere else, or should I file a separate bug to fix the height and/or gap?

pmahnke:
@yaili I assume this is what @deadlight and @grahambancroft are working on...

yaili:
@deadlight @grahambancroft can you confirm before I +1 this?

yaili:
- [ ] The chevron here is really bold and looks different - is there another demo link I can compare with that has a more final version of it? I'm confused as to which demos have the correct things

pmahnke:
@yaili I think @barrymcgee fixed the chevron over my fix of the chevron... so it was there twice!

PR #1863, opened 2017-05-28
1612 /download/cloud/autopilot to vbt

### LINKED ISSUE #1612: /cloud/openstack/autopilot/ to VBT

pmahnke:
## Done

* update /download/cloud/autopilot to vbt

## QA

- View the site locally in your web browser at: [page](http://0.0.0.0:8001/download/cloud/autopilot)
- compare to the [design](https://www.dropbox.com/work/00_web_team/_ubuntu.com/new%20projects/vanilla%20update/Patterns%20inventory/9.%20Download?preview=download-cloud-autopilot.png) and the [demo](http://www.ubuntu.com-1612-cloud-openstack-autopilot.demo.haus/download/cloud/autopilot)

## Issue / Card

Fixes #1612

### DIFF
--- a/static/sass/styles-v1.scss
+++ b/static/sass/styles-v1.scss
@@ -219,11 +219,15 @@ body.has-background {
 
 /// XXX fix nested bullets to use default styles
 ul.p-list  ul.p-list {
-  list-style-type: disc;
+  list-style-type: circle;
 
   ul.p-list {
-    list-style-type: circle;
+    list-style-type: disc;
   }
+}
+
+.p-list-step__content ul.p-list {
+  list-style-type: disc;
 }
 
 /// XXX fix two-digit padding in .p-list-step__bullet
--- a/static/sass/styles-v1.scss
+++ b/static/sass/styles-v1.scss
@@ -216,3 +216,17 @@ body.has-background {
     background-image: url(#{$assets-path}f8a323a7-image-background-paper.png);
   }
 }
+
+/// XXX fix nested bullets to use default styles
+ul.p-list  ul.p-list {
+  list-style-type: disc;
+
+  ul.p-list {
+    list-style-type: circle;
+  }
+}
+
+/// XXX fix two-digit padding in .p-list-step__bullet
+.p-list-step__bullet {
+  padding: .65rem .4rem;
+}
--- a/templates/download/cloud/autopilot.html
+++ b/templates/download/cloud/autopilot.html
@@ -1,269 +1,257 @@
-{% extends "download/_base_download.html" %}
-
+{% extends "download/_base_download-v1.html" %}
 {% block title %}Build OpenStack with Autopilot | Download{% endblock %}
-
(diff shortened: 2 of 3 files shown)


yaili:
- [ ] The nested tick list should have non-ticks in the nested lists, according to [this issue](https://github.com/ubuntudesign/vanilla-brochure-theme-design/issues/10) it sounds like this should have been done before
- [ ] Remove hero bottom border
- [ ] Does it make more sense to use normal lists in the stepped list (instructions) instead of ticked? Because they are not really "completed" or "positive" yet
- [ ] The items in the instructions section that have 2 digits don't align properly in the circle background (see image) - bug in Vanilla?
- [ ] Can we make the text column in the "Before you start" section col-7 instead of col-6?
- [ ] The p around the image in "Before you start" I think also needs the u-align--center class
- [ ] There is something wrong with the longer headings in the "Installation instructions" and it might be the Vanilla Brochure theme/Vanilla implementation of the pattern: there is some indentation which means the text doesn't align left and the line height of the heading is too high
- [ ] In S screen, the ebook form should be aligned left to the rest of the content, it's a bit indented

The last few items are exactly the same as https://github.com/canonical-websites/www.ubuntu.com/pull/1862 just added them for completeness.

screen shot 2017-05-28 at 22 23 27

pmahnke:
@yaili updated most

* removed padding on ebook fieldset (doesn't work on a contact us form, so I think just locally)
* made two-digit instructions fit in the circle (not very elegantly)
* columns size updates

can't really do ' something wrong with the longer headings' easily... I think it should be a vbt issue, I also think we give too little space to the instructions in the detailed pattern

yaili:
- [ ] Does the tick list in step 2 need to be there?
- [ ] The text below the 3rd code snippet in step 2 should have same left padding as text above first code snippet
- [ ] Same for text in step 3 and 4, and after the first paragraph in step 5
- [ ] In step 6 of the instructions, the text below the screenshot has more left padding than the text above, it should be the same

I've submitted a VBT bug for the list stepped detailed headings.

pmahnke:
@yaili fixed

static/sass/styles-v1.scss
```diff
@@ -178,3 +178,21 @@ fieldset {
   font-size: 0.875rem;
   text-transform: uppercase;
 }
+
+/// XXX fix nested bullets to use default styles
+ul.p-list  ul.p-list {
```
barrymcgee: Has this been logged upstream?

templates/download/cloud/autopilot.html
```diff
@@ -1,269 +1,249 @@
...
+{% include "download/shared/_get_ebook-v1.html"%}
+
+<div class="p-strip--light is-deep is-bordered" id="instructions">
+  <div class="row">
+    <div class="col-12">
+      <h2>Installation instructions</h2>
+
+      <ol class="p-stepped-list--detailed">
+
+        <li class="p-list-step__item col-12">
```
barrymcgee: `col-12` should not be used like this - and in this case (and all other `.p-list-step__item`), it seems redundant anyway?

pmahnke:
@barrymcgee ... looks ok to me locally and the demo... but the ./run script is borked, so perhaps the scss isn't building?

PR #1882, opened 2017-05-31
Ensure search icon is always consistent

### LINKED ISSUE #1738: Search icon should be the same everywhere
In M and S screen, the search icon in the nav is different from the one inside the input field. The one in the nav is correct.

barrymcgee:
## Done

Ensure search icon is always consistent by using same SVG is all instances, only changing color from white/black where appropriate

## QA

- View the site locally in your web browser at: [http://0.0.0.0:8001/about](http://0.0.0.0:8001/about))
- Verify search icon is the same (apart from color) at all breakpoints

## Issue / Card

Fixes #1738

## Demo

http://www.ubuntu.com-1738-search-icon.demo.haus/about/


yaili:
If the search results page is not included as part of this bug it's got design+1, otherwise search results still needs updated.

barrymcgee:
@yaili Didn't realise search results was part of this and it wasn't mentioned in the original issue. Will change there too and add to this PR.

yaili:
Thanks!

barrymcgee:
@yaili Search results icon updated; http://www.ubuntu.com-1738-search-icon.demo.haus/search?q=hello

PR #1883, opened 2017-05-31
Update to latest ./run

WillMoggridge:
Update ./run, and it should be compatible for PR demos!

Hopefully a comment will appear below about a demo...


barrymcgee:
YES @WillMoggridge 💥

PR #1894, opened 2017-06-02
Remove u-vertical-spaced from contextual footers.

### LINKED ISSUE #1865: Remove vertically-spaced from contextual footers
Is it part of Vanilla? If so, it should be documented and the issue moved there (I can do that).

Or is it just for ubuntu.com?

Either way, it's adding extra top and bottom padding to the wrapper which adds to the padding of the strips - can we review this?

deadlight:
## Done

Removed u-vertically-spaced from the contextual footer template on pages migrated to VBT.

## QA

- View the site locally in your web browser at: [http://0.0.0.0:8001/support](http://0.0.0.0:8001/support))
- See the reduced padding in the contextual footer.

## Issue / Card

Fixes #1865

### DIFF
--- a/static/sass/_v1_pattern_strip.scss
+++ b/static/sass/_v1_pattern_strip.scss
@@ -85,10 +85,3 @@
     }
   }
 }
-
-@mixin ubuntu-p-strip--x-light {
-  .p-strip--x-light {
-    @include strip;
-    background-color: $color-x-light;
-  }
-}
--- a/templates/shared-v1/contextual_footers/_contextual_footer.html
+++ b/templates/shared-v1/contextual_footers/_contextual_footer.html
@@ -1,6 +1,6 @@
 <div class="p-strip--x-light contextual-footer">
   <div class="row">
-    <div class="p-divider u-vertically-spaced">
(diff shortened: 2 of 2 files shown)


yaili:
The vertical padding is too tight now, because there is a rule ".p-strip--x-light" which is changing the default padding to 2rem, which is not right, the strip colour should do nothing to the padding. Should this be changed as part of this PR or separately?

deadlight:
@yaili I'll look into it here.

deadlight:
@yaili it's fixed now - it was an old overide. Nobody was trying to set specific padding for a colour override; it was just a quirk of vanilla inheritance.
fixed-contextual

yaili:
@deadlight can I see small screen too?

deadlight:
@yaili here's small
contextual-small

yaili:
@deadlight contact us button is orange in small screen but neutral in large - how is that possible?

deadlight:
@yaili there was some weirdness there. You're right, that isn't possible. Here's a fixed demo: http://www.ubuntu.com-1865-remove-vertically-spaced.demo.haus/support

PR #1940, opened 2017-06-07
Revert "Added canonical-global-nav as a dependency"

deadlight:
Reverts canonical-websites/www.ubuntu.com#1729

This caused some issues on merging which will require fixing.

## QA

Ensure that the old global nav still works at all viewports.


deadlight:
@yaili a demo: http://www.ubuntu.com-revert-1729-global-nav.demo.haus/

deadlight:
The issue @yaili noticed with the "more" dropdown is an existing issue. I'll record it separately.

PR #1942, opened 2017-06-07
Intelligently generate top menu and footer nav

nottrobin:
Introducing a new `navigation.yaml` document, which describes the pages included in the navigation and their relationships to each other.

This is then used to generate the top menu and the footer nav.

This is the first step in fixing #1944 

Demo: http://www.ubuntu.com-navigation-menu-footer.demo.haus/

QA
--

`./run`, then browse around the site and check the top menu and the footer nav work as expected. Especially:

- That pages not included in the menu (e.g. contact-us pages) still show everything correctly
- That the current section is highlighted in the top nav
- Try changing something in `navigation.yaml`, check the menu responds as you would expect (you'll need to reload the server to see changes).

### DIFF
--- a/navigation.yaml
+++ b/navigation.yaml
@@ -1,0 +1,374 @@
+cloud:
+  title: Cloud
+  path: /cloud
+
+  children:
+    - title: Overview
+      path: /cloud
+    - title: OpenStack
+      path: /cloud/openstack
+
+      children:
+        - title: Autopilot
+          path: /cloud/openstack/autopilot
+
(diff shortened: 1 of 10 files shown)


nottrobin:
Working demo here: http://www.ubuntu.com-navigation-menu-footer.demo.haus/server/maas

barrymcgee:
@nottrobin There's no design changes proposed, right? If so, you can remove 'Review: Design needed' and merge.

PR #1943, opened 2017-06-07
Generate breadcrumbs from navigation.yaml

### LINKED ISSUE #1944: Restructure navigation & breadcrumb logic in views
The way navigation and breadcrumbs are currently being generated in the templates is complex and confusing.

This should really be logic handled in the views, rather than in the templates. We should simplify the templates.

nottrobin:
Use the new navigation.yaml to generate breadcrumbs smartly, so you don't need to specify them on every page.

**NB:** Make sure #1942 is merged before reviewing this one..

Demo: http://www.ubuntu.com-navigation-breadcrumbs.demo.haus/

Fixes #1944.

QA
--

Read the HACKING.md section about navigation, check it makes sense.

`./run` and browse the site. Check breadcrumbs behave as expected. Especially try:

- contact-us and thank-you pages. These are "hidden" from other navigations, and shouldn't display their breadcrumb siblings.
- "child" pages like `/cloud/managed-cloud`
- "child" pages with children, like `/cloud/openstack`
- "grandchild" pages like `/download/cloud/conjure-up`

- Try changing something in `navigation.yaml`, check the breadcrumbs respond as you would expect (you'll need to reload the server to see changes).


barrymcgee:
Adding 'Don't merge' label until #1942 is merged.

nottrobin:
@barrymcgee Thanks, that's a good idea. A "don't review" label would be more accurate but we don't have one. Because this one really just represents all the work, rather than half of it. It's split for ease of reviewing.

barrymcgee:
No design changes so removing 'Design needed' label

nottrobin:
@barrymcgee I've only added body class to old base.html, not the v1 one. Yes it still breaks some old.layout pages to remove it.

PR #1984, opened 2017-06-14
remove pre-v1 content, rename -v1 conent

pmahnke:
## Done

* remove all pre Vanilla v1 pages
* remove ‘-v1’ from all directories and file names
* remove ‘-v1’ from all html and scss files

## QA

- look at the homepage and at least one page from every section
- try searching
- try filling out a form
- try making a contribution
- see the site looks ok

## Note

* we don’t seem to have any ie.css files or print.css is that correct?
* are there more pages to delete?


deadlight:
Sorry, @pmahnke, I think that we'll need to do these as smaller PRs.

PR #2014, opened 2017-06-23
Added canonical-global-nav as a dependency

anthonydillon:
## Done
Added canonical-global-nav as a dependency and removed all the old global navigation code.

## QA
- Pull branch and run `./run`
- Go to http://0.0.0.0:8001/
- See that the global navigation is there and works
- Check small screens

Demo: http://www.ubuntu.com-global-nav-component.demo.haus


pmahnke:
I had to

1. ./run clean
2. rm -rf static/components/canonical-global-nav
3. ./run

as I had the old version of the non-bower canonical-global-nav

on live, the global in in the footer inline with the other footer lists...

image


in the demo, it is at the bottom of the page

image


is that expected?

PR #2058, opened 2017-07-12
Feed check

deadlight:
## Done

 - Made the feed template tags return `False` to the template if loading the feed throws and error
 - Modified templates loading insights feeds 

## QA

  - see that the insights news items load
- View the site locally in your web browser at: [http://0.0.0.0:8001/server](http://0.0.0.0:8001/server)
  - see that the "further reading" news items appear in the contextual footer.
- Edit `webapp/templatetags/feeds.py` and add `raise Exception('Simulate error')` as line 17
  - see that the insights news items do not load and that there is an error message
- View the site locally in your web browser at: [http://0.0.0.0:8001/server](http://0.0.0.0:8001/server)
  - see that the "further reading" news items do not load and that there is an error message

### DIFF
--- a/templates/shared/_insights_news_strip.html
+++ b/templates/shared/_insights_news_strip.html
@@ -21,12 +21,16 @@
 
       {# insights news items #}
       <div class="row">
-        {% for item in insights_feed %}
-          <div class="col-3">
-            <h3 class="p-heading--four u-less-margin--bottom"><a href="{{ item.link }}" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'insights', 'eventAction' : '{{ gtm_event_label }} news link', 'eventLabel' : '{{ item.title|escapejs }}', 'eventValue' : '{{ item.link }}' });">{{ item.title }}</a></h3>
-            <p><time pubdate datetime="{{ item.updated }}">{{ item.updated_datetime|date:"j F Y" }}</time></p>
-          </div>
-        {% endfor %}
+        {% if insights_feed is False %}
+          Feed error.
+        {% else %}
+          {% for item in insights_feed %}
+            <div class="col-3">
+              <h3 class="p-heading--four u-less-margin--bottom"><a href="{{ item.link }}" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'insights', 'eventAction' : '{{ gtm_event_label }} news link', 'eventLabel' : '{{ item.title|escapejs }}', 'eventValue' : '{{ item.link }}' });">{{ item.title }}</a></h3>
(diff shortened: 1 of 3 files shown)


templates/shared/_insights_news_strip.html
```diff
@@ -9,40 +9,58 @@
...
+          <a href="https://insights.ubuntu.com/" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'insights', 'eventAction' : 'clicks insights feed link', 'eventLabel' : '{{ gtm_event_label }}', 'eventValue' : undefined });">
+            Latest news from Insights
           </a>
         </h2>
-        {% for item in spotlight_feed %}
-          <h3 class="p-heading--four u-less-margin--bottom"><a href="{{ item.link }}" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'insights', 'eventAction' : '{{ gtm_event_label }} spotlight article', 'eventLabel' : '{{ item.title|escapejs }}', 'eventValue' : '{{ item.link }}' });">{{ item.title }}</a></h3>
-          <p><time pubdate datetime="{{ item.updated }}">{{ item.updated_datetime|date:"j F Y" }}</time></p>
-        {% endfor %}
+
+        {# insights news items #}
```
nottrobin: Couldn't the `if insights_feed` conditional be here? So we don't need to duplicate the whole `<section>`?

templates/templates/_further_reading_links.html
```diff
@@ -5,7 +5,13 @@
 {% endif %}
 
 <ul class='p-list'>
-{% for item in feed %}
+{% if feed is False %}
```
nottrobin: It's not terribly important, but shouldn't this conditional be one line higher, so we don't end up with a `<div>` inside a useless `<ul>`? I tried it out in the inspector and it seems to display exactly the same.

templates/shared/_insights_news_strip.html
nottrobin: Here's a diff to illustrate my suggestion. Seems to work okay. https://pastebin.canonical.com/193214/

deadlight: I had considered this and the options tiggered by the flags `strip_classes` and `spotlight_feed` can alter the markup in unexpected ways. This way the error-state markup is kept encapsulated.

templates/templates/_further_reading_links.html
deadlight: Yes, that's an oversight, I'll switch them around.

templates/shared/_insights_news_strip.html
nottrobin: OK

Issue #1956, opened 2017-06-09
Please add a hover-state to the 'platform' strip on homepage


frankiegranato:
I would suggest a colour change to the icons on hover. Lets consult with Lil or Joana as to colour choice. https://www.ubuntu.com/

---
*Reported from: https://www.ubuntu.com/*

pmahnke:
@yaili or @joanasa89 ?

joanasa89:
Could be orange. @yaili do you have any other suggestion?

yaili:
Sounds good. Could we just do a demo first to make sure it's not too jarring?

Caleb-Ellis:
For the demo I've just duplicated the SVG icons and changed the colour to orange, so it's serving an entirely different image on hover (which is why there's a slight delay on the first hover). I haven't figured out a way to change the colour of SVGs in CSS but I imagine it would be a million times better.

anthonydillon:
@Caleb-Ellis you would have to put the SVG inline.

For example:
https://codepen.io/anthonydillon/pen/dVWEyJ

PR #2246, opened 2017-09-28
demo of orange hover state for homepage platform strip

Caleb-Ellis:
## Done

- Added orange hover state for homepage platform strip icons and text

## QA

- Check that the icons and text on the platform strip change colour to orange on hover

## Issue / Card

Issue #1956


Caleb-Ellis:
@anthonydillon I tried writing the SVG icons inline which made the hover effect very easy to implement, but found a lot of problems with the formatting and layout when I did. I think a lot of the SASS refers to the icons as background-images, so changing them to inline SVGs would require quite a bit more rewriting of the CSS rules. For now I've embedded the SVG icons as background-images in the CSS which addresses the jumpiness. Thoughts?

PR #2247, opened 2017-09-28
Remove duplicate links from vanilla upgrade

deadlight:
## Done

Removed the duplicate set of links/buttons under Kubernetes consulting packages

## QA

- View the site locally in your web browser at: [http://0.0.0.0:8001/kubernetes](http://0.0.0.0:8001/kubernetes)
- See that there is only one version of the link block containing "Download the datasheet"


nottrobin:
Okay fine, I've QAed it too.

PR #2299, opened 2017-10-16
Update vanilla v1.6.0

anthonydillon:
## Done
Updated to Vanilla-framework@1.6.0. Checked through the site and removed redundant code. Also added a few hotfixes for Vanilla which should all have issues as comments.

## QA

- Look through the site and make sure it looks ok


frankiegranato:
Logos are too small on /download/alternative-downloads

frankiegranato:
Can we fix the positioning of the Recognised security graphic on /desktop/government (probably not a vanilla bug but still its seems to be effected)

frankiegranato:
image
RE ‘template’ pages like /cloud/foundation-cloud and /server/maas: can we make the “Get started” heading the same size as other headings. (I realise this is probably not a vanilla bug, but I though tI’d mention it anyway)

frankiegranato:
Generally looks much better than live. I have added a few small issues though.

PR #2319, opened 2017-10-19
Changed laptop image back to one with dock on left

### LINKED ISSUE #2321: Homepage Showcases an Incorrect Portrayal of Ubuntu 17.10
## Summary

The Ubuntu 17.10 spot on the Ubuntu website incorrectly portrays what a user should expect from the OS as it is installed.

## Process

Navigate to https://www.ubuntu.com

The image displayed in the laptop graphic should be more representative of Ubuntu 17.10 as it is installed, not after it has been modified. This misrepresents what the user is to expect when installing this OS.

## Screenshot

image

Caleb-Ellis:
## Done

- Changed laptop image back to one with dock on left

## QA

- Check image of laptop on takeover has the dock on the left instead of the bottom

Fixes #2321

### DIFF
--- a/templates/takeovers/_1710_takeover.html
+++ b/templates/takeovers/_1710_takeover.html
@@ -6,13 +6,13 @@
         <p class="intro-text">Brand new Ubuntu desktop!<br>
         Same amazing experience!</p>
         <p class="u-hidden u-visible--small">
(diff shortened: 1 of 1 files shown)


iainlane:
Please do this, people keep asking us (desktop) what's going on with this image. We can argue again later on if needed.

anthonydillon:
Why does this have don't merge on it? Can we land and get live please?

PR #2459, opened 2017-11-23
Add topic filtering to the resources list

anthonydillon:
## Done
Add topic filtering to the resources list. Edit the display to be cards.

## QA
- Check the feed shows a list of resources
- Filter by the topic and check it works

## Issue / Card
Fixes https://github.com/ubuntudesign/web-squad/issues/263

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -36,39 +36,66 @@ class ResourcesView(TemplateView):
     INSIGHTS_URL = 'https://insights.ubuntu.com'
     API_URL = INSIGHTS_URL + '/wp-json/wp/v2'
     RESOURCE_FILTER = 'categories=1172,1509,1187'
+    GROUPS = {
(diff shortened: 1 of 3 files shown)


webapp/views.py
```diff
@@ -26,6 +30,59 @@ def get_context_data(self, **kwargs):
...
+                category_id=str(category_id),
+                resource_filter=self.RESOURCE_FILTER,
+            )
+
+        resources = get_json_feed_content(api_url)
+        return resources
+
+    def get_context_data(self, **kwargs):
+        """
+        Get the context form insights for the resources page
```
Lukewh: for/ form?

frankiegranato:
Looks lovely! I think we's better try separating the content by adding in topic headings with a rule in between each group. I'd suggest groups of three articles per topic in the first instance.

frankiegranato:
Please move 'Resources' to the penultimate position in the global nav bar, so that 'Downloads' stays at the end.

PR #2468, opened 2017-11-27
Add content type filtering to the resources list

anthonydillon:
## Done
Added a select element to filter the content type. Added the functionality to filter by type, topic and both. Moved the navigation item to the left of downloads.

## QA
- Filter some and check it works

## Issue / Card
Fixes https://github.com/ubuntudesign/web-squad/issues/264

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -41,6 +41,11 @@ class ResourcesView(TemplateView):
         'desktop': {'id': 1479, 'name': 'Desktop'},
         'internet-of-things': {'id': 1666, 'name': 'Internet of things'},
     }
+    CATEGORIES = {
+        'case-studies': {'id': 1172, 'name': 'case studies'},
+        'videos': {'id': 1509, 'name': 'videos'},
+        'webinars': {'id': 1187, 'name': 'webinars'},
+    }
 
     def _get_categories_by_slug(self, slugs=[]):
         if slugs:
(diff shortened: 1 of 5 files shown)


anthonydillon:
@frankiegranato has seen this branch and :ok: it.

webapp/views.py
```diff
@@ -69,20 +74,55 @@ def _normalise_resources(self, resources):
...
-            )
-            feed_items[topic] = {}
-            feed_items[topic]['items'] = get_json_feed_content(api_url)
-            feed_items[topic]['group_name'] = self.GROUPS[topic]['name']
+        if topic or content:
+            if not content:
+                category_id = str(self.GROUPS[topic]['id'])
+                api_url = (
+                    '{api_url}/posts?_embed'
+                    '&group={category_id}&{resource_filter}'
```
grahambancroft: Do the ampersands in the urls need to be escaped?

grahambancroft:
Couple of issues on small, arrow on right hides covers links and the content dropdown is difficult to get to
 
screen shot 2017-11-28 at 10 29 47

grahambancroft:
Code 👍

webapp/views.py
anthonydillon: No, they are a string that represent the URI to the data endpoint.

anthonydillon:
Thanks @grahambancroft updated the filtering section to work on all viewports.

grahambancroft:
I'm still seeing the same issue with the chevron on the right covering the links

PR #2478, opened 2017-11-29
Defer or asynchronously load blocking resources

### LINKED ISSUE #656: Try to defer or asynchronously load blocking resources
We have a number of javascript assets blocking the page render. Lets try and remove where possible and if required try to defer or async to load.

Here is a list of assets:
- https://assets.ubuntu.com/v1/b0601acd-modernizr.2.7.1.js
- https://assets.ubuntu.com/v1/2c8fd8c9-Imager.js
- https://assets.ubuntu.com/v1/5d7e5bbf-jquery-2.2.0.min.js
- http://www.google-analytics.com/…api.js?experiment=IbIACoOrTnG8eSnRY64nWQ
- https://assets.ubuntu.com/v1/c954db4a-jfeed_prm.js

grahambancroft:
## Done

- Remove Modernizr
- Add HTML5 shiv inline in head
- Add no-js replacement inline in head
- Place the minified contents of polyfills.js at the bottom of the page, before any real JavaScript is loaded
- Move all other JavaScript to the bottom of the page, below the now inline polyfills
- In core.js, use hasFeature instead of Modernizr.svg (Oper mini supports svg as of Mar 2015)

## QA

- Check everything works as expected

## Issue / Card

Fixes #656


static/js/core.js
```diff
@@ -68,5 +53,5 @@ core.mobileNav = function () {
     }
 };
 
-core.svgFallback();
+core.supportsSvg();
```
anthonydillon: Why is this necessary it will just return a boolean?

templates/templates/base.html
```diff
@@ -43,7 +43,12 @@
     <!-- <![endif]-->
 
     <!-- javascript -->
-    <script src="{{ ASSET_SERVER_URL }}b0601acd-modernizr.2.7.1.js"></script>
+    {% include "shared/_html5_shiv.html" %}
+    <!-- replace no-js -->
+    <script>
+      var html = document.documentElement
+      html.className = html.className.replace(/\bno-js\b/, 'js')
```
anthonydillon: You could use classList.replace here https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

templates/shared/_polyfill.html
```diff
@@ -0,0 +1,7 @@
+    <script>
```
anthonydillon: Can you add a comment here to state it is the js polyfills

anthonydillon:
Google PageSpeed is still complaining that jquery should be async: https://developers.google.com/speed/pagespeed/insights/?url=http%3A%2F%2Fwww.ubuntu.com-pr-2478.run.demo.haus%2F&tab=desktop

templates/templates/base.html
grahambancroft: Done

templates/shared/_polyfill.html
grahambancroft: Done

static/js/core.js
grahambancroft: Only in modern browsers, it works okay in older browsers - https://stackoverflow.com/questions/43317691/why-document-implementation-hasfeature-always-return-true

grahambancroft:
@anthonydillon Added async to jQuery instances

static/js/core.js
anthonydillon: I understand it will only return on modern browsers but it will just return `true` which does nothing else? Is this function used by anything?

anthonydillon:
Thanks, think we need to remove `async` from jQuery as we are failing on forms. Please remove and we should be good to land.

PR #2511, opened 2017-12-06
 Add search functionality

anthonydillon:
## Done
Add global search functionality to the resources page.

## QA

- Search in the search box
- Check that the resources are search results for that search query

## Issue / Card
https://github.com/ubuntudesign/web-squad/issues/288

### DIFF
--- a/templates/resources/index.html
+++ b/templates/resources/index.html
@@ -12,7 +12,7 @@
     <form action="/resources" method="get" class="p-form p-form--inline p-form--search">
       <div class="p-form__group">
         <label for="search-input" class="u-off-screen">Search resources</label>
-        <input id="search-input" type="search" name="search" value="{{ search_slug }}" class="p-form__control"/>
+        <input id="search-input" type="search" name="q" value="{{ search_slug }}" class="p-form__control"/>
       </div>
       <button type="submit" alt="search" class="p-button--positive u-align--left">Search resources</button>
     </form>
--- a/templates/resources/index.html
+++ b/templates/resources/index.html
@@ -9,6 +9,13 @@
 <section class="p-strip--accent is-shallow">
   <div class="row">
     <h1>Resources</h1>
+    <form action="/resources" method="get" class="p-form p-form--inline p-form--search">
+      <div class="p-form__group">
+        <label for="search-input" class="u-off-screen">Search resources</label>
+        <input id="search-input" type="search" name="search" value="{{ search_slug }}" class="p-form__control"/>
+      </div>
+      <button type="submit" alt="search" class="p-button--positive u-align--left">Search resources</button>
+    </form>
   </div>
 </section>
 
@@ -39,7 +46,7 @@
 <section class="p-strip is-shallow">
   <header class="row">
     <div class="col-12 u-align--right">
-      <form action="?foo=bar" method="get" class="p-form p-form--inline">
+      <form action="/" method="get" class="p-form p-form--inline">
         <div class="p-form__group">
           <label for="full-name-inline" aria-label="Filter the resources by type" class="p-form__label">Filter:</label>
           <select class="js-submit-on-change p-form__control" name="content" aria-label="Filter by content type">
--- a/templates/resources/index.html
+++ b/templates/resources/index.html
@@ -6,16 +6,20 @@
 
 {% block content %}
 
-<section class="p-strip--accent is-shallow">
+<section class="p-strip--accent is-dark is-shallow">
   <div class="row">
-    <h1>Resources</h1>
-    <form action="/resources" method="get" class="p-form p-form--inline p-form--search">
-      <div class="p-form__group">
-        <label for="search-input" class="u-off-screen">Search resources</label>
-        <input id="search-input" type="search" name="q" value="{{ search_slug }}" class="p-form__control"/>
-      </div>
-      <button type="submit" alt="search" class="p-button--positive u-align--left">Search resources</button>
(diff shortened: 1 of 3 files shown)


webapp/views.py
```diff
@@ -151,6 +172,9 @@ def get_context_data(self, **kwargs):
         context['items'] = self._get_resources()
         context['topic_slug'] = self.request.GET.get('topic')
         context['content_slug'] = self.request.GET.get('content')
+        context['search_slug'] = ''
+        if (self.request.GET.get('search')):
```
Lukewh: Bind this to a variable so you don't call the `GET.get('search')` twice (if that's a thing in Python)?

webapp/views.py
```diff
@@ -75,7 +75,28 @@ def _normalise_resources(self, resources):
     def _get_resources(self):
         topic = self.request.GET.get('topic')
         content = self.request.GET.get('content')
+        search = self.request.GET.get('search')
```
WillMoggridge: I think this should be `?q=` and not `?search=`. It is the defacto standard for search across the web.

Lukewh: Good spot @WillMoggridge

anthonydillon: Done

frankiegranato:
Functionality is fine but I think it looks a bit visually clunky. Can we try the following: 

- [ ] Can you try ranging the Search right (aligned with search in primary nav)
- [ ] Please ensure the field is the same hight as the button
- [ ] Decrease the gap in-between the field and the button. See my jpg attached 
- [ ] Please could we try the search button as neutral (ie. transparent with white outline) - Just to see how it looks
- [ ] Also, can you try to align all the elements to the far left (ie. with the ubuntu home in the main nav). See my jpg attached. 
Sorry if I'm being too fussy and designer-y.

anthonydillon:
> Decrease the gap in-between the field and the button. See my jpg attached

This is set in Vanilla I could "hack" it but I would rather not.

anthonydillon:
>  Please could we try the search button as neutral (ie. transparent with white outline) - Just to see how it looks

@frankiegranato there is no such button as transparent with white outline. Neutral is a white background with grey border and turns grey on hover: https://github.com/ubuntudesign/vanilla-design/blob/master/Buttons/buttons.md#neutral

webapp/views.py
```diff
@@ -113,7 +113,28 @@ def _get_resources(self):
...
+        if search:
+            api_url = (
+                    '{api_url}/posts?_embed'
+                    '&search={search}&{resource_filter}'
+                )
+            api_url = api_url.format(
+                    api_url=self.API_URL,
+                    search=search,
+                    resource_filter=self.RESOURCE_FILTER,
+                )
```
tbille: all this block has an indentation problem

tbille:
```python
api_url = (
                    '{api_url}/posts?_embed'
                    '&search={search}&{resource_filter}'
                ).format(
                    api_url=self.API_URL,
                    search=search,
                    resource_filter=self.RESOURCE_FILTER,
                )
```
instead of instanciating 2 times the variable

anthonydillon: Done.

webapp/views.py
```diff
@@ -113,7 +113,28 @@ def _get_resources(self):
...
+        if search:
+            api_url = (
+                '{api_url}/posts?_embed'
+                '&search={search}&{resource_filter}'
+            )
+            api_url = api_url.format(
+                api_url=self.API_URL,
+                search=search,
+                resource_filter=self.RESOURCE_FILTER,
+            )
```
tbille: I added the comment before as a reply but:
```
api_url = (...).format(...)
```
to avoid instanciate `api_url` 2 times

anthonydillon:
@tbille thanks, ready for another review

anthonydillon:
@WillMoggridge did you want to look?

webapp/views.py
anthonydillon: Done, good spot.

templates/resources/index.html
```diff
@@ -39,7 +50,7 @@ <h1>Resources</h1>
 <section class="p-strip is-shallow">
   <header class="row">
     <div class="col-12 u-align--right">
-      <form action="?foo=bar" method="get" class="p-form p-form--inline">
+      <form action="/" method="get" class="p-form p-form--inline">
```
WillMoggridge: This redirect to the root path. It is probably best being empty.

anthonydillon: Done.

frankiegranato:
screen shot 2017-12-14 at 09 59 43
@spencerbygraves this page is still looking a bit off to me. Can you suggest any small amends that might lift it or tidy it up a bit?

spencerbygraves:
I think this should have had some design input, but I understand we don't have a dedicated web team designer at the moment.

The Snappy team recently created this [discover page](https://snapcraft.io/discover/) with search in the top area. Can we try using the same layout here please.

anthonydillon:
@spencerbygraves thanks that is the design I was working from before by having the search below the title.

@frankiegranato think the snappy design is ok to go with our should we change it?

frankiegranato:
This does need more design attention but for now it will do. We'll look at it in more detail next iteration. I have added my +1 and have removed the design review label as we have no design resource to attend to this at the moment :-(

spencerbygraves:
I understand if you want to come back to it later, the design is already in place in the form of the Snappy page.

PR #2522, opened 2017-12-08
Add whitepapers to resources

anthonydillon:
## Done
Add whitepapers to resources

## QA

- Filter by whitepapers
- Check it all works

## Issue / Card
Fixes https://github.com/canonical-websites/www.ubuntu.com/issues/2488

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -52,6 +52,7 @@ class ResourcesView(TemplateView):
         'case-studies': {'id': 1172, 'name': 'case studies'},
         'videos': {'id': 1509, 'name': 'videos'},
         'webinars': {'id': 1187, 'name': 'webinars'},
+        'whitepapers': {'id': 1485, 'name': 'whitepapers'},
     }
 
     def _get_categories_by_slug(self, slugs=[]):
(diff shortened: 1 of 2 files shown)


Lukewh:
We should really show some form of message when there are no results:
http://www.ubuntu.com-pr-2522.run.demo.haus/resources?topic=internet-of-things&content=whitepapers

webapp/views.py
```diff
@@ -198,13 +199,14 @@ def _get_resources(self):
...
                 posts = get_json_feed_content(api_url, limit=3)
-                posts_length = len(posts)
-                if 'posts' not in feed_items.keys():
-                    feed_items['posts'] = {}
-
-                feed_items['posts'][group_name] = {}
-                feed_items['posts'][group_name]['posts'] = posts
-                feed_items['posts'][group_name]['group_name'] = group['name']
+                if posts:
+                    posts_length = len(posts)
```
Lukewh: This doesn't seem to be used so can be removed.

anthonydillon: Its passed to the `_generate_pagination_queries` function below.

anthonydillon:
@Lukewh Cheers, fixed the no results notification.

Lukewh:
@anthonydillon has there been a regression in vanilla? Notifications seems to be non-full-width? I noticed this when doing some dashboard work too:

screen shot 2017-12-13 at 11 45 33

webapp/views.py
Lukewh: Line 100 checks the length - what happens if posts_length is not defined? You should maybe define it at the top of the function? Not sure how python handles things like that.

anthonydillon: It creates the var if not defined. I will set it at the top though.

anthonydillon:
@Lukewh The notification issue is a Vanilla issue. Not sure what changed.

PR #2541, opened 2017-12-19
Disable 17.10 download

anthonydillon:
## Done
Replaced the 17.10 download button with a notice.

_Landing without review for speed, this should be reverted shortly_

### DIFF
--- a/templates/download/desktop/index.html
+++ b/templates/download/desktop/index.html
@@ -55,7 +55,7 @@
         </div>
         <div class="col-4">
           <p class="p-card__content">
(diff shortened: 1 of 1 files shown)


feren:
I'm not a member of the team, or anything, but I think it'd be a good idea to link to the bug report itself too, for those curious.

edwinksl:
From the bug report at https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1734147, it looks like the fix has been released. Any idea when the download link will be enabled again?

PR #2587, opened 2018-01-15
2580: Build resource page enhancements

### LINKED ISSUE #2580: Build resource page enhancements

Caleb-Ellis:
## Done

- Updated resources page, which includes:
  - New header strip
  - New styling for the articles cards
  - Removal of search box (for now)
  - Removal of tab label in lieu of an 'All Topics' tab

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/resources
- Check that the styling is in line with the [new design](https://github.com/canonical-websites/www.ubuntu.com/issues/2565)
- Check that filtering by Topic works as expected
- Check that filtering by Content Type works as expected
- Check that everything looks ok for different screen sizes and browsers (as per [document](https://github.com/canonical-webteam/practices/blob/master/coding/browser-support.md))

## Issue / Card

Fixes #2580 

## Screenshots
New design:

image

### DIFF
--- a/static/sass/_pattern_card.scss
+++ b/static/sass/_pattern_card.scss
@@ -45,8 +45,8 @@
   }
 
   %resources-card-header {
-    color: $color-x-light;
-    padding: $sp-medium;
+    border-radius: 2px;
+    padding: .875rem $sp-medium;
   }
 
   .p-card--resources {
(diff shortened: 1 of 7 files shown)


Caleb-Ellis:
## TODO
- Make coloured top border have corner radius
- Dotted line width on topic header to match content type
- Make image heights uniform
- Look into removing breadcrumb navigation

frankiegranato:
Looks lovely. 
@Caleb-Ellis there is one small enhancement I'm thinking of:
Is there a way to have the type un-pluralised? Eg. instead of Case studies -> Case Study?  
No probs if that is tricky.

Caleb-Ellis:
@frankiegranato New demo has the cards with singular category now if you want to check it out
@spencerbygraves I've fixed a couple of the design issues you had. The breadcrumb nav at the top I think should be a separate issue, just because the way it's implemented is sort of automated from the urls and miiiight be a little work. As for the image heights, I just want to check how you'd like it to be implemented.

## Same image height
This would make all image heights the same as the smallest in the row. Left and right margins are affected.
same height

## Same image container height
This would keep the images themselves the same, but the container the image sits in would have a set height for all cards in the row (I'm thinking percentage based). Bottom margin is affected.
same container height

## Same image container height, justified center
The same as above, but the images now have a common baseline. Top and bottom margins are affected.
same container height centered

spencerbygraves:
@Caleb-Ellis That's great, thanks for the examples. I think the second version is safest, same image container height.

frankiegranato:
Love it. The singular "type" makes more sense. Thank you, for your carfeul investigations Caleb.

templates/resources/index.html
```diff
@@ -6,78 +6,80 @@
...
+      <input type="submit" value="Send Request" class="u-hide">
+      {% if topic_slug %}
+      <input type="hidden" name="topic" value="{{ topic_slug }}">
+      {% endif %}
     </div>
-  </header>
+  </form>
+</div>
 
+<section class="p-strip is-shallow u-no-padding--top" style="margin-bottom: 2rem;">
```
anthonydillon: I think we should remove this margin-bottom

static/sass/_pattern_card.scss
```diff
@@ -43,4 +43,66 @@
...
+
+  .p-card__header {
+    &--cloud-and-server {
+      @extend %resources-card-header;
+      border-top: 3px solid #2b939c;
+    }
+
+    &--desktop {
+      @extend %resources-card-header;
+      border-top: 3px solid #66a2c1;
```
anthonydillon: Desktop and Cloud and server colours are very similar.

anthonydillon:
If you go to http://www.ubuntu.com-pr-2587.run.demo.haus/resources?topic=desktop&content=videos the resources are all desktop but one shows cloud and server so maybe its worth just grabbing the slug from the URL so its not confusing.

anthonydillon:
Once complete I will browser test.

static/sass/_pattern_card.scss
Caleb-Ellis: The secondary colour palette has been signed off already. I believe Spencer and Frankie will look into it later this/next week, but for now I think we can leave it. I do agree though that with just the top border colour they look very similar.

anthonydillon:
Browser testing:
:+1: Chrome Latest
:+1: Firefox Latest
:+1: Edge Latest
:+1: IE Latest
:-1: Safari 5.1 Latest Win 10
bs_win10_safari_5 1

Caleb-Ellis:
@anthonydillon shouldn't we really only be testing Safari on MacOS? Most of www.ubuntu.com is broken on Safari 5.1 for Windows.

anthonydillon:
@Caleb-Ellis fair enough, Safari 11 is :+1:

PR #2592, opened 2018-01-18
Fixed 500 error page

pmahnke:
## Done

- Fixed 500 error page by removing nav and footer and use of {{ asset server url }}

## QA

- View the site locally in your web browser at: [500 page](http://0.0.0.0:8001/500)
- see that the page looks good

## Issue / Card

Fixes # 2591

## Screenshots
image


spencerbygraves:
@pmahnke does it need the padding-top on large screens?

templates/500.html
```diff
@@ -1,4 +1,4 @@
-{% extends "templates/one-column.html" %}
+{% extends "templates/one-column_no_nav.html" %}
```
anthonydillon: Why did you remove the navigation on the 500 page? This is get the user stuck there. Just because the current page is 500'ing it doesn't mean other pages wont work as expected.

pmahnke: @anthonydillon oddly, it wasn't executing, and showing empty navigation - and it didn't parse the {{asset_server}} variable - see https://github.com/canonical-websites/www.ubuntu.com/issues/2591

anthonydillon: Makes sense, dynamic (YAML) navigation will not work in a server error situation. If we hard code it, it will get out of date. So the Ubuntu logo is ok for now.

pmahnke:
@spencerbygraves should be ok for a re-review - http://www.ubuntu.com-pr-2592.run.demo.haus/500

pmahnke:
@spencerbygraves you are right, I have updated it

anthonydillon:
:+1:

Issue #2583, opened 2018-01-12
In the alternate downloads page there's no link to minimal iso


mymike00:
In the Alternative download page I expect to see a section regarding minimal iso, that can be found [here](https://help.ubuntu.com/community/Installation/MinimalCD).

---
*Reported from: https://www.ubuntu.com/download/alternative-downloads*

pmahnke:
@mymike00 they are actually the 'Network installer' images... I will try and work the wording 'minimal iso' into the text.

mymike00:
Oh, I thought they were different things... Thanks

pmahnke:
@mymike00 they are the same... it installed a really basic ubuntu and asks you about loads of options and languages, etc...

cbrownstein:
Resolved in #2608

PR #2605, opened 2018-01-29
Describe the network installer better

cbrownstein:
I explain that the network installer includes only a minimal set of packages needed to install Ubuntu, and that the rest of the packages are downloaded over the network.

Ref: #2583


pmahnke:
@cbrownstein this is excellent.  I did a minor additional tweak, but couldn't push to your branch for some reason, so I created a new PR #2606 ... thanks so much!

PR #2711, opened 2018-02-14
Upgrade to get-feeds v0.2.2 - more accurate timing

nottrobin:
The latest version of get-feeds uses native requests library features to
measure the request times it reports to prometheus. This is better.

QA
--

``` bash
./run exec --expose-port 8001 --expose-port 9990 --env DJANGO_DEBUG=false ./manage.py runserver --noreload 0.0.0.0:8001
```

Browse around the site a bit at http://0.0.0.0:8001

Now go to http://0.0.0.0:9990 and check the `feed_request_latency_seconds_bucket` and `feed_request_latency_seconds_bucket` measurements look reasonable. Hopefully most requests are taking considerably less than a second.


WillMoggridge:
I see link checker has failed but I don't think it should block this PR.

PR #2752, opened 2018-02-26
2730: Update /download/cloud page

### LINKED ISSUE #2730: Build /download/cloud page
## Summary

we are moving the existing pages to /openstack

here is the new page's [copy doc](https://docs.google.com/document/d/1kp7QQXst1EbCjZjzWZP3YZwfmG0EIXLfQoRUVaMEZng/edit#)

Caleb-Ellis:
## Done

- Built new /download/cloud page according to [copydoc](https://docs.google.com/document/d/1kp7QQXst1EbCjZjzWZP3YZwfmG0EIXLfQoRUVaMEZng/edit#)

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/download/cloud
- Check that the page matches the [copydoc](https://docs.google.com/document/d/1kp7QQXst1EbCjZjzWZP3YZwfmG0EIXLfQoRUVaMEZng/edit#)

## Issue / Card

Fixes #2730

### DIFF
--- a/navigation.yaml
+++ b/navigation.yaml
@@ -205,13 +205,6 @@ download:
       path: /download
     - title: Cloud
       path: /download/cloud
-
-      children:
-        - title: Try OpenStack
-          path: /download/cloud/try-openstack
-        - title: Build OpenStack
-          path: /download/cloud/build-openstack
-
     - title: Core
       path: /download/core
 
--- a/static/sass/_pattern_card.scss
+++ b/static/sass/_pattern_card.scss
@@ -106,4 +106,11 @@
       border-top: 3px solid #ff8936;
     }
   }
+
+  .p-card__image-container {
+    align-items: center;
+    display: flex;
+    justify-content: center;
+    min-height: 10rem;
+  }
 }
--- a/templates/download/cloud/index.html
+++ b/templates/download/cloud/index.html
@@ -1,88 +1,133 @@
 {% extends "download/_base_download.html" %}
 
-{% block title %}Build OpenStack with conjure-up | Download{% endblock %}
-
+{% block title %}Use Ubuntu in public clouds | Download{% endblock %}
+{% block meta_description %}How to use Ubuntu in the public cloud and where to find our customised cloud images for development.{% endblock %}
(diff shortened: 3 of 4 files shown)


frankiegranato:
I love the use of the cards, but some of the logos look huge compared to others. Can we get them to look more consistent? Nb @kwm14

frankiegranato:
@Caleb-Ellis see a few other notes in the copydoc.

spencerbygraves:
If we start to use these cards in more places, I think the link text at the bottom should be 16px. 14px was fine when it was only a topic label in resources. I also think we could remove the margin top.

Can we also make the amazon logo smaller, I set it to max-height: 4rem;

Thanks

Here's an example of the changes:
image

frankiegranato:
@Caleb-Ellis I'm sorry - I've messed up the contextual footer. Can you please replace the content in the middle column with the content I've specified in the copydoc?

pmahnke:
@spencerbygraves @Caleb-Ellis @kwm14  @frankiegranato I worry that we are using these cards too much

I think the plain Vanilla ones look better frankly for this

image
image
(I added some min-height to the bottom one)

pmahnke:
@Caleb-Ellis also, it still has the build and try links in the nav, unless you are moving these in another PR

spencerbygraves:
@pmahnke thanks for pointing this out, you're right that we shouldn't overuse them. As we discussed just now, let's keep this card style for Resources and Insights. @frankiegranato is that ok with you?

Caleb-Ellis:
@pmahnke @spencerbygraves I've updated it now so that the top three cards are highlighted and the bottom three are just normal cards (all Vanilla). I've also removed the openstack stuff from the nav (and I'll put them in the right place with the openstack PRs)

pmahnke:
@Caleb-Ellis last link should be to /support and the title needs a 'the'

Ubuntu Advantage for **the** public cloud
Flexible Ubuntu Virtual Guest options designed for virtualised workloads on Ubuntu-certified public clouds.
[Get support >](/support)

PR #2761, opened 2018-02-28
Swap uappexplorer link for snap store.

felicianotech:
## Done

- replaced UAppExplorer link on the Snappy page to the Snap Store now that it exists

## QA

- ran locally, link works and appears as expected


pmahnke:
@felicianotech thanks, great catch!

PR #2781, opened 2018-03-06
Change Insights RSS feed URL

WillMoggridge:
## Done

Update Insights RSS URL to admin.insights.ubuntu.com

## QA

View the homepage and make sure the Insights feed loads


WillMoggridge:
We have decided to solve this with a different solution. I will close this.

PR #2802, opened 2018-03-12
Add snap installation instructions on try-openstack

### LINKED ISSUE #2718: snap installation instructions for Ubuntu Server (not Desktop) - UX
## Summary

One of our customers was trying to use conjure-up on an Ubuntu **Server** system which does not come with snapd debian package installed by default.

https://github.com/canonical-websites/www.ubuntu.com/blob/e4ee6ca0fe298e1b6e3d3e69d620fbaba4ca84d0/templates/download/cloud/try-openstack.html

This can be seen by manifests associated with desktop and server images:
http://releases.ubuntu.com/16.04/
http://releases.ubuntu.com/16.04/ubuntu-16.04.3-desktop-amd64.manifest
snapd	2.25

no snapd here
http://cdimage.ubuntu.com/ubuntu-server/xenial/daily/current/xenial-server-amd64.manifest

After snapd package installation the current shell does not get updated with a new PATH entry for /snap/bin (this may be somewhat difficult to achieve via an external process due to the fact that environment variables are passed in exec system calls when a process is created, including bash). A re-login is required to be able to use installed snaps, e.g. a user will get 

`conjure-up: command not found`

even if snap list shows conjure-up as installed.

snapd package installs /etc/profile.d/apps-bin-path.sh which gets added to a shell environment, so

`$ source /etc/profile.d/apps-bin-path.sh`

Is enough to solve the problem without re-login or reboot.

The lack of those instructions may be frustrating for users unfamiliar with those details.

It would be great to update  https://github.com/canonical-websites/www.ubuntu.com/blob/e4ee6ca0fe298e1b6e3d3e69d620fbaba4ca84d0/templates/download/cloud/try-openstack.html with relevant warnings.

pmahnke:
## Done

- Add snapd installation instructions on /download/cloud/try-openstack
- Also adding to the beta

## QA

- View the site locally in your web browser at: [/download/cloud/try-openstack](http://0.0.0.0:8001/download/cloud/try-openstack)
- compare it to the [copy doc](https://docs.google.com/document/d/1QIXOAlCDV2cyQqU5uEW2kpITobf7OkFHr-tsEvjd_1Q/edit)

## Issue / Card

Fixes #2718

### DIFF
--- a/templates/download/cloud/try-openstack.html
+++ b/templates/download/cloud/try-openstack.html
@@ -17,7 +17,7 @@
           <div class="col-6 p-divider__block">
             <h3>Requirements</h3>
             <ul class="p-list">
-              <li class="p-list__item is-ticked">Laptop, desktop or virtual machine with 16GB RAM, Ubuntu 16.04 LTS and <a href="/containers/lxd" title="More information about LXD">LXD</a> installed</li>
+              <li class="p-list__item is-ticked">Laptop, desktop or virtual machine with 16GB RAM, Ubuntu 16.04 LTS, snapd and <a href="/containers/lxd" title="More information about LXD">LXD</a> installed</li>
               <li class="p-list__item is-ticked">An hour of your time</li>
             </ul>
           </div>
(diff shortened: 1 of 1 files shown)


pmahnke:
@frankiegranato can you look at the 'note' on http://www.ubuntu.com-pr-2802.run.demo.haus/download/cloud/try-openstack for a +1?

frankiegranato:
image
I've read it, and looked back at dshcherb's recommendation and it seems to make sense. 
Looks great. 
+1

PR #2832, opened 2018-03-15
Simplify _insights_news_strip include

nottrobin:
The `_insights_news_strip` include is only used twice, in `/` and in `/cloud`.

It used to take a number of different options, including the whole URL
for requesting the spotlight feed and the normal feeds, and a limit.

However, the only thing that actuall varies is the topic_id - apart from
that we always request the feed from insights, and we always request 3
articles + a spotlight, or 4 articles.

So I've made it more opinionated, so you can call it with no arguments,
or if you want to filter it, just pass a topic_id - which is all the
/cloud page needs.

I've also moved the `<section>` outside the include, so we don't need to
pass through the `strip_classes`.

This means that:

``` html
{% include "shared/_insights_news_strip.html" with
  insights_spotlight_url="https://admin.insights.ubuntu.com/wp-json/wp/v2/posts?topic=1477&sticky=true&per_page=1"
  insights_news_url="https://admin.insights.ubuntu.com/wp-json/wp/v2/posts?topic=1477&sticky=false&per_page=5"
  strip_classes="p-strip--light is-deep is-bordered"
  gtm_event_label="ubuntu.com cloud overview" %}
```

Can become simply:

``` html
<section class="p-strip is-deep is-bordered">
  {% include "shared/_insights_news_strip.html" with gtm_event_label="ubuntu.com homepage" %}
</section>
```

QA
--

`./run` and go to http://127.0.0.1:8001/ and http://127.0.0.1:8001/cloud/, and check the insights feed sections show the correct articles.

### DIFF
--- a/templates/cloud/index.html
+++ b/templates/cloud/index.html
@@ -380,7 +380,9 @@
   </div>
 </section>
 
-{% include "shared/_insights_news_strip.html" with insights_spotlight_url="https://admin.insights.ubuntu.com/wp-json/wp/v2/posts?topic=1477&sticky=true&per_page=1" insights_news_url="https://admin.insights.ubuntu.com/wp-json/wp/v2/posts?topic=1477&sticky=false&per_page=5" strip_classes="p-strip--light is-deep is-bordered" gtm_event_label="ubuntu.com cloud overview" %}
+<section class="p-strip--light is-deep is-bordered">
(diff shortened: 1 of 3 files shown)


anthonydillon:
Good work, much better. Code :ok: QA :+1:

Issue #2739, opened 2018-02-23
Beta: Make primary nav sticky 


frankiegranato:

frankiegranato:
As discussed with the gang: Lets remove "sticky" nav when drown down is expanded but lets make sticky nav appear on scroll-up. Please keep nav sticky in all other instances (ie. when the drown down is not in use). Sorry - about all the complicated logic that I'm sure this will involve (I do think its worth doing though - I think its the only way we can really assess if these menu interactions are usable or not).

spencerbygraves:
@Caleb-Ellis @frankiegranato from the review...

The sticky header causes the open navigation to not scroll on screens shorter than the menu length. Can we try having the navigation appear only when you scroll back up the page please. And if possible to keep some white space to make sense of the selected state.

Thanks

frankiegranato:
@Caleb-Ellis  lets catch up on this on Thursday. I think we said we should leave this until we've tested the beta.

Caleb-Ellis:
After discussing this with @frankiegranato  and @spencerbygraves, we all agree that its best to leave the interaction as it is for now. There is no point doing an involved and time-consuming investigation into how to make a sticky nav work with expanded drop downs, before we've tested the BETA. Also, we've heard rumours that @pmahnke has received some direction from Mark about exploring the use of a darker primary nav (which might have beneficial implications for the drown downs) so we'll come back to this discussion once that exploration was been done. The original PR and demo have been left open as a point of reference for future discussion on the topic.

PR #2843, opened 2018-03-20
Sticky beta navigation demo

Caleb-Ellis:
## Done

- Demo of first pass at making the beta nav sticky. Looking for Design + UX feedback before I tidy the code and bugs up.
- On scroll down, nav disappears - on scroll up, nav comes into view and sticks to the top of the page as per discussion on #2739. @kwm14 using Vanilla easing of course :)
- Main page content disappears when nav dropdown is open. I think this is fine because the nav is so huge and text-heavy that you'd barely see the page content anyway
- Ignore the code. It's a buggy mess but will be fixed once Design/UX is closer to finalised


spencerbygraves:
@Caleb-Ellis thanks, looking good.

When I first scroll down, the nav is sticky for a little while until I scroll a bit further. Is that deliberate? It seems a bit odd to me and could be removed I think.

Ideally i think the page content should remain, so a user could scroll beyond the open menu and click off to close the menu. I thin we should ask @frankiegranato what she thinks on this.

Thanks

Caleb-Ellis:
@spencerbygraves The initial stickiness was deliberate. I thought it might be a nice UX thing to let the user know that the nav _is_ actually sticky, and it disappears when you scroll down. But I've removed it now and agree it feels better without hahaha

frankiegranato:
After discussing this with @Caleb-Ellis and @spencerbygraves, we all agree that its best to leave the interaction as it is for now. There is no point doing an involved and time-consuming investigation into how to make a sticky nav work with expanded drop downs, before we've tested the BETA. Also, we've heard rumours that @pmahnke has received some direction from Mark about exploring the use of a darker primary nav (which might have beneficial implications for the drown downs) so we'll come back to this discussion once that exploration was been done. 
For now, we've advised @Caleb-Ellis to remove the bottom margin from dropdown window and leave it at that for now.

frankiegranato:
It looks good with removed bottom margin from dropdown window. 
Lets close this issue for now but please keep this demo running.

PR #3250, opened 2018-05-31
Move contribute to thank you

anthonydillon:
## Done
Move the contribution section to the download thank you page and removed the logical with the pre-download step.

## QA
- Follow the download flow to desktop
- See that you don't get the contribute page before the thank you page
- See that the thank you page contains the contribute section
- Try out the contribution
- Return to /download/desktop/thank-you (which is the return URL) and see that the messaging is different and a download does not start.


redirects.yaml
```diff
@@ -268,6 +268,7 @@ developers/?: "/desktop/developers"
 developer-terms-and-conditions/?: "/legal/terms-and-policies/developer-terms-and-conditions"
 digital-signage/?:  "/internet-of-things/digital-signage"
 donations/?: "/download/desktop/contribute/"
+/download/desktop/contribute/?: "/download/desktop/thank-you/"
```
pmahnke: remove the starting '/'

anthonydillon:
@pmahnke Done

PR #3317, opened 2018-06-13
3273 nav anchors

### LINKED ISSUE #3273: beta updates: add anchor functionality to primary nav
- when the menu is expanded and you leave the page and then return via the back-button in the browser, the menu is still expanded on the page
- perhaps use anchor links

steverydz:
## Done

Remember nav state when leaving page

This PR is branched from https://github.com/canonical-websites/www.ubuntu.com/pull/3315 and therefore should be merged after that PR

## QA

- Open one of the navigation tabs, then click a link to another page
- Hit the browsers back button and see that the previously selected menu is open
- Close the menu by clicking the active tab or the overlay beneath it
- Go to another page
- Hit the back button and see that the menu is still closed

## Issue / Card

Fixes #3273

### DIFF
--- a/static/js/navigation.js
+++ b/static/js/navigation.js
@@ -22,6 +22,10 @@ navDropdowns.forEach(function(dropdown) {
         } else {
           dropdown.classList.add('is-selected');
           dropdownContent.classList.remove('fade-animation', 'u-hide');
+
+          if (window.history.pushState) {
+            window.history.pushState(null, null, '#' + dropdown.id);
+          }
         }
       } else {
         dropdown.classList.remove('is-selected');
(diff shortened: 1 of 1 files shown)


steverydz:
This PR should be merged after https://github.com/canonical-websites/www.ubuntu.com/pull/3315

anasereijo:
@steverydz looks good to me 👍 
@frankiegranato do we want to do the same on Products?

steverydz:
@Caleb-Ellis I think that will be horrendously complicated - we'd have to find a way of storing the state of the menu with JS in that case.

Essentially this is why I don't think this persistent menu functionality is a good thing at all - it's not what dropdowns typically do and it creates a whole load of other UX issues.

frankiegranato:
@Caleb-Ellis & @steverydz after a discussion with Peter and Spencer and Caleb, we have agreed to go ahead with this version (even though, I can see that the browser-back/forth functionality is a bit messed up). Lets add this to the 'fix-it-later' list :-)  +1 from me.

PR #3496, opened 2018-07-02
Merge beta

steverydz:
## Done

- Branch from master
- Merge beta-1.7.0 into branch
- Resolve conflicts

## QA

- All code has been QA'd so it's mostly a case of making sure nothing looks broken


templates/cloud/index.html
```diff
@@ -0,0 +1,510 @@
+{% extends "cloud/base_cloud.html" %}
```
nottrobin: this page doesn't exist any more right? redirects to `/openstack`. So I'm assuming this file shouldn't exist?

templates/containers/lxd.html
```diff
@@ -77,9 +77,6 @@ <h2 class="p-muted-heading u-align--center">Supported guest operating systems</h
       <li class="p-inline-images__item">
```
nottrobin: Does this page exist in the new IA? Shouldn't this maybe be deleted too? If not, just ignore me.

templates/download/cloud/build-openstack.html
```diff
@@ -0,0 +1,226 @@
+{% extends "download/_base_download.html" %}
```
nottrobin: Another page being created here - is this the same thing? Should this remain deleted 'cos it's not in the new IA?

steverydz:
@nottrobin Thanks - it's expected to fail at the moment as there were some links to pages we haven't made yet. We'll get that cleaned up as part of the QA.

PR #3575, opened 2018-07-04
Add hover state to breadcrumb title

### LINKED ISSUE #3532: Status indicator is needed for "Index tab" in the primary nav bar
We should implement some status indicators on the index "tab" to teach users that the tab is part of the navigation.

My initial suggesions as to how we do this are as follows: 
Hover state: Underline
Selected state: Colour is dark grey 

I'm open to other suggestions though. Especially @anasereijo might have a view.
primary nav - index indicator

---
*Reported from: https://beta.ubuntu.com/openstack*

steverydz:
## Done

Add hover state to breadcrumb title

@frankiegranato If by selected state you mean when you are on that specific page, if we highlight in that case then it will be highlight when you're on one of the sub pages too, e.g.

If on /openstack it is highlighted
If on /openstack/features it is highlight as you are still within the openstack section

Here is what it would look like:
screenshot from 2018-07-04 11-04-17

## QA

- View the site locally in your web browser at: [http://0.0.0.0:8001/openstack](http://0.0.0.0:8001/openstack)
- See that the breadcrumb title (orange box) has a hover state

## Issue / Card

Fixes #3532

### DIFF
--- a/static/sass/_pattern_navigation.scss
+++ b/static/sass/_pattern_navigation.scss
@@ -198,7 +198,8 @@
         font-weight: 300;
 
         &:hover {
-          text-decoration: none;
+          text-decoration: underline;
+          color: $color-x-light;
         }
       }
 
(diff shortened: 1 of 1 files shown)


steverydz:
@Caleb-Ellis Thanks, fixed

PR #3634, opened 2018-07-06
Update Google Optimize container ID

caldav:
## Done

- Update to the new Optimize container ID
- Update page-hiding snippet to match Optimize proposed defaults (also matches what was used on the blog.ubuntu.com homepage experiment) 

## QA

If you have access to the right Google Optimize account, you should be able to verify experiments without errors.

**Update:** Successfully created an experiment on the demo website. :+1:


caldav:
Please ping me when this gets deployed, so I can check again on prod.

caldav:
Just confirmed that there was no delay when the experiment was turned off.

PR #3787, opened 2018-07-24
3782: Make beta primary nav sticky

### LINKED ISSUE #3782: BETA: Make primary nav sticky
- not including the global nav
- the primary dropdown slides behind the sticky nav on a scroll

Caleb-Ellis:
## Done

- Separated the original `navigation.js` file into `navigation.js`, `global-navigation.js` and `ga-events.js`
- Refactored `navigation.js` slightly into discrete functions
- Made the beta primary nav stick to the top of the viewport on large screens
- Added slight fade transitions to the dropdowns and overlay

## QA

- Check that the primary nav sticks to the top of the screen when you scroll down
- Check that clicking a dropdown will open the content at the right place
- Check that the dropdown closes automatically if you scroll up or down far enough
- Open the global nav and check that it closes if you scroll down too far

## Issue / Card

Fixes #3782

### DIFF
--- a/static/js/ga-events.js
+++ b/static/js/ga-events.js
@@ -1,0 +1,61 @@
+var origin = window.location.pathname;
+
+function addGANavEvents(target, category){
+  var t = document.querySelector(target);
+  if (t) {
+    t.querySelectorAll('a').forEach(function(a) {
+      var destination = a.href.replace('https://www.ubuntu.com', '').split("?")[0];
+      a.addEventListener('click', function() {
+        dataLayer.push({
+          'event' : 'GAEvent',
+          'eventCategory' : category,
+          'eventAction' : "from:" + origin + " to:" + destination,
+          'eventLabel' : a.text,
+          'eventValue' : undefined
+        });
+      });
+    });
+  }
+}
+
+function addGAContentEvents(target){
+  var t = document.querySelector(target);
+  if (t) {
+    t.querySelectorAll('a').forEach(function(a) {
+      var destination = a.href.replace('https://www.ubuntu.com', '').split("?")[0];
+      if (a.className.includes('p-button--positive')) {
+        var category = 'www.ubuntu.com-content-cta-0';
+      } else if (a.className.includes('p-button')) {
+        var category = 'www.ubuntu.com-content-cta-1';
+      } else {
+        var category = 'www.ubuntu.com-content-link';
+      }
+      if (!destination.startsWith("#")){
+        a.addEventListener('click', function() {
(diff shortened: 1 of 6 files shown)


nottrobin:
Shouldn't the global nav also close if you scroll too far?

Caleb-Ellis:
Before I do any more work on this, I'm going to list some things I think should be figured out.

First of all, the content in the Developer and Enterprise dropdowns is huge - larger than the vertical space on most viewports. Therefore, the dropdown content can't be sticky *and* bigger than the viewport height, because then you wouldn't be able to scroll through the content that extends beyond the bottom of the viewport.

One solution would be to limit the height of the dropdown and have a scrollbar. It would be very easy to implement, but inline scrollbars are generally considered bad UX.

We could also make the viewport jump to the top of the screen when you open the navigation, but I imagine this is also bad UX.

Another solution would be to hide the contents of the page when you open the navigation - kind of like you're opening a new page (except not). It would still be sticky normally, and you would return to the original position on the page when you close it. This is slightly more difficult to implement, but not too bad. The problem with this is that the navigation would then no longer act like a dropdown, because you wouldn't be able to interact with the dropdown content and page content simultaneously.

One more solution could be: the menu is sticky when you scroll up, and not sticky when you scroll down. This is unbelievably problematic, because we would need to set up listeners that would constantly be checking if the user is scrolling up or down, which would definitely hit page performance. It would mean that the dropdown content just kind of floats in the middle of the page when you scroll down. Therefore, we would constantly be alternating between calculating the location of the dropdown content relative to the viewport (i.e. the top of your browser window) and relative to the page content (i.e. the top of the page). Anyway, I can't imagine making this work in a way that would perform well, be accessible and be cross-browser/OS compatible.

Basically, we need to figure out how we want the sticky navigation to work that *isn't* alternating stickiness/non-stickiness, because I'd rather not put in a bunch of work for something that we end up not wanting.

nottrobin:
I quite like the way you've currently done it, where the navigation is sticky but the drop-down is static, and then disappears when you scroll away.

Although, personally I'm generally not a big fan of these sticks navigations that disappear on scrolling down and then reappear when scrolling up, because on my mobile, I find they're always massively getting in my way when I want to go back up the page to find something I just scrolled past. However, that may not be as annoying on large screen, so if it's only on large screen that might be fine.

frankiegranato:
I think we've reached the conclusion that this sticky nav is too complicated to implement and to use. Carla and I will discuss on thurs and decide how to proceed.

PR #3794, opened 2018-07-25
Netlify & Markdown legal pages

nottrobin:
WIP work from hack day


nottrobin:
Doesn't look like this is going to progress any time soon, so I'll close it. Unless @pmahnke or @frankiegranato would like to outline some way in which it could be brought into reality?

PR #3856, opened 2018-08-02
Increase feed timeout to 2 seconds

### LINKED ISSUE #3801: Resources do not appear on the resource page
screen shot 2018-07-26 at 09 38 36

## Summary

Resources do not appear on the resource page

## Process

visit https://www.ubuntu.com/resources, on a random basis the feed from blog does not appear

Expecting to see the list of resources, but this does not seem to always show

## Screenshot

See screenshot

nottrobin:
Feeds from blog.ubuntu.com are usually failing when the cache is empty, because the 1.2 second timeout is not long enough.

It's risky to increase the timeouts because there's a possible cascading effect that could bring down the whole website - see https://portal.admin.canonical.com/C113355/. But as long as the total response time remains below 5 seconds we should be okay.

Given that the largest number of feeds on any page on the site is two, it should be safe to make the timeout 2 seconds, so the app should always respond within slightly over 4 seconds.

And 2 seconds seems to be sufficient in local testing.

Fixes #3801.

QA
--

``` bash
$ rm cache.sqlite  # Make sure the cache is clear
$ ./run
```

Now browse to e.g.: http://127.0.0.1:8001/resources?topic=cloud-and-server, and see that resources load properly.


webapp/views.py
```diff
@@ -69,17 +69,6 @@ class ResourcesView(TemplateView):
         1485: {'slug': 'whitepapers', 'name': 'Whitepaper'},
     }
 
-    def _get_categories_by_slug(self, slugs=[]):
```
Lukewh: How come this has been deleted?

nottrobin: It wasn't used anywhere.

PR #3863, opened 2018-08-07
3838: Keep primary nav open on mobile screens

### LINKED ISSUE #3838: On mobile the header tabs should always be displayed
## Summary

On mobiles we currently treat the top-level tabs as a menu (Enterprise, Developer, Community, Download etc). These should actually be treated as a header - displayed by default, together with the search. Ideally, it would be a semi-sticky header that emerges when one starts scrolling back up the page, but it would be sufficient just to display it by default on page load.

## Process

Go to beta.ubuntu.com on a phone browser. You'll see the mobile rendering of the site. The top level header tabs are not displayed, but there is a 'Menu'  which reveals that header.

The 'Menu' text should disappear. Instead, the search and top-level tabs should be displayed always on page load. On scrolling down, these can disappear off the top of the page, when scrolling up it would be nice if they emerged almost immediately (rather than waiting to scroll all the way up).

Caleb-Ellis:
## Done

- Made the primary nav open by default on mobile screens, and removed the Menu button

## QA

- Resize your browser to mobile size
- Check that the primary nav is open by default and the "Menu" toggle is gone

## Issue / Card

Fixes #3838

### DIFF
--- a/static/js/navigation.js
+++ b/static/js/navigation.js
@@ -1,7 +1,6 @@
 var navDropdowns = document.querySelectorAll('.p-navigation__dropdown-link');
 var dropdownWindow = document.querySelector('.dropdown-window');
 var dropdownWindowOverlay = document.querySelector('.dropdown-window-overlay');
-var closeMenuLink = document.querySelector('.p-navigation__toggle--close');
 var navigationThresholdBreakpoint = 900;
 
 navDropdowns.forEach(function(dropdown) {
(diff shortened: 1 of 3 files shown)


frankiegranato:
Looks great. The primary nav is open by default and the "Menu" toggle is gone.

pmahnke:
when I open the developer tab, I am getting a navigation.js error

image

might not be related, but it stops me from clicking on the menu

Caleb-Ellis:
@pmahnke good catch - it was to do with auto-opening the tab when going back in the browser. Anyway, should be fixed now :)

PR #3914, opened 2018-08-17
Add appstore whitepaper landing page

caldav:
## Done

* Add appstore whitepaper landing page
* [Copydoc](https://docs.google.com/document/d/1PM7HoXgECZQSPDsMOM9h-4YsEvHhiilUV5iZovqFuxY)

## QA

- View the site locally in your web browser at: [http://0.0.0.0:8001/engage/iot-business-model](http://0.0.0.0:8001/engage/iot-business-model)

## Notes

* Form ID + success page changes needed before merging

### DIFF
--- a/templates/engage/iot-business-model.html
+++ b/templates/engage/iot-business-model.html
@@ -167,40 +167,32 @@
               <input required  id="Phone" name="Phone" maxlength="255" class="mktoField mktoTelField  mktoRequired"  type="tel" aria-label="Phone Number">
             </li>
             <li  class="mktField">
-              <label for="utm_campaign" class="u-hide">utm_campaign</label>
-              <input name="utm_campaign" class="mktoField  mktoFormCol"  value="{{utm_campaign}}" type="hidden">
+              <input name="utm_campaign" aria-label="utm_campaign" class="mktoField  mktoFormCol"  value="{{utm_campaign}}" type="hidden">
             </li>
             <li  class="mktField">
-              <label for="utm_medium" class="u-hide">utm_medium</label>
-              <input name="utm_medium" class="mktoField  mktoFormCol"  value="{{utm_medium}}" type="hidden">
+              <input name="utm_medium" aria-label="utm_medium" class="mktoField  mktoFormCol"  value="{{utm_medium}}" type="hidden">
             </li>
           <li  class="mktField">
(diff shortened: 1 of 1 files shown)


kwm14:
@caldav looking good, just one small update/bug on the field input:

- [ ] border missing on bottom of fieldset

screen shot 2018-08-17 at 11 17 40

barrymcgee:
@caldav  Could we resolve the accessibility issues that are introduced by this changeset, please?

barrymcgee:
I get a 404 both locally and at the demo link?

screen shot 2018-08-23 at 18 00 17

anthonydillon:
@barrymcgee these are hidden inputs so we don't want/need labels for them.

barrymcgee:
@anthonydillon If we can't fix the error, we fix the linter - ignoring warnings and errors will inevitably lead to genuine concerns being ignored as they will increasingly be seen as just noise.

nottrobin:
I have 2 QA points, but I'm happy for someone to tell me these are out-of-scope for this PR:

- If you enter an invalid email you see this, which hardly stands out:
  image
  Don't we have a better error style?
- We've turned off HTML5 validation with `novalidate="novalidate"`, but then we don't implement our own validation for the "phone number" `type="tel"` field. Shouldn't we at least be going for feature parity with HTML5 validation if we're going to disable it?

barrymcgee:
Reading this article: https://developer.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/

...I believe the correct way to add a hidden input is to include the attribute `hidden` like so;

`<input name="cr" aria-label="cr" class="mktoField " value="" type="hidden" hidden>`

..as this will remove the element from the accessibility tree.

anthonydillon:
@barrymcgee `type="hidden"` is used through the website. To make this change we could do it as a separate branch.

nottrobin:
Interesting, I'd not heard of the `hidden` attribute before: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden

nottrobin:
Thanks @deadlight, I've filed https://github.com/canonical-websites/www.ubuntu.com/issues/3941 and https://github.com/canonical-websites/www.ubuntu.com/issues/3942

PR #3936, opened 2018-08-22
Only use development Talisker settings in local development

nottrobin:
Production is using development settings, which is bad.

QA
--

Open up `webapp/views.py` and at line 62 (just below `def search(request):`) add:

``` bash
    raise Exception("We're just testing")
```

Then run the site in the normal way.

``` bash
./run
```

See pretty coloured Talisker output in the terminal. See that http://127.0.0.1:8001 works. Now go to http://127.0.0.1:8001/search and see the `NotImplementedError` displayed with a full debugging stack trace.

Now run it like production:

``` bash
docker build --tag ubuntu --build-arg COMMIT_ID=`git rev-parse --short HEAD` .
docker run --rm -p 8901:80 ubuntu
```

Now go to http://127.0.0.1:8901 and see that it works. Go to http://127.0.0.1:8901/search and see that you see the production "500 Server error" page, rather than the development stack trace.

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,3 @@
 PORT=8001
+DJANGO_DEBUG=true
+DEVEL=true
--- a/entrypoint
+++ b/entrypoint
@@ -2,5 +2,9 @@
 
 set -e
 
-talisker.gunicorn.gevent webapp.wsgi --reload --log-level debug --timeout 9999 --access-logfile - --workers 3 --worker-class gevent --error-logfile - --bind $1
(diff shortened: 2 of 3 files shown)


webapp/views.py
```diff
@@ -59,6 +59,8 @@ def search(request):
     Get search results from Google Custom Search
     """
 
+    raise NotImplementedError()
```
tbille: oups push

nottrobin: Fixed, I think

PR #3944, opened 2018-08-26
Corrects conjure-up link

### LINKED ISSUE #3527: /support "Conjure up" link goes to openstack index, where Conjure up is not mentioned once
https://beta.ubuntu.com/support

On /support, under the "Systems management tool: Landscape" heading, there is a "Conjure up" link that links to /openstack, but the openstack page does not mention conjure up once.

We should either add conjure up to the openstack page, or link to conjure-up.io directly.

matthewpaulthomas:
## Done

In this “Landscape” section, changes the “conjure-up” link from the Openstack page (which doesn’t mention conjure-up) to the conjure-up site. Fixes #3527.

I would have linked to [ubuntu.com/openstack/install](https://www.ubuntu.com/openstack/install), which mentions both conjure-up and Openstack … but unfortunately that page doesn’t mention Landscape, which is more important in this context.

Demo: http://www.ubuntu.com-pr-3944.run.demo.haus/support


matthewpaulthomas:
The `linkchecker` error is a 503, not a 4xx, and it is not in the file that I changed in this PR.

templates/shared/_systems_management_tool.html
```diff
@@ -21,7 +21,7 @@ <h2>Systems management tool: Landscape</h2>
                   <li>
                       Fully automated deployment of our reference architecture
                       OpenStack cloud with
-                      <a href="/openstack/">conjure-up</a>
+                      <a class="external" href="https://conjure-up.io/">conjure-up</a>
```
anthonydillon: Thanks, the class name should be `p-link--external`

nottrobin:
I've updated this - see:

https://github.com/canonical-websites/www.ubuntu.com/pull/3944/commits/c568ebece368367a6f2221a4ed8ce21e33aa41b4

WillMoggridge:
@nottrobin This looks like it doesn't change the link, which the original description mentions.

PR #3946, opened 2018-08-28
Rotate takeovers

caldav:
## Done

* Rotate takeovers on refresh / new session
* Refresh CTA text of AI wbn takeover and landing page
* Drop redundant event from Rigado CTA

## QA

- On the homepage, ensure takeovers are rotating following this loop (you may enter in the middle of the loop): 
```
{% include "takeovers/_tackling_iot.html" %}
{% include "takeovers/_ai_webinar.html" %}
{% include "takeovers/_rigado_webinar.html" %}
```
- Close the tab, open a new one, ensure you are seeing the next takeover in the rotation

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -8,6 +8,25 @@
 {% block takeover_content %}
 
 {% include "takeovers/_tackling_iot.html" %}
+{% include "takeovers/_ai_webinar.html" %}
+{% include "takeovers/_rigado_webinar.html" %}
+
+<script>
+  if (window.localStorage && window.sessionStorage) {
+    var takeovers = document.getElementsByClassName('takeover');
+    if (!localStorage.getItem('TAKEOVER')) {
+      var i = 0;
+      localStorage.setItem('TAKEOVER', i);
+    } else {
+      var i = parseInt(localStorage.getItem('TAKEOVER')) + 1;
+      if (i >= takeovers.length) {
+        i = 0;
+      }
+    }
+    localStorage.setItem('TAKEOVER', i);
+    takeovers[i].classList.remove('u-hide');
+  }
+</script>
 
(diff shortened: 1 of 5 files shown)


templates/index.html
```diff
@@ -8,6 +8,25 @@
...
+    } else {
+      var i = parseInt(localStorage.getItem('TAKEOVER')) + 1;
+      if (i >= takeovers.length) {
+        i = 0;
+      }
+    }
+    localStorage.setItem('TAKEOVER', i);
+    takeovers[i].classList.remove('u-hide');
+  }
+</script>
```
anthonydillon: Beware this will result in no takeover without JavaScript.

templates/takeovers/_ai_webinar.html
```diff
@@ -1,10 +1,10 @@
-<section class="p-strip is-deep p-takeover--ai-webinar u-hide js-primary-takeover">
+<section class="p-strip is-deep p-takeover--ai-webinar u-hide takeover">
```
anthonydillon: Could you please prefix the`takeover` class with js to indicate it adds functionally and not styling.

templates/index.html
anthonydillon: Also browsers that do not support local storage.

caldav:
@ant, some changes:

- Class name change as requested
- First takeover visible is randomised, but the order of the loop is followed afterwards. The rationale is for all takeovers to have the same exposure (as opposed to the initial logic of the PR), but without giving a feeling of randomisation or mess in what is displayed. The same user refreshing the page will see a loop.
- A noJS fallback mechanism. It's working as expected, but it's not the most elegant solution, so if you have any suggestions to improve, happy to change things

frankiegranato:
Nice work @caldav

PR #3948, opened 2018-08-28
Removes non-link items from Developer navigation menu

matthewpaulthomas:
## Done

Removes, from the “Developer” navigation, these items that do not link anywhere and don’t seem to have anywhere on ubuntu.com that they could link to.

“Develop on Ubuntu” items:
  - GCC, CLANG
  - Go
  - Python, Ruby, Node.js
  - Deb, snap, and charm packages
  - Android development
  - Architectures and ISAs
  - Kernel selection & lifecycle

“Publish apps” items:
  - Cross-distro testing of snaps
  - Rigorous snap security model
  - Integration between snaps

“Containers” item:
  - Test CentOS apps on Ubuntu with LXD

(Probably we _should_ have pages for all of these things, but showing them as unclickable items is a poor way of achieving this. The vast majority of people who see the items won’t be in a position to publish the pages.)

## QA

- Open the “Developer” navigation and check that every remaining item is clickable

### DIFF
--- a/templates/templates/_navigation-developer-h.html
+++ b/templates/templates/_navigation-developer-h.html
@@ -13,13 +13,6 @@
             <a href="/ai">Develop on Ubuntu&nbsp;&rsaquo;</a>
           </h4>
           <ul class="p-text-list--small is-bordered">
-            <li class="p-list__item">GCC, CLANG</li>
-            <li class="p-list__item">Go</li>
-            <li class="p-list__item">Python, Ruby, Node.js</li>
-            <li class="p-list__item">Deb, snap, and charm packages</li>
-            <li class="p-list__item">Android development</li>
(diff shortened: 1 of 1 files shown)


matthewpaulthomas:
A drawback is that the “Develop on Ubuntu” section is left looking somewhat ridiculous, consisting of:
- a “Develop on Ubuntu” heading linking to `/desktop/developers`
- a “Multipass” link
- a “Develop with Ubuntu” button, which doesn’t do anything except take you to the same page that the heading does.

One solution to this, which I think would still be an improvement even once any of the missing pages are written:
- de-link the heading, as in the “Devops” section;
- convert the button to a link, as in the “Devops” section.

frankiegranato:
@matthewpaulthomas This was Mark's directive.

PR #3965, opened 2018-08-29
Set up D3 and create basic JS helper functions for desktop stats page

Caleb-Ellis:
## Done

- Renamed desktop-statistics.js to stickyNav.js (and left as-is)
- Added a dummy data file for testing before hooking up to Grafana API
- Added a class that makes the charts always appear below the title+text on mobile

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/desktop/statistics
- Check that the sticky nav still works
- Check that the cards with percentage values in them have been calculated from the dummy data (you can test by editing the values in dummyData.js)
- Check that on mobile the charts always appear below the title+text


anthonydillon:
@Caleb-Ellis looking good but the charts are not responsive.

screenshot_20180829-200442

PR #4058, opened 2018-09-13
Added a form validation field to the contact-us forms

### LINKED ISSUE #4059: spam on contact-us forms
We've been heavily spammed recently on this contact form, I checked and recaptcha is in place but still... Maryan mentioned a lot of spam coming through livechat too. Do you know if there's any way to block these or anything else we can do?

pmahnke:
## Done

- Because of spam, we have decided to add a form validation field to the contact-us forms to attempt to filter out spammers in marketo

## QA

- View the site locally in your web browser at: [http://0.0.0.0:8001/internet-of-things/contact-us?product=iot-overview](http://0.0.0.0:8001/internet-of-things/contact-us?product=iot-overview)
- view the source and see there is a hidden field called 'formValidation' with no value

Fixes #4059

### DIFF
--- a/templates/shared/_client-contact-us-form.html
+++ b/templates/shared/_client-contact-us-form.html
@@ -68,18 +68,18 @@
               <button type="submit" class="mktoButton p-button--positive" onclick="dataLayer.push({'event' : 'GAEvent', 'eventCategory' : 'Form', 'eventAction' : 'iot contact-us', 'eventLabel' : '{{product}}', 'eventValue' : undefined });">Submit</button>
             </li>
           </ul>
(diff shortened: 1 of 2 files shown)


caldav:
I'd like to suggest using `aria-hidden`, to avoid spamming screen readers with hidden fields.

pmahnke:
@caldav it doesn't seem to like it without aria-label

caldav:
`aria-hidden` is used to hide an element from screen readers when there is `aria-label` on it. We can either avoid `aria-label` for things that will always be hidden, or use `aria-hidden="true"` in addition to `aria-label`, when we need to set `aria-label` on a hidden element.

PR #4091, opened 2018-09-27
Takeover Language Detection

### LINKED ISSUE #4081: Detect browser language for takeovers

inuwa:
## Done
Takeovers now change based on the users browser default language. Takeovers are now only display for a particular region based on the language detected on the browser.
## QA

- Changed the default browser language to German (Deustche) or German
- Refresh the multiple times to German Takeover

## Issue / Card
Fixes #4081

### DIFF
--- a/static/js/navigation.js
+++ b/static/js/navigation.js
@@ -163,7 +163,7 @@ function closeMainMenu() {
   }
 }
 
-var origin = window.location.href;
+var origin = window.location.pathname;
 
 addGANavEvents('.global-nav__row', 'www.ubuntu.com-nav-0');
 addGANavEvents('#canonical-products', 'www.ubuntu.com-nav-0-products');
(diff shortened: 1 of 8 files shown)


inuwa:
Hi @caldav, yes that is the intended behaviour. There is only one takeover for german language for now. If another German Language Takeover is added then there will be a switch after the refresh. Hope that helps.

anthonydillon:
@caldav yes that is the MVP, we will add the lang stacking in a follow up.

caldav:
@inuwa got it.

The intended design is to also have "en" takeovers displayed to german speakers (to have more or less the same kind of rotation as the english version, for the sake of not locking this audience on a single campaign and exposing diverse messages). 

But that's not a blocker, your proposal is an elegant solution and a step up from what we had. If you have the time to look into adding the concept of international takeovers that would be part of the rotation regardless of the lang, it would be great. In any case, +1 from me.

PR #4173, opened 2018-10-11
Redirect for kubernetes docs

nottrobin:
For @evilnick

QA
--

`./run` and go to http://0.0.0.0:8001/kubernetes/docs - see you're redirected to the beautiful new Kubernetes docs at http://blog.juju.solutions/kubernetes-docs/.

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -683,3 +683,5 @@ usn/?: "https://usn.ubuntu.com/"
 (?P<page>.*)\.html: "/{page}"
 telco/?: "/telecommunications"
 telcos/?: "/telecommunications"
+kubernetes/docs/?: "http://blog.juju.solutions/kubernetes-docs/"
+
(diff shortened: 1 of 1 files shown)


redirects.yaml
```diff
@@ -683,3 +683,5 @@ usn/?: "https://usn.ubuntu.com/"
 (?P<page>.*)\.html: "/{page}"
 telco/?: "/telecommunications"
 telcos/?: "/telecommunications"
+kubernetes/docs/?: "http://blog.juju.solutions/kubernetes-docs/"
```
anthonydillon: Shouldn't we link to https?

evilnick: good point, I think that would be preferable

nottrobin: Oh I didn't realise that works

nottrobin: Fixed

PR #4371, opened 2018-11-09
Import takeovers from a <template> element

### LINKED ISSUE #4143: Avoid requesting images for hidden takeovers
Currently, visitors to the www.ubuntu.com homepage are downloading the images for all the takeovers, even though the majority of them are hidden:

image

We should make sure images from takeovers that won't be displayed aren't downloaded.

nottrobin:
Fixes #4143

This keeps all takeover markup inside a `<template>` element,
so as to avoid loading images, scripts and styles from unused takeovers.

This should improve performance for the client.

QA
--

`./run`, load the homepage. See that the "base" takeover gets replaced by the nicer CUTTLEFISH takeover.

Play around a bit, add some more takeovers, check they get chosen. Maybe try changing your language to German.

If you're feeling extra awesome, you might try testing in IE11.

### DIFF
--- a/templates/base_index.html
+++ b/templates/base_index.html
@@ -7,8 +7,11 @@
 
 {% block outer_content %}
   {% block content %}
-    {% block takeover_content %}
-    {% endblock takeover_content %}
+    <template id="takeovers" class="u-hide">
+      {% block takeover_content %}
+      {% endblock takeover_content %}
+    </template>
+
     <section class="p-strip--accent is-shallow is-dark">
       <div class="row">
         <div class="col-12">
--- a/templates/base_index.html
+++ b/templates/base_index.html
@@ -149,44 +149,63 @@
         }
       };
     </script>
+
     <script>
       if (window.localStorage && window.sessionStorage) {
-        var takeovers = document.getElementsByClassName('js-takeover');
-        for (t = 0; t < takeovers.length; ++t) {
-            // hide all takeovers
-            takeovers[t].classList.add('u-hide');
-        }
+        /**
+         * Choose a takeover
+         * ===
+         * 
+         * From the list of provided takeovers in the #takeovers template,
+         * choose one (that matches the client's language), and replace the
+         * base template with it.
+         */
+
+        var baseTakeover = document.querySelector('section#takeover');  // The base takeover element
+        var takeovers = document.querySelector('template#takeovers');  // The list of current takeovers to choose from
+
+        // For browsers that support <template> (all but IE), grab the template's ".content"
+        takeovers = 'content' in takeovers ? takeovers.content: takeovers;
+
         // get the users language and remove any extra detail suffix (e.g. -gb)
         var browserLanguage = detectLanguage();
         var languageClass = browserLanguage || 'en';
         languageClass = languageClass.split('-')[0];
 
-        // set the default takeover or if use the first one
-        var defaultTakeovers = document.querySelectorAll(
-          '.js-takeover[data-default="true"]'
-        );
-        if (!defaultTakeovers) {
-          defaultTakeovers = takeovers[0];
-        }
-
         // find any takeovers that match the users language
         var queryLang = ".js-takeover[data-lang*='"+languageClass+"']";
         var queryAll = ".js-takeover[data-lang*='all']";
-        var takeoversFiltered = (document.querySelectorAll([queryLang,queryAll].join(',')));
+        var takeoversFiltered = (takeovers.querySelectorAll([queryLang,queryAll].join(',')));
 
-        takeoversFiltered = (takeoversFiltered.length) ? takeoversFiltered : defaultTakeovers;
+        // If we have any takeovers
+        if (takeoversFiltered) {
+          var selectedIndex = null;
 
-        if (!localStorage.getItem('TAKEOVER')) {
-          var i = Math.floor(Math.random() * takeoversFiltered.length);
-          localStorage.setItem('TAKEOVER', i);
-        } else {
-          var i = parseInt(localStorage.getItem('TAKEOVER')) + 1;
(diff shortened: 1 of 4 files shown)


nottrobin:
I was wondering if this would effect SEO - because we probably want Google to index our homepage for all things mentioned in the takeovers, as long as Google doesn't judge us for trying to trick it with "hidden content".

In fact, it seems like using the `<template>` element is probably the best of both worlds - for the time being, Google [does index content](https://stackoverflow.com/a/47288775/613540) inside a `<template>` element, so our homepage could show up in search results for anything mentioned in any of our takeovers.

And, if in the future, Google's algorithms became smart enough notice that this content was "hidden", they shouldn't penalise us, because we've already signified that the content isn't a part of the main page by putting it in a `<template>` element.

templates/base_index.html
```diff
@@ -135,44 +149,63 @@ <h3><a href="/desktop">Fast, secure and simple, Ubuntu powers millions of PCs wo
...
+        /**
+         * Choose a takeover
+         * ===
+         * 
+         * From the list of provided takeovers in the #takeovers template,
+         * choose one (that matches the client's language), and replace the
+         * base template with it.
+         */
+
+        var baseTakeover = document.querySelector('section#takeover');  // The base takeover element
```
anthonydillon: Why not use `getElementById`?

templates/takeovers/_german_takeover_k8.html
```diff
@@ -1,4 +1,4 @@
-<div data-lang="{{ locale }}" lang="de" class="p-strip--image is-dark js-takeover {% if not default %}u-hide{% endif %}" style="background-color:#326DE6">
+<div data-lang="{{ locale }}" class="p-strip--image is-dark js-takeover" style="background-color:#326DE6">
```
anthonydillon: Why have you removed the `lang` here?

templates/takeovers/_german_takeover_k8.html
```diff
@@ -1,4 +1,4 @@
-<div data-lang="{{ locale }}" lang="de" class="p-strip--image is-dark js-takeover {% if not default %}u-hide{% endif %}" style="background-color:#326DE6">
```
inuwa: @nottrobin thank you for taking the time out. I noticed you removed the lang attribute from the takeover. I believe it is used to generate to filter for the takeovers. The idea of adding it was to be able to add English Language takeovers to German sites for example.

templates/takeovers/_german_takeover_openstack_made_easy.html
```diff
@@ -1,4 +1,4 @@
-<div data-lang="{{ locale }}" lang="de" class="p-strip--image js-takeover p-takeover--german-openstack-made-easy {% if not default %}u-hide{% endif %}">
```
inuwa: @nottrobin same issue here. Cheers🙂

templates/base_index.html
nottrobin: Only because `querySelector` allows me to be more expressive to include the `section` element name. Although there's no need to be so opinionated.

I'm happy to change it to `getElementById` if you care at all.

templates/takeovers/_german_takeover_k8.html
nottrobin: It's not actually used by JavaScript, only the `data-lang` is. But maybe we need the `lang` attribute too, I'll continue that discussion on ant's comment: https://github.com/canonical-websites/www.ubuntu.com/pull/4371#discussion_r232564652

templates/takeovers/_german_takeover_k8.html
nottrobin: I did this because I couldn't tell what purpose it was serving. I don't like the duplication between `data-lang` and `lang`, it seems redundant.

However, thinking about it, I think maybe it isn't redundant. The `lang` attribute semantically marks the content as having a primary language of e.g. German, which is worthwhile. It would be nice if we could use the `lang` attribute for everything, but the main thing preventing this, as far as I can see, is that `lang` doesn't support multiple languages, it is only intended to demark the *primary* language.

So I'll reinstate the `lang` attribute, but I'd like to propose that instead of `data-lang` we have `data-lang-extra`. The implication here is that we use the value of the `lang` attribute in JavaScript to determine the default language, and then we can *optionally* add extra languages.

(For takeovers that are intended for *all* languages, we should indicate this by simply removing the `lang` attribute).

nottrobin: Actually, inspecting the existing takeovers, none of them have used multiple languages. So actually the `data-lang-extra` attribute could remain theoretical at this point, to be implemented in the future if we need it. And for now, we just use the existence or non-existence of the `lang` attribute.

nottrobin: I've now implemented this. Please re-review.

templates/base_index.html
nottrobin: I've changed this.

nottrobin:
@anthonydillon, @inuwa I've addressed all the points - please re-review when you get a chance.

templates/takeovers/_1810-webinar-takeover.html
```diff
@@ -1,4 +1,4 @@
-<section lang="en" data-lang="{{ locale }}" class="p-strip--image is-dark is-deep u-image-position js-takeover p-takeover--1810--webinar {% if not default %}u-hide{% endif %}" {% if default %}data-default="true"{% endif %}>
+<section data-lang="{{ locale }}" class="p-strip--image is-dark is-deep u-image-position js-takeover p-takeover--1810--webinar">
```
anthonydillon: Please remove

anthonydillon:
QA: Can you add back the incremental index, please.

PR #4678, opened 2019-02-11
Updated all typos 'premise' to 'premises'

deadlight:
## Done

Fixed all instances of "premise", replacing them with "premises"

## QA

### DIFF
--- a/templates/kubernetes/managed.html
+++ b/templates/kubernetes/managed.html
@@ -11,7 +11,7 @@
         <h1>
           Fully managed private Kubernetes clusters. Cost effective. Pure upstream.
         </h1>
-        <p>Focus on your business while Canonical builds and manages your Kubernetes clusters. Take advantage of your on-premise assets and leverage public clouds for your multi-cloud strategy.</p>
+        <p>Focus on your business while Canonical builds and manages your Kubernetes clusters. Take advantage of your on-premises assets and leverage public clouds for your multi-cloud strategy.</p>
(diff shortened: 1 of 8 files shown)


templates/kubernetes/managed.html
```diff
@@ -48,7 +48,7 @@ <h2 class="p-heading--three">Reliable</h2>
       </div>
       <div class="col-4 p-divider__block">
         <h2 class="p-heading--three">Pure upstream</h2>
-        <p>We work directly with Google and the community to provide a pure Kubernetes experience on-premise and maximise compatibility across all public Kubernetes services such as Google GKE, Amazon EKS and Azure AKS.</p>
+        <p>We work directly with Google and the community to provide a pure Kubernetes experience on-premises and maximise compatibility across all public Kubernetes services such as Google GKE, Amazon EKS and Azure AKS.</p>
```
pmahnke:
```suggestion
        <p>We work directly with Google and the community to provide a pure Kubernetes experience on premises and maximise compatibility across all public Kubernetes services such as Google GKE, Amazon EKS and Azure AKS.</p>
```

deadlight:
Very unlikely. If that's the case then Google will treat them as synonyms.

I get a negative impression of poorly-written copy, and I think that
appealing to humans is a better goal than speculative SEO hacking.

I'm not a prescriptivist in language but when this is a case where using
the wrong word can completely derail a sentence.

PR #4680, opened 2019-02-11
Update link in core full disk encryption takeover

pmahnke:
## Done

- updated the link to - /engage/iot-disk-encryption?utm_source=takeover&utm_campaign=CY19_IOT_Ubuntu Core_Whitepaper_Secure booth/FDE

## QA

- See that the takeover goes to the above engage page


hatched:
There is a link lower down on the page that wasn't updated. Did this need to be updated as well?

Because it was `#` before, does that mean that there is a js handler somewhere intercepting the click so this link won't actually go anywhere?

hatched:
Heh nm my comment about the second link. I see you updated it while I was typing that message :-)

anthonydillon:
@hatched good spot

PR #4752, opened 2019-02-26
Disable Google Optimize script

barrymcgee:
## Done

Disable Google Optimize script as it's not being used.

## QA

### DIFF
--- a/templates/templates/_tag_manager.html
+++ b/templates/templates/_tag_manager.html
@@ -1,10 +1,12 @@
         <!-- Google Optimize page-hiding -->
+        {% comment %}
         <style>.async-hide { opacity: 0 !important} </style>
         <script>(function(a,s,y,n,c,h,i,d,e){s.className+=' '+y;h.start=1*new Date;
         h.end=i=function(){s.className=s.className.replace(RegExp(' ?'+y),'')};
         (a[n]=a[n]||[]).hide=h;setTimeout(function(){i();h.end=null},c);h.timeout=c;
         })(window,document.documentElement,'async-hide','dataLayer',1000,
         {'GTM-N2MDH37':true});</script>
+        {% endcomment %}
         <!-- End Google Optimize page-hiding -->
 
         <!-- Google Analytics and Google Optimize -->


templates/templates/_tag_manager.html
```diff
@@ -1,10 +1,10 @@
         <!-- Google Optimize page-hiding -->
-        <style>.async-hide { opacity: 0 !important} </style>
+        <!-- <style>.async-hide { opacity: 0 !important} </style>
```
pmahnke:
```suggestion
        <style>.async-hide { opacity: 0 !important} </style>
```

templates/templates/_tag_manager.html
```diff
@@ -1,10 +1,10 @@
...
         <script>(function(a,s,y,n,c,h,i,d,e){s.className+=' '+y;h.start=1*new Date;
         h.end=i=function(){s.className=s.className.replace(RegExp(' ?'+y),'')};
         (a[n]=a[n]||[]).hide=h;setTimeout(function(){i();h.end=null},c);h.timeout=c;
         })(window,document.documentElement,'async-hide','dataLayer',1000,
-        {'GTM-N2MDH37':true});</script>
+        {'GTM-N2MDH37':true});</script> -->
```
pmahnke:
```suggestion
        {'GTM-N2MDH37':true});</script>
{% endcomment %}
```

templates/templates/_tag_manager.html
```diff
@@ -1,10 +1,10 @@
         <!-- Google Optimize page-hiding -->
```
pmahnke:
```suggestion
{% comment %}
        <!-- Google Optimize page-hiding -->
```

nottrobin:
So unfortunately we do still get a flash of the default font, which looks pretty jarring:

image

I think we need to make a call right now about whether this unpleasant experience for people's first visit more, or the ~0.5s delay in showing the page more.

Either way, I think we should find a way to inline these fonts in our CSS as data-uris, possibly using some of @hatched's pruning tricks.

barrymcgee:
@nottrobin Added your FOUT issue here: https://github.com/canonical-websites/www.ubuntu.com/issues/4756

nottrobin:
Thanks @barrymcgee. I'm happy to merge this without fixing the FOUT issue, but it would be good if we could focus on the issue as a priority.

Having [discovered](https://github.com/canonical-websites/www.ubuntu.com/issues/4756#issuecomment-467406286) that the preconnect stuff seemingly doesn't have any impact at all, I really do think the only actual solution is to inline our fonts inside the CSS, however we do that.

PR #4757, opened 2019-02-26
Use cookie policy package

huwshimi:
## Done

- Replaced the cookie policy code with the cookie-policy npm package.

## QA

- Check that the cookie policy is shown (you might need to use a private browser window).


static/sass/styles.scss
```diff
@@ -7,6 +7,9 @@
 // custom mixins
 @import "mixins/full-width-breakout";
 
+// import cookie policy
+@import "cookie-policy/build/css/cookie-policy";
```
anthonydillon: I don't think you need these. I believe the js brings in the CSS.

anthonydillon:
Can you bump the timer from 3 seconds to something more.

static/sass/styles.scss
huwshimi: I just double checked the JS, but without the css include it appeared unstyled.

huwshimi:
@anthonydillon I've updated it to appear for 5 seconds (same as blog.ubuntu.com). Sound OK?

huwshimi:
@anthonydillon I've updated the timeout to be 10 seconds, to match the current timeout on ubuntu.com.

.gitignore
```diff
@@ -60,5 +60,10 @@ env[23]/
 # Project-specific ignores
 etc/ubuntu-mirrors-rss.xml
 .search-api-key
+<<<<<<< HEAD
```
anthonydillon: Cornflakes 🤪

PR #4764, opened 2019-02-26
Remove and ignore global nav

nottrobin:
QA
--

`./run`, check global nav still works, and the git work tree appears untouched.


.gitignore
```diff
@@ -62,3 +62,4 @@ etc/ubuntu-mirrors-rss.xml
 .search-api-key
 static/global-nav
 
+statis/js/global-nav.js
```
pmahnke: is that a typo?
```suggestion
static/js/global-nav.js
```

barrymcgee: Good catch.

PR #4775, opened 2019-03-01
Replace images

nottrobin:
Draft: Replace images, using https://github.com/nottrobin/image-template/blob/replacer/scripts/replace-images.py


cristinadresch:
move this

Issue #4723, opened 2019-02-19
Download pages do not offer checksums


pmatulis:
There are no links to checksum files for the stuff that people download such as Server and Desktop ISO files. The user has to stumble their way to a mirror in order to find them. This is bad UX. Simply add a link to a reliable and trusted mirror page such as (for Bionic):

http://releases.ubuntu.com/bionic/

---
*Reported from: https://www.ubuntu.com/download/server*

pmahnke:
@pmatulis  perhaps not the best UX, but on the thank-you pages we have a Verify your download CTA with a link to this tutorial - https://tutorials.ubuntu.com/tutorial/tutorial-how-to-verify-ubuntu

I will try and put a better UXer on the project!

pmatulis:
@pmahnke That's not what people expect, so the vast majority will never go looking for it.

The tutorial uses a standard mirror listing of files so I'm not sure if that's the best resource to point people to. It can even be a source of confusion (see below). 

This issue (lack of checksums) was raised on the [ubuntu-doc mailing list](https://lists.ubuntu.com/archives/ubuntu-doc/2019-February/020642.html).

mlda065:
> on the thank-you pages we have a Verify your download CTA with a link to this tutorial

That's a great start (I can never remember the gpg commands). But it's not good enough. The tutorial tells the user that the checksum and signature can be downloaded on the same page as the iso. This is a false statement.

@pmatulis just FYI, I'm the person who wrote that initial email. I never actually received your reply. I can see it [here](https://lists.ubuntu.com/archives/ubuntu-doc/2019-February/020643.html). So I think there's a problem with the mailing list. (I recall spotting a confirmation email in my spam filter several days after that. So perhaps my subscription was put on hold? I don't know.)

mlda065:
Actually, I stand corrected. It seems the [verification tutorial](https://tutorials.ubuntu.com/tutorial/tutorial-how-to-verify-ubuntu?_ga=2.64884655.1248685753.1547162864-1813460345.1547162864#2) has been updated since I sent that email.

[...]

PR #4800, opened 2019-03-09
add checksum links to download thank-you page

mlda065:
## Done

I added links to the checksum and signature files on the download thank you page.

For now I've just done it for server and desktop. 
If you guys are happy with this, I can try adding it to all the other download pages too.

## QA

## Issue / Card

#4723 

## Screenshots

Screenshot from 2019-03-09 13-37-37

Screenshot from 2019-03-09 13-37-51

### DIFF
--- a/templates/download/desktop/thank-you.html
+++ b/templates/download/desktop/thank-you.html
@@ -47,6 +47,19 @@ Thank you for your contribution
         <a href="http://releases.ubuntu.com/{{ version }}/ubuntu-{{ version }}-desktop-{{ architecture }}.iso">download now</a>.
         {% endif %}
       </p>
+      <p>
+         You can <a href="https://tutorials.ubuntu.com/tutorial/tutorial-how-to-verify-ubuntu">verify</a>
+         your image using the
+         {% if architecture == 'amd64+mac' %}
+         <a href="http://cdimage.ubuntu.com/releases/{{ version }}/release/SHA256SUMS">SHA256 checksum</a>
+         and
+         <a href="http://cdimage.ubuntu.com/releases/{{ version }}/release/SHA256SUMS.gpg">signature</a>.
+         {% else %}
+         <a href="http://releases.ubuntu.com/{{ version }}/SHA256SUMS">SHA256 checksum</a>
+         and
+         <a href="http://releases.ubuntu.com/{{ version }}/SHA256SUMS.gpg">signature</a>.
+         {% endif %}
+      </p>
       {% else %}
       <h1>Thank you for your contribution</h1>
       <p>It will help further the open source development of Ubuntu.</p>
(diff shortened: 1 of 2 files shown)


mlda065:
By the way, I just want to say that the whole `./run` server-in-a-container thing is great! It makes it so easy to contribute.
Keep up the good work!

nottrobin:
Thanks so much for taking the time @mlda065, and I'm really happy the `./run` script works well for you =)

Unfortunately, simply linking to `http://releases.ubuntu.com/{{ version }}/SHA256SUMS.gpg` isn't ideal because that domain is served over unencrypted HTTP, and so the checksums could be manipulated in transit (e.g. with a [MITM](https://en.wikipedia.org/wiki/Man-in-the-middle_attack)) - which makes them effectively useless as one of their main points is to protect you from an Ubuntu download that may have been manipulated in transit.

The simplest secure solution would be to link to https://tutorials.ubuntu.com/tutorial/tutorial-how-to-verify-ubuntu instead, which explains how to verify your download in a secure way, but that puts a fair burden on the user. An even better solution would be to actually provide the correct checksums directly on these thank-you pages (which are served over encrypted HTTPS).

So if you have time, it would be very helpful if you can update your PR to either:

1. Link to https://tutorials.ubuntu.com/tutorial/tutorial-how-to-verify-ubuntu; or:
2. Communicate the appropriate checksums directly for each OS version we currently provide (not trivial...)

I'll check with the team if we think we can take on the burden of keeping live checksums up-to-date for new images, but I think we should be able to do that.

Thanks again for your help.

mlda065:
> that domain is served over unencrypted HTTP, and so the checksums could be manipulated in transit (e.g. with a MITM) - which makes them effectively useless

The first half of that is true, the second half is false.

The whole point of the cryptography stuff is to verify the hash file. The ISO files are not cryptographically signed.  The hash files are. You can download the hash files from an untrustworthy source. As long as the gpg commands say it's ok, then it is ok.

Even if that wasn't true, the url you are asking me to link to does not contain the hash files. *It is not possible to securely obtain the hash files or signatures.* (please provide an https url to prove me wrong.) You can only obtain them through the same insecure locations as the ISOs. Then you cryptographically verify them using the Ubuntu signing key, which can be obtained over https from that tutorial page.

I did already link to the url you are asking for. The word "verify" is a link to that. Perhaps I should rephrase that to make it more obvious that it links to instructions on how to verify?

That tutorial page says:

> Alongside the actual ISO files containing the Ubuntu image you downloaded, all Ubuntu mirrors publish some extra files

That statement is currently false. This pull request makes it true.

> you can fetch these files from the http://releases.ubuntu.com page for the relevant release. For example, for the current version of Ubuntu, if you check in http://releases.ubuntu.com/bionic/ you will find these files near the top.

So the tutorial page just tells the user to download the signatures from the http location my pull request links to.

Therefore I believe my pull request is still valid. It solves a problem in a reasonable and straightforward way, with zero drawbacks.

I'm not familiar enough with the structure of this website and all the mirrors and infrastructure to implement option 2. Since that is non-trivial, this pull request should be merged in the interim.

PR #4803, opened 2019-03-11
Fetch test of javascript based dropdown navigation

### LINKED ISSUE #4527: Navigation doesn't work without JavaScript
If one has JavaScript turned off, the main navigation on www.ubuntu.com looks like it should work (the chevrons are there) but it does nothing at all.

Couldn't we at least make it so the top-level items (Enterprise, Developer, Community, Download) are clickable so they can take you *somewhere* when you don't have JavaScript, so you can begin to browse the site? Also, maybe remove the chevrons if JavaScript isn't available?

pmahnke:
## Done

- Loading dropdowns with fetch onmouseover, one time only

## QA

- Test the dropdowns
- Test with no script

## Issues

Fixes #4527


anthonydillon:
Nice work, works really well.

## QA issues
 - [ ] Some buttons are missing text in the download section:
Screenshot_2019-03-11 The leading operating system for PCs, IoT devices, servers and the cloud Ubuntu
 - [ ] If you refresh once one is open it does not fetch the menu
 - [ ] This makes the navigation strictly JS only. Maybe you could add a noscript simple navigation. Maybe included from the footer.

Caleb-Ellis:
@pmahnke You should add a listener for on focus as well, otherwise people navigating by keyboard won't get anything

Issue #4773, opened 2019-03-01
Reduce number of JS HTTP requests


barrymcgee:
## Summary

Speedcurve reports the Ubuntu homepage requesting 24 distinct HTTP requests for JS files. Can this be reduced? Can files be merged? If not, are they properly deferred?

## Process

- Go to www.ubuntu.com
- Use network tab in dev tools to monitor JS requests

## Current and expected result

Can we try and reduce this number by 5 to start with?

## Screenshot

<img width="1233" alt="screenshot 2019-03-01 at 10 05 02" src="https://user-images.githubusercontent.com/505570/53627737-85487000-3c09-11e9-8690-67dd20a02878.png">

PR #4829, opened 2019-03-15
Minify and combine js

huwshimi:
## Done

- This work is to reduce the number of scripts loaded as well as minifying the JS.
- Add Webpack and Babel.
- Moved some files to modules where needed.
- Removed global nav/cookie policy script copying.

## QA

- Check that all Javascript works as before, the following covers instances of all bundled/minified files:
  - Check global nav, cookie policy, main nav dropdowns etc. on all pages
  - Check charts are shown on /about/release-cycle
  - Check charts and sticky nav work on /desktop/statistics
  - Check the contact us button on the bottom of pages like /openstack/consulting opens the form
  - Check the country select box works on /openstack/contact-us?product=foundation-cloud

## Issue / Card

- Fixes: #4773

### DIFF
--- a/static/js/cookie-proxy.js
+++ b/static/js/cookie-proxy.js
@@ -1,3 +1,3 @@
-var cpNs = require('exports-loader?cpNs!./cookie-policy.js');
+var cpNs = require('exports-loader?cpNs!cookie-policy/build/js/cookie-policy');
 
 export default cpNs;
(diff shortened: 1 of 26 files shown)


static/js/cookie-proxy.js
```diff
@@ -0,0 +1,3 @@
+var cpNs = require('exports-loader?cpNs!cookie-policy/build/js/cookie-policy');
```
huwshimi: This file is not currently set up as a module, so we use the special loader to expose the global variable that we need.

PR #4844, opened 2019-03-16
Add /legal/partner-portal-terms page

### LINKED ISSUE #4841: Legal - Partner portal - partner-portal-terms & conditions

pmahnke:
## Done

- Added a new t&c page for the partner portal
- Created a copy doc
- Added the page to the navigation

## QA

- View the site locally in your web browser at: [/legal/partner-portal-terms](http://0.0.0.0:8001/legal/partner-portal-terms)
- Compare it to the [copy doc](https://docs.google.com/document/d/1CqJbwC3KpEJeJaCUVjAIN0y6ZLmoc7gVlFyc-BeFMYw/edit#)

## Issue / Card

Fixes #4841

### DIFF
--- a/navigation.yaml
+++ b/navigation.yaml
@@ -420,6 +420,8 @@ legal:
           path: /legal/systems-information-notice
         - title: Ubuntu font
           path: /legal/font-licence
+        - title: Partner Portal
+          path: /legal/partner-portal-terms
 
           children:
             - title: FAQ
--- a/navigation.yaml
+++ b/navigation.yaml
@@ -402,19 +402,19 @@ legal:
         - title: Intellectual property - 14th May 2013
           path: /legal/intellectual-property-policy/2013-05-14
           hidden: True
-        - title: Online account terms
+        - title: Online accounts
           path: /legal/online-account-terms
         - title: U1 Terms of Service
           path: /legal/terms-of-service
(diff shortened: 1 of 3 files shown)


nottrobin:
The space above the "VERSION" line looks very odd to me:

image

Can anything be done about that?

pmahnke:
@nottrobin 	that's the design from many other pages, we could remove the top padding?  @anasereijo  I have asked you to look please.

anasereijo:
@nottrobin thanks for your comment! I agree it does look odd but as @pmahnke it is the same padding we use on other pages. 

I think if we have the title first and then the version it might help. See below:
Screenshot 2019-03-20 at 15 07 29

@pmahnke I compared to other pages and they are all `class="p-strip is-deep" ` but this one for some reason looks slightly deeper than the other pages. Can you please check they all match?

Finally, is it possible to make the `hr` 12 col?

Thank you

pmahnke:
@anasereijo 	it was the same as on the data-privacy section... I make a new row now with col-12

I think it is that the second level nav is three levels deep that is the real issue.

Any more fixing will require coming out of markdown or hacking... so let's chat if this doesn't work.

anasereijo:
@pmahnke I think it's already looking better with the 12 col line

PR #4967, opened 2019-04-17
Update to Django2

b-m-f:
## Done

- uses sync workers instead of gevent (https://docs.gunicorn.org/en/stable/design.html)
- upgrades to Django 2.2 LTS (https://www.djangoproject.com/weblog/2019/apr/01/django-22-released/)
- removes unsupported `django-unslashed` middleware
## QA

## Issue / Card

Fixes https://github.com/canonical-web-and-design/base-squad/issues/454

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -720,6 +720,7 @@ certification(?P<page>(/.*)?): "https://certification.ubuntu.com{page}"
 usn/?: "https://usn.ubuntu.com/"
 usn/rss\.xml: "https://usn.ubuntu.com/rss.xml"
 (?P<page>.*)\.html: "/{page}"
+(?P<page>.+)/: "/{page}"
 telco/?: "/telecommunications"
 telcos/?: "/telecommunications"
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@ canonicalwebteam.http==0.1.6
 whitenoise==4.1.2
 statsd==3.3.0
 prometheus_client==0.6.0
-talisker[gunicorn]==0.14.2
+talisker[gunicorn]==0.14.3
 django-yaml-redirects==0.5.4
 django-asset-server-url==0.1
 django-static-root-finder==0.3.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,10 +7,9 @@ canonicalwebteam.yaml-responses[django]==1.1.0
 canonicalwebteam.get-feeds==0.2.4
 canonicalwebteam.http==0.1.6
 whitenoise==4.1.2
-gunicorn[gevent]
 statsd==3.3.0
 prometheus_client==0.6.0
-talisker==0.14.2
+talisker[gunicorn]==0.14.2
 django-yaml-redirects==0.5.4
 django-asset-server-url==0.1
 django-static-root-finder==0.3.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-Django==1.11.20
+Django==2.2
 django-prometheus==1.0.15
 canonicalwebteam.custom_response_headers==0.2.0
 canonicalwebteam.versioned_static==1.0.2
@@ -14,7 +14,6 @@ talisker==0.14.2
 django-yaml-redirects==0.5.4
 django-asset-server-url==0.1
 django-static-root-finder==0.3.1
-django-unslashed==0.3.0
 feedparser==5.2.1
 requests==2.21.0
 requests-cache==0.4.13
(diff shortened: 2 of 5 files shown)


pmahnke:
I put a don't merge label on, because it is release day.

b-m-f:
Remoed **Dont merge** and changed the target branch to  `canonical-web-and-design:upgrade-to-django2`

webapp/settings.py
```diff
@@ -112,6 +110,10 @@
 
 ASSET_SERVER_URL = "https://assets.ubuntu.com/v1/"
 
+DATABASES = {
```
tbille: why adding this?

tbille:
can you run `./scripts/get-mirrors-rss`

this will give you the list of mirrors to download ubuntu.

Then go on: 0.0.0.0:8001/download/desktop/thank-you?version=18.10&architecture=amd64

you will see that you get a 500.

in `webapp/views.py` you need to update:

```
        try:
            with open(mirrors_path) as rss:
                mirrors = parse(rss.read()).entries
                                ^^^^^^^^^^
        except IOError:
            mirrors = []
```

webapp/settings.py
nottrobin: Yeah, it's a strong requirement that this site doesn't have a database.

b-m-f: What do you think about the testing approach here?

Since we have no DB I have found the following approach on [Stackoverflow](https://stackoverflow.com/questions/5917587/django-unit-tests-without-a-db/):

>  from django.test.runner import DiscoverRunner
> 
>  class NoDbTestRunner(DiscoverRunner):
>    """ A test runner to test without database creation/deletion """
> 
>    def setup_databases(self, **kwargs):
>      pass
> 
>    def teardown_databases(self, old_config, **kwargs):
>      pass
> 
> Use like that :
> 
> python manage.py test app --testrunner=app.filename.NoDbTestRunner
> 

The only problem here is that the tests have to be called with explicitly specifying the test runner. Can still easily be called in the `package.json` of course.

jpmartinspt: I think you can set the test_runner in settings.py!

redirects.yaml
```diff
@@ -717,5 +717,6 @@ certification(?P<page>(/.*)?): "https://certification.ubuntu.com{page}"
 usn/?: "https://usn.ubuntu.com/"
 usn/rss\.xml: "https://usn.ubuntu.com/rss.xml"
 (?P<page>.*)\.html: "/{page}"
+(?P<page>.+?)\/: "/{page}"
```
nottrobin: I'm not quite sure what the `?` is doing in this expression. Is `.+?` the same as `.*`? If not, what does it mean? I also don't think `/` needs to be escaped.

I think this should work:

```suggestion
(?P<page>.+)/: "/{page}"
```

webapp/settings.py
```diff
@@ -66,21 +65,18 @@
...
     "whitenoise.middleware.WhiteNoiseMiddleware",
     "canonicalwebteam.custom_response_headers.Middleware",
-    "unslashed.middleware.RemoveSlashMiddleware",
 ]
 
 # Prometheus
 if not DEBUG:
     INSTALLED_APPS.append("django_prometheus")
-    MIDDLEWARE_CLASSES.insert(
+    MIDDLEWARE.insert(
```
nottrobin: @tbille should prometheus now be covered with Talisker? Or not?

redirects.yaml
b-m-f: Works! Thank you

requirements.txt
```diff
@@ -7,14 +7,12 @@ canonicalwebteam.yaml-responses[django]==1.1.0
 canonicalwebteam.get-feeds==0.2.4
 canonicalwebteam.http==0.1.6
 whitenoise==4.1.2
-gunicorn[gevent]
 statsd==3.3.0
 prometheus_client==0.6.0
-talisker==0.14.2
+talisker[gunicorn]==0.14.2
```
nottrobin: This should now be `0.14.3` I think. @tbille could confirm.

b-m-f: Updated it. Its the latest version

webapp/views.py
```diff
@@ -407,7 +407,7 @@ def get_context_data(self, **kwargs):
 
         try:
             with open(mirrors_path) as rss:
-                mirrors = parse(rss).entries
+                mirrors = parse(rss.read()).entries
```
nottrobin: Actually, @tbille added this in his PR, which is already merged to the django2 branch. If you rebase, I think you'll find this is already there.

webapp/settings.py
nottrobin: Let's treat this as out-of-scope =)

nottrobin: Issue: https://github.com/canonical-web-and-design/www.ubuntu.com/issues/5036

PR #4968, opened 2019-04-17
Download mirror list on build

tbille:
## Done

Fixes https://github.com/canonical-web-and-design/base-squad/issues/459
Download mirror list at build time

## QA

- `./run`
- make sure new directory created with the list of mirrors: `$PROJECT/etc/ubuntu-mirrors-rss.xml`
- If you run again, the script **shouldn't** download again the mirrors
- `./run clean`
- Make sure `$PROJECT/etc/` is deleted

### DIFF
--- a/scripts/get-mirrors-rss
+++ b/scripts/get-mirrors-rss
@@ -12,10 +12,10 @@ XML_FILE=${PROJECT_DIR}/etc/ubuntu-mirrors-rss.xml
 
 mkdir -p ${PROJECT_DIR}/etc
 rm -f $TEST_FILE
-wget -q -O $TEST_FILE https://launchpad.net/ubuntu/+cdmirrors-rss
 
+curl https://launchpad.net/ubuntu/+cdmirrors-rss -o $TEST_FILE
 RESULT=$(grep "Ubuntu CD Mirrors Status" $TEST_FILE)
 
 if [ "$RESULT" ]; then
-     mv $TEST_FILE $XML_FILE
+    mv $TEST_FILE $XML_FILE
 fi
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -407,7 +407,7 @@ class DownloadView(UbuntuTemplateFinder):
 
         try:
             with open(mirrors_path) as rss:
-                mirrors = parse(rss).entries
+                mirrors = parse(rss.read()).entries
         except IOError:
             mirrors = []
 
--- a/package.json
+++ b/package.json
@@ -8,8 +8,9 @@
     "build-css": "node-sass --include-path node_modules static/sass --source-map true --output-style compressed --output static/css && postcss --use autoprefixer --replace 'static/css/**/*.css' --no-map",
     "build-js-bundle": "webpack",
     "build": "yarn run build-css && yarn run build-js-bundle",
+    "get-mirrors": "./scripts/get-mirrors-rss",
     "watch": "watch -p 'static/sass/**/*.scss' -p 'node_modules/vanilla-framework/scss/*.scss' -c 'yarn run build'",
-    "clean": "rm -rf node_modules yarn-error.log css static/css *.log *.sqlite _site/ build/ .jekyll-metadata .bundle static/js/build",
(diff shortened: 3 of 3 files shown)


tbille:
I am checking what is the issue when accessing http://0.0.0.0:8001/download/desktop/thank-you?version=18.04.2&architecture=amd64
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/dist-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "/usr/local/lib/python3.6/dist-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/dist-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.6/dist-packages/django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/django/views/generic/base.py", line 88, in dispatch
    return handler(request, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/django/views/generic/base.py", line 155, in get
    context = self.get_context_data(**kwargs)
  File "/home/tbmb/Documents/projects/www.ubuntu.com/webapp/views.py", line 410, in get_context_data
    mirrors = parse(rss).entries
  File "/usr/local/lib/python3.6/dist-packages/feedparser.py", line 3922, in parse
    data, result['encoding'], error = convert_to_utf8(http_headers, data)
  File "/usr/local/lib/python3.6/dist-packages/feedparser.py", line 3574, in convert_to_utf8
    xml_encoding_match = RE_XML_PI_ENCODING.match(tempdata)
TypeError: cannot use a bytes pattern on a string-like object

```

webapp/views.py
```diff
@@ -407,7 +407,7 @@ def get_context_data(self, **kwargs):
 
         try:
             with open(mirrors_path) as rss:
-                mirrors = parse(rss).entries
+                mirrors = parse(rss.read()).entries
```
tbille: @nottrobin we need to try this on python 2 I am not sure if it will be handled.

scripts/get-mirrors-rss
```diff
@@ -12,10 +12,12 @@ XML_FILE=${PROJECT_DIR}/etc/ubuntu-mirrors-rss.xml
 
 mkdir -p ${PROJECT_DIR}/etc
 rm -f $TEST_FILE
-wget -q -O $TEST_FILE https://launchpad.net/ubuntu/+cdmirrors-rss
 
-RESULT=$(grep "Ubuntu CD Mirrors Status" $TEST_FILE)
+if [ ! -f $XML_FILE ] ; then
```
nottrobin: I can see why you did this - you don't want the mirror file to be downloaded every time in local development.

Unfortunately, this won't work for Jenkins. In Jenkins we need the mirrors file to refresh every time we build.

We could solve this by doing `rm etc/ubuntu-mirrors-rss.xml` before `./run build` in Jenkins - but that's a bit of weird structure, because we're expecting Jenkins to have deep knowledge of how this script works, but not letting it run the script directly.

If the time taken to download the mirror file every time is a concern (which I'd say it probably is), then I'd suggest this:

-  We remove the `if [ ! -f £XML_FILE ]` here, so the script downloads the mirror file every time
- We remove `./scripts/get-mirrors-rss` from `build`, and instead put it in a new `package.json` script called `get-mirrors`
- We document that, if you want the mirrors file for local development, you can run manually `./run exec yarn run get-mirrors`
- In the Jenkins build job we run `./run build && ./run exec yarn run get-mirrors`

Wadayathink?

tbille: Aye sink it maykes cense

webapp/views.py
nottrobin: Now that this is merging into "update-to-django2" branch, we presumably don't care if it works on python2.

PR #5203, opened 2019-05-24
Update stale to focus on ignored issues

pmahnke:
per discussion with @anthonydillon


nottrobin:
This looks interesting. What does it do?

pmahnke:
@nottrobin see https://github.com/probot/stale it will push us to deal with issues in a more timely fashion

PR #5218, opened 2019-05-29
5123 improve blog card template

sowasred2012:
## Done

- Ensured a group label is always present in a blog card header (defaults to "Ubuntu" if one isn't returned from the API)
- Added logic from blog.ubuntu.com to ensure a category label is always present in a blog card footer
- When a newsletter card is shown, ensured the blog cards on the same row display their post excerpts to reduce the amount of white space
- Updated newsletter card template to more closely match the one on blog.ubuntu.com

## QA

- View the site locally in your web browser at: [/blog](http://0.0.0.0:8001/blog)
- Ensure that blog cards on the same row as the newsletter display their summary
- Ensure blog cards listed throughout the blog each has:
  - a group label in its header
  - a category label in its footer

## Issue / Card

Fixes https://github.com/canonical-web-and-design/www.ubuntu.com/issues/5213

## Screenshots

Screenshot 2019-05-29 at 14 20 30

Screenshot 2019-05-29 at 14 20 53

### DIFF
--- a/static/sass/_pattern_blog-card.scss
+++ b/static/sass/_pattern_blog-card.scss
@@ -84,6 +84,7 @@
 
   .blog-p-card__header {
     @extend %post-card-header;
+    border-top: 3px solid #666; //sass-lint:disable-line no-color-literals
 
     &--cloud-and-server {
       @extend %post-card-header;
--- a/static/sass/_pattern_blog-card.scss
+++ b/static/sass/_pattern_blog-card.scss
@@ -82,43 +82,40 @@
     }
   }
 
-  .blog-p-card__header {
+  // a fallback to catch any instances where a post.group exists,
+  // but we haven't yet defined a BEM variant to match
+  [class*="blog-p-card__header"] {
     @extend %post-card-header;
-    border-top: 3px solid #666; //sass-lint:disable-line no-color-literals
+    border-top: 3px solid $color-mid-dark;
+  }
+
+  .blog-p-card__header {
+    &--canonical-announcements {
+      border-color: #ff8936; //sass-lint:disable-line no-color-literals
+    }
 
     &--cloud-and-server {
-      @extend %post-card-header;
-      border-top: 3px solid #a87ca0; //sass-lint:disable-line no-color-literals
+      border-color: #a87ca0; //sass-lint:disable-line no-color-literals
     }
 
     &--desktop {
-      @extend %post-card-header;
-      border-top: 3px solid #faba54; //sass-lint:disable-line no-color-literals
+      border-color: #faba54; //sass-lint:disable-line no-color-literals
     }
 
     &--internet-of-things {
-      @extend %post-card-header;
-      border-top: 3px solid #8db255; //sass-lint:disable-line no-color-literals
-    }
-
-    &--canonical-announcements {
-      @extend %post-card-header;
-      border-top: 3px solid #ff8936; //sass-lint:disable-line no-color-literals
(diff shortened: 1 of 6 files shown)


templates/blog/blog-card-upcoming.html
```diff
@@ -1,6 +1,6 @@
 {% for article in upcoming %}
   <div class="col-4 blog-p-card--post">
-    <header class="blog-p-card__header--{{article.group.slug}}">
+    <header class="blog-p-card__header blog-p-card__header--{{article.group.slug}}">
```
anthonydillon: Single classes are much easier to maintain then having to chain them with knowledge of the knock on effect. Could you revert back to extending `%post-card-header` on each modifier please

sowasred2012: I've reverted that change, but remembered why I made it as I was doing so - in [blog-card.html](https://github.com/canonical-web-and-design/www.ubuntu.com/blob/24b4ff93622b3df8d16e4d6b14b14b02141f6b7a/templates/blog/blog-card.html#L3), that class chain means it falls back to the default border color if we for some reason haven't defined a `blog-p-card__header--` variant based on the group slug. I could look into a better solution for that so we're not chaining there too?

anthonydillon: I would suggest if an article has no group to have the class `blog-p-card__header` or simple `blog-p-card__header--default` that can contain the default styling

anthonydillon:
Could you remove the duplicated card header classes. I know you didnt add them but we should just use a single class for BEM mods.

anasereijo:
Looks good design wise!

templates/blog/blog-card.html
```diff
@@ -1,6 +1,6 @@
 <div class="col-4 blog-p-card--post">
   {% if article.group %}
-  <header class="blog-p-card__header blog-p-card__header--{{ article.group.slug }}">
+  <header class="blog-p-card__header--{{ article.group.slug }}">
```
anthonydillon: Couldn't we add some logic here. If there is a group slug then `--{{ article.group.slug }}` is prepended to the class?

sowasred2012: Good point - updating that now

static/sass/_pattern_blog-card.scss
```diff
@@ -82,8 +82,21 @@
...
+  // a fallback to catch any instances where a post.group exists,
+  // but we haven't yet defined a BEM variant to match
+  [class*="blog-p-card__header"] {
+    @extend %post-card-header;
+    border-top: 3px solid #666; //sass-lint:disable-line no-color-literals
+  }
+
   .blog-p-card__header {
     @extend %post-card-header;
+    border-top: 3px solid #666; //sass-lint:disable-line no-color-literals
```
Caleb-Ellis: Since you've already got the fallback you can remove the `@extend`s for each variant - not gonna result in any less css but it's kind of redundant. Also, all you need is `border-color` for each variant because the width and style is already defined. Also also, Vanilla has `#666` as one of its defaults (`$color-mid-dark`) so you can just use that: https://docs.vanillaframework.io/en/settings/color-settings

So it would look like:
``` scss
[class*="blog-p-card__header"] {
  @extend %post-card-header;
  border-top: 3px solid $color-mid-dark;
}

.blog-p-card__header {
  &--category1 {
    border-color: $color-category-1;
  }

  &--category2 {
    border-color: $color-category-2;
  }
...
}
```

anthonydillon:
Thanks good to go

Issue #4756, opened 2019-02-26
Flash of unstyled text on page load


barrymcgee:
## Summary

Due to a performance issue, a new user will see text unstyled before the Ubuntu font kicks in.

nottrobin:
Unfortunately this isn't fixed by https://github.com/canonical-websites/www.ubuntu.com/pull/4750, as we thought it might be. I just tested it, it doesn't seem to have any effect at all.

pmahnke:
@barrymcgee and @nottrobin next steps?  should I close this for now?

barrymcgee:
@pmahnke Yes, I think we exhausted any options we had in Malta at the time - is that your view @nottrobin?

nottrobin:
I dunno. It still unsettles me. I don't think this issue is high priority, but I would like to leave it open if we can. I'm not quite sure we definitely exhausted all options. I feel that there there should be a way to hold off on first paint until the font has downloaded, or maybe inline the fonts, or at least make it so the titles display at the same time as the rest of the text.

nottrobin:
This could maybe be addressed through [the `font-display` rule](https://font-display.glitch.me/), but it doesn't look like https://github.com/canonical-web-and-design/www.ubuntu.com/pull/5226#issuecomment-500450043 implements it correctly.

barrymcgee:
@pmahnke It works!

matthewpaulthomas:
The `font-display` rule was removed in #5886.

barrymcgee:
@matthewpaulthomas A local override has been added via https://github.com/canonical-web-and-design/ubuntu.com/pull/5893 until the change can be made upstream in Vanilla.

SirSamTumless:
I am closing this. If you need to re-open it please make actions for it.

nottrobin:
:cry:

PR #5226, opened 2019-05-30
Add font-display attribute to font loading

barrymcgee:
## Done

Add `font-display: swap` attribute to font loading.

> The font swap period occurs immediately after the font block period. During this period, if the font face is not loaded, any element attempting to use it must instead render with a fallback font face. If the font face successfully loads during the swap period, the font face is then used normally.

Ref: https://developers.google.com/web/updates/2016/02/font-display

## QA

- Throttle the network on Chrome DevTools using a mobile emulator and observe fast font loading than in production.

Related: #4756

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -68,6 +68,12 @@
 @import "layout-whitepapers";
 @include layout-whitepapers;
 
+// Performant web font loading
+// https://developers.google.com/web/updates/2016/02/font-display
+body {
+  font-display: swap;
+}
+
 // Bug fixes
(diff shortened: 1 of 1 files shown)


nottrobin:
I don't think this is doing anything, I think it might be implemented wrongly:

image

Even though it [should be supported](https://caniuse.com/#search=font-display) in my Firefox 67.

I think this is because this property is meant for the `@font-face` block:

> @font-face {
>   font-family: 'my-font';
>   font-display: optional;
>   src: url(my-font.woff2) format('woff2');
> }

Per: https://font-display.glitch.me/

PR #5255, opened 2019-06-05
[RFC] Extra metadata handling for engage pages

caldav:
This pull request is a proposal for a new way to handle metadata for pages such as engage pages, where having an extra layer of metadata for social sharing is critical.

Currently we surface things such as page description through template blocks, but when the content needs to duplicated across tags, this doesn't scale, especially since Django doesn't allow duplicating blocks in the extended template.

This proposal introduces:
*  an "extends_with_args" template tag, that wraps around "extends" and allows setting context variables from extension pages.
* an `_extra_metadata.html` which makes use of these args to build social sharing metadata for Twitter, FB, etc.
* an example (`/engage/starting-with-ai`)

The benefits of this approach is that it doesn't conflict with our existing template layers and metadata surfacing, it's simply a new method that cloud be the preferred one for engage pages.

## QA

- Using the demo, test the following pages on https://socialdebug.com/:
  * `/engage/19-04-webinar`
  * `/engage/starting-with-ai`
  * `/desktop/developers`
- For each page, ensure the "overall score" for the demo is not regressing compared to prod
- Test the same pages on https://cards-dev.twitter.com/validator
- ~and on https://developers.facebook.com/tools/debug/sharing~ / it's using the canonical URL, hence failing on the demo.

### DIFF
--- a/templates/engage/starting-with-ai.html
+++ b/templates/engage/starting-with-ai.html
@@ -1,8 +1,4 @@
-{% extends "engage/base_engage.html" %}
-
-{% block title %}Getting Started with AI{% endblock %}
-
-{% block meta_description %}From the smallest startups to the largest enterprises alike, organizations are using Artificial Intelligence and Machine Learning to make the best, fastest, most informed decisions to overcome their biggest business challenges.{% endblock %}
+{% extends_with_args "engage/base_engage.html" with title="Getting started with AI" meta_image="https://assets.ubuntu.com/v1/83facbfb-getting-started-with-ai.png" meta_description="From the smallest startups to the largest enterprises alike, organizations are using Artificial Intelligence and Machine Learning to make the best, fastest, most informed decisions to overcome their biggest business challenges." %}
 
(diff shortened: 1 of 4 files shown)


anthonydillon:
@caldav I'm getting F's for all social scores and errors in the twitter card validator with the following URL: https://www-ubuntu-com-canonical-web-and-design-pr-5255.run.demo.haus/engage/19-04-webinar

caldav:
> @caldav I'm getting F's for all social scores and errors in the twitter card validator with the following URL: https://www-ubuntu-com-canonical-web-and-design-pr-5255.run.demo.haus/engage/19-04-webinar

@ant that's normal, the only page that uses the proposed mechanism is `/engage/starting-with-ai`, this one is there for checking potential regressions of our current score against prod.

nottrobin:
This is really neat, thanks @caldav 

@pmahnke I think this is basically the solution we were looking for in https://github.com/canonical-web-and-design/www.ubuntu.com/pull/5247, so this PR should replace that one, or if that has other work in it then that PR should be based on this one.

@caldav you just need to fix the formatting of your python with `black --line-length 79 webapp/` to pass the python linter.

caldav:
@nottrobin done!

PR #5260, opened 2019-06-06
Merge Django 2

nottrobin:
This is a huge branch, not worth checking manually (everything going into this branch should have been checked - except 00e2e9ae)

QA
--

`./run`, roughly check the site runs looks the same as master.


nottrobin:
Merging this myself, as it's going to be thoroughly QAed at the other end anyway, and we need it merged.

PR #5351, opened 2019-06-20
Bump blog version to 2.0.6

tbille:


nottrobin:
need `.webcache_blog/` in .gitignore

PR #5422, opened 2019-07-04
Fix int casting from string

tbille:
# Summary

Fixes int casting when parameters are != than int. (eg. `int("abc")`)

- https://ubuntu.com/search?q=kernel&num=10&start=asdf
- https://ubuntu.com/search?q=kernel&num=asdf&start=1
- https://ubuntu.com/resources?content=webinars&page=asdf
- https://ubuntu.com/resources?page=asdf

Fixes https://sentry.is.canonical.com/canonical/ubuntu-com/issues/2280/?query=is%3Aunresolved

# QA 

- `./run`
- Visit the pages previsouly mentionned
- All those page should change the parameter to it's default value (in most case 1)

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -26,6 +26,13 @@ if settings.SEARCH_API_KEY:
     )
 
 
+def str_to_int(string, default_value=1):
+    try:
+        return int(string)
+    except ValueError:
+        return default_value
+
+
 def _get_search_results(query, start, num, site=None):
     """
     Query the Google Custom Search API for search results
@@ -60,8 +67,8 @@ def search(request):
 
     query = request.GET.get("q")
     site = request.GET.get("siteSearch") or request.GET.get("domain")
-    num = int(request.GET.get("num", "10"))
-    start = int(request.GET.get("start", "1"))
+    num = str_to_int(request.GET.get("num", 10), 10)
+    start = str_to_int(request.GET.get("start", "1"))
 
     context = {"query": query, "start": start, "siteSearch": site, "num": num}
 
@@ -70,7 +77,7 @@ def search(request):
             context["results"] = _get_search_results(query, start, num, site)
 
             if "searchInformation" in context["results"]:
-                context["estimatedTotal"] = int(
+                context["estimatedTotal"] = str_to_int(
                     context["results"]["searchInformation"]["totalResults"]
                 )
             else:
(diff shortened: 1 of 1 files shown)


webapp/views.py
```diff
@@ -70,7 +77,7 @@ def search(request):
             context["results"] = _get_search_results(query, start, num, site)
 
             if "searchInformation" in context["results"]:
-                context["estimatedTotal"] = int(
+                context["estimatedTotal"] = str_to_int(
```
nottrobin: How could this ever not be a number?

webapp/views.py
```diff
@@ -60,8 +67,8 @@ def search(request):
 
     query = request.GET.get("q")
     site = request.GET.get("siteSearch") or request.GET.get("domain")
-    num = int(request.GET.get("num", "10"))
-    start = int(request.GET.get("start", "1"))
+    num = str_to_int(request.GET.get("num", 10), 10)
```
nottrobin: So the question here is, what is the right thing to do when someone writes an erroneous URL like `/search?num=fish`.

Technically, I think we should probably be returning a [`400 Bad Request` status](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400).

webapp/views.py
tbille: yeah here it shouldn't be an issue. I mostly did this for all of castings. It can be removed

webapp/views.py
tbille: Well here to "keep the experience" I thought that returning the default page without breaking the app.

nottrobin: But who are these people going to `/search?num=asdf`? I think this is one of two cases:

- Someone's hacking on the URL in their browser
- HTML on some site is badly formed leading to the URL being badly formed

In both cases, I think it's more helpful to convey clearly that we didn't understand the request than to simply show them the first page of results. If they're hacking on the URL it's useful for them to know they got it wrong. If their website's HTML is malformed then they're probably not linking to the page they thought they were linking to, and it would be useful if that was made clear to them when they test clicking around their website.

This is an extension of the broader point I keep trying to make that I really don't think it's that helpful to silently handle errors, in basically any context.

nottrobin: The current scenario, where https://ubuntu.com/search?q=kernel&num=10&start=asdf shows a 500, also doesn't seem to me to be the worst thing in the world. If doing the work to make it return the slightly more accurate 400 status seems like more trouble than it's worth, I would also be fine with leaving it as it is and simply "Ignore"ing the error in Sentry.

PR #5428, opened 2019-07-04
Remove un-needed front-matter

pmahnke:
removing this 
```
---
name: Default Pull Request
about: Default PR
title: ''
labels: ''
assignees: ''
---
```


solazio:
@pmahnke the PR template works now, however I get the same template for engage pages and takeovers as for any other PR. Are those supposed to be different?

PR #5444, opened 2019-07-12
Center align and max-width for wordpress figures

### LINKED ISSUE #5443: Blog figures are left aligned and too big
See

Wordpress is allowing figure tags for images with has a width of 100% and are left aligned, resulting in some strange looking images

image

image

image

-- mentioned by Thibaut Rouffineau

pmahnke:
## Done

- Center align and max-width for wordpress figures

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/blog/deploying-kubernetes-at-the-edge-part-i-building-blocks
- See that the figures are center aligned and fit in the column

## Issue / Card

Fixes #5443

## Screenshots
|Before|After|
|---|---|
|image|
image|
|image|
image|
|image|
image
|

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -242,3 +242,9 @@ abbr[title] {
   height: auto !important;
   width: 100% !important;
 }
+
+// Override for wordpress use of <figure>
+.blog-article figure {
+  max-width: 37rem;
+  text-align: center;
+}
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -245,6 +245,6 @@ abbr[title] {
 
 // Override for wordpress use of <figure>
 .blog-article figure {
-  max-width: 37rem;
+  @extend %max-width--p;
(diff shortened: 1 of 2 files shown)


static/sass/styles.scss
```diff
@@ -242,3 +242,9 @@ abbr[title] {
   height: auto !important;
   width: 100% !important;
 }
+
+// Override for wordpress use of <figure>
+.blog-article figure {
+  max-width: 37rem;
```
anthonydillon: Can you use `map-get($max-widths, default);` so it stays inline with Vanilla updates.

pmahnke: that map doesn't exist on ubuntu.com yet

deadlight: @pmahnke You're right, the map is introduced in v2. For future-proofing, I'd use `%max-width--p` and it'll get caught when upgrading to v2

PR #5637, opened 2019-08-01
Ua prototype

anthonydillon:
## Done
Prototype of the UA flow.

- Start at the /esm page
- There is a button to advantage page
- Click around and get your token


matthewpaulthomas:
Nice work!

- [ ] On both pages, “Get your free ~~developer~~ token” (it’s not just for developers)

- [ ] On /advantage, “Get your free token” and “Purchase Ubuntu Advantage” should be green

With “Personal” selected:
- [ ] I think the “Standard” and “Advanced” column _text_ should be greyed out. White-on-dark-grey does not look disabled to me.
- [ ] “Knowledge base access” should not be checked.
- [ ] After accepting the Ts & Cs, no need to provide the token in its own copyable field. People will want to copy the two commands, one at a time, so I think they belong in separate copyable fields.

matthewpaulthomas:
> I think the “Standard” and “Advanced” column text should be greyed out

I guess {opacity: 0.5} or similar would be the easiest way of greying out both the text and the checkmark graphics.

anthonydillon:
@matthewpaulthomas I've made the changes above. Let me know what you think.

matthewpaulthomas:
@anthonydillon Good progress. Most things in the list above are still to do, though. When I said “no need to provide the token in its own copyable field”, I meant, no need to provide it in its own field at all — I’m not aware of any case where it’s used other than with `ua attach`.

Besides those, I think the next most important/distracting things are, in order of importance:
- [ ] Mark asked that until buy.ubuntu.com is revised, “Purchase Ubuntu Advantage” should be replaced by a “Contact us” button — so, “[ Contact us ] about buying an enterprise subscription”
- [ ] “Knowledge base access” is now missing — sweeping this under the carpet makes it harder to judge whether Personal (that excludes knowledge base, and probably others too) really makes sense as an “Essential” subscription
- [ ] The Superhuman text on /esm
- [ ] Drop “Developer” from the terms & conditions heading (just a minor distraction)

I’d forgotten that ESM is enabled by default on attach. Good idea to point this out specifically.

anthonydillon:
@matthewpaulthomas could you point me to the T&C we should use instead of the developer ones now?

matthewpaulthomas:
@anthonydillon Sorry, I’ve never seen such terms. ubuntu.com has [terms for Livepatch but not for Ubuntu Advantage in general](https://ubuntu.com/legal/terms-and-policies). And [the “ubuntu-advantage-service-terms” copydoc](https://drive.google.com/open?id=0B2_PyaH0PaVmVVBJRHZxWFlhbkk) applies to cases where “(i) …Canonical receives a purchase order… (ii) Customer pays the Fees to Canonical, or (iii) Canonical's reseller orders the Services” — none of which applies to someone getting a free personal subscription.

I suggest replacing the developer terms with magenta lorem ipsum, to highlight that the terms haven’t been provided yet.

anthonydillon:
Also, did we decide to only link to the t&c or do we still want the popup?

matthewpaulthomas:
@anthonydillon That’s another legal question, rather than a design question, sorry. I guess we need to schedule some legal input.

If you do need to accept the terms affirmatively, it would then be an interesting question whether we should prompt for that before or after you’re signed in. I guess doing it after would be easier to implement — no need to use a separate cookie, and no need for different code to handle accepting in the case where you’re signed in already. But doing it before might result in fewer people bailing at account registration, because they’ve already invested the effort in reading and accepting the terms.

anthonydillon:
No longer required

PR #5742, opened 2019-09-02
Fix logic for http tests

pmahnke:
## Done

- Updated logic for testing for http in the change to jinja2

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/engage/apparmor-intro
- See that there is a header image

### DIFF
--- a/templates/engage/shared/_header-case-study.html
+++ b/templates/engage/shared/_header-case-study.html
@@ -6,7 +6,7 @@
       </h1>
     </div>
     {% if image %}<div class="col-5 u-align--center u-vertically-center">
-      <img src="{% if image != "http" %}https://assets.ubuntu.com/v1/{% endif %}{{ image }}" alt="image for (( title ))" >
+      <img src="{% if 'http' not in image %}https://assets.ubuntu.com/v1/{% endif %}{{ image }}" alt="image for (( title ))" >
     </div>{% endif %}
   </div>
 </section>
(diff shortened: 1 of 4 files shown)


templates/engage/shared/_header-case-study.html
```diff
@@ -6,7 +6,7 @@ <h1 class="u-no-margin--bottom">
       </h1>
     </div>
     {% if image %}<div class="col-5 u-align--center u-vertically-center">
-      <img src="{% if image != "http" %}https://assets.ubuntu.com/v1/{% endif %}{{ image }}" alt="image for (( title ))" >
+      <img src="{% if 'http' not in image %}https://assets.ubuntu.com/v1/{% endif %}{{ image }}" alt="image for (( title ))" >
```
nottrobin: @pmahnke my preference here would be to remove the conditional entirely and make `{{ image }}` always of the same format - either with or without the domain prefix. Probably just making it contain the full URL makes most sense.

```suggestion
      <img src="{{ image }}" alt="image for (( title ))" >
```

pmahnke: I think this would break pages on the site and would require a clean-up... if I understand what you are saying.

nottrobin: yeah I'm suggesting this is done in this PR, but if you want to merge this first and do that work as a follow-up that's fine too.

PR #5745, opened 2019-09-03
Restructure views

nottrobin:
Tidy up `app.py` to make it more readable by exporting views into
`views.py` and `login.py`.

QA
--

Check things still work - `/`, `/blog`, `/advantage`.

### DIFF
--- a/webapp/login.py
+++ b/webapp/login.py
@@ -1,0 +1,68 @@
+# Packages
+import flask
+import flask_openid
+import requests
+from pymacaroons import Macaroon
+
+# Local
+from webapp.macaroons import MacaroonRequest, MacaroonResponse
+
+
+open_id = flask_openid.OpenID(
+    stateless=True, safe_roots=[], extension_responses=[MacaroonResponse]
+)
+
+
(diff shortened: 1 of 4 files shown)


webapp/login.py
```diff
@@ -0,0 +1,68 @@
...
+def logout():
+    if "openid" in flask.session:
+        flask.session.pop("macaroon_root", None)
+        flask.session.pop("macaroon_discharge", None)
+        flask.session.pop("openid", None)
+        flask.session.pop("user_shared_snaps", None)
+
+    return flask.redirect(
+        "https://login.ubuntu.com/+logout"
+        f"?return_to={flask.request.url_root}/advantage&return_now=True"
```
anthonydillon: Could you use request to return to the page you were on. This would make logout work across the entire site.

nottrobin: Yeah let's do that.

nottrobin: Of course if we have a page that automatically kicks you over to login, then you'll end up in a loop. But we'll just have to handle that case when it comes up.

PR #5769, opened 2019-09-05
Ubuntu Advantage dashboard

nottrobin:
Demo of the `advantage` feature branch, containing the work for the new `/advantage` section.


anthonydillon:
### Screenshots
Signed out
Screenshot_2019-10-08 Ubuntu Advantage for Infrastructure Ubuntu

Signed in with paid subscriptions
Screenshot_2019-10-08 Ubuntu Advantage for Infrastructure Ubuntu(1)

Signed in with no paid subscriptions
Screenshot_2019-10-08 Ubuntu Advantage for Infrastructure Ubuntu(2)

mhilton:
I'm seeing a KeyError when accessing the demo
```
builtins.KeyError

KeyError: 'accounts'
Traceback (most recent call last)

    File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 2295, in wsgi_app

    response = self.handle_exception(e)

    File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1741, in handle_exception

    reraise(exc_type, exc_value, tb)

    File "/usr/local/lib/python3.6/dist-packages/flask/_compat.py", line 35, in reraise

    raise value

    File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 2292, in wsgi_app

    response = self.full_dispatch_request()

    File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1815, in full_dispatch_request

    rv = self.handle_user_exception(e)

    File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1718, in handle_user_exception

    reraise(exc_type, exc_value, tb)

    File "/usr/local/lib/python3.6/dist-packages/flask/_compat.py", line 35, in reraise

    raise value

    File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1813, in full_dispatch_request

    rv = self.dispatch_request()

    File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1799, in dispatch_request

    return self.view_functions[rule.endpoint](**req.view_args)

    File "/srv/run.demo.haus-demos/ubuntu-com-canonical-web-and-design-pr-5769.run.demo.haus/webapp/views.py", line 75, in advantage

    ).json()["accounts"]

    KeyError: 'accounts'
```

This looks to be happening when receiving a 401 error from the API 
```
{"level":"info","ts":"2019-10-08T17:03:54.316Z","msg":"invalid macaroon: verification failed: macaroon not found in storage","method":"GET","path":"/v1/accounts","trace-id":"vZQqGJPuqjPxCWH-S1UlIlBOFA1OcJvX","status":401}
```

invalid credentials should probably be handled more gracefully.

anthonydillon:
@mhilton you need to clear your SSO cookie as its linked with prod. This is just while we are switching.

mhilton:
@anthonydillon I still think that an error response should be handled more gracefully than dumping a stack trace on the user. you could just ditch the cookie and return to a logged out state.

nottrobin:
@mhilton you won't see the stack trace in staging or production, but the demos (currently) run in development mode. In production you'd see a prettier 500 page.

I'm not sure logging someone out is the most logical response to receiving a 401 from the API. I've received a number of 401 responses before for endpoints that simply weren't enabled yet.

If the server, after believing you logged in, gets an unexpected response from the API then a server error seems logical. If, however, a 401 is absolutely an *expected* response then we should do something more intelligent than simply logging the user out, which could be very surprising.

nottrobin:
@mhilton under what circumstances should we expect a 401, apart from you having been previously logged in to staging (which is a weird edge case only relevant in local development)? Will we get that if e.g. your API thinks the macaroon has expired?

mhilton:
@nottrobin thanks for clarifying that the stack traces are in debug mode only. You can expect a 401 response (when you've sent a macaroon) if the macaroon is expired (the timeout is currently 1 week) or if the macaroon is considered invalid because it wasn't issued by the contract server or if it's been tampered with. These two are unlikely. Also possibly if it's not valid for the operation being performed but that would be unlikely from your current page.

pmahnke:
My visual comments

### Logged in

image

- remove the box and as much text on the right sign-in bits
- remove the --light 
- right align the table headings for livepatch, esm, etc...
- move the t&c link below the box and add * superscript
- add a nbsp between 'for' and 'Infrastructure' 
- add a is-border to the Free personal subscription box

#### Alt view

image

#### Alt view 2

image

nottrobin:
@mhilton OK so the important case to handle is the possibility of the macaroon having expired. Do you know if it's possible to renew the macaroon without sending the user back to login.ubuntu.com? As in, is it possible to request a new macaroon and bind it with an existing discharge from login.ubuntu.com? Or does login.ubuntu.com need to be the party performing the binding?

mhilton:
@nottrobin each third-party caveat has it's own caveatID that is encrypted with login.ubuntu.com's public key. So the binding happens in the client but the discharge must be created separately on login.ubuntu.com. If you have a compelling reason / A better justification that my (a week sounds about right) to change the default macaroon timeout just say and we can change it.

nottrobin:
@mhilton I don't think there's anything wrong with a week, but presumably whatever the limit, there's a possibility of it having expired. I was just wondering if it's possible to renew without a new round-trip (as long as your auth from login.ubuntu.com is still valid). but if not, that's fine, I'll just send the user through another login.ubuntu.com loop if the macaroon has expired.

nottrobin:
@mhilton is there any way for me to tell the difference between an expired macaroon 401, vs another type of 401 response from the API?

mhilton:
@nottrobin not from the API, it is theoretically possible to decode the macaroon caveats and check them for an expiry time. But I'm not certain that's worth the hassle. I can help you if you would like to do that.

nottrobin:
@mhilton my main thought is that if we get a `401` from the API, it seems not outside the bounds of possibility that it might be because something changed in the API to make an endpoint unavailable to the public, rather than that the user's macaroon expired.

I really don't want to send the user into a login loop when the actual problem is that the API is returning the wrong response. In that scenario, the right solution is definitely to throw a 5xx server error, which would mean we'd be able to pick it up in sentry etc..

nottrobin:
@mhilton Basically I think I would like to know the difference between "we can't verify who you are" and "you don't have access to this". Which might just be the difference between 401 and 403? But I think we would ideally have some mechanism or other.

anthonydillon:
## Screenshots
WIth subscriptions
Screenshot_2019-10-11 Ubuntu Advantage for Infrastructure Ubuntu(1)

PR #5772, opened 2019-09-06
Improve logout

nottrobin:
Logout doesn't redirect back to the site properly.

Here I've changed the logic of the `/logout` view, so it also can accept a `return_to` paramter, and so the "Sign out" link on `/advantage` now passes a `return_to` parameter to bring you back to `/advantage`.

This is because in general I don't think it's sensible for `/logout` to send you back to the page you came from. In many cases the page you came from would only be available with authentication, and so you'd end up in a redirect loop.

If you want the user to come back to the existing page, you should explicitly set `return_to`, otherwise the user will be sent to the homepage, which I think is sensible.

QA
--

Don't QA this yet.

### DIFF
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -21,7 +21,7 @@
     <div class="col-10">
       <h3>You are signed in</h3>
       <p>{{ openid.fullname }} ({{ openid.email }})</p>
-      <a href="/logout" class="p-button--neutral u-no-margin--bottom">Sign out</a>
+      <a href="/logout?return_to={{ request.host_url }}advantage" class="p-button--neutral u-no-margin--bottom">Sign out</a>
     </div>
   </div>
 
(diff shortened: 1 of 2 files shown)


nottrobin:
> @anthonydillon ... I would expect any page that requires auth would display as a not logged in view

For any system that relies on auth to any significant degree this doesn't tend to be true. There might be a landing page that introduces you to the authed section, but there are usually plenty of pages that are only available to logged-in users, and visiting those pages would prompt you to sign in. E.g. https://github.com/settings/profile or https://snapcraft.io/snaps.

I really think this is vastly superior as a default model.

nottrobin:
@anthonydillon I suppose we could instead make the interface something like `/logout?return=True`, if we don't think there'd ever be a case where you want to return somewhere else?

But I'm anticipating the possible existence of something like `/advantage/my-tokens`, where you'd want the same "logout" button, which would return you to `/advantage`.

anthonydillon:
@nottrobin Yeah that makes sence :+1:

PR #5786, opened 2019-09-10
Move homepage blog feed client-side

nottrobin:
This uses the `/blog/latest-news` JSON endpoint provided by the blog
module, and relies on a couple of fixes in v3.1.1:
https://github.com/canonical-web-and-design/canonicalwebteam.blog/pull/122

Fixes https://github.com/canonical-web-and-design/base-squad/issues/714

QA
--

Check the homepage blog feed works as expected.


templates/takeovers/templates.html
```diff
@@ -88,11 +88,7 @@ <h3>Details</h3>
 {% include "takeovers/_template.html" %}
 {% endwith %}
 
-<section class="p-strip is-deep is-bordered">
```
nottrobin: I removed the blog lists from this page because they seem unnecessary, and it's tricky to make multiple blog lists work on the same page with JavaScript & includes - 'cos we'd have to have unique IDs for the elements, which is troublesome and not worth solving.

templates/base_index.html
```diff
@@ -28,9 +28,35 @@ <h1 style="font-weight: 100;">Ubuntu {{ releases.latest.short_version }} is here
     {% endblock notices_content %}
 
     <section class="p-strip is-deep is-bordered">
-      {% with gtm_event_label="ubuntu.com homepage" %}
-        {% include "shared/_insights_news_strip.html" %}
```
nottrobin: Deleted the `_insights_news_strip` include because the homepage seems to be the only places really using it (`takeovers/templates` doesn't count).

static/js/latest-news.js
```diff
@@ -0,0 +1,63 @@
+(function() {
```
nottrobin: This is copied from cn.ubuntu.com, and then heavily modified. For one thing, I made it IE11 compatible by removing `s and `=>` functions, and I wrapped it in an anonymous function, and made it so we could pass through the gtmEventLabel.

templates/base_index.html
```diff
@@ -28,9 +28,35 @@ <h1 style="font-weight: 100;">Ubuntu {{ releases.latest.short_version }} is here
...
-      {% endwith %}
+      <div class="row p-divider">
+        <div class="col-9">
+          <h3>Latest news from our blog</h3>
+          <div id="latest-articles" class="row">
+            <div style="min-height: 9.1rem"><img src="https://assets.ubuntu.com/v1/2eb3f9f4-spinner-big.gif"></div>
```
anthonydillon: Missing alt on the image and the min-height could be quite large on small screens

anthonydillon: Please use Vanilla loading icon and animation:
```html
<i class="p-icon--spinner u-animation--spin">Loading...</i>
```

static/js/latest-news.js
```diff
@@ -0,0 +1,63 @@
...
+  function _articleDiv(article, gtmEventLabel) {
```
anthonydillon: Your not using the `gtmEventLabel` param here. Do you need it? It seems to be hard coded in the template.

nottrobin: I have now just implemented it, sorry.

static/js/latest-news.js
```diff
@@ -0,0 +1,90 @@
...
+      const containerForLatestNews = document.querySelector(latestNewsContainer);
+
+      // Clear out latest news container
+      while (containerForLatestNews.hasChildNodes()) {
+        containerForLatestNews.removeChild(containerForLatestNews.lastChild);
+      }
+
+      const data = JSON.parse(event.target.responseText);
+
+      const latestPinned = data.latest_pinned_articles[0];
```
anthonydillon: This will error if offline as the error response will not contain `latest_pinned_articles`

anthonydillon:
Don't we want error notifications?

templates/base_index.html
nottrobin: Thanks I've used that Vanilla spinner.

But the min-height is because I want to try to prevent the page jumping. On large screens, this height is the same height as smallest the row will be (it might be bigger if some article titles flow onto 4 rows). On small screens, this min-height will be much less than the total height when the articles arrive, so it doesn't seem unreasonable to reserve this amount of space.

static/js/latest-news.js
```diff
@@ -0,0 +1,90 @@
...
+    div.innerHTML = header + date;
+
+    div.querySelector('a').onclick = function() {
+      dataLayer.push(
+        {
+          'event': 'GAEvent',
+          'eventCategory': 'blog',
+          'eventAction': gtmEventLabel + ' news link',
+          'eventLabel': _slugify(article.title),
+          'eventValue': article.link
```
anthonydillon: eventValue is the numerical value of the click like $1. The label should cover the tracking.

static/js/latest-news.js
nottrobin: If offline...

nottrobin: What would you expect to happen other than an error?

nottrobin: I deliberately removed the `try .. catch` statements [here](https://github.com/canonical-web-and-design/cn.ubuntu.com/blob/85057516f6f27f02325f948ceeb580fe73c20781/static/js/latest-news.js#L96) as they simply catch an error to then send a different error alert. They have no idea which error they're catching and masking.

If I was anyone - a user, a developer - and I knew about JavaScript errors, I would find the native error much more useful.

static/js/latest-news.js
nottrobin: What should I do? Remove `eventValue`? Or return some property from the  `event` object?

anthonydillon: Oh just leave it `undefined`

nottrobin: Ok done

static/js/latest-news.js
nottrobin: I guess we *could* actually update the text the user sees. That might have value, but I'd like to suggest we don't bother in this first pass. Let's see how often it breaks. If it's often then we could consider giving users more info.

PR #5787, opened 2019-09-10
Fix the default suru on small screens

anthonydillon:
## Done
Add a background on small screens.

## QA
- On Chrome on your phone check the /internet-of-things page

### DIFF
--- a/static/sass/_pattern_strip.scss
+++ b/static/sass/_pattern_strip.scss
@@ -56,6 +56,7 @@
 
       @media (min-width: $breakpoint-medium) {
         $padding: 6rem;
+        background-image: linear-gradient(-89deg, #e95420 0%, #772953 85%);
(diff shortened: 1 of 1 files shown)


static/sass/_pattern_strip.scss
```diff
@@ -24,6 +24,7 @@
   .p-strip--suru {
     @extend %vf-strip;
     background-blend-mode: multiply, multiply, normal, normal;
+    background-color: #772953;
```
sowasred2012: Just checked this out on Chrome mobile, looks good, but it makes the suru on large screens look a lot darker - is that intended?

anasereijo:
same comment as @sowasred2012 regarding colours.

static/sass/_pattern_strip.scss
pmahnke: perhaps do it only in a media query

PR #5788, opened 2019-09-10
Update the readme and hacking pages

pmahnke:
## Done

- Updated the readme and the hacking pages for github users

## QA

- look at the files, perhaps hit the ... and try and look at a preview.
- the key thing is the text... does it make sense, it is something helpful and usefule.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,6 +1,8 @@
 # Working on the ubuntu.com project
 
-The ubuntu.com codebase is a [Flask](https://flask.palletsprojects.com/en/1.1.x/) app, which also includes some [Yarn](https://yarnpkg.com/lang/en/)-based tools and tasks for building static files like CSS.
+The ubuntu.com codebase is a [Flask](https://flask.palletsprojects.com/) app, which builds on our own [flask-base](https://pypi.org/project/canonicalwebteam.flask-base/), [templatefinder](https://pypi.org/project/canonicalwebteam.templatefinder/), [blog](https://pypi.org/project/canonicalwebteam.blog/) and [search](https://pypi.org/project/canonicalwebteam.search/) packages.
+
+We use [Yarn](https://yarnpkg.com/lang/en/) for building static files like CSS through [`package.json` scripts](https://yarnpkg.com/lang/en/docs/package-json/#toc-scripts).
 
 ## Running the site
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,6 +1,6 @@
 # Working on the www.ubuntu.com project
 
-The www.ubuntu.com codebase is a [Django](https://www.djangoproject.com/) app, which also includes some [NPM](https://www.npmjs.com/)-based tools and tasks for building static files like CSS.
+The www.ubuntu.com codebase is a [Flask](https://flask.palletsprojects.com/en/1.1.x/) app, which also includes some [Yarn](https://yarnpkg.com/lang/en/)-based tools and tasks for building static files like CSS.
 
 ## Running the site
 
@@ -23,6 +23,8 @@ The basic options are:
 ./run build  # Build the CSS
 ./run watch  # Watch and build the CSS whenever Sass changes
 ./run clean  # Remove created files and docker containers
+./run stop   # Stop any running containers
+
 ```
 
 #### Watching in the background
@@ -41,10 +43,10 @@ To check if the watcher daemon is running, use `docker ps`. Then you can use `do
 
 ### Running the site with native python
 
-Since the site is basically a Django app, you can also run the site in the traditional way using [python 2.7](https://www.python.org/download/releases/2.7/) and [virtualenv](http://docs.python-guide.org/en/latest/dev/virtualenvs/):
+Since the site is basically a Flask app, you can also run the site in the traditional way using [python 3](https://docs.python.org/3/) and [venv](https://docs.python.org/3/library/venv.html?highlight=venv#module-venv):
 
 ``` bash
-virtualenv env
+python3 -m venv env
 source env/bin/activate
 pip install -r requirements.txt
 ./entrypoint 0.0.0.0:8001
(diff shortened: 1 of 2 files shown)


HACKING.md
```diff
@@ -41,50 +43,47 @@ To check if the watcher daemon is running, use `docker ps`. Then you can use `do
...
 
 ``` bash
-virtualenv env
+python3 -m venv env
 source env/bin/activate
 pip install -r requirements.txt
 ./entrypoint 0.0.0.0:8001
 ```
 
 Now browse to the site at <http://127.0.0.1:8001>. If it looks a bit odd, it's probably because you haven't built sass - see below.
```
solazio: Is it not http://0.0.0.1:8001 ?

pmahnke: well, I don't think it matters, but for consistencies sake, they should be the same.

HACKING.md
```diff
@@ -23,6 +23,8 @@ The basic options are:
 ./run build  # Build the CSS
 ./run watch  # Watch and build the CSS whenever Sass changes
 ./run clean  # Remove created files and docker containers
+./run stop   # Stop any running containers
```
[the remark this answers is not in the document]
pmahnke: @nottrobin 	?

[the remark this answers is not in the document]
nottrobin: Just ones for this project

HACKING.md
```diff
@@ -41,50 +43,47 @@ To check if the watcher daemon is running, use `docker ps`. Then you can use `do
...
-# or
 yarn run build  # Build the Sass to CSS then exit
 yarn run watch  # Dynamically watch for Sass changes and build CSS
 ```
 
-### Overriding NPM modules
+### Overriding Yarn modules
 
-You can use the `./run` script to use NPM modules from a local folder on a one-time basis, instead of the modules declared in `package.json`, as follows:
+You can use the `./run` script to use Yarn NPM modules from a local folder on a one-time basis, instead of the modules declared in `package.json`, as follows:
```
nottrobin:
```suggestion
You can use the `./run` script to use Yarn modules from a local folder on a one-time basis, instead of the modules declared in `package.json`, as follows:
```

HACKING.md
```diff
@@ -41,50 +43,47 @@ To check if the watcher daemon is running, use `docker ps`. Then you can use `do
...
 
 ``` bash
-./run --node-module $HOME/projects/vanilla-framework watch  # Build CSS dynamically, using a local version of vanilla-framework
+./run [-r|--root] [-p|--expose-port PORT] <script-name>
```
nottrobin: Why change this? I think the original version (`--node-module`) works and is what you want here.

HACKING.md
```diff
@@ -114,7 +113,7 @@ If a child is "hidden", then it won't be displayed in the navigation menus, eith
 
 #### How it works
 
-The `navigation.yaml` file is read in `webapp/settings.py` to create a Django settings objects `django.conf.settings.NAV_SECTIONS`.
+The `navigation.yaml` file is read in `webapp/settings.py` to create a Flask settings objects `django.conf.settings.NAV_SECTIONS`.
```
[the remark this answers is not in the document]
nottrobin: `webapp/context.py`

HACKING.md
solazio: @nottrobin thanks for the explanations. I didn't know that. Every day is a school day :+1:

README.md
```diff
@@ -1,12 +1,18 @@
+![ubuntu.com](https://assets.ubuntu.com/v1/9f61b97f-logo-ubuntu.svg "Ubuntu")
```
nottrobin: Wow I'd never seen this syntax before. Cool.

HACKING.md
nottrobin: Just to add a little more flavour - the practical use for `0.0.0.0` is for when you want to tell a server which requests to listen for. If you tell it to listen at `127.0.0.1` then it will only listen for incoming requests with `Host: 127.0.0.1` headers. If you tell it to listen at `0.0.0.0` then it will listen for incoming requests with any `Host` header.

This is why we run Talisker with `--bind 0.0.0.0:8001`, and why what gets output on the command-line is `http://0.0.0.0:8001`. And the fact that that pattern is so common in local development is why browsers tend to assume that people visiting `http://0.0.0.0` probably want the local machine.

HACKING.md
```diff
@@ -41,50 +43,47 @@ To check if the watcher daemon is running, use `docker ps`. Then you can use `do
...
-Since the site is basically a Django app, you can also run the site in the traditional way using [python 2.7](https://www.python.org/download/releases/2.7/) and [virtualenv](http://docs.python-guide.org/en/latest/dev/virtualenvs/):
+Since the site is basically a Flask app, you can also run the site in the traditional way using [python 3](https://docs.python.org/3/) and [venv](https://docs.python.org/3/library/venv.html?highlight=venv#module-venv):
 
 ``` bash
-virtualenv env
+python3 -m venv env
 source env/bin/activate
 pip install -r requirements.txt
-./entrypoint 0.0.0.0:8001
+./entrypoint 127.0.0.0:8001
```
nottrobin: This needs to be either `./entrypoint 127.0.0.1:8001` or `./entrypoint 0.0.0.0:8001`. I'd recommend the latter just 'cos that's what we use in `./run`.

nottrobin:
"Design needed"?

HACKING.md
```diff
@@ -1,6 +1,6 @@
-# Working on the www.ubuntu.com project
+# Working on the ubuntu.com project
 
-The www.ubuntu.com codebase is a [Django](https://www.djangoproject.com/) app, which also includes some [NPM](https://www.npmjs.com/)-based tools and tasks for building static files like CSS.
+The ubuntu.com codebase is a [Flask](https://flask.palletsprojects.com/en/1.1.x/) app, which also includes some [Yarn](https://yarnpkg.com/lang/en/)-based tools and tasks for building static files like CSS.
```
nottrobin:
```suggestion
The ubuntu.com codebase is a [Flask](https://flask.palletsprojects.com/en/1.1.x/) app, which builds on our own [flask-base](https://pypi.org/project/canonicalwebteam.flask-base/), [templatefinder](https://pypi.org/project/canonicalwebteam.templatefinder/), [blog](https://pypi.org/project/canonicalwebteam.blog/) and [search](https://pypi.org/project/canonicalwebteam.search/) packages.

We use [Yarn](https://yarnpkg.com/lang/en/) for building static files like CSS through [`package.json` scripts](https://yarnpkg.com/lang/en/docs/package-json/#toc-scripts).
```

anasereijo:
LGTM

PR #5818, opened 2019-09-18
Initial commit of new docs section

pmahnke:
## Done

- created a new docs section
- wrote a new strips doc
- wrote engage markdown doc
- updated takeovers and old engage headings fit style

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/_docs
- These are living docs... please read and focus if they will be helpful or not... mor than the exact content


anasereijo:
Thanks for doing this, I think it will be very useful especially after we populate it with more content and explanations. 
I left some visual things to change and also a few suggestions that we can discuss together.

### Navigation

- **Strips**: not sure if this name will lead people to think the new suru strips can be used down the pages. They were designed to be used as hero strips (hero banners or heading strips - not sure what's the best name yet). That said, I suggest changing the name. On the other hand, we will have suru strips to use down the pages soon (not sure when though) so maybe then naming this section "Strips" will make sense. >>> _we can discuss this_

- [ ]  **Engage pages vs Engage headings**: maybe this can be just one section called "Engage pages"

- [ ] **Order of links**: In my opinion it makes more sense having the following order: Overview / Strips / Takeovers / Engage pages

### Overview

- [ ] Use this suru strip instead: `class="p-strip--suru is-dark is-deep"`

- [ ] remove the list of links. We can have a more visual section instead >>> _to be designed_

- Would be nice to have some intro text explaining the purpose of this section. >>> _we can discuss this_

### Takeovers, Suru strips, Engage pages

I have a few ideas, let's discuss this in person as it easier to explain.

anasereijo:
- [ ] `p-strip--suru`: This strip is meant to be used only on overview pages (or main ones). That said, I suggest changing the copy to something like: "This strip should be used on overview/main pages and is meant to have an illustration on the right side."

- [ ] `p-strip--suru-topped`: the illustration doesn't seem to be positioned correctly... Should be vertically centred with the text.
Can we have some text: "This strip should be followed by a light-grey section or at least have a border to indicate a separation between the heading and the next section."

- [ ] `p-strip--suru-half-and-half`: there's an issue with this strip - it should have a bottom padding, Ant fixed it could you please update it?
copy says "This strip should have an image on the right." but not necessarily, so we can say "This strip can have an illustration on the right."

- [ ] `p-strip--suru-half-and-half-reversed`: is this the same template as this [one](https://ubuntu.com/desktop/organisations)? If so we could say: "This strip has room for an image or an illustration." and show both options.

- [ ] `p-strip-suru-half-top`: should we say: "This strip should be used on overview/main pages and is meant to be followed by a light-grey section."

- [ ]` p-strip--suru-background`: "This strip should be used on overview/main pages and expects the use of a background image."

- [ ] `p-strip--suru-image`: "An image or illustration can be used on the right. It also requires a fair bit of height to look right."

I was also thinking it might be nicer to have the overview ones first and then the secondary ones. Could be: p-strip--suru, p-strip-suru-half-top, p-strip--suru-background, and then the other ones.



 

- [ ]

pmahnke:
@anasereijo  updated

anasereijo:
Looking good, thanks @pmahnke !

on the `p-strip--suru-half-and-half` the bottom padding is still missing. I thought @anthonydillon fixed it

pmahnke:
@anasereijo regarding `p-strip--suru-half-and-half` might be fixed when this is merged.

PR #5821, opened 2019-09-19
Update og metadata and current url"

pmahnke:
## Done

- added prefix to all og tags
- updated the canonical url and the og:url

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/ and other pages
- See that all og: meta data has a prefix and the canonical and og:url is correct

### DIFF
--- a/templates/templates/base.html
+++ b/templates/templates/base.html
@@ -1,6 +1,6 @@
 <!doctype html>
 
-<html class="no-js{% block extra_html_class %}{% endblock %}" lang="en" dir="ltr">
+<html prefix="og: http://ogp.me/ns#" class="no-js{% block extra_html_class %}{% endblock %}" lang="en" dir="ltr">
   <head>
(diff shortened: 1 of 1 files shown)


templates/templates/base.html
```diff
@@ -25,17 +25,17 @@
     <meta name="theme-color" content="#E95420">
     <meta name="twitter:account_id" content="4503599627481511">
     <meta name="twitter:site" content="@ubuntu">
-    <meta property="og:type" content="website">
-    <meta property="og:url" content="https://ubuntu.com{{ request.get_full_path }}">
-    <meta property="og:site_name" content="Ubuntu">
+    <meta prefix="og: http://ogp.me/ns#" property="og:type" content="website">
```
nottrobin: I believe you can put this namespace on the `<html>` element, can't you? https://ogp.me/

PR #5878, opened 2019-10-02
Address advantage feedback

anthonydillon:
## Done
Addressed some of the feedback on the advantage page.

## QA
- Go to /advantage
- See the page looks ok
- Login and check that works
- Check the page looks ok

### DIFF
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -550,65 +550,49 @@
 {% endif %}
 
 {% else %}
-<section class="p-strip--suru is-dark">
+<section class="p-strip--suru is-dark" style="padding-bottom: 10rem !important;">
   <div class="row">
-    <div class="col-8">
+    <div class="col-6">
       <h1>Ubuntu Advantage for Infrastructure</h1>
       <p>Ubuntu Advantage for Infrastructure is the most comprehensive Linux enterprise subscription, covering all aspects of open infrastructure.</p>
     </div>
-  </div>
-</section>
-
-<section class="p-strip is-shallow">
-  <div class="row u-equal-height">
-    <div class="col-12 u-vertically-center">
-      <h2>Already subscribed?</h2>
+    <div class="col-6 p-card">
+      <h2 class="p-heading--four">Sign in or register</h2>
       <p class="u-no-margin--bottom">
-        <a href="/login" class="p-button--positive" onclick="dataLayer.push({
-          'event' : 'GAEvent',
-          'eventCategory' : 'Advantage',
-          'eventAction' : 'Authentication',
-          'eventLabel' : 'Sign in',
-          'eventValue' : undefined
-        });">Sign in</a> to access your personal or paid subscriptions.
+        <a href="/login" class="p-button--positive" style="width: 6rem;"
+          onclick="dataLayer.push({
+            'event' : 'GAEvent',
+            'eventCategory' : 'Advantage',
+            'eventAction' : 'Authentication',
+            'eventLabel' : 'Sign in',
+            'eventValue' : undefined
+          });">Sign in</a> to access your personal or paid subscriptions.
+      </p>
+      <p class="u-no-margin--bottom">
+        <a href="/login" class="p-button--neutral"  style="width: 6rem;"
+          onclick="dataLayer.push({
+            'event' : 'GAEvent',
+            'eventCategory' : 'Advantage',
+            'eventAction' : 'Authentication',
+            'eventLabel' : 'Register',
+            'eventValue' : undefined
+          });">Register</a> to get your free subscription.
       </p>
     </div>
   </div>
 </section>
 
-<section class="p-strip is-shallow">
-  <div class="row">
-    <div class="col-12">
-      <h2>Free for personal use</h2>
-      <p>Anyone can use UA Infrastructure Essential for free on up to 3 machines (limitations apply).  All you need is an Ubuntu One account. And if you’re a recognised <a href="https://wiki.ubuntu.com/Membership" class="p-link--external">Ubuntu community member</a>, it’s free on up to 50 machines.</p>
-      <p class="u-no-margin--bottom">
-        <a href="/login" class="p-button--neutral" onclick="dataLayer.push({
-          'event' : 'GAEvent',
-          'eventCategory' : 'Advantage',
-          'eventAction' : 'Authentication',
-          'eventLabel' : 'Sign in',
-          'eventValue' : undefined
-        });">Register</a> to get your free subscription.
-      </p>
-    </div>
-  </div>
-</section>
-
-<section class="p-strip is-shallow is-bordered">
+<section class="p-strip is-shallow is-bordered u-no-padding--top">
   {% include "advantage/_table.html" %}
 </section>
 {% endif %}
 
-<section class="p-strip">
-  <div class="row u-equal-height">
-    <div class="col-6 u-align--center u-hide--small">
-      <img src="https://assets.ubuntu.com/v1/9c7bda7d-contact-us.svg" alt="">
-    </div>
-    <div class="col-6 u-vertically-center">
-      <div>
-        <h2>Got questions?</h2>
-        <p class="u-no-margin--bottom"><a href="/support/contact-us" class="p-button--positive js-invoke-modal">Contact us</a></p>
-      </div>
+<section class="p-strip--light is-shallow">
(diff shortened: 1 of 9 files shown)


anthonydillon:
## Screenshots
Revealing a service
Screenshot_2019-10-03 Ubuntu Advantage for Infrastructure Ubuntu

Many personal contracts
Screenshot_2019-10-03 Ubuntu Advantage for Infrastructure Ubuntu(1)

matthewpaulthomas:
[To-do list](https://github.com/CanonicalLtd/Ubuntu-design/issues/190).

templates/advantage/index.html
```diff
@@ -226,318 +140,38 @@ <h2>Your enterprise subscriptions</h2>
...
-                  'eventValue' : undefined
-                });">Copy to clipboard</button>
-              </div>
-            </section>
-          </li>
-        </ul>
-      </aside>
+        <p>To attach a machine:</p>
+        <pre style="border: 0; padding: 0;"><code>sudo ua attach {{ personal_account.free_token }}</code></pre>
+        <p>For further options <a href="">visit the ua manpage</a>.</p>
```
matthewpaulthomas: “Further” suggests that nothing more will be mentioned here. So move this line after the instructions for individual services.

templates/advantage/index.html
```diff
@@ -226,318 +140,38 @@ <h2>Your enterprise subscriptions</h2>
...
-            </div>
-          </div>
-        </li>
-        {% endif %}
-      </ul>
+      {% if personal_account['entitlements']['livepatch'] %}
+      <div class="row">
+        <div class="col-1">Livepatch:</div>
+        <div class="col-6"><pre style="border: 0; padding: 0;"><code>sudo ua enable livepatch</pre></code></div>
+      </div>
```
matthewpaulthomas: This is a table. Any reason not to use `<tr>` and `<td>`? And put the `attach` line in the same layout.

templates/advantage/index.html
```diff
@@ -550,65 +184,49 @@ <h4>EAL2</h4>
...
-    <div class="col-6 u-vertically-center">
-      <div>
-        <h2>Got questions?</h2>
-        <p class="u-no-margin--bottom"><a href="/support/contact-us" class="p-button--positive js-invoke-modal">Contact us</a></p>
-      </div>
+<section class="p-strip is-shallow">
+  <div class="row">
+    <div class="col-8">
+      <h2>Got questions?</h2>
+      <p>Let our cloud experts help you get started in the right direction with our Ubuntru Advantage offering.</p>
```
matthewpaulthomas:
```suggestion
      <p>Let our experts help you in the right direction with Ubuntu Advantage.</p>
```

templates/advantage/index.html
```diff
@@ -7,18 +7,19 @@
...
 
 {% if openid %}
-<section class="p-strip--suru-topped">
+<section class="p-strip--suru-topped is-shallow" style="padding-top: 6rem">
   <div class="row">
-    <div class="col-12">
-      <h1 class="u-no-max-width">Ubuntu Advantage for Infrastructure</h1>
+    <div class="col-6">
+      <h1>Ubuntu Advantage for Infrastructure</h1>
+      <p>By using Ubuntu Avantage you agree to our <a href="/legal/ubuntu-advantage-service-description">terms and conditions</a>.</p>
```
matthewpaulthomas: This statement is necessary, but makes sense mainly as something you review just before you get your free token. Which tells me:
- we still need a one-off button for that specific purpose (“Get my free token”)
- this statement should be above it.

matthewpaulthomas:
Ok, here’s my laundry list 🧺 for achieving the goals of “denser and classier”. ✨

- [ ] Move the account widget back below the intro paragraph, but drop the “Already subscribed?” heading, merging those words into the single line:
  > [ Sign in ] if you’re subscribed already, to see your machine status and tips.

  (I really don’t think the account widget works in a floated box, because (a) it’s secondary to the main heading, and (b) the account state controls what’s across the full width of the page below it. This change instead makes it more compact.)

- [ ] Restore the “Free for personal use” section. (Most people coming to this page will be coming here specifically for that, so we can’t just not have it!)

- [ ] Make “By using Ubuntu Avantage you agree to our terms and conditions” appear only if (a) I’m signed in _and_ (b) I don’t yet have a subscription. (That makes the page shorter and less legalistic in every other case. People can still get to the Ts&Cs from the contextual footer.)

- [ ] Prepare a variant with instructions embedded in each paid-sub cell, and a variant with expandable cells like we have now.

- [ ] Add an “Attached” column after the “UA” column, with the number of attached machines. Each “Attached” cell has instructions (embedded or expanded, depending on variant) for attaching to that subscription, with the relevant token.

- [ ] In the variant with expandable cells, expand the first “Attached” by default. (This makes it much less likely you’ll need to expand anything at all, and if you will, it hints at where to find it.)

- [ ] No background tint for the instructions in the paid subs table.

- [ ] Move “Add another subscription” back opposite the “Your enterprise subscriptions” heading where it used to be. (Sorry @cassiocassio, you were logically right, but now reducing scrolling is more important.)

- [ ] In the “Your free personal subscription” section, make the instructions a table that lines up exactly with the “Your enterprise subscriptions” table above.

- [ ] Restore the Copy button for instructions, but classy: no noisy border, no `$` prompt, consisting only of a simple icon image that appears only on mouseover.

- [ ] “Let our experts help you in the right direction with Ubuntu Advantage.” (I will never understand why anyone thinks calling something an “offering” sounds good.)

- [ ] Make “Contact us” a neutral button. (So we don’t have two primary buttons on the same page.)

- [ ] Make “Got questions?” the first section of the contextual footer, moving Landscape into the second section.

- [ ] Make the second section of the footer “Legal info” (“• Terms and conditions” “• Ubuntu Advantage Service Description”), so that people can still access the Ts&Cs after accepting them.

- [ ] Make the third section “Further reading” that is [actually about Ubuntu Advantage](https://ubuntu.com/blog/tag/ubuntu-advantage), rather than just “Random reading”.

cassiocassio:
- I feel that we have to at least try the floated box top right, to show alongside any preferred alternative solution, so good for @anasereijo to explore.

- Repetition: if the UA column reads "Essential" and the name of the subscription, is "Ubuntu Advantage - Esential" is there any way to reduce the repetition? Or can users name subs whatever and we need to show Essentila twice?

matthewpaulthomas:
> if the UA column reads "Essential" and the name of the subscription, is "Ubuntu Advantage - Esential" is there any way to reduce the repetition?

Yes, we should report that to CPC as a bug to be fixed both for new and existing subscriptions. It’s completely redundant with data outside that field. Even renaming them “Subscription A”, “Subscription B” etc would be better, because at least that would be unique. (It mightn’t be unique once we allow multiple admins per subscription, but hopefully we can allow+require renaming before that happens.)

templates/advantage/index.html
anthonydillon: Done

templates/advantage/index.html
anthonydillon: Done

templates/advantage/index.html
anthonydillon: Done

anthonydillon:
@mhilton thanks, addressed.

mhilton:
> > if the UA column reads "Essential" and the name of the subscription, is "Ubuntu Advantage - Esential" is there any way to reduce the repetition?
> 
> Yes, we should report that to CPC as a bug to be fixed both for new and existing subscriptions. It’s completely redundant with data outside that field. Even renaming them “Subscription A”, “Subscription B” etc would be better, because at least that would be unique. (It mightn’t be unique once we allow multiple admins per subscription, but hopefully we can allow+require renaming before that happens.)

This is not the only location in which a contract name is used, not all of those locations would include the support level immediately afterwards. I'll happily change it if you think that it's necessary. What this would result in is all commercial subscriptions having the name "Ubuntu Advantage" which really means you can't tell the difference between them at all.

anthonydillon:
### Screenshots
Signed out
Screenshot_2019-10-08 Ubuntu Advantage for Infrastructure Ubuntu

Signed in with paid subscriptions
Screenshot_2019-10-08 Ubuntu Advantage for Infrastructure Ubuntu(1)

Signed in with no paid subscriptions
Screenshot_2019-10-08 Ubuntu Advantage for Infrastructure Ubuntu(2)

PR #5958, opened 2019-10-17
Add Python testing infrastructure

jkfran:
## Done

- Have a `tests` folder
- Have at least one test in the `tests` folder, based on `unittest`, to test that the `/` homepage returns a `200` response
- `./run test` run `lint-python` and `test-python`
- `./run lint-python` run `flake8` and `black --check --line-length 79` against all Python code (including tests)
- `./run test-python` run the `tests`
- Set up CircleCI with a job called `test-python`, which run the tests
- Added codecov integration on CircleCI

## QA

- Run the command `./run test`, it should end correctly.
- Run the command `./run lint-python`, it should end correctly.
- Run the command `./run test-python`, it should end correctly.


tests/fixtures/flask.py
```diff
@@ -0,0 +1,7 @@
+from webapp.app import app
```
jpmartinspt: I would maybe put this in `tests/client.py`. Thoughts @tbille?

jkfran: I'm trying to replicate the same structure as the example @nottrobin give me:
https://github.com/canonical-web-and-design/canonicalwebteam.yaml-responses/blob/master/tests/test_flask.py

But I'm not sure if this is an good case to use test fixtures. Please let me know to maintain the same structure in all projects. @tbille ?

tbille: yeah I like the test/client.py

because it feels that fixtures/flask.Py implies that there could be other clients.

PR #5991, opened 2019-10-22
add table of contents to /about/release-cycle

### LINKED ISSUE #5990: Add table of contents to /about/release-cycle page
From kiko, per [copy doc](https://docs.google.com/document/d/1-l3B7AoyDoETaL80UhlfEWCVKmMWPdy3zoF5LpsGcYQ/edit
)

Please do not use the horizontal pattern, but a sidebar like on https://ubuntu.com/legal/ubuntu-advantage-service-description

sowasred2012:
## Done

- Added short table of contents to /about/release-cycle

## QA

- See that there is a table of contents that links to three different sections on the page

## Issue / Card

Fixes #5990 

## Screenshots

Screenshot 2019-10-22 at 15 23 43

### DIFF
--- a/static/sass/_pattern_table-of-contents.scss
+++ b/static/sass/_pattern_table-of-contents.scss
@@ -1,0 +1,7 @@
+@mixin ubuntu-p-table-of-contents {
+  .p-table-of-contents {
+    @media only screen and (max-width: $breakpoint-navigation-threshold) {
+      padding-left: 0;
+    }
+  }
+}
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -54,6 +54,7 @@ $font-base-family: '"UbtuOverride", -apple-system, "Segoe UI", "Roboto", "Oxygen
 @import "pattern_strip-photos";
 @import "pattern_takeunders";
 @import "pattern_takeovers";
+@import "pattern_table-of-contents";
 @import "pattern_table";
 @import "pattern_ubuntu_intro";
 @import "takeovers/financial-services";
@@ -84,6 +85,7 @@ $font-base-family: '"UbtuOverride", -apple-system, "Segoe UI", "Roboto", "Oxygen
 @include ubuntu-p-pie-chart;
 @include ubuntu-p-separator;
 @include ubuntu-p-strip-photos;
+@include ubuntu-p-table-of-contents;
 @include ubuntu-p-tables;
 @include ubuntu-p-takeunders;
 @include ubuntu-p-takeovers;
--- a/templates/about/release-cycle.html
+++ b/templates/about/release-cycle.html
@@ -24,6 +24,21 @@
       <h2>Version numbers are YY.MM</h2>
       <p>Releases of Ubuntu get a development codename (‘Breezy Badger’) and are versioned by the year and month of delivery - for example Ubuntu 17.10 was released in October 2017.</p>
     </div>
+
+    <div class="col-4 u-hide--small">
+      <aside class="p-table-of-contents">
+        <div class="p-table-of-contents__section">
+          <h4 class="p-table-of-contents__header">CONTENTS</h4>
+          <nav class="p-table-of-contents__nav">
+            <ul class="p-list">
+              <li><a href="#ubuntu-kernel-release-cycle">Ubuntu kernel release cycle (UA-I)</a></li>
+              <li><a href="#ubuntu-openstack-release-cycle">Ubuntu OpenStack release cycle</a></a></li>
+              <li><a href="#charmed-kubernetes-release-cycle">Charmed Kubernetes<sup>&reg;</sup> release cycle</a></a></li>
+            </ul>
+          </nav>
+        </div>
+      </aside>
+    </div>
   </div>
 </section>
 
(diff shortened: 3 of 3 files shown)


sowasred2012:
Open to design suggestions - I did consider adding it to the "Version numbers are YY.MM" strip, but it made it larger than it needed to be:

Screenshot 2019-10-22 at 15 22 59

Could we swap the order of the strips, so "Version numbers are YY.MM" is second, and the strip with the contents is first?

anasereijo:
Hi @matthewpaulthomas, I'm reviewing this issue but I think we need a UX review too, could you help us?

Is it odd and confusing to the user to click on one of the links then scrolling down and not having the other 2 links to click on or a way to quickly go up? Or do you think the user will only want to see one of the sections and is not interested in reading the other ones?

I would suggest using the [sticky nav](https://ubuntu.com/desktop/statistics) but Mark doesn't want us to use it. I think @sowasred2012 took this "pattern" from the [legal page](https://ubuntu.com/legal/ubuntu-advantage-service-description#uasd-professional-support) (which in my opinion isn't working well either - should be sticky at least).

Happy to design a solution with you if needed.

Thank you

sowasred2012:
@matthewpaulthomas @anasereijo yep, that's where the pattern came from. It's also in use on https://ubuntu.com/legal/ubuntu-advantage-service-description#uasd-support-levels, where there's the added benefit of the "Back to top" CTA between each section, maybe that would be useful?

anasereijo:
@matthewpaulthomas ignore my previous comment for now. I just spoke to @clagom about it. Thanks anyway

anasereijo:
@sowasred2012 could you try using the [Table of contents pattern](https://docs.vanillaframework.io/patterns/table-of-contents/)?

- should be on the right side
- make everything fit the grid (you'll have to resize some things)
- should be sticky
- remove the numbers
- I think it makes sense adding the 'Long term support and interim releases' to the list as the user might want to go back to that first section.

sowasred2012:
@pmahnke @clagom 

I've been trying a few different approaches for slotting this sidebar table of contents in there, using a col-8 (content)/col-4 (sidebar) layout, but the js charts are causing a bit of a problem, so would love your thoughts:

- without any changes to the chart, it's very squashed and the x-axis labels are unreadable
Screenshot 2019-10-23 at 11 06 12

- I then tried adding `overflow: hidden` and making the chart wider, but it's not obvious that it can be scrolled:
table

- I tried removing the whitespace between the y-axis labels and the chart, but it still looks cluttered, and breaks another chart further down the page (we could potentially update the chart js to take a width from a data attribute, as the same chart looks a little cluttered on the [live site](https://ubuntu.com/about/release-cycle) too):
Screenshot 2019-10-23 at 11 42 34

Screenshot 2019-10-23 at 11 44 49

- Finally, I also tried a col-9/col-3 layout, but it's still not great:
Screenshot 2019-10-23 at 11 16 33

Screenshot 2019-10-23 at 12 00 59

pmahnke:
@anasereijo, @clagom and @sowasred2012 
I agree, the charts need the space.

What about just have it static, not sticky on the light row?

image

anasereijo:
@pmahnke your solution is definitely the easy and quick one but I wonder if the user will want to go through the other links as well? It can be a bit overwhelming and annoying scrolling up to see the table of contents again. Unless we think the user will not be interested clicking through all the links? I think this is a UX question. @clagom

anasereijo:
@pmahnke @clagom 

**a)** What if instead of having the years like 2011, 2012, 2013, etc we abbreviate them to something like '11, '12, '13... Or do we think this is confusing?

**b)** could we put the years like this?

Ubuntu release cycle   Ubuntu

**c** do we need the repetition of Ubuntu for all releases?

Ubuntu release cycle   Ubuntu (1)

**d** as per @sowasred2012 suggestion we can have a col-9/col-3 layout, but we'll need to change the text to fit

Screenshot 2019-10-23 at 12 53 19

clagom:
These are good points

> @pmahnke @clagom
> 
> **a)** What if instead of having the years like 2011, 2012, 2013, etc we abbreviate them to something like '11, '12, '13... Or do we think this is confusing?
> 
> **b)** could we put the years like this?
> 
> Ubuntu release cycle Ubuntu
> 
> **c)** do we need the repetition of Ubuntu for all releases?
> 
> Ubuntu release cycle Ubuntu (1)
> 
> **d)** as per @sowasred2012 suggestion we can have a col-9/col-3 layout, but we'll need to change the text to fit
> 
> Screenshot 2019-10-23 at 12 53 19

sowasred2012:
Are we saying we want these all implemented?

> These are good points
> 
> > @pmahnke @clagom
> > **a)** What if instead of having the years like 2011, 2012, 2013, etc we abbreviate them to something like '11, '12, '13... Or do we think this is confusing?
> > **b)** could we put the years like this?
> > Ubuntu release cycle Ubuntu
> > **c)** do we need the repetition of Ubuntu for all releases?
> > Ubuntu release cycle Ubuntu (1)
> > **d)** as per @sowasred2012 suggestion we can have a col-9/col-3 layout, but we'll need to change the text to fit
> > Screenshot 2019-10-23 at 12 53 19

clagom:
I think these are improvements, not sure about the priorities, @pmahnke? But I would iterate on those points yes

> Are we saying we want these all implemented?
>

pmahnke:
Personally, I think redefining the chart is a losing battle, it is a lot of work and the chart might always grow and shrink. Getting an extra three columns seems hard.

We have an issue with the sticky horizontal nav, we should focus on that longer-term.

For now, the request seems to be about clients needing to be taken via an anchor to a certain chart. Not move around the page.

I would propose we try to solve with the simpliest solution that doesn't require D3 chart work or a new pattern and think about the best way to fix this larger problem when we have more time.

clagom:
Sounds good as a first step

sowasred2012:
> I think the links should be blue? They don't look very clickable at the moment

OK, I'll get this updated 👍

sowasred2012:
@clagom @anasereijo @pmahnke this is ready for review again when you have a moment

anasereijo:
@sowasred2012 is it me or it's not showing on mobile>?

sowasred2012:
@anasereijo I added `u-hide--small`, without it looked like this:

Screenshot 2019-10-24 at 11 12 00

What do you think?

PR #6063, opened 2019-11-04
Refactor to better handle expired tokens

jpmartinspt:
## Done
Moved API calls into a separate file, so we can add some extra logic to detect expired macaroons.
Add an handler to clear session and re-login when we detect an expired macaroon.

## QA

- Make sure `/advantage` still works

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/6056

### DIFF
--- a/webapp/api/advantage.py
+++ b/webapp/api/advantage.py
@@ -1,0 +1,100 @@
+import os
+
+from requests import Request, Session
+from pymacaroons import Macaroon
+
+from webapp.macaroons import binary_serialize_macaroons
+
+
+API_URL = os.getenv("ADVANTAGE_API", "https://contracts.canonical.com/")
+api_session = Session()
+
+
+class UnauthorizedRequest(Exception):
+    pass
+
+
+def _prepare_request(method, path, data=None, session=None):
+    request = Request(method=method, url=f"{API_URL}{path}")
+
+    if session:
+        root = Macaroon.deserialize(session["macaroon_root"])
+        discharge = Macaroon.deserialize(session["macaroon_discharge"])
+        bound = root.prepare_for_request(discharge)
+        token = binary_serialize_macaroons([root, bound]).decode("utf-8")
+        request.headers["Authorization"] = f"Macaroon {token}"
+
+    if data is not None:
+        request.json = data
+
+    return request.prepare()
+
+
+def _send(request, timeout=3):
+    response = api_session.send(request, timeout=timeout)
+
+    if response.status_code == 401:
+        raise UnauthorizedRequest
+
+    return response
+
+
+def get_macaroon():
+    response = _send(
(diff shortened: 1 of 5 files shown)


webapp/api/advantage.py
```diff
@@ -0,0 +1,102 @@
...
+    request = Request(method=method, url=f"{API_URL}{path}")
+
+    if session:
+        root = Macaroon.deserialize(flask.session["macaroon_root"])
+        discharge = Macaroon.deserialize(flask.session["macaroon_discharge"])
+        bound = root.prepare_for_request(discharge)
+        token = binary_serialize_macaroons([root, bound]).decode("utf-8")
+        request.headers["Authorization"] = f"Macaroon {token}"
+
+    if data is not None:
```
anthonydillon: Can you not do `if data:` here? Why check for None here and not on line 22 above?

webapp/api/advantage.py
```diff
@@ -0,0 +1,102 @@
...
+        _prepare_request(
+            method="post",
+            path=f"v1/contracts/{contract_id}/token",
+            data={},
+            session=session,
+        )
+    )
+
+    payload = response.json()
+    return payload.get("contractToken")
```
anthonydillon: Why not provide a second param here?

jkfran:
It might be my fault but I got the following error during the QA steps:

```
ERROR gunicorn.error "Exception in worker process"
ModuleNotFoundError: No module named 'werkzeug.middleware'
```
@jpmartinspt ?

webapp/api/advantage.py
```diff
@@ -0,0 +1,102 @@
+import os
+
+import flask
```
tbille: not a big fan of putting `flask` dependencies here.
Can you pass the session object that you have to the function that needs it directly.

jpmartinspt: Leftover import, already taking the session as a parameter(will remove the places where it's still using flask.session by mistake)

webapp/api/advantage.py
jpmartinspt: No because we need to call an endpoint with data={} and `{}` would evaluate to `False`.

webapp/views.py
```diff
@@ -60,82 +59,64 @@ def advantage():
...
+                            elif entitlement["type"] == "cc-eal":
+                                entitlements["cc-eal"] = True
+                            elif entitlement["type"] == "support":
+                                contract["supportLevel"] = entitlement[
+                                    "affordances"
+                                ]["supportLevel"]
+                        contract["entitlements"] = entitlements
+                        enterprise_contracts.append(contract)
+        except advantage_api.UnauthorizedRequest:
+            # We got an unauthorized reqest, so we likely
```
tbille: small typo: `s/reqest/request`

anthonydillon:
QA :+1:

PR #6067, opened 2019-11-05
Capture advantage API exceptions in Sentry

jpmartinspt:
## Done
Capture advantage API exceptions in Sentry

## QA

- Set SENTRY_DSN=`<askme>`
- Go to `/advantage` and login
- Swap to use the staging contracts API
- Go to `/advantage`
- See you have a new exception logged in sentry

### DIFF
--- a/webapp/api/advantage.py
+++ b/webapp/api/advantage.py
@@ -10,10 +10,6 @@ API_URL = os.getenv("ADVANTAGE_API", "https://contracts.canonical.com/")
 api_session = Session()
 
 
-class UnauthorizedRequest(Exception):
-    pass
-
-
 def _prepare_request(method, path, data=None, session=None):
     request = Request(method=method, url=f"{API_URL}{path}")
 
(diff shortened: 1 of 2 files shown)


webapp/api/advantage.py
```diff
@@ -11,7 +11,12 @@
 
 
 class UnauthorizedRequest(Exception):
-    pass
+    def __init__(self, url=None, payload=None):
+        self.url = url
+        self.payload = payload
+
+    def as_dict(self):
+        return {"url": self.url, "payload": self.payload}
```
nottrobin: Another option would be to simply do `response.raise_for_status` lower down and then catch the resulting `HTTPError`, which would presumably contain all the information you need when it's passed through to sentry?

nottrobin: Well it's available at `HTTPError.response.json` or something, so you could simply update your code at the point where you `captureException` to extract this data.

jpmartinspt: Ah, sorry I though you were suggesting something different. I'll update

Issue #6070, opened 2019-11-05
U.c/advantage should link to the wiki FAQ


BryanQuigley:
We should add a prominent link to the UAclient wiki / FAQ.

https://wiki.ubuntu.com/UAclient

---
*Reported from: https://ubuntu.com/advantage*

matthewpaulthomas:
I appreciate the effort that’s gone into writing the wiki page. It may be a useful source for developing contextual help. However, it’s really not appropriate to link to the page directly, for four reasons:

1. If there’s evidence of people having a particular issue, it would be addressed much more effectively by adjusting the page design, or adding contextual help at the relevant place in ubuntu.com/advantage, rather than hoping people will follow a link and find the answer somewhere in one long separate page.
2. Even if it was a good idea to provide help on one long separate page, that page should be on the same site (for example, ubuntu.com/advantage/help), so that “Advantage” is still present in the navigation, not on a different site.
3. Even if it was a good idea to have help on a different site, it should not be on the Ubuntu wiki, which is for Ubuntu contributors, not end users. (help.ubuntu.com and docs.ubuntu.com would both be less-bad choices.)
4. Even if it was a good idea to put help on the Ubuntu wiki, the text would need substantial revision to be appropriate for end users.

pmahnke:
@matthewpaulthomas I agree that we should do your no. 2 - how can we get that created?

BryanQuigley:
It was decided to make the Wiki page the FAQ for this - because we could do it in the shortest time.  Again I'm happy for these to be owned elsewhere - but there are MANY pages used for these purposes on the wiki such as:

https://wiki.ubuntu.com/Kernel/Livepatch
https://wiki.ubuntu.com/SecurityTeam/KnowledgeBase (and all subpages)

We also want these to be updated quickly.

Both pages I maintain Livepatch and this one are locked down so only Canonical employees can edit.

matthewpaulthomas:
@pmahnke I’ve reported #6099 for creating a help page on ubuntu.com.

[...]

PR #6078, opened 2019-11-06
Add a wiki link to the advantage page

anthonydillon:
## Done
Had a punt at putting a link to the wiki on the logged in version of the dashboard.

## QA
- Go to /advantage
- Log in
- Check the link to the wiki under the token section.

## Issue / Card
Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/6070

## Screenshots
Screenshot_2019-11-06 Ubuntu Advantage for Infrastructure Ubuntu

### DIFF
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -194,7 +194,7 @@
       <dl>
         <dt>To attach a machine:</dt>
         <dd class="u-no-margin--left"><pre style="border: 0; padding: 0;" class="u-no-margin--bottom"><code>sudo ua attach {{ personal_account.free_token }}</code></pre></dd>
-        <dt>To check status:</dt>
+        <dt style="border-top: 0;">To check status:</dt>
         <dd class="u-no-margin--left"><pre style="border: 0; padding: 0;" class="u-no-margin--bottom"><code>sudo ua status</pre></code></dd>
       </dl>
       <hr style="margin: 1.25rem 0 1.5rem 0;" />
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -187,6 +187,7 @@
           <td><pre style="border: 0; padding: 0;" class="u-no-margin--bottom"><code>sudo ua status</code></pre></td>
         </tr>
       </table>
+      <p>For further options <a href="https://wiki.ubuntu.com/UAclient" class="p-link--external">visit the UA wiki page</a>.</p>
     </div>
     <div class="col-12 u-hide--medium u-hide--large">
       <dl>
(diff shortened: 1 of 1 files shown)


anasereijo:
@anthonydillon why is the link on a different strip?

anasereijo:
@anthonydillon why is the link on a different strip? Could it be in the same strip as "Your Personal subscription"?

anthonydillon:
@anasereijo yes it could, although it makes the "Your Personal subscription" strip a little more cluttered.

anasereijo:
@anthonydillon I don't think it makes it more cluttered. Having another strip will waste more vertical space and makes it look very "sliced"... could we try what we originally designed? I know the link is different but layout wise should work. (see below)

Screenshot 2019-11-06 at 16 07 58

anasereijo:
@anthonydillon I don't think it makes it more cluttered. Having another strip will waste more vertical space and makes it look very "sliced"... could we try what we originally designed? (see below). If needed you can have a bit more padding between link and status

Screenshot 2019-11-06 at 16 07 58

anthonydillon:
@anasereijo done

anasereijo:
If we think it's not prominent enough we can try something like this

Screenshot 2019-11-06 at 16 16 12

anthonydillon:
@anasereijo :+1: done

matthewpaulthomas:
This is a good-faith implementation of the request, but I think the request is fundamentally mistaken. If we have evidence of people on this page who need help with using UA, we should address it in the relevant place in the UI. It’s a scrollable page with plenty of scope for contextual help, so it isn’t be necessary to send people to a different page, let alone a completely different site.

It would be especially inappropriate to send people using UA to the Ubuntu wiki, [which is for Ubuntu contributors](https://wiki.ubuntu.com/), not end users.

anasereijo:
@anthonydillon on mobile it's showing an extra `hr`. please delete it

Screenshot 2019-11-07 at 09 42 36

anthonydillon:
This PR in now blocked due to the UX feedback.

anthonydillon:
@anasereijo removed the border on small screens.

anthonydillon:
@matthewpaulthomas do you think a link to the manpage would work instead?

matthewpaulthomas:
@anthonydillon Linking to the man page instead would be better in one way, that manpages.ubuntu.com’s navigation is less bewildering than wiki.ubuntu.com’s. However:
- It would have the same general problem of being a long page of stuff on a separate site.
- [The ua man page is currently published for 18.04 LTS, 19.04, and 19.10, but not for 14.04 LTS](http://manpages.ubuntu.com/cgi-bin/search.py?q=ua). We could link to the 19.10 version, and the info would be correct, but it would look like a mistake.
- There is, surprisingly, very little overlap between topics covered by the man page and topics covered by the wiki page. So linking to the man page probably wouldn’t address any particular problem that motivated #6070.

Once the man page is published for 14.04 LTS, maybe we could link it for completeness in the contextual footer, where its format and destination would be less surprising. But that shouldn’t distract us from providing contextual help within the page.

anthonydillon:
@matthewpaulthomas Ok, closing for now until we have something to link too.

PR #6118, opened 2019-11-13
Add headers to advantage page

### LINKED ISSUE #6081: Signing in to /advantage doesn’t work
Starting from a signed-out state:

1\. Go to [ubuntu.com/advantage](https://ubuntu.com/advantage).
2\. Click “Sign in” or “Register”.
3\. Click “Sign in” or “Register”.

What happens:
2\. After you sign in, the Advantage page loads in exactly the same state as before, with “Sign in” and “Register” buttons.
3\. The page reloads in exactly the same state.

What should happen:
2\. After you sign in, you see your subscriptions.

---
*Reported from: https://ubuntu.com/advantage*

jpmartinspt:
## Done

Add cache-control headers to advantage page

## QA

- Check `Cache-Control` is set to `no-cache`

## Issue / Card

Fixes #6081

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -60,6 +60,7 @@ def advantage():
     enterprise_contracts = []
     entitlements = {}
     openid = flask.session.get("openid")
+    headers = {"Cache-Control": "no-cache"}
 
     if auth.is_authenticated(flask.session):
         try:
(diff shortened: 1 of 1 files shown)


webapp/views.py
```diff
@@ -60,6 +60,7 @@ def advantage():
     enterprise_contracts = []
     entitlements = {}
     openid = flask.session.get("openid")
+    headers = {"Cache-Control": "no-store"}
```
nottrobin: It occurred to me that `private` is probably more accurate.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Cacheability

jpmartinspt: updated

PR #6213, opened 2019-12-02
Fix bugs in eol charts

### LINKED ISSUE #6139: Years overlap on Ubuntu kernel release cycle graph
## Issue

- If you go to https://ubuntu.com/about/release-cycle, scroll down to the `Ubuntu kernel release cycle` section and make your window size smaller (medium size) you will notice an overlap of the years at the bottom of the graph. Perhaps we can show only every other year instead.

P.S. I know that a screenshot would have been very helpful but for some reason I cannot take screenshots of my screen.
---
*Reported from: https://ubuntu.com/about/release-cycle

steverydz:
## Done

- Reduce year axis to every other year when labels begin to overlap
- Update breakpoint variable so graph doesn't try to render when table is viewed instead causing errors in the console

## QA

- Resize the browser and see that when the labels on the year axis on the release charts start to get close to each other they change to every other year, before they turn into tables

## Issue / Card

Fixes #6139

## Screenshots

eol-chart

### DIFF
--- a/static/js/chart.js
+++ b/static/js/chart.js
@@ -207,7 +207,7 @@ export function createChart(chartSelector, taskTypes, taskStatus, tasks) {
   var width = document.querySelector(chartSelector).clientWidth - margin.right - margin.left;
   var tickSize = 1;
 
-  if (window.innerWidth < 1000) {
+  if (window.innerWidth < 1080) {
     tickSize = 2;
   }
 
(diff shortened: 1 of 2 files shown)


static/js/chart.js
```diff
@@ -205,6 +205,11 @@ export function createChart(chartSelector, taskTypes, taskStatus, tasks) {
   var timeDomainEnd = d3.time.year.offset(tasks[tasks.length - 1].endDate, +1);
   var height = taskTypes.length * rowHeight;
   var width = document.querySelector(chartSelector).clientWidth - margin.right - margin.left;
+  var tickSize = 1;
+
+  if (window.innerWidth < 1000) {
```
anthonydillon: I would suggest increasing the threshold as the years get very close.

steverydz: Thanks. Done

PR #6368, opened 2020-01-13
Use coreutils sha256sum instead of Perl's shasum

baryluk:
Current download verify dialog shows this:

echo "96a8095001d447bbb9078925d72f7a77a3f62fbd78460093759af4394ce83d79 *ubuntu-19.10-desktop-amd64.iso" | shasum -a 256 --check

but the `shasum` is a program implemented in Perl, which might not be
installed. On my distro, I am able to have entire system installed
without any Perl at all, and no shasum program by default at all.

However, it is very likely to have GNU coreutils installed. It would
be actually hard not to have it installed.

So the dialog should show instead:

echo "96a8095001d447bbb9078925d72f7a77a3f62fbd78460093759af4394ce83d79 *ubuntu-19.10-desktop-amd64.iso" | sha256sum --check

As a bonus, sha256sum is slightly faster: I also done minor benchmarks,
and sha256sum was about 6% faster.
(Perl shasum -a 256: 8.68s, coreutils sha256sum: 8.13s; real time).

The input / output format is exactly the same.

## Done

[List of work items including drive-bys]

## QA

## Issue / Card

Fixes #

## Screenshots

[if relevant, include a screenshot]

### DIFF
--- a/templates/download/shared/_verify-checksums.html
+++ b/templates/download/shared/_verify-checksums.html
@@ -5,7 +5,7 @@
     <span class="p-contextual-menu__dropdown" id="menu-4" aria-hidden="true" aria-label="Verification instructions:" style="width: 600px; max-width: 70vw; padding: 1rem;">
       <span class="p-contextual-menu__group">
         <small>Run this command in your terminal in the directory the iso was downloaded to verify the SHA256 checksum:</small><br />
-        <code style="display: block; margin-top: 1rem;">echo "{{ releases.checksums[system][version] }}" | shasum -a 256 --check</code><br />
+        <code style="display: block; margin-top: 1rem;">echo "{{ releases.checksums[system][version] }}" | sha256sum --check</code><br />
         <small>You should get the following output:</small><br />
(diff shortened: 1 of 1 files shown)


matthewpaulthomas:
@baryluk Thanks for the PR.

We originally did use `sha256sum`. [I changed it to `shasum -a 256` last July](https://github.com/canonical-web-and-design/ubuntu.com/pull/5312) so that it would work for more of the people who arrive at these pages. This is because a non-trivial fraction of those people are using macOS (maybe so that they can use Ubuntu in a VM, or dual-boot). And `shasum`, besides working for most Linux OS users, also works on macOS, while `sha256sum` does not.

I didn’t understand the technicalities at the time, but your explanation of one being in GNU coreutils and the other being in Perl makes sense. And that `shasum` is a fraction of a second slower is less important than that it works for more people.

Now, this is fiddling at the margins — because most of the people arriving at these pages are using Windows, and for most of them, _neither_ of these commands work! So, ideally we would use OS detection to provide the most appropriate command depending on the visitor: `sha256sum` on Linux OSes, `shasum -a 256` on macOS, and `certutil -hashfile`…`SHA256` on Windows. Until anyone has time to work on that, though, we’ll stick with `shasum`.

PR #6502, opened 2020-01-29
Remove anchor on /advatage link on livepatch page

### LINKED ISSUE #6110: /advantage table row anchors hide the purpose of the table
1\. Go to any of these URLs:
- https://ubuntu.com/advantage/#esm
- https://ubuntu.com/advantage/#livepatch
- https://ubuntu.com/advantage/#fips

What happens:
👍 The relevant row in the “Plans for enterprise use” table is highlighted.
👎 The page is scrolled so that nothing is visible above that table row.

What’s wrong with this:
- Linking to an anchor on a separate page is unusual, so we need to be more careful than usual about ensuring people understand where they’ve ended up.
- Unless you scroll, you can’t even see the table headers, so you don’t know what the table columns are for.
- Unless you scroll, you can’t see the section heading, so you don’t know what the entire table is for.

What should happen:
👍 The relevant row in the “Plans for enterprise use” table is highlighted.
👍 The page is scrolled so that the “Plans for enterprise use” heading and table headers are visible.

[Encountered while reviewing #6085.]

pmahnke:
## Done

- Remove anchor on /advatage link on livepatch page

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/livepatch
- See that the button 'Get Ubuntu Advantage' no longer has a anchor

## Issue / Card

Fixes #6110

### DIFF
--- a/templates/livepatch/index.html
+++ b/templates/livepatch/index.html
@@ -50,7 +50,7 @@
       <p><a class="p-button--positive"  href="https://auth.livepatch.canonical.com/">Get Livepatch</a></p>
       <h2>Part of Ubuntu Advantage</h2>
       <p>Buy as part of Ubuntu Advantage from Canonical. Buy online.</p>
-      <p><a class="p-button--positive"  href="/advantage#livepatch">Get Ubuntu Advantage</a></p>
+      <p><a class="p-button--positive"  href="/advantage">Get Ubuntu Advantage</a></p>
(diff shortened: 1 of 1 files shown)


matthewpaulthomas:
UX +1 because it’s an improvement, but I’ve disconnected it from #6110 because (as Ant says) it doesn’t fix that problem.

PR #6675, opened 2020-02-17
WIP: USN API

jpmartinspt:
It's WIP

Please consider
- Are you OK with using marshmallow?
- I feel like it would be good to split USN into it's own files.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,3 +13,4 @@ pybreaker==0.6.0
 Flask-WTF==0.14.2
 requests==2.22.0
 sqlalchemy==1.3.10
+marshmallow
--- a/webapp/security/load_fixtures.py
+++ b/webapp/security/load_fixtures.py
@@ -1,0 +1,18 @@
+import yaml
+
+from webapp.database import db_session
+from webapp.models import Release
+
+
+with open("webapp/security/fixtures/releases.yaml") as file:
+    document = yaml.full_load(file)
+    for codename, r in document["releases"].items():
+        release = Release(
+            name=r["name"],
+            version=r["version"],
+            codename=codename,
+            lts=r["lts"],
+        )
+
+        db_session.add(release)
+        db_session.commit()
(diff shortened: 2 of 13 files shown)


requirements.txt
```diff
@@ -13,3 +13,4 @@ pybreaker==0.6.0
 Flask-WTF==0.14.2
 requests==2.22.0
 sqlalchemy==1.3.10
+marshmallow
```
carkod: No version? 3.4.0 is the last one

webapp/views.py
```diff
@@ -313,3 +316,79 @@ def notices():
             total_results=total_results,
         ),
     )
+
+
+# USN API
```
tbille: It feels that this could be in it's own file (`webapp/security/views.py` ?) 
I think it could help splitting a bit the different views that provides templates, and the others that are APIs.

webapp/security/load_fixtures.py
```diff
@@ -0,0 +1,18 @@
...
+    for codename, r in document["releases"].items():
+        release = Release(
+            name=r["name"],
+            version=r["version"],
+            codename=codename,
+            lts=r["lts"],
+        )
+
+        db_session.add(release)
+        db_session.commit()
```
tbille: so every time we start the application we are updating the releases db?

webapp/views.py
jpmartinspt: Yeah I wrote that in the description of the PR, with ubuntu.com growing so much it feels like we need to split things up.

jkfran: I agree, but also is it fine to define the database schema in the view?

webapp/security/load_fixtures.py
jpmartinspt: No, every time you run that script, it was mostly a helper thing for me to bootstrap the DB

jpmartinspt: I can wrap it in a function if we decide to keep it

webapp/views.py
jpmartinspt: No, it should live in it's own file in my opinion

jkfran: Ah okay. So it is here provisionally or something?

jpmartinspt: Yeah, until we decide if it's ok to split this into it's own module

jpmartinspt:
Landed as part of #6769

PR #6691, opened 2020-02-19
Integrate cypress

tbille:
## Done

- Integrate cypress in ubuntu.com test suite
- Add first simple test
- Add configuration to avoid cypress to generate default files

## QA

- In another terminal run `PORT=8001 yarn run cypress:run`
- Make sure tests pass.
- In another terminal run `PORT=8001 yarn run cypress:open`
- Have fun with the interface that can run the tests that cypress provides

## Issue / Card

Fixes https://github.com/canonical-web-and-design/web-squad/issues/2361


tests/cypress/form_spec.js
```diff
@@ -0,0 +1,12 @@
+/// <reference types="cypress" />
+
+context('Navigation', () => {
+    beforeEach(() => {
+        cy.visit('http://0.0.0.0:8001')
```
tbille: Note to me. This should change to use the base url function that cypress provides

PR #6703, opened 2020-02-20
Add inline utility class

steverydz:
## Done

Add utility for making an element inline

## QA

Check that a class of `.u-inline` has been added


anthonydillon:
Why is this required?

steverydz:
@anthonydillon The image template wraps images in a `div` so we need a way to make them inline when required

PR #6708, opened 2020-02-20
Use image_template on /about/release-cycle page

pmahnke:
## Done

Replaced images with the image_template module

## QA

- See that the images are still there (compare to ubuntu.com)

### DIFF
--- a/templates/about/release-cycle.html
+++ b/templates/about/release-cycle.html
@@ -13,7 +13,17 @@
       <p>Canonical publishes new releases of Ubuntu on a regular cadence, enabling the community, businesses and developers to plan their roadmaps with certainty of access to newer open source upstream capabilities.</p>
     </div>
     <div class="col-5 u-vertically-center u-align--center u-hide--small">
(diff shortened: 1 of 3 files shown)


templates/about/release-cycle.html
```diff
@@ -13,7 +13,17 @@ <h1>The Ubuntu lifecycle and release cadence</h1>
...
-      <img src="https://assets.ubuntu.com/v1/ed348358-logo-cof.svg" width="200" alt="">
+      {{
+        image(
+        url="https://assets.ubuntu.com/v1/ed348358-logo-cof.svg",
+        alt="",
+        width="200",
+        height="200",
+        hi_def=True,
+        loading="auto",
+        attrs={"class": "", "id": ""},
```
sowasred2012: If these are empty it's fine to leave the `attrs` object out

```suggestion
```

barrymcgee: We'll grep the codebase and remove these all in bulk at the end.

Issue #6710, opened 2020-02-20
/advantage: Add “CIS benchmark” (cis-audit)


matthewpaulthomas:
1\. Go to [/advantage](https://ubuntu.com/advantage)
2\. Look for info about the CIS benchmark
3\. Sign in
4\. Look for instructions about how to turn the CIS service on/off

What happens: 2, 4. There is no such info.

What should happen: CIS is covered in all the same places that FIPS and CC-EAL are.

This will involve finding out what exactly happens when you “enable” `cis-audit`. Presumably it doesn’t just download the benchmarks PDF. So what does it do?

[Based on [a comment by @lechsandecki in the copydoc](https://docs.google.com/document/d/1AW8wQu1Uvq0xFtd_c2JwNdic075MPOTx7k2KHqFPPeo/edit?disco=AAAAGN2UYVE).]

lechsandecki:
Hi Matthew, thank you for taking on this one. We are currently providing CIS benchmark to all UA Infra customers. The process today goes by the support team. In the future w will address it via the UA Client. Technically the product exists (owned by the security team), and can be provisioned (manually, by the support team). I edited copy docs and asked ubuntu-editors to update the following pages:
UASD: https://docs.google.com/document/d/1vQNdVmpsB3iQP7SnJHxT3A8rbD5jSAzTEoYUOepR-SM/edit#

pricing: https://docs.google.com/document/d/1Ynk0iLa9hjrdich6ATTsybtPbDwW_k4vRI0nk_HNhcI/edit#

security certifications:
https://docs.google.com/document/d/1CiF-dPqG1Y--h4FWfgKB1ft4G3nC2Boa8NT_JG7D6dQ/edit#

I'd like to ask you to update the /advantage page accordingly, please

many thanks,
Lech

matthewpaulthomas:
[Specification updated](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.rvsupctzkbga) for this issue and #6154.

PR #6711, opened 2020-02-20
Changes “cis/?” to “cis-audit/?”

matthewpaulthomas:
`ua help` refers to https://ubuntu.com/cis-audit, which does not exist. But `/fips`, `/cc-eal`, and `/cis` all redirect to `/security/certifications`. So it looks like the `/cis` redirect was mistaken.

Discovered while investigating #6710.

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -128,7 +128,7 @@ category/product/ubuntu-advantage/?: /support
 category/product/ubuntu-light/?: http://lubuntu.net/
 cc/?: "/security/certifications"
 cc-eal/?: "/security/certifications"
-cis/?: "/security/certifications"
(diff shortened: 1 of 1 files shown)


anthonydillon:
@matthewpaulthomas should we have both?

matthewpaulthomas:
@anthonydillon Well, it’s just one extra line, but then, that could excuse the list of redirects growing without bound … My hypothesis is that the current redirect is either a typo, or a misunderstanding between ubuntu.com and UA teams about what the URL would be. This could be tested by traffic logs for `/cis` for any period before this morning. It doesn’t show up in Google Analytics, but I guess a 404 wouldn’t.

PR #6732, opened 2020-02-20
Image builder copy update

anthonydillon:
## Done
Update the copy from the [copy doc](https://docs.google.com/document/d/1Hw_8dWAAUpHHm5Z_XNUVJaSVE5k5i2Vt6WANcJYUZvY/edit#)

## QA

- Check that the copy mostly matches the copy doc

## Issue / Card

Fixes https://github.com/canonical-web-and-design/web-squad/issues/2367

### DIFF
--- a/templates/core/build.html
+++ b/templates/core/build.html
@@ -5,151 +5,180 @@
 {% block meta_copydoc %}https://docs.google.com/document/d/1Hw_8dWAAUpHHm5Z_XNUVJaSVE5k5i2Vt6WANcJYUZvY/edit{% endblock meta_copydoc %}
 
 {% block content %}
-
-{% if not openid %}
-<section>
-  <center><strong>You need to <a href="/login?next=/core/build">login</a> mate</strong></center>
-</section>
-{% else %}
-<section class="p-strip--suru-topped u-no-padding--bottom">
+<section class="p-strip--suru-topped">
   <div class="row">
-    <div class="col-6">
-      <h1>Create an image</h1>
-      <p>Build your own image by selecting your board, Operating System and applications to pre-install.</p>
+    <div class="col-7">
+      <h1>Build an Ubuntu IoT image</h1>
+      <p>Choose your board, Ubuntu version, and the apps you want installed, and in a few minutes we’ll build an Ubuntu image ready for your IoT development.</p>
     </div>
+    {% if openid %}
+    <div class="col-5 col-start-large-8">
+      <div class="u-no-margin--bottom p-card">
+        <p>You are signed in as {{ openid.fullname }}<br />
+        <small style="color:#666">({{ openid.email }})</small></p>
+        <p class="u-no-margin--bottom">
+          <a href="/logout?return_to={{ request.host_url }}core/build" class="p-button--neutral u-no-margin--bottom" onclick="dataLayer.push({
+            'event' : 'GAEvent',
+            'eventCategory' : 'Advantage',
+            'eventAction' : 'Authentication',
+            'eventLabel' : 'Sign out',
+            'eventValue' : undefined
+          });">Sign out</a>
+        </p>
+      </div>
+    </div>
+    {% else %}
+    <div class="col-5 col-start-large-8">
+      <div class="u-no-margin--bottom p-card">
+        <p>
+           Sign in to access your personal or paid subscriptions.
+        </p>
+        <p class="u-no-margin--bottom">
+          <a href="/login" class="p-button--neutral u-no-margin--bottom"
+            onclick="dataLayer.push({
+              'event' : 'GAEvent',
+              'eventCategory' : 'Advantage',
+              'eventAction' : 'Authentication',
+              'eventLabel' : 'Sign in',
+              'eventValue' : undefined
+            });">
+            Sign in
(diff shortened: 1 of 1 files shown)


templates/core/build.html
```diff
@@ -5,151 +5,180 @@
...
-<section class="p-strip is-shallow">
-  <div class="row">
-    <div class="col-8">
-      <h2>What version of Ubuntu do you want? <a href=""><i class="p-icon--help">Help me choose</i></a></h2>
-      <p>Select the version of Ubuntu to install on the device.</p>
+    {% else %}
+    <div class="col-5 col-start-large-8">
+      <div class="u-no-margin--bottom p-card">
+        <p>
+           Sign in to access your personal or paid subscriptions.
```
matthewpaulthomas: Copy-paste error?

templates/core/build.html
```diff
@@ -5,151 +5,180 @@
...
-        <input type="search" class="p-search-box__input" name="search" placeholder="Search for snaps" required="">
-        <button type="reset" class="p-search-box__reset" alt="reset"><i class="p-icon--close"></i></button>
-        <button type="submit" class="p-search-box__button" alt="search"><i class="p-icon--search"></i></button>
-      </form>
-      <div class="js-snap-results"></div>
+    <div class="row js-boards">
+      <button class="col-2 p-card js-selection">
+        <img class="p-card__thumbnail" src="https://assets.ubuntu.com/v1/8542c958-Raspberry+Pi+2.png" alt="A photo the the Raspberry Pi 2">
+        <hr class="u-sv1">
+        <p class="p-card__content">Raspberry Pi2</p>
```
matthewpaulthomas: “One What board are you using? A photo the the Raspberry Pi two. Raspberry Pi two” 😕

Here, “[the icon is supplemental to a text label conveying the same meaning. In those cases, the alt attribute must be present but must be empty.](https://html.spec.whatwg.org/multipage/images.html#a-short-phrase-or-label-with-an-alternative-graphical-representation:-icons,-logos)”

The same for all the other boards, and all the Ubuntu versions.

templates/core/build.html
```diff
@@ -5,151 +5,180 @@
...
-        <dt>Operating system</dt>
-        <dd class="js-summary-os"></dd>
-        <dt>Snaps</dt>
-        <dd><span class="js-summary-snaps">0</span> pre-installed snaps</dd>
-      </dl>
+  <section class="p-strip is-shallow">
+    <div class="row">
+      <div class="col-8">
+        <h2>3 What snaps do you want preinstalled?</h2>
+        <p>Select the applications that will be pre-installed on the device. For Ubuntu Core this will be snaps, for Ubuntu Classic it will be snaps or packages.</p>
```
matthewpaulthomas: I didn’t know this. We’ll need to figure out how to distinguish them when there’s a snap and a .deb package with the same name.

Snaps are a kind of package, so “or packages” doesn’t really work. I think you need to specify “.deb packages”.

templates/core/build.html
```diff
@@ -5,151 +5,180 @@
...
+      <h2>Summary</h2>
+    </div>
+    <div class="row p-divider">
+      <div class="col-6 p-divider__block">
+        <dl>
+          <dt>Board</dt>
+          <dd class="js-summary-board"></dd>
+          <dt>Operating system</dt>
+          <dd class="js-summary-os"></dd>
+          <dt>Snaps</dt>
```
matthewpaulthomas: Packages

templates/core/build.html
```diff
@@ -5,151 +5,180 @@
...
+          <dd><span class="js-summary-snaps">0</span> pre-installed snaps</dd>
+        </dl>
+      </div>
+      <div class="col-6 p-divider__block">
+        <p>This will trigger a build in Launchpad to create your image. Once the image has been built we will contact us via email on <strong>{{ openid.email }}</strong>. This build will only be available for a few days so please do not rely on Launchpad for future downloads</p>
```
matthewpaulthomas: Where did this text come from? “We will contact us”??

templates/core/build.html
anthonydillon: done

templates/core/build.html
anthonydillon: I agree, done

templates/core/build.html
anthonydillon: I stole this from [the prototype](https://image.iotdevice.io/boards/raspberrypi2/core16). This might not be the case if we only support core.

templates/core/build.html
anthonydillon: I wrote it as placeholder text. Fixed the typo at least.

templates/core/build.html
anthonydillon: Done

anthonydillon:
@matthewpaulthomas All address, ready for another review.

templates/core/build.html
matthewpaulthomas: Not for this PR, but: If .deb packages are an open question, or something that might change later, that suggests `core` maybe shouldn’t be baked into the URL. (Another reason is that this page would be quite at home in ubuntu.com’s IoT section.)

PR #6753, opened 2020-02-20
Image builder snap search

anthonydillon:
## Done
Make the snap search real but limited to 12.

## QA
- Scroll to the snap search and search for something
- See that renders the result and allows you to select them.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/web-squad/issues/2382

### DIFF
--- a/static/js/imageBuilder.js
+++ b/static/js/imageBuilder.js
@@ -29,6 +29,10 @@
         e.preventDefault();
         triggerSearch();
       });
+      snapSearch.addEventListener('submit', e => {
+        e.preventDefault();
+        triggerSearch();
+      });
     }
   }
 
@@ -89,12 +93,18 @@
       if (Object.entries(responce).length !== 0) {
         responce.forEach((item, index) => {
           item.icon_url = (item.icon_url)?item.icon_url:'https://assets.ubuntu.com/v1/be6eb412-snapcraft-missing-icon.svg';
+          item.validation_icon = (item.developer_validation === 'verified')?`<span class="p-tooltip p-tooltip--top-center" aria-describedby="${item.package_name}-tooltip">
+          <img src="https://assets.ubuntu.com/v1/75654c90-rosette.svg">
+          <span class="p-tooltip__message u-align--center" role="tooltip" id="${item.package_name}-tooltip">Verified account</span>
+        </span>`:'';
           results.insertAdjacentHTML('beforeend',
             `<div class="p-media-object" data-container-index="${index}">
               <img src="${item.icon_url}" alt="" class="p-media-object__image">
               <div class="p-media-object__details">
                 <h1 class="p-media-object__title">${item.title}</h1>
-                <p class="p-media-object__content">${item.developer_name}</p>
+                <p class="p-media-object__content">
+                  ${item.publisher} ${item.validation_icon}
+                </p>
                 <a href="" class="p-button--neutral js-${buttonText.toLowerCase()}-snap" data-index="${index}">${buttonText}</a>
               </div>
             </div>`
--- a/static/js/imageBuilder.js
+++ b/static/js/imageBuilder.js
@@ -41,7 +41,7 @@
           return response.json();
         })
         .then((json) => {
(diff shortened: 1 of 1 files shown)


anthonydillon:
image

matthewpaulthomas:
I’m assuming that this PR is just for search behaviour and results, not for the layout that we discussed this morning.

- In Firefox, if I happen to move the pointer even one pixel after clicking the search field, I get a tooltip “Please fill out this field.”, which is nagging since I haven’t had a chance to start typing yet. This isn’t your fault, it’s Firefox overreacting to `required=`, but I don’t see that we get any benefit in return from that attribute here.
- Relatedly, I’d much prefer it if we didn’t have to type Enter to perform a search. Is it practical to make it a live filter instead?
- The search returns many false positives. For example if I search for “apache” I get 12 results, but the fifth one is the only one that mentions Apache. (Maybe it’s mentioned in the description? That’s a mystery if I have no way of accessing a snap’s description from this page.)
- When a snap doesn’t have an icon, I get ``. Shouldn’t the API provide the generic snap icon in that case? You shouldn’t be having to provide it yourself. (What if it changes?)
  image
- All snap icons should have `alt=""`, since they are [supplemental to a text label conveying the same meaning](https://html.spec.whatwg.org/multipage/images.html#a-short-phrase-or-label-with-an-alternative-graphical-representation:-icons,-logos).
- For every snap, the publisher name appears as “undefined”, and when the publisher is verified the verified badge is missing. (I don’t know whether that’s one issue or two.)
- When I “Add” a snap it disappears from the search results. That would be fine if it was a small list with no search. But I think it’s a bit confusing for a _search_ to not return a particular search result merely because that result happens to be in another list at the moment. I think it would be more understandable just to disable the Add button.

nottrobin:
> Relatedly, I’d much prefer it if we didn’t have to type Enter to perform a search. Is it practical to make it a live filter instead?

Yes, I think we would all prefer live search, but (I'm sure I speak for @anthonydillon) not this is not practical for the first pass. It would be a nice enhancement, but we definitely shouldn't delay releasing the first version for this.

> The search returns many false positives. For example if I search for “apache” I get 12 results, but the fifth one is the only one that mentions Apache. (Maybe it’s mentioned in the description? That’s a mystery if I have no way of accessing a snap’s description from this page.)

I assume the search results are the same as on https://snapcraft.io/search?q=apache. There is no way we can make this search more effective for this piece of work. This should be a bug against the Snap Store API, and I'm sure it wouldn't be trivial for them to solve either.

The other points sound fair, and are hopefully fairly easy to fix.

anthonydillon:
> In Firefox, if I happen to move the pointer even one pixel after clicking the search field, I get a tooltip “Please fill out this field.”, which is nagging since I haven’t had a chance to start typing yet. This isn’t your fault, it’s Firefox overreacting to required=, but I don’t see that we get any benefit in return from that attribute here.

I am unsure what you are experiencing but I only see the tooltip if you blur the input with nothing entered into it.

> Relatedly, I’d much prefer it if we didn’t have to type Enter to perform a search. Is it practical to make it a live filter instead?

Done :rocket: 

> The search returns many false positives. For example if I search for “apache” I get 12 results, but the fifth one is the only one that mentions Apache. (Maybe it’s mentioned in the description? That’s a mystery if I have no way of accessing a snap’s description from this page.)

Unfortunately, that is the results from the API.

> When a snap doesn’t have an icon, I get . Shouldn’t the API provide the generic snap icon in that case? You shouldn’t be having to provide it yourself. (What if it changes?)

Added the default icon for snaps without icons :rocket: 

> For every snap, the publisher name appears as “undefined”, and when the publisher is verified the verified badge is missing. (I don’t know whether that’s one issue or two.)

Fixed and all done :rocket: 

> When I “Add” a snap it disappears from the search results. That would be fine if it was a small list with no search. But I think it’s a bit confusing for a search to not return a particular search result merely because that result happens to be in another list at the moment. I think it would be more understandable just to disable the Add button.

I will do this as a follow-up as the class I need is in the step by step branch

matthewpaulthomas:
Nice work!

PR #6830, opened 2020-02-21
Remove cypress

tbille:
Removing cypresss from dependencies. Since it downloads a lot of stuff that we don't need when deploying.


.github/workflows/cypress.yml
```diff
@@ -1,8 +1,6 @@
 name: Run cypress at 7am every day
```
barrymcgee: Should this not be removed also?

tbille: so I didn't added a description I am just removing from the package json but I will keep the jobs

PR #6845, opened 2020-02-24
Remove the summary section

anthonydillon:
## Done
Remove the summary section

## QA
- Make a section and see there is no summary at the bottom
- Check it matched [the copy doc](https://docs.google.com/document/d/1Hw_8dWAAUpHHm5Z_XNUVJaSVE5k5i2Vt6WANcJYUZvY/edit#)

## Issue / Card
Fixes https://github.com/canonical-web-and-design/web-squad/issues/2401

## Screenshots
image

### DIFF
--- a/templates/core/build.html
+++ b/templates/core/build.html
@@ -149,12 +149,12 @@
   <section class="p-strip--light is-shallow">
     <div class="row">
       <div class="col-8">
+        <p>We’ll email you at {{ openid.email }} once the build is ready.</p>
         <p>
           <input class="mktoField" value="yes" id="canonicalUpdatesOptIn" name="canonicalUpdatesOptIn" type="checkbox">
           <label class="mktoLabel mktoHasWidth" for="canonicalUpdatesOptIn">I agree to receive information about Canonical’s products and services.</label>
         </p>
         <p><button class="p-button--positive js-build-button" aria-disabled="true" disabled="">Build image</button></p>
-        <p>We’ll mail you at {{ openid.email }} once the build is ready.</p>
       </div>
     </div>
   </section>
--- a/templates/core/build.html
+++ b/templates/core/build.html
@@ -148,26 +148,13 @@
 
   <section class="p-strip--light is-shallow">
     <div class="row">
-      <h2>Summary</h2>
-    </div>
-    <div class="row p-divider">
-      <div class="col-6 p-divider__block">
-        <dl>
-          <dt>Board</dt>
-          <dd class="js-summary-board"></dd>
-          <dt>Operating system</dt>
-          <dd class="js-summary-os"></dd>
-          <dt>Packages</dt>
-          <dd><span class="js-summary-snaps">0</span> pre-installed packages</dd>
-        </dl>
-      </div>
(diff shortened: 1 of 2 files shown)


anasereijo:
Thanks for doing this, @anthonydillon!
It's not great, but to be honest I'm not sure what else I can propose unless we have a title saying Final steps, or Build your image? any thoughts @matthewpaulthomas ?

anasereijo:
@anthonydillon I much preferred having the Summary section as before, but I guess there were valid reasons to remove it? @matthewpaulthomas

matthewpaulthomas:
Thanks @anthonydillon. 

> I much preferred having the Summary section as before, but I guess there were valid reasons to remove it? @matthewpaulthomas

@anasereijo All else being equal, simpler designs are easier to learn and remember. So I think it should be the other way around: we should need a reason to _include_ anything in a design. Can you elaborate on why you preferred having it?

A summary is sometimes used at the end of multi-page forms, where you may want to double-check what you’ve entered, but you can’t use the browser’s Back for fear of losing what you entered, or if going Back would be much slower than reading all your answers in a single table. Neither applies here. This is a very simple form: 9~12 data controls (depending on your board selection), on one single short page, and you can scroll up to see what you entered in exactly the format that you entered it. In comparison, [our BootStack contact form](https://ubuntu.com/openstack/managed#get-in-touch) has _99_ data controls (not counting “I’m not a robot”), spread across three pages — and even then, it _still_ doesn’t need a summary, or “Final steps”, or anything like that. In that case, just a “Previous” button is enough.

I’m sorry if this is a surprise. I deliberately did not include a summary in the original copydoc, and I said it was unnecessary during the review last Thursday.

templates/core/build.html
```diff
@@ -148,26 +148,13 @@ <h4 class="p-muted-heading">Pre-installed snaps</h4>
...
-      </div>
-      <div class="col-6 p-divider__block">
-        <p>This will trigger a build in Launchpad to create your image. Once the image has been built we will contact you via email on <strong>{{ openid.email }}</strong>. This build will only be available for a few days so please do not rely on Launchpad for future downloads</p>
+      <div class="col-8">
         <p>
           <input class="mktoField" value="yes" id="canonicalUpdatesOptIn" name="canonicalUpdatesOptIn" type="checkbox">
           <label class="mktoLabel mktoHasWidth" for="canonicalUpdatesOptIn">I agree to receive information about Canonical’s products and services.</label>
         </p>
         <p><button class="p-button--positive js-build-button" aria-disabled="true" disabled="">Build image</button></p>
+        <p>We’ll mail you at {{ openid.email }} once the build is ready.</p>
```
nottrobin: I don't like "We'll mail you". That sounds very archane language to me. I get that in this day and age email is so common that you would almost think it would be the default form of "mail"ing, but still, I don't think "mail" is a common term in modern parlance.

I'd suggest "We'll email you" or "We'll contact you".

I'm approving this anyway.

anthonydillon: @matthewpaulthomas what do you think?

PR #6847, opened 2020-02-24
Style the board and OS selection sections

anthonydillon:
## Done
Style the board and OS selection sections

## QA
- Check the select board and OS sections match [the design](https://app.zeplin.io/project/5e33f5075368883e5da9d0bf/screen/5e4fe76e44a1ee65293d3d7f) 

## Issue / Card
Fixes https://github.com/canonical-web-and-design/web-squad/issues/2362

## Screenshots
image


anasereijo:
Hi @anthonydillon, a few things to be changed if possible:

- [ ] instructions are missing below each step but I guess that's because you are following Matthews copy doc?

- [ ] make hover state `#f7f7f7` instead of the current `#e6e6e6`

- [ ] align cards content to the top. see below

Screenshot 2020-02-24 at 16 51 08

- [ ] gutter is looking too wide. I removed the `margin-right: 1rem;` from each card and looked much better, making them wider.

Screenshot 2020-02-24 at 17 04 38
Screenshot 2020-02-24 at 16 59 33

- [ ] mobile: I suggest deleting the board images on small screens

matthewpaulthomas:
Agreed with all Ana’s points. Besides those, my comments are the same as [I wrote yesterday in Zeplin](https://zpl.io/V4xO40N):

- It’s _very_ unusual — actually I don’t recall a single example — for text+image buttons to have the text above the image, or to have horizontal rules inside them. I think the text would look much better below the image, with no horizontal rule, as it is [on snapcraft.io](https://snapcraft.io/first-snap/ros) for example.

- There are three clues that something is badly wrong with the layout of the OS version buttons:
  - they’re much taller than they are wide, which hardly ever happens with buttons
  - “64-bit” is being split across two lines (and if it wasn’t, the whole label would wrap to three lines!)
  - worst of all, the space for the captions is so narrow that most of the lines fit only a single word.

  One way to improve this would be to move the captions outside of the buttons (so that they can use the full width of the column, rather than having the button’s padding subtracted), and using the caption font size (as used for your e-mail address at the top) rather than body text size.

  If that’s not enough, I suggest switching to a vertical stack of plain Vanilla buttons, with captions to their right.

anasereijo:
> Agreed with all Ana’s points. Besides those, my comments are the same as [I wrote yesterday in Zeplin](https://zpl.io/V4xO40N):
> 
> * It’s _very_ unusual — actually I don’t recall a single example — for text+image buttons to have the text above the image, or to have horizontal rules inside them. I think the text would look much better below the image, with no horizontal rule, as it is [on snapcraft.io](https://snapcraft.io/first-snap/ros) for example.
> * There are three clues that something is badly wrong with the layout of the OS version buttons:
>   
>   * they’re much taller than they are wide, which hardly ever happens with buttons
>   * “64-bit” is being split across two lines (and if it wasn’t, the whole label would wrap to three lines!)
>   * worst of all, the space for the captions is so narrow that most of the lines fit only a single word.
>   
>   One way to improve this would be to move the captions outside of the buttons (so that they can use the full width of the column, rather than having the button’s padding subtracted), and using the caption font size (as used for your e-mail address at the top) rather than body text size.
>   If that’s not enough, I suggest switching to a vertical stack of plain Vanilla buttons, with captions to their right.

@matthewpaulthomas I'm sorry, for some reason I didn't get any notifications from Zeplin! I will have a look tomorrow and read your feedback. Thanks

anasereijo:
> * It’s _very_ unusual — actually I don’t recall a single example — for text+image buttons to have the text above the image, or to have horizontal rules inside them. I think the text would look much better below the image, with no horizontal rule, as it is [on snapcraft.io](https://snapcraft.io/first-snap/ros) for example.

@anthonydillon can you try Matthew's suggestion? Text below the image


> * There are three clues that something is badly wrong with the layout of the OS version buttons:
>   
>   * they’re much taller than they are wide, which hardly ever happens with buttons
>   * “64-bit” is being split across two lines (and if it wasn’t, the whole label would wrap to three lines!)
>   * worst of all, the space for the captions is so narrow that most of the lines fit only a single word.
>   
>   One way to improve this would be to move the captions outside of the buttons (so that they can use the full width of the column, rather than having the button’s padding subtracted), and using the caption font size (as used for your e-mail address at the top) rather than body text size.
@anthonydillon if it doesn't take you much time, could you quickly try this to see how it looks? otherwise, it would be a great opportunity to break the grid eheh joking!

anthonydillon:
I've given all @matthewpaulthomas suggestions a whirl. What do you think?
image

matthewpaulthomas:
Ahhhh the board buttons look so much better now, thank you.

For the OS buttons, by “captions to their right” I meant to the right of the buttons, rather than inside. But I can see that putting them inside provides a larger click target. Happy to defer to Ana on that one.

matthewpaulthomas:
Ana said in standup that she really doesn’t like the OS selector now. So here’s a quick sketch of what I meant by “a vertical stack of plain Vanilla buttons, with captions to their right”:

os-selector

@anasereijo Would that be better?

anasereijo:
- [ ]  make hover state #f7f7f7 instead of the current #e6e6e6 --- any reason why this wasn't done?

- [ ] mobile: please add spacing between buttons

Screenshot 2020-02-25 at 11 22 31

Regarding step number 2, I designed a few more options. I still think the 1st is the best one, but I don't mind the 2nd option either. Let us know your thoughts @matthewpaulthomas 

Image builder - signed in  Copy

anasereijo:
> Ana said in standup that she really doesn’t like the OS selector now. So here’s a quick sketch of what I meant by “a vertical stack of plain Vanilla buttons, with captions to their right”:
> 
> os-selector
> 
> @anasereijo Would that be better?

just saw this now :) I actually ended up designing the same thing! see above

anthonydillon:
@anasereijo 
> make hover state #f7f7f7 instead of the current #e6e6e6 --- any reason why this wasn't done?

This is the default hover for Vanilla buttons.

>  mobile: please add spacing between buttons

This is very difficult with the amount of bespoke styling. I tried for 30mins.

@matthewpaulthomas 
Updated the board selection section:
image

anthonydillon:
I just spent another night working on this section so I'm worried to point this out but I find it difficult to see what is selected.

anthonydillon:
Resetting all labels as the design and code has changed so much.

anasereijo:
> I just spent another night working on this section so I'm worried to point this out but I find it difficult to see what is selected.

I really don't like that grey, I much prefer the `#f7f7f7` but I understand it's vanilla, so maybe try using that colour as hover and selected.

anasereijo:
It's a shame how the buttons look on mobile, but if it's not possible we'll have to live with that.

anthonydillon:
@anasereijo Done, ready for review.

matthewpaulthomas:
Looking good to me now, thanks for all the revisions. 🌟

templates/core/build.html
```diff
@@ -62,80 +62,96 @@ <h1>Build an Ubuntu IoT image</h1>
...
-        <hr class="u-sv1">
-        <p class="p-card__content">Raspberry Pi2</p>
-      </button>
-      <button class="col-2 p-card js-selection">
-        <img class="p-card__thumbnail" src="https://assets.ubuntu.com/v1/c4a007b6-Raspberry+Pi+3.png" alt="">
-        <hr class="u-sv1">
-        <p class="p-card__content">Raspberry Pi3</p>
+      <button class="col-3 p-card js-selection u-no-margin--right" style="width:100%;height:100%;">
+        <img src="https://assets.ubuntu.com/v1/8542c958-Raspberry+Pi+2.png" alt="" style="height:130px;" class="u-hide--small">
+        <h4 class="js-name u-no-margin--bottom">Raspberry Pi2</h4>
```
sowasred2012: Forgive the nitpick - these titles are descendants of a `h2`, so should be `h3`.

```suggestion
        <h3 class="p-heading--four js-name u-no-margin--bottom">Raspberry Pi2</h4>
```

PR #6848, opened 2020-02-24
Usn pagination improvements

jpmartinspt:
## Done
Small improvements related to pagination for the security notices page.

## QA

- Make sure pagination works fine, and that weird page parameter values don't cause unexpected behaviour.

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -100,6 +100,9 @@ def notices():
     total_pages = ceil(total_results / page_size)
     offset = page * page_size - page_size
 
+    if page < 1 or 1 < page > total_pages:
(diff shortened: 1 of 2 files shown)


webapp/security/views.py
```diff
@@ -100,6 +100,9 @@ def notices():
     total_pages = ceil(total_results / page_size)
     offset = page * page_size - page_size
 
+    if page < 1 or 1 < page > total_pages:
```
jkfran: The first part of the second condition looks a bit redundant. Because you are using `or`, it means the number is greater than 1, so the second condition can just be `page > total_pages` without `1 < page`, right?

PR #6851, opened 2020-02-24
Restyle the snap selection section

anthonydillon:
## Done
Restyle the snap selection section

## QA
- Select your board and OS
- Check the snap selection functions by searching then adding and removing some snaps
- Check it matches [the design](https://app.zeplin.io/project/5e33f5075368883e5da9d0bf/screen/5e4fe76ec1b203753a5fa78b)

## Issue / Card
Fixes https://github.com/canonical-web-and-design/web-squad/issues/2402

## Screenshots
image

### DIFF
--- a/static/js/imageBuilder.js
+++ b/static/js/imageBuilder.js
@@ -67,7 +67,7 @@
       addButton.addEventListener('click', e => {
         e.preventDefault();
         const button = (e.target.classList.contains('js-add-snap'))?e.target:e.target.closest('.js-add-snap');
-        if (!lookup(snapSearchResults[button.dataset.index].package_name, 'package_name', STATE.snaps)) {
+        if (lookup(snapSearchResults[button.dataset.index].package_name, 'package_name', STATE.snaps) === false) {
           const selectedSnapContainer = button.closest('.js-snap-search-container');
           if (selectedSnapContainer) {
             selectedSnapContainer.classList.add('u-disable');
@@ -221,7 +221,7 @@
   function lookup(name, key, arr) {
     for (var i = 0, len = arr.length; i < len; i++) {
       if (arr[i][key] === name) {
-        return true;
+        return i;
       }
     }
     return false;
--- a/static/js/imageBuilder.js
+++ b/static/js/imageBuilder.js
@@ -111,7 +111,7 @@
         </span>`:'';
           results.insertAdjacentHTML('beforeend',
             `<div class="row js-snap-search-container" data-container-index="${index}">
-              <div class="col-5">
+              <div class="col-5 col-medium-5 col-small-3">
                 <div class="p-media-object u-no-margin--bottom" data-container-index="${index}">
                   <img src="${item.icon_url}" alt="" class="p-media-object__image">
                   <div class="p-media-object__details">
@@ -122,7 +122,7 @@
                   </div>
                 </div>
               </div>
-              <div class="col-1">
+              <div class="col-1 col-medium-1 col-small-1">
                 <a href="" class="p-button--base js-${buttonText.toLowerCase()}-snap" data-index="${index}"><i class="p-icon--${buttonIcon}">${buttonText}</i></a>
               </div>
             </div>
--- a/static/js/imageBuilder.js
+++ b/static/js/imageBuilder.js
@@ -57,8 +57,7 @@
   }, 250);
 
   function clearSearch() {
-    changeState('snaps', []);
-    renderSnapList(STATE.snaps, snapResults, 'Add');
+    renderSnapList([], snapResults, 'Add');
   }
 
   function addSnapHandler() {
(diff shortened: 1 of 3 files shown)


anasereijo:
- [ ] Most common additions missing

Screenshot 2020-02-25 at 09 58 57


- [ ] No matching snaps missing
Screenshot 2020-02-25 at 09 59 48

- [ ] I think it's weird when you select a snap and you still see it on the left (even though it's with opacity). Would be nice to have some kind of animation giving you the idea it moved to the other side?

Screenshot 2020-02-25 at 10 02 54

- [ ] when you 'clean' the search input I think you should also clean the results

Screen-Recording-2020-02-25-at-10 53

- [ ] if you search for something you searched before you get the same results even if you have selected some of them. With Docker it's even possible to add it twice

Screen-Recording-2020-02-25-at-11 10

- [ ] on mobile can you move the + and - to the right side to save some space?

Screenshot 2020-02-25 at 11 17 36

- [ ] on mobile can we show something like 4 results and also have the option to scroll?

matthewpaulthomas:
This is a huge improvement. UX–1 for _one_ issue:

- When a snap is already added, it’s faded out, and I can’t click “+” any more … but I can still focus “+” with the keyboard and press Enter, as many times as I like! The button needs to be _really_ disabled.

  image

_Non-blocker_ issues:

- The empty states aren’t implemented (as Ana said while I was writing this).

- Placeholders in text fields should be [avoided if at all possible](https://adamsilver.io/articles/placeholders-are-problematic/). This field has a search icon and an “Available snaps:” label above, so the placeholder “Search for snaps:” is gratuitous.

- With a board or OS version button focused, pressing Space selects that option, as I’d expect. But with a “+” or “–” button focused, pressing Space … scrolls the page.

- Clearing the search field doesn’t clear the search results.

  image

- As I said in Zeplin, I don’t think it’s obvious that “–” at the end of a row is a button. So I’d rather the “+” and “–” buttons have borders around them like normal Vanilla buttons. They’ll probably need borders anyway if/when items become highlightable.

And for the future:

- Animation when you choose an item (as Ana said above)

- I don’t think you should have to press Tab 13 times to get from the left list to the right. I think you should be able to press Tab just once, and therefore, navigating through list items should be done with arrow keys instead, like it is in native listboxes. (That would require a highlight, as in native listboxes, to show the currently selected item.)

anthonydillon:
> Most common additions missing

This is not possible (yet)

> I think it's weird when you select a snap and you still see it on the left (even though it's with opacity). Would be nice to have some kind of animation giving you the idea it moved to the other side?

You and @matthewpaulthomas need to work out if you want it to hide or disable a selection?

> if you search for something you searched before you get the same results even if you have selected some of them. With Docker it's even possible to add it twice

For now I have fixed the bug of adding the same snap.

@matthewpaulthomas I have covered all the main items and some of the wish list.

anasereijo:
- [ ] when you remove snaps from your list they should go back to the available snaps

Screen-Recording-2020-02-27-at-10 36

- [ ] can't remove a snap from my list after I clear the search

Screen-Recording

- [ ] after adding a snap to my list, if I search for something else and add another snap it removes the one I added previously

Screen

from my last comment, this is still happening:
- [ ] if you search for something you searched before you get the same results even if you have selected some of them. With Docker it's even possible to add it twice

Screen-Recording-2020-02-25-at-11 10

anthonydillon:
@anasereijo 
> if you search for something you searched before you get the same results even if you have selected some of them. With Docker it's even possible to add it twice

I cant add the same one twice but what would you expect to happen if you search again?

matthewpaulthomas:
> I cant add the same one twice but what would you expect to happen if you search again?

Which snaps are in the “Your chosen snaps” list should always determine:
- whether a snap in the “Available” pane is disabled.

Which snaps are in the “Your chosen snaps” list should _not_ affect:
- which snaps show up for a search
- the order in which snaps show up for a search
- anything else about the “Available” pane.

matthewpaulthomas:
UX –1 now for two points:

- [ ] Ana’s last point above — a currently-chosen snap, in a new search, appears enabled when it isn’t.

- [ ] Clearing the search field shows a spinner, then says “No matching snaps”. Neither should happen. Until “Most common additions” becomes implementable, just leaving the list empty (as it is initially) would be more coherent than referring to “matching” with nothing.

New, but non-blocking, problems:

- The spinner is a good addition, but I think “Loading snaps” could be confused with loading them onto the board/image. The context makes obvious what the task is, so I think the spinner would be fine by itself without any text.

- The search is needlessly redone (with fast Internet+device, you may just see the results flickering) if you do any of these things:
  - switch to another tab, then switch back
  - Tab out of the search field
  - Tab into the search field
  - press the Left, Right, Up, or Down arrow keys.

anthonydillon:
@matthewpaulthomas I have addressed all your points. Ready for another round.

matthewpaulthomas:
Looks good!

Non-blocker issues:
- Reloading the page persists the search field contents (probably shouldn’t), but results for that search aren’t shown.
- Clearing the search defocuses it — but that seems to be a Vanilla issue, so I’ve reported it as [vanilla-framework#2866](https://github.com/canonical-web-and-design/vanilla-framework/issues/2866).

PR #6881, opened 2020-02-26
Move JS to make linting easier

### LINKED ISSUE #6880: Reorganise JS to make linting easier
We don't want to lint build files or third-party so the JS directory should be organised in a way where we can point the linter at one directory only

steverydz:
## Done

- Restructure JS directory to make linting easier
- Delete unused files (cookie-proxy.js and rtp.js)

## QA

- Run `./run` and make sure the build doesn't fail

## Issue / Card

Fixes #6880


templates/base_index.html
```diff
@@ -55,7 +55,7 @@ <h4><a class="article-link article-title"></a></h4>
         </div>
       </template>
 
-      <script src="{{ versioned_static('js/latest-news.js') }}"></script>
+      <script src="{{ versioned_static('js/src/src/latest-news.js') }}"></script>
```
sowasred2012: I think this might be a typo?

steverydz: Thanks, good catch!

PR #6902, opened 2020-02-27
Update the thank you page

anthonydillon:
## Done
Update the thank you page

## QA
- Make a selection and click build
- See that you land on a page that matches [the copy doc](https://docs.google.com/document/d/1Hw_8dWAAUpHHm5Z_XNUVJaSVE5k5i2Vt6WANcJYUZvY/edit#) and [the design](https://github.com/canonical-web-and-design/web-squad/issues/2370#issuecomment-590955286)

## Issue / Card
Fixes https://github.com/canonical-web-and-design/web-squad/issues/2363

## Screenshots
image

### DIFF
--- a/static/js/imageBuilder.js
+++ b/static/js/imageBuilder.js
@@ -71,9 +71,8 @@
           const selectedSnapContainer = button.closest('.js-snap-search-container');
           if (selectedSnapContainer) {
             selectedSnapContainer.classList.add('u-disable');
-            e.target.disabled = true;
           }
-          changeState('snaps', snapSearchResults[e.target.dataset.index], 'ADD');
+          changeState('snaps', snapSearchResults[button.dataset.index], 'ADD');
           renderSnapList(STATE.snaps, preinstallResults, 'Remove');
           removeSnapHandler();
         }
@@ -92,10 +91,9 @@
           const revealItem = snapResults.querySelector(`[data-container-index="${searchIndex}"]`)
           if (revealItem) {
             revealItem.classList.remove('u-disable');
-            e.target.disabled = false;
           }
         }
-        changeState('snaps', e.target.dataset.index, 'REMOVE');
+        changeState('snaps', button.dataset.index, 'REMOVE');
         renderSnapList(STATE.snaps, preinstallResults, 'Remove');
         removeSnapHandler();
       });
@@ -120,7 +118,7 @@
           results.insertAdjacentHTML('beforeend',
             `<div class="row js-snap-search-container ${disable}" data-container-index="${index}">
               <div class="col-5 col-medium-5 col-small-3">
-                <div class="p-media-object u-no-margin--bottom" data-container-index="${index}">
+                <div class="p-media-object u-no-margin--bottom" data-index="${index}">
                   <img src="${item.icon_url}" alt="" class="p-media-object__image">
                   <div class="p-media-object__details">
                     <h1 class="p-media-object__title" style="line-height: 1.4rem">${item.title}</h1>
--- a/static/js/imageBuilder.js
+++ b/static/js/imageBuilder.js
@@ -1,4 +1,4 @@
-(function () {
+(function() {
   // State management
   class StateArray extends Array {
     push(item) {
@@ -21,9 +21,9 @@
   class StateManager {
     constructor() {
       this.states = {
-        "board": new StateArray(),
-        "os": new StateArray(),
-        "snaps": new StateArray()
+        board: new StateArray(),
+        os: new StateArray(),
+        snaps: new StateArray()
       };
     }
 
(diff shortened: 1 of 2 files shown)


anasereijo:
I don't have much to say because I am waiting for @matthewpaulthomas to review the [design options](https://github.com/canonical-web-and-design/web-squad/issues/2370#issuecomment-590955286) I've done. Then I can review this.

anasereijo:
@anthonydillon please update with the latest design https://github.com/canonical-web-and-design/web-squad/issues/2370

anthonydillon:
@anasereijo done.

Screenshot_2020-02-27 Ubuntu image builder Ubuntu

anasereijo:
@anthonydillon please add the board as the first item of the list as per @matthewpaulthomas comment in my design issue

anthonydillon:
@nottrobin is it possible to know the board selected on the thank you page?

matthewpaulthomas:
@anthonydillon You have to know the board to implement the list at all, because the list depends on which board you choose. Sorry I hadn’t made this clear enough in the copydoc — I’ve just split out the lists for Pi 2/3 and Pi 4 for starters.

matthewpaulthomas:
–1 because it gives you Raspberry-Pi-specific instructions when you chose the Intel NUC or Dragonboard. For those two boards, where we don’t have a checklist yet, just omit the whole section.

anthonydillon:
@anasereijo @matthewpaulthomas I have addressed the comments and wired-up the states build to be posted to the backend.

anasereijo:
@anthonydillon can we see a screenshot?

anthonydillon:
From snap dragon:
Screenshot_2020-02-28 Ubuntu image builder Ubuntu

From raspberrypi:
image

anasereijo:
@anthonydillon I thought @matthewpaulthomas asked for the board to be the first item on the list?

anthonydillon:
@matthewpaulthomas are you happy without the board name on the thank you page for now until I have all the data passed through?

matthewpaulthomas:
@anthonydillon Yes, that’s fine.

templates/core/build.html
```diff
@@ -175,15 +191,15 @@ <h4>Your chosen snaps:</h4>
...
+        <form method="POST" id="build-form">
           <p>We’ll email you at {{ openid.email }} once the build is ready.</p>
           <p>
             <input class="mktoField" value="yes" id="canonicalUpdatesOptIn" name="canonicalUpdatesOptIn" type="checkbox">
             <label class="mktoLabel mktoHasWidth" for="canonicalUpdatesOptIn">I agree to receive information about Canonical’s products and services.</label>
           </p>
-          <input type="hidden" name="board" value="cm3" />
-          <input type="hidden" name="system" value="core16" />
-          <input type="hidden" name="snaps" value="toto,code,toto-fran" />
+          <input type="hidden" name="board" value="" />
```
nottrobin: Weren't you gonna make these text fields / radio buttons or something?

anthonydillon: No, I think its easier to update if you just add a new card instead of both the card and the value in a radio button.

static/js/imageBuilder.js
```diff
@@ -204,12 +205,42 @@
 
   function render() {
     renderSummary();
+    updateForm();
+    checkDisabled();
   }
 
-  function changeState(key, value) {
-    STATE[key] = value;
+  function changeState(key, value, action) {
```
nottrobin: It seems to me you're writing a lot of code to turn the interface from a declarative method-based one into an obscure function that expects strings with very specific required values that are hard to guess.

How are these functions meaningfully different from the default options available on an array?

``` javascript
const STATE = {
  'board': [],
  'os': [],
  'snaps': []
};

# UPDATE
STATE.board = ["new", "list"];

# ADD
STATE.board.push("newvalue")
```

I can't quite tell what you're intending with "REMOVE" - are you expecting `value` to contain the *index* of an item to remove? That seems odd, but if so then `STATE.board.splice(3, 1)` seems potentially clearer than `changeState("board", 3, "REMOVE")`.

If you wanted to be able to do `STATE.board.remove("somevalue")`, which seems like a reasonable thing to want to do, then that could be achieved with:

``` javascript
class StateArray extends Array{
  remove(itemValue) {
    const index = array.indexOf(5);
    if (index > -1) {
      this.splice(index, 1);
    }
  }
}

const STATE = {
  'board': StateArray(),
  'os': StateArray(),
  'snaps': StateArray()
};
```

But I also have a few other questions:

- Why is `STATE` upper-case? I know it's technically a `const` but you're hardly treating it as generally immutable, which is normally what upper case would indicate...
- Do you need state at all? You could simply do e.g. `boardState = StateArray()`
- Why do your initial values for `STATE` include 2 strings and 1 array? Surely you'd want all your states to be of the same basic type...

templates/core/build.html
nottrobin: Oh so there isn't a JS-less option?

static/js/imageBuilder.js
anthonydillon: After trying to implement these changes for a while today I ran into a number of issues:
- There is no native reassignment function on arrays therefore I created an update function but that basically called remove and then push which seemed awkward.
- The board and os variables are only ever a single value and so I would have to get the values with `state.board[0]` throughout the code which I really don't like.

If you feel strongly @nottrobin I am happy to give it another whirl but this branch has been in for a while. Let me know what you think?

nottrobin: I played around with `watch`, which is deprecated, and then `__defineSetter__` etc., but the best I could come up with was the more straightforward approach, something like this:

``` javascript
class StateArray extends Array {
  push(item) {
    const result = super.push(item);
    render();
    return result;
  }
  pop() {
    const item = super.pop();
    render();
    return item;
  }
  remove(item) {
    const index = this.indexOf(item);
    if (index > -1) {
      const item = this.splice(index, 1);
    }
    render();
    return item;
  }
}

class StateManager {
  states = {};

  set(name, items) {
    this.states[name] = new StateArray(...items);
    render();
  }

  get(name) {
    return this.states[name];
  }
}

state = new StateManager();


function render() {console.log("rendered", state.states);}


state.set("board", [1,2,3,4]);
state.set("os", [1,2,3,4]);
state.set("snaps", [1,2,3,4]);


state.get("board").remove(3);
state.get("snaps").pop(3);
state.get("os").push(10);
```

Maybe this? If you don't think this works then let's just forget this for now and continue with your code.

templates/core/build.html
nottrobin: ^ ?

anthonydillon: I think I need a follow up card to make this UI work without JS as I need to work out the snaps input too.

anthonydillon:
@nottrobin applied the new state management system. Ready for another look.

PR #6933, opened 2020-03-02
K8s charm pages

evilnick:
## Done

Add 18 or so new pages of charm reference.

## QA


evilnick:
@pmahnke ah yes, i didn't notice that, seems to be when I made a change to fix up the page that was broken. I will take another look thanks

PR #6973, opened 2020-03-05
Add JS test command to build steps

steverydz:
## Done

Add JS testing as a build step

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,16 @@ jobs:
           command: |
             coverage run  --source=. -m unittest discover tests
             bash <(curl -s https://codecov.io/bash) -cF python
+  test-js:
+    <<: *defaults
+    steps:
+      - checkout
+      - run:
+          name: Install dependencies
+          command: yarn
+      - run:
+          name: Test JS
+          command: yarn test-js
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


steverydz:
@nottrobin now we have a (very) modest JS test suite it makes sense to add it as a step in the build process. Based on the other entries in the CircleCI config I'm sure this is fine but I don't know how to test it.

nottrobin:
Nice! Where's the actual test code? I'm struggling to find it.

steverydz:
> Nice! Where's the actual test code? I'm struggling to find it.

There's only two at the moment. Anything with `.test.js` at the end basically:

https://github.com/canonical-web-and-design/ubuntu.com/tree/master/static/js/src

.circleci/config.yml
```diff
@@ -86,6 +86,16 @@ jobs:
...
+      - checkout
+      - run:
+          name: Install dependencies
+          command: yarn
+      - run:
+          name: Test JS
+          command: yarn test-js
 workflows:
   version: 2
   build:
```
nottrobin: To get this running in CI you need to add `- test-js` to the `jobs:` list below.

steverydz: done

PR #6996, opened 2020-03-12
Disable notices API

tbille:
## Done

- To get a first draft of the notices we are disabling the api for now.

## QA

- run `./run --env FLASK_APP=webapp.app:app exec flask routes`
- There should be no POST routes for `/security/notices`


nottrobin:
This might be a case where I'd suggest commenting out instead - 'cos we presumably want to easily understand how to turn this back on?

PR #6695, opened 2020-02-19
Image builder

anthonydillon:
## Done
This branch contains the entire Image builder feature.

## QA

QA at http://imagebuilder.staging.demo.haus/build

- Go to /build
- Sign in
- Select a board, OS and optionally add some snaps.
- Click Build
- You should see a thank you page
- You will **not** receive the completed email as the webhook is targeting production. (It's likely @nottrobin will get the email which he may kindly forward for your QA)
- Once you have the email click the links in the email and check they work

Fixes https://github.com/canonical-web-and-design/web-squad/issues/2387

PR #7028, opened 2020-03-18
Image builder address design review comments

anthonydillon:
## Done
- Restyle the login section
- Fix the disable overlap
- Always display Got an app that isn’t in the Snap Store
- Conditionally display email in build page
- Update the footer content 

## QA

- Check I have addressed al lthe comment here: https://github.com/canonical-web-and-design/ubuntu.com/pull/6695#issuecomment-600053523

### DIFF
--- a/templates/core/build.html
+++ b/templates/core/build.html
@@ -11,7 +11,7 @@
   <div class="row">
     <div class="col-8 u-sv3">
       <h1>Your Ubuntu image is building</h1>
-      <p>We’ll email you at {{ openid.email }} when we’ve finished. A build usually takes less than 30 minutes.</p>
+      <p>We’ll email you {% if openid.email %}at {{ openid.email }} {% endif %}when we’ve finished. A build usually takes less than 30 minutes.</p>
       <p><a href="{{ build_info.web_link }}" class="p-link--external">Build log</a></p>
     </div>
   </div>
--- a/templates/core/build.html
+++ b/templates/core/build.html
@@ -220,6 +220,8 @@
       </div>
     </div>
   </div>
+</section>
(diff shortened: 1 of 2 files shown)


templates/core/build.html
```diff
@@ -31,239 +31,226 @@ <h1>Your Ubuntu image is building</h1>
...
-          <input type="hidden" name="FullName" value="{{ openid.fullname }}" />
-          <input type="hidden" name="Email" value="{{ openid.email }}" />
-          <p><button class="p-button--positive js-build-button">Build image</button></p>
-        </form>
-      </div>
+<section class="p-strip--light is-shallow  {% if not openid %}u-disable{% endif %}">
+  <div class="row">
+    <div class="col-8">
+      <form method="POST" id="build-form">
+        <p>We’ll email you at {{ openid.email }} once the build is ready.</p>
```
solazio: I don't think the text makes too much sense if there is no emali.

anthonydillon:
Oh sorry you need the lp token to trigger builds but that part has been tested and hasn't changed in this branch.

templates/core/build.html
anthonydillon: When did you see it without a email?

matthewpaulthomas: Yep, in #6695 I said “It’s a bit unpolished that when signed out, we say ‘We’ll email you at once the build is ready.’ Could we drop the ‘at’ in that case?’”

solazio: @anthonydillon it's at the bottom of the page.

anthonydillon: Ah, I see. I have hidden the message if there is no email.

PR #7038, opened 2020-03-19
Add github action to test docker images

carkod:
## Done

- Add github action to test docker images

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - name: Update pip
+      run: pip install setuptools --upgrade
     - name: Build image
       run: docker build --tag ubuntu-com .
     - name: Run image
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,16 @@
+name: Test docker image
+on: pull_request
+env:
+  DOCKER_BUILDKIT: 1
+  DATABASE_URL: sqlite:///
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build image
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Update pip
+      run: pip install setuptools --upgrade
+    - name: Build image
+      run: docker build --tag ubuntu-com .
+    - name: Run image
+      run: |
+        docker run -d -p 80:80 ubuntu-com
+        sleep 1 && curl --head --fail --retry-delay 5 --retry 10  --retry-connrefused http://localhost
```
tbille: Just in case we don't have issues with possible requests that could hang could you curl this endpoint instead: `http://localhost/_status/check`

.github/workflows/docker.yml
```diff
@@ -0,0 +1,18 @@
...
+env:
+  DOCKER_BUILDKIT: 1
+  DATABASE_URL: sqlite:///
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Update pip
+      run: pip install setuptools --upgrade
```
tbille: Can you explain why this has to be added?

.github/workflows/docker.yml
tbille: Also just to be sure. Vanilla run the commands in 2 different jobs: https://github.com/canonical-web-and-design/vanilla-framework/blob/master/.github/workflows/pr_docker_build.yml#L13-L19

It seems cleaner and maybe the `sleep 1` might not be necessary.

.github/workflows/docker.yml
carkod: That was an attempt to fix the `bdist_wheel error` I am having

carkod:
somebody else did it.

PR #7042, opened 2020-03-20
UA renewal info markup

sowasred2012:
# Done

- Updated the layout of /advantage based on the feedback received below. When a renewal is due to expire soon, there is now information to that effect in the table indicating how many days are left. This PR doesn't provide a way for the user to do anything about it, that will come with the [Stripe integration work](https://github.com/canonical-web-and-design/ubuntu.com/issues/6202).

# QA

**QAing this is dependent on you having some subscriptions on the staging contracts service**

- Check out this branch, or visit the demo link
- Visit /advantage
- Log out, if you aren't already logged out
- See that the page looks good
- Click the sign in button at the top of the page
- If you have subscriptions associated with your account that have "effective to" dates, see that there is warning information indicating how many days until they expire.
- See that the page looks good

Fixes https://github.com/canonical-web-and-design/web-squad/issues/2481

## Screenshot

FireShot Capture 002 - Ubuntu Advantage for Infrastructure - Ubuntu_ - ubuntu-com-canonical-web-and-design-pr-7042 run demo haus

### DIFF
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -22,7 +22,7 @@
       <div class="col-12">
         <h1>Ubuntu Advantage for&nbsp;Infrastructure</h1>
         <p>You are signed in as {{ openid.fullname }}<br />
-        <small style="color:#666">({{ openid.email }})</small></p>
+        <small style="color: #666;">({{ openid.email }})</small></p>
         <p class="u-no-margin--bottom">
           <a href="/logout?return_to={{ request.host_url }}advantage" class="p-button--neutral u-no-margin--bottom" onclick="dataLayer.push({
             'event' : 'GAEvent',
@@ -99,7 +99,7 @@
                               {{ contract['contractInfo']['name'] }} &nbsp;<i class="p-icon--contextual-menu">Open</i>{% if expiring_soon %}<div class="" style="color: #666; font-size: 14px;"><i class="p-icon--warning"></i>&nbsp;&nbsp;Ends in {{ contract["contractInfo"]["daysTillExpiry"] }} d</div>{% endif %}
                           </a>
                       </td>
-                      <td style="text-transform: capitalize"><span class="u-truncate" style="padding-top:.35rem;">{{ contract['supportLevel'] }}</span></td>
+                      <td style="text-transform: capitalize;"><span class="u-truncate" style="padding-top: 0.35rem;">{{ contract['supportLevel'] }}</span></td>
                       <td class="u-align--center {% if loop.index == 1 and outer_loop.index == 1 %}p-table--open{% endif %}">
                         <button class="u-toggle" aria-controls="#expanded-row-token-{{ outer_loop.index }}-{{ loop.index }}" aria-expanded="{% if loop.index == 1 and outer_loop.index == 1 %}true{% else %}false{% endif %}" data-shown-text="Hide" data-hidden-text="Show">
                           {{ contract['machineCount'] }} &nbsp;<i class="p-icon--contextual-menu {% if loop.index == 1 and outer_loop.index == 1 %}u-rotate{% endif %}">Open</i>
@@ -112,7 +112,7 @@
                           <span class="u-truncate"><i class="p-icon--success"></i> {% if contract['entitlements']['esm-infra'] == True %}Infra{% else %}Apps{% endif %} <i class="p-icon--contextual-menu">Open</i></span>
                         </button>
                         {% else %}
-                        <span style="padding-top:.35rem;">N/A</span>
+                        <span style="padding-top: 0.35rem;">N/A</span>
                         {% endif %}
                       </td>
                       <td class="u-align--center">
@@ -121,7 +121,7 @@
                           <span class="u-truncate"><i class="p-icon--success"></i> &nbsp;<i class="p-icon--contextual-menu">Open</i></span>
                         </button>
                         {% else %}
-                        <span style="padding-top:.35rem;">N/A</span>
+                        <span style="padding-top: 0.35rem;">N/A</span>
                         {% endif %}
                       </td>
                       <td class="u-align--center">
@@ -130,7 +130,7 @@
                           <span class="u-truncate"><i class="p-icon--success"></i> &nbsp;<i class="p-icon--contextual-menu">Open</i></span>
                         </button>
                         {% else %}
-                        <span style="padding-top:.35rem;">N/A</span>
+                        <span style="padding-top: 0.35rem;">N/A</span>
                         {% endif %}
                       </td>
                       <td class="u-align--center">
@@ -139,7 +139,7 @@
                           <span class="u-truncate"><i class="p-icon--success"></i> &nbsp;<i class="p-icon--contextual-menu">Open</i></span>
                         </button>
                         {% else %}
-                        <span style="padding-top:.35rem;">N/A</span>
+                        <span style="padding-top: 0.35rem;">N/A</span>
                         {% endif %}
                       </td>
 
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -90,7 +90,7 @@
                   {% if contract["contractInfo"]["status"] != 'expired' %}
 
                     {% if contract["contractInfo"]["daysTillExpiry"] %}
-                      {% set expiring_soon = contract["contractInfo"]["daysTillExpiry"] < 3000 %}
+                      {% set expiring_soon = contract["contractInfo"]["daysTillExpiry"] <= 30 %}
                     {% endif %}
 
                     <tr class="p-table__row">
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -87,8 +87,8 @@
                     {% endif %}
 
                     <tr class="p-table__row">
-                      <td class="u-no-padding--left">
-                          <a href="#" class="p-button u-toggle u-toggle--full-width u-align--left" aria-controls="#expanded-details-{{ outer_loop.index }}-{{ loop.index }}" aria-expanded="false" data-shown-text="Hide" data-hidden-text="Show" style="padding-left: 0.5rem;">
+                      <td>
+                          <a href="#" class="p-button u-toggle u-toggle--full-width u-align--left u-no-padding--left" aria-controls="#expanded-details-{{ outer_loop.index }}-{{ loop.index }}" aria-expanded="false" data-shown-text="Hide" data-hidden-text="Show" style="padding-left: 0.5rem;">
                               {{ contract['contractInfo']['name'] }} &nbsp;<i class="p-icon--contextual-menu">Open</i>{% if expiring_soon %}<div class="u-toggle__supplemental" style="color: #666; font-size: 14px;"><i class="p-icon--warning"></i>&nbsp;&nbsp;Ends in {{ contract["contractInfo"]["daysTillExpiry"] }} days</div>{% endif %}
                           </a>
                       </td>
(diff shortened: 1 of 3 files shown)


matthewpaulthomas:
I think you’re exactly right to add warning icons in the first column. Unfortunately this makes more obvious an existing problem, which I guess I hadn’t noticed before, because in Ant’s sampledata all the plans were identical…

image

The warning icons should really line up — and so should the chevrons. So, the chevrons should be at the right end of the cell. (One way to implement that would be for the chevrons to secretly have their own `<td>`, with the header being `<th colspan="`…`">`.) That in turn requires realigning the header cells, to left align with the text rather than centering over it.

cassiocassio:
@lyubomir-popov has done work on positioning status icons before items (when they all have them) and as suffixes (when they are rare)...

Our standard for data is that numbers (and their headers) range right (so that units, tens, hundreds line up, because place-value, because maths), and that text ranges left

UXwise, the most importantly discoverable item in the whole table  is the chevron - and I guess we got somewhat lucky with the green tick dragging your eye there.

lyubomir-popov:
Just thought I'd link to the icons [spec](https://discourse.ubuntu.com/t/icon-positioning-within-tables/14435/2) @cassiocassio mentioned. Happy to help address the points Martin mentioned unless someone else is already working on this.

sowasred2012:
@cassiocassio I've just pushed an update to the icon alignment based on a conversation @matthewpaulthomas had that doesn't take into account which way numbers should range (but should). We settled on aligning the chevrons to the right of their column and aligning everything else to the left, but we weren't completely happy with it.

Although UX is not quite my wheelhouse, I wondered if this sketch I hastily threw together would tackle a few of your points:

1: Although it wouldn't tell you when each one ended, you could draw attention to the fact that the user needs to take action by having a notification saying something like "One or more of your subscriptions is about to end/expire". The warning symbol then appears (as it currently does in this demo) to create the association between that message and the subscriptions that require action.

2: Labelling - we could nest a table in the dropdown?

3: If the subscription is expiring soon, we could do something as simple as, again, adding the warning icon to the cell that contains the impending expiry date. Another approach is to set a different background color on the cell, and include a second notification above this nested table stating exactly how many days there are until expiry.

4: I'll leave that you guys ;p

5: If we use the nested table, we could just show/hide an "Actions" column (or the column heading stays empty), and in the cell include a "Renew" button.

sowasred2012:
I've pushed another update to have the icons follow the spec @lyubomir-popov mentioned

lyubomir-popov:
note to myself, wireframe [here](https://docs.google.com/presentation/d/1Og6UeBaxAfBrcoD79PBLl-HyO4reeVNhNa2nhcumfZ0/edit#slide=id.g6db70580c6_1_62)

matthewpaulthomas:
> I wish we could see "subscription ends in 2 days without clicking

Maybe this could appear conditionally, depending on the cell width and name length, so that it’s next to the icon that it’s explaining:

subscription-ending-soon

If it would seldom appear that way, given current page widths and name lengths, an alternative layout would be small text below the subscription name.

> The labels for "Created on:" etc seem very heavy, more important than (3). the data they label

Agreed.

> Could the button be made to belong to the thing it's renewing?

Yes. Regardless of where buttons end up going generally, I think _this_ button should go immediately below the end date:

subscription-ending-soon-button

No need for a separate Vanilla notification, just draw attention to what’s already there.

cassiocassio:
If there was a separate Vanilla notification, perhaps it's one of those more global ones at the top that says "You have an subscription/s expiring in n days"

The colour would be enough to then understand what the warning icons were about even if the text was truncated.

-

Separate question:

Have we got the whole flow for this?

I know Ant's POC used a lightbox for what happens next when you click "Renew" which isn't insane, but is just one option.

cassiocassio:
many thanks @sowasred2012 :  UX is in everyone's wheelhouse! This is the design team! We all care about the user's experience. It's just an unfortunate quirk of job titles that UX has become a verb.

lyubomir-popov:
Just noticed that clicking on the "Add another subscription" button opens a strange slightly broken panel, so maybe it could reuse whatever container the Renew button triggers?

cassiocassio:
@Lyubomir Popov <lyubomir.popov@canonical.com> can you share with @Matthew
Thomas <mpt@canonical.com> the work you and @Lilyana Videnova
<lilyana.videnova@canonical.com> have done with single page "cards" in MAAS
settings as a _possibility_ for the "You're now renewing / paying"

matthewpaulthomas:
> Have we got the whole flow for this?

Reported [web-squad#2515](https://github.com/canonical-web-and-design/web-squad/issues/2515) for that.

> Just noticed that clicking on the "Add another subscription" button opens a strange slightly broken panel, so maybe it could reuse whatever container the Renew button triggers?

It’s strange in that it’s full-width, so that the table inside it can be the same width as the same table when it appears outside the panel. (Following the general principle that it’s more important to be consistent with things nearby than things far away.)

It’s slightly broken partly because I’ve seldom been able to test it, because the UA team haven’t been able to provide test subscriptions. I’ve now reported #7065 after testing with this demo.

I don’t think it would would be a good idea for adding a new subscription, and renewing an existing subscription, to use the same panel, whatever its format. They differ, massively, in how much they contain.

In fact, the only new piece of information in the renewal panel is the price. So if we can’t include/embed payment fields in that panel, I’d rather not have it at all. Instead, we could simply put the price next to the “Renew” button in the expanded row.

lyubomir-popov:
@cassiocassio @matthewpaulthomas @sowasred2012 @anthonydillon I've tried to address most of the comments on this thread, plus a few concerns of my own. I've done a lot of modifications in the browser and then ovelrayed a sketch mockup of the table. You can find it on zeplin - [web link](zpl://screen?sid=5e7b72117eef528de4b59d25&pid=5e70c2e3d8982a16dfc6ce94), [app link](https://zpl.io/bzoZggl):

In order from top to bottom:
- I find Matthew's hand drawn version of the hero element (no card) better in the wireframe than the pr, so I've dropped the card
- bigger hero strip, for a clear separation between page "head" and then the content
- dropped the grey background
- used fixed width `<hr>`s between sections
- `<h4>`s instead of h2s for section headings
- tried to improve the spacing within sections, currently there's a lot of white space before the second h2, not so much before the first one...

and then the table itself:
- none of the content in the expanded panels looks good when placed within panels this wide; in maas we have the concept of in-table actions on the machine listing, which I think would work better here. The contextual menu container is only as wide as it needs to be to fit the content
- If we Matthew insists on aligning the chevrons, then we need a background at least on the td that houses the chevron that was clicked. That's because the chevron is closer to the text in the next cell, so just to disambiguate. I can see that's how Matthew has originally specced it in his wireframe.
- I've introduced a second row so the "Ends in 2 days" notification is not in danger of colliding with the Subscription name regardless of screen width

cassiocassio:
Thanks. Comments on the Zeplin.

matthewpaulthomas:
In Friday’s review call, and the Zeplin, I’ve described why I think we should continue using expanded rows and not have a dedicated column for icons. But Lyubo’s layout gave me an idea for a design that makes the reminder text fully visible in collapsed state, and places it in context in the expanded state, without duplication.

•|**Collapsed**|**Expanded**
-|-|-
**Unrenewable** (e.g. too soon)|unrenewable-0|unrenewable-1
**Renewable**|renewable-0|renewable 1
**Imminent**|imminent-0|imminent-1

So, when you expand an imminently-expiring subscription, the caution icon+text scoot out from under the name, and flank the normal end-date display.

lyubomir-popov:
@matthewpaulthomas I've added a [mockup](https://zpl.io/2Gl8N3E) that closely follows the wireframes above.

lyubomir-popov:
@matthewpaulthomas @sowasred2012 

I've added some comments on top of the screenshot:

UA expanded panel - QA feedback

lyubomir-popov:
Link to double row spec mentioned in the comments on the image: https://docs.google.com/document/d/1GqH96hAymC3KQCZPhBdzWChxWsjy72szrRqoX9Ep-Pg/edit#bookmark=id.23xtbrjr1qny

matthewpaulthomas:
- I know it’s copying the mockup, but I don’t see any reason for “You are signed in as {name} {address} [ Sign out ]” to be in two font sizes and colours across three lines. That’s visually noisy, and scrolls needlessly. Someday that part of the UI won’t be in the main content area at all (it’ll be in the header), so it’ll take up even less space.

- I know it’s copying the mockup, but … `<h2 class="p-heading--four">`? Why? I think the `<h2>`s should be exactly the same size, when you’re signed in, as when you’re signed out.

  - And if they were, the section separators wouldn’t be necessary, which would further reduce noise.

- None of the column headings are lined up with their contents. That isn’t critical for this branch, I’m just mentioning it because you fixed it earlier, so I’m wondering where the fixes went.

  image

- The name cell’s secondary text says “d”, which should be “days”. (I hope that isn’t because I positioned a Zeplin note obscuring the “ays”!)

  image

- As I noted both above and in Zeplin, when the row is expanded, the name cell’s secondary text should be hidden, avoiding duplication.

  image

- For the same reason, and for brevity, “Effective to:” should be “Ends:”

(To be continued…)

lyubomir-popov:
> * I know it’s copying the mockup, but I don’t see any reason for “You are signed in as {name} {address} [ Sign out ]” to be in two font sizes and colours across three lines. That’s visually noisy, and scrolls needlessly. Someday that part of the UI won’t be in the main content area at all (it’ll be in the header), so it’ll take up even less space.

just moved what was already there, happy with @matthewpaulthomas 's suggestion

> * I know it’s copying the mockup, but … `<h2 class="p-heading--four">`? Why? I think the `<h2>`s should be exactly the same size, when you’re signed in, as when you’re signed out.

just tried to make it take up less vertical space, felt like too many headings, but you're right it shouldn't change based on whether you're signed in. If smaller headings don't work in the signed-out case, then happy to do what you say.

cassiocassio:
The logged in box is a bit a hack that none of us were that proud of, happy to clean it up 

(the real solution is some indicator of logged in identity in the global navigation - out of the scope of this piece of work)

matthewpaulthomas:
Beautiful work. One issue remaining:

- [ ] The contents of the expanded panel is way off-center — it’s in a pair of flex boxes that’s centered, but that’s no good if the text will never be close to reaching the left edge of the left box. The expanded cells for the other columns don’t have this problem.

  image

> the free section is less related to the footer than it is to the paid section, so it should be at least as far away from the footer as it is from the paid section.

This is fixed now, and looks fixed whenever the table’s last row is expanded, but looks unfixed whenever it’s collapsed. I guess that’s a problem in Vanilla: the lack of a bottom border, in the collapsed state, leads the eye to measure from the bottom of the row’s text rather than from the bottom of the row.

matthewpaulthomas:
Lots of polish to come later, but the renewal info looks good, thank you.

templates/advantage/index.html
```diff
@@ -39,257 +33,264 @@ <h1>Ubuntu Advantage for&nbsp;Infrastructure</h1>
...
+                  <th class="u-hide">&nbsp;</th>
+                  <th class="u-hide">&nbsp;</th>
+                </tr>
+              </thead>
+              <tbody>
+                {% for contract in enterprise_contracts[account] %}
+                  {% if contract["contractInfo"]["status"] != 'expired' %}
+
+                    {% if contract["contractInfo"]["daysTillExpiry"] %}
+                      {% set expiring_soon = contract["contractInfo"]["daysTillExpiry"] <= 3000 %}
```
anthonydillon, line 86: I would suggest moving this logic to the python side and adding a status of `expiring`

PR #7073, opened 2020-03-25
Build db models

### LINKED ISSUE #7072: [CVEs] Setup models and schema
## Summary

Build models and schema for CVEs to create tables.

carkod:
## Done

- Models and schema (no API yet)

## QA

- No QA really, just want to make sure the models are correct and I am creating the relationships correctly because, once this is run, we'll start messing the db if it is not correct.

## Issue / Card

Fixes #7072

### DIFF
--- a/webapp/security/models.py
+++ b/webapp/security/models.py
@@ -37,17 +37,54 @@ notice_releases = Table(
     Column("release_id", Integer, ForeignKey("release.id")),
 )
 
+cve_bugs = Table(
+    "cve_bugs",
+    Base.metadata,
+    Column("cve_id", String, ForeignKey("cve.id")),
+    Column("bug_id", Integer, ForeignKey("bug.id")),
+)
+
+cve_references = Table(
+    "cve_references",
+    Base.metadata,
+    Column("cve_id", String, ForeignKey("cve.id")),
+    Column("cve_reference_id", Integer, ForeignKey("cve_reference.id")),
+)
+
+package_release_status = Table(
+    "package_release_status",
+    Base.metadata,
+    Column("package_id", String, ForeignKey("package.id")),
+    Column("release_id", Integer, ForeignKey("release.id")),
+)
+
 
 class CVE(Base):
     __tablename__ = "cve"
 
-    id = Column(String, primary_key=True)
+    id = Column(String, primary_key=True, unique=True)
+    candidate = Column(String, unique=True)
+    public_date = Column(DateTime)
+    public_date_usn = Column(DateTime)
+    crd = Column(String)
+    description = Column(String)
+    ubuntu_description = Column(String)
+    notes = Column(String)
+    mitigation = Column(String)
+    priority = Column(String)
+    discovered_by = Column(String)
+    assigned_to = Column(String)
+    approved_by = Column(String)
+    cvss = Column(String)  # CVSS 3 and Base Score
+    references = relationship("CVEReference", secondary=cve_references)
+    bugs = relationship("Bug", secondary=cve_bugs)
+    packages = relationship("Package", secondary=package_release_status)
 
 
 class Notice(Base):
     __tablename__ = "notice"
 
-    id = Column(String, primary_key=True)
+    id = Column(Integer, primary_key=True)
     title = Column(String)
     published = Column(DateTime)
     summary = Column(String)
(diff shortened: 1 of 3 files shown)


webapp/security/models.py
```diff
@@ -37,17 +37,54 @@
...
+    references = relationship("CVEReference", secondary=cve_references)
+    bugs = relationship("Bug", secondary=cve_bugs)
+    packages = relationship("Package", secondary=package_release_status)
 
 
 class Notice(Base):
     __tablename__ = "notice"
 
-    id = Column(String, primary_key=True)
+    id = Column(Integer, primary_key=True)
```
jpmartinspt: The id of a notice is a `String`

webapp/app.py
```diff
@@ -146,6 +148,11 @@ def utility_processor():
 app.add_url_rule("/security/notices/<notice_id>", view_func=notice)
 
 
+# cve section
+app.add_url_rule("/security/cve", view_func=cve_index)
+app.add_url_rule("/security/<regex('cve-d{4}-d{4,7}'):cve_id>", view_func=cve)
```
jpmartinspt: why not just `security/cve/<cve_id>`

webapp/security/models.py
carkod: OK, I thought that was a mistake, because the others are all Integers?

jpmartinspt: Nop it should be a String

webapp/app.py
carkod: It was requested by [design]
(https://docs.google.com/document/d/1VuAElUpFay7MBFNmE3NhSyY9TkXiqHcA1ZHQ11AYO-Q/edit#heading=h.ua7qthpl074)

webapp/security/models.py
```diff
@@ -37,11 +60,50 @@
...
+    notes = Column(String)
+    mitigation = Column(String)
+    priority = Column(String)
+    discovered_by = Column(String)
+    assigned_to = Column(String)
+    approved_by = Column(String)
+    cvss = Column(String)  # CVSS 3 and Base Score
+    references = relationship("CVEReference", secondary=cve_references)
+    bugs = relationship("Bug", secondary=cve_bugs)
+    packages = relationship("Package", secondary=package_release_status)
```
jpmartinspt: I think what you want here is a `secondary=cve_packages` here(You need to create that table).

Then you probably want to add `releases_status = relationship("Release", secondary=package_release_status)` to the Packages models.

Does that make sense?

webapp/security/models.py
```diff
@@ -37,11 +60,50 @@
...
+    "cve_references",
+    Base.metadata,
+    Column("cve_id", String, ForeignKey("cve.id")),
+    Column("cve_reference_id", Integer, ForeignKey("cve_reference.id")),
+)
+
+package_release_status = Table(
+    "package_release_status",
+    Base.metadata,
+    Column("package_id", String, ForeignKey("package.id")),
```
jpmartinspt: This should be an `Integer`

webapp/security/models.py
carkod: Ok yeah. So adding that relationship will link it to the existent releases table?

jpmartinspt: Yep

webapp/security/models.py
```diff
@@ -7,13 +7,36 @@
     JSON,
     String,
     Table,
+    Enum,
 )
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
-
+import enum
```
jpmartinspt: stdlib imports at the top please.

.env
```diff
@@ -2,7 +2,7 @@ PORT=8001
 FLASK_DEBUG=true
 DEVEL=true
 ADVANTAGE_API=https://contracts.canonical.com/
-DATABASE_URL=sqlite:///usn.sqlite3
+DATABASE_URL=sqlite:///
```
jpmartinspt: Any reason we are changing this? is quite useful to persist on disk so we have the same data in between re-runs.

carkod: Ok, I thought this was the right way to do it. So should i keep it as `sqlite:///usn.sqlite3`?

webapp/security/models.py
```diff
@@ -37,11 +61,59 @@
...
+    Column("status", Enum(PackageStatus)),
+    extend_existing=True,
+)
+
 
 class CVE(Base):
     __tablename__ = "cve"
 
-    id = Column(String, primary_key=True)
+    id = Column(String, primary_key=True, unique=True)
```
jpmartinspt: I believe `unique=True` is not necessary here case the `primary_key=True` constraint already enforces uniqueness.

carkod: ok, didn't know that

jpmartinspt:
This needed a migration.

PR #7113, opened 2020-04-01
Switch to using postgres in local dev

nottrobin:
QA
--

Follow postgres instructions in the readme

Go to /security/notices, see it works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ If you have found a bug in the Ubuntu OS itself, the please file it [here](https
 
 ## Local development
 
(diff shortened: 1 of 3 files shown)


nottrobin:
Oh.... this will break demos :(

nottrobin:
I wonder if there's an easy way to support both?

jpmartinspt:
I would say leave the .env to point to sqlite. And leave instructions to setup postgres. Most people that dev on ubuntu don't care about using postgres at all.

nottrobin:
OK I've updated it to provide specialist postgres instructions, so that the `./run` will continue to work for demos.

PR #7117, opened 2020-04-01
Switch CI to GitHub actions, with improvements

nottrobin:
Move all CircleCI checks to GitHub actions, with the following changes:

- Use `--immutable` with `yarn install` so we know
  if the lockfile is out of date
- Add `run-image` check, to ensure the docker image builds and runs
- Change `run-site` to `run-dotrun` to check the site works with the
  dotrun snap
- Move linkchecker to a workflow that's run every morning against
  master, like cypress

The docker build will help to avoid a release issue we had last night,
the `immutable` change will help to avoid a cypress issue we had this
morning.

QA
--

See new GitHub CI steps pass


nottrobin:
Again?

PR #7118, opened 2020-04-01
Switch CI to GitHub actions, with improvements

nottrobin:
Move all CircleCI checks to GitHub actions, with the following changes:

- Use `--immutable` with `yarn install` so we know
  if the lockfile is out of date
- Add `run-image` check, to ensure the docker image builds and runs
- Change `run-site` to `run-dotrun` to check the site works with the
  dotrun snap
- Move linkchecker to a workflow that's run every morning against
  master, like cypress

The docker build will help to avoid a release issue we had last night,
the `immutable` change will help to avoid a cypress issue we had this
morning.

QA
--

See new GitHub CI steps pass


.github/workflows/linkchecker.yaml
```diff
@@ -0,0 +1,26 @@
+name: Check links
+
+on:
+  schedule:
+    - cron: "10 7 * * *"
```
tbille, line 5: Maybe in the same way that we did for the webbot, this could run during weekdays? ` cron: "10 7 * * 1-5"`

.github/workflows/linkchecker.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  run-cypress:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v1
```
tbille, line 13: `uses: actions/checkout@v2`

nottrobin: Is 2 better than 1?

.github/workflows/linkchecker.yaml
nottrobin: You think? Unlike with your pull request notifier, I think we want to know when tests fail any day of the week.

.github/workflows/linkchecker.yaml
nottrobin: Done

.github/workflows/linkchecker.yaml
```diff
@@ -0,0 +1,26 @@
...
+        uses: actions/checkout@v2
+
+      - name: Build image
+        run: DOCKER_BUILDKIT=1 docker build --tag ubuntu-com .
+
+      - name: Run linkchecker against the server
+        run: |
+          docker run -d -p 80:80 ubuntu-com
+          sleep 5
+          linkchecker --threads 2 --ignore-url /usn --ignore-url /search --ignore-url ".*(?:png|jpg|jpeg|gif|bmp|svg|js)$" --ignore-url /q_auto --ignore-url /fl_sanitize --ignore-url /w_* --ignore-url /h_* --ignore-url /blog --ignore-url /server/docs --no-warnings http://127.0.0.1
```
tbille, line 22: I think we should add ` --check-extern  ` parameter to check also sites outside that we point to.

.github/workflows/linkchecker.yaml
tbille: hopefully yeah. I am fine in running it every day.

.github/workflows/linkchecker.yaml
nottrobin: let's experiment with this later. I'm worried it will always fail with that parameter and spam us all with emails every morning.

.github/workflows/linkchecker.yaml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+
+      - name: Build image
+        run: DOCKER_BUILDKIT=1 docker build --tag ubuntu-com .
+
+      - name: Run linkchecker against the server
+        run: |
+          docker run -d -p 80:80 ubuntu-com
```
tbille, line 20: I think you need to add here `DATABASE_URL=sqlite:///db.sqlite3 ubuntu-com `

.github/workflows/linkchecker.yaml
```diff
@@ -0,0 +1,26 @@
+name: Check links
```
tbille, line 1: I have done this work also here https://github.com/canonical-web-and-design/cn.ubuntu.com/blob/master/.github/workflows/linkchecker.yaml (just sharing)

.github/workflows/linkchecker.yaml
nottrobin: nope. it will read `.env`, which is part of the test

.github/workflows/linkchecker.yaml
tbille: ok, if we create an issue for it I am fine with it.

.github/workflows/linkchecker.yaml
tbille: I don't think it reads from .env here in the docker build step

PR #7125, opened 2020-04-02
Add renewal to context

tbille:
## Done

- Add renewal object to context

## QA

- Should work as usual

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -204,6 +204,11 @@ def advantage_view():
                         ):
                             contract["contractInfo"]["status"] = "expired"
 
+                    if "renewals" in contract["contractInfo"]:
+                        contract["renewal"] = contract["contractInfo"][
+                            "renewals"
+                        ][0]
+
                     enterprise_contracts.setdefault(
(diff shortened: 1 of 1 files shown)


webapp/views.py
```diff
@@ -204,6 +204,11 @@ def advantage_view():
                         ):
                             contract["contractInfo"]["status"] = "expired"
 
+                    if "renewals" in contract["contractInfo"]:
+                        contract["renewal"] = contract["contractInfo"][
+                            "renewals"
+                        ][0]
```
anthonydillon, line 210: Will this mask renewals if there are multiple? Why not just access `contract["contractInfo"]["renewals"]` instead of `contract["renewal"]`. I'm worried the more we adapt the data the more we have to maintain and introduce bugs.

tbille: @anthonydillon I asked exactly that question but:
- at the moment we only support 1 renewal, even if they give an array, they get ready in a maybe future to have multiple
- I was actually going to adapt the object that I give until I realized that I was bascially passing everything to the template... It felt really useless to do it.

PR #7127, opened 2020-04-02
Improved data insertion and fixed relationships

### LINKED ISSUE #7030: [CVEs] Migrations to create CVE data tables

carkod:
## Done

- An improved version of the data migration PR. If this works, we should close #7100 
- All data I could add has been added to the first CVE-2020

## QA

- Run the site with dotrun snap
- `docker-compose up`
- Visit page `/security/cve`
- Repopulate data with `dotrun exec python3 webapp/security/fixtures/load_sample_cve.py`

## Issue / Card

Fixes #7030


static/sass/styles.scss
```diff
@@ -19,12 +19,7 @@
   font-style: normal;
   font-weight: 300;
   src:
-<<<<<<< HEAD
```
jpmartinspt, line 22: Needs fixing.

webapp/security/models.py
```diff
@@ -7,13 +8,40 @@
...
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
 
 
 Base = declarative_base()
 
+# Enums
+
+
+class CveStatus(enum.Enum):
```
jpmartinspt, line 22: I though you were going to remove the enums?

webapp/security/views.py
```diff
@@ -263,8 +273,107 @@ def api_create_notice():
 # CVE views
 # ===
 def cve_index():
-    return flask.render_template("security/cve/index.html")
+
+    # Temporary fix for data to be displayed
+    cves_query = db_session.query(CVE).first()
+    if not cves_query and os.environ["FLASK_DEBUG"] == "true":
```
jpmartinspt, line 279: Don't do this in where. Just add an helper script somewhere that loads dummy data for you.

webapp/security/views.py
```diff
@@ -263,8 +273,107 @@ def api_create_notice():
...
+    if not cve:
+        flask.abort(404)
+    return flask.render_template("security/cve/cve.html", cve=cve)
+
+
+# CVE API
+# ===
+
+
+def api_create_cve():
```
jpmartinspt, line 303: This is a views file and this looks more like an helper function to load things into the DB. This is probably what your helper script looks like.

webapp/security/models.py
carkod: No, just temporarily removing it from the insertion, although I will try again when I have time, because the problem was with alembic.

webapp/security/models.py
```diff
@@ -77,3 +149,37 @@ class Release(Base):
...
+
+
+class Package(Base):
+    __tablename__ = "package"
+
+    id = Column(Integer, primary_key=True)
+    name = Column(String, unique=True)
+    type = Column(Enum(PackageType))
+    releases_status = relationship(
+        "CVERelease", secondary=package_release_status
```
jpmartinspt, line 175: is it package_release_status or cve_release_status?

webapp/security/models.py
jpmartinspt: What insertion? If the problem is with alembic does that not mean we are going to run into it whenever we need to apply a migration? Will things not fail when sqlalchemy tries to do things with the enums and the enum type is not actually defined on the DB schema?

webapp/security/models.py
carkod: So I created a different table called CVERelease, which is similar to USN release table, we can maybe later talk about this, but the data model didn't really fit into mine (I have a status and I have a "status description"), so I decided to create a new one to get it out of the way, so I can append data to the CVEs and let front-end carry on with work. We can revisit this.

webapp/security/views.py
carkod: Ok. Do you mean creating a `helper.py` in this folder and importing it?

webapp/security/models.py
jpmartinspt: Sure, but the secondary table is still named `package_release_status` probably because this relationship used to be with the `Package` table.

webapp/security/views.py
jpmartinspt: I mean create a `scripts/load_cves.py`, make it executable and run it `dotrun exec scripts/load_cves.py` whenever you want to load sample data.

webapp/security/models.py
carkod: I removed the enums. Basically the problem is that autogenerate does not create the types.

webapp/security/models.py
carkod: Is that not correct? And the `package_release_status` is related to the `cve_release` table. Things seem to be working fine here.

webapp/security/views.py
carkod: mm I added it here on purpose, so that people didn't have to do that... Because things will not work if there is no data at all

static/sass/styles.scss
carkod: it's a github bug

webapp/security/views.py
carkod: so `/scripts/load_cves.py` in the root folder can't use `import webapp.security`

jpmartinspt:
There are conflicts, probably needs a rebase.

jpmartinspt:
@carkod There are conflicts, can you rebase before I review?

jpmartinspt:
You have two docker-compose files with different extensions.

webapp/security/views.py
```diff
@@ -263,8 +269,32 @@ def api_create_notice():
...
+
+    return flask.render_template("security/cve/index.html", list_cve=list_cve)
 
 
 def cve(cve_id):
-    return flask.render_template("security/cve/cve.html")
+
+    # Temporary fix for data to be displayed
+    cves_query = db_session.query(CVE).first()
+    if not cves_query and os.environ["FLASK_DEBUG"] == "true":
```
jpmartinspt, line 286: Can we please remove this? I suspect this will actually fail the second time you load the page, since it will attempt to insert a CVE with an ID that already exists. It is acceptable to just execute the helper script before you start developing anything CVE related. IMO we should not be loading data at runtime.

webapp/security/helper.py
```diff
@@ -0,0 +1,94 @@
+from webapp.security.database import db_session
```
jpmartinspt, line 1: Just move this into `webapp/security/fixtures/load_sample_cve.py`

webapp/security/helper.py
```diff
@@ -0,0 +1,94 @@
...
+from webapp.security.models import (
+    CVE,
+    CVEReference,
+    Bug,
+    CVERelease,
+    Package,
+)
+
+
+def create_data():
```
jpmartinspt, line 11: load_sample_cve()

webapp/security/helper.py
```diff
@@ -0,0 +1,94 @@
...
+    for ref in reference_query:
+        cve.references.append(ref)
+
+    for bug in bugs_query:
+        cve.bugs.append(bug)
+
+    db_session.add(cve)
+    db_session.commit()
+    db_session.flush()
+    return
```
jpmartinspt, line 94: Append

```
if name == "__main__":
    load_sample_cve()
```

carkod: I added `__name__ == "__main__" lol took me some time to realise why this was not found

webapp/security/helper.py
jpmartinspt, line 1: This can be remove now?

webapp/security/models.py
```diff
@@ -77,3 +120,41 @@ class Release(Base):
...
+    launchpad = Column(String)
+    ubuntu = Column(String)
+    debian = Column(String)
+    type = Column(String)
+    releases_status = relationship(
+        "CVERelease", secondary=package_release_status
+    )
+
+
+class CVERelease(Base):
```
jpmartinspt, line 154: We are calling this `CVERelease` but this seems then establishing a relationship with `Package` instead?

carkod: So,`cve_release` has a relationship with `package_release_status`  which in turn has a relationship with `package` and this one has the relationship with `cve`
https://docs.google.com/drawings/d/1dsU5KT0iC3QSIFcE4xlf1q8SsgeAuyypkKpU7qVB0Ww/edit

jpmartinspt: From that schema, "Release" looks way more like a "PackageRelease" than a "CVERelease"?
Also does one CVE only link to one package?

carkod: Yeah, when I named it, I was thinking of making a difference between notices and CVEs. No, CVEs can have multiple packages

static/sass/styles.scss
```diff
@@ -635,4 +635,4 @@ summary {
   position: relative;
   top: 2px;
   width: 1em;
-}
```
carkod, line 638: @jpmartinspt this is because I copy pasted from upstream CVE feature branch

steverydz:
When executing `dotrun exec webapp/security/fixtures/load_sample_cve.py` I get a permissions error from `dotrun`: https://pastebin.canonical.com/p/5FByZh8XG2/

PR #7128, opened 2020-04-02
Add linkchecker tests

nottrobin:
I've added the following scheduled checks, and added badges for them to the README.

- "Links in master": Check internal links against master branch
   - Run daily at 7:10
- "Links on live": Check internal links on https://ubuntu.com
   - Run daily at 7:20
- "External links": Check external links from https://ubuntu.com
   - Run weekly at 7:30 on Monday

The "external links" check fails and shows many existing 404s. I can't fix them all in this branch, so we should address this when we can. Fortunately we'll be reminded every Monday ;)

I've fixed about a million 404s from tutorials and server docs, and I've fixed some from this codebase, but there are definitely quite a few more.

QA
--

Don't think this can really be QAed, more than reading the code. We'll just have to see how it goes when we merge it.

### DIFF
--- a/.github/workflows/external-links.yaml
+++ b/.github/workflows/external-links.yaml
@@ -1,0 +1,23 @@
+name: Links on live
+
+on:
+  schedule:
+    - cron: "30 7 * * 1"
+
+jobs:
+  check-links:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+
+      - name: Install linkchecker
+        run: sudo apt update && sudo apt install linkchecker
+
+      - name: Run linkchecker
+        run: linkchecker --check-extern --ignore-url '.*&start=.*' --ignore-url 'https://ubuntuforums.org' --ignore-url 'https://res.cloudinary.com' --ignore-url /q_auto --ignore-url /fl_sanitize --ignore-url /w_* --ignore-url /h_* --ignore-url '/blog/*' --no-warnings http://localhost
+
+      - name: Send message on failure
+        if: failure()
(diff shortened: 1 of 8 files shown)


.github/workflows/external-links.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v2
+
+      - name: Install linkchecker
+        run: sudo apt update && sudo apt install linkchecker
+
+      - name: Run linkchecker
+        run: linkchecker --check-extern --ignore-url '.*&start=.*' --ignore-url 'https://ubuntuforums.org' --ignore-url 'https://res.cloudinary.com' --ignore-url /q_auto --ignore-url /fl_sanitize --ignore-url /w_* --ignore-url /h_* --ignore-url '/blog/*' --no-warnings http://localhost
```
anthonydillon, line 19: Should we ignore so much if this is run out of dev cycle? Can we retry an error link a few times?

nottrobin: All of these ignores are very deliberate. I checked them all. Ubuntuforums.org is on many pages but fails in linkchecker cos of a HTTPS cert issue. Res.cloudinary.com is mentioned in domain prefetching, but doesn't have an index page. The /fl_sanitize etc. are not really linked to buy are found through errors in linkcheckers parser. Blog posts are ignored cos there are many old posts that we know have 404s that we CBA to fix, and &start is there to avoid linkchecker going infinitely deep on search results pages and never finishing.

nottrobin: I'd love it if we could fix 404s in old blog posts, but we have quite a task right now just fixing 404s in tutorials and docs, so I think let's not try to do blog posts just yet.

PR #7130, opened 2020-04-02
Add endpoints for renewals

tbille:
## Done

In every endpoint send the session so we can  authenticate the user. 

- Add endpoints:
  - Set payment method for account
```
Returns what is returned by the API: 
POST /advantage/payment-method
{
  "stripe_method_id": "<STRIPE_METHOD_ID>",
  "account_id": "<ACCOUNT_ID>"
}
```
   - Accept renewal
```
POST      /advantage/renewals/{renewal_id}/process-payment
```
   - Get renewal (returns what is returned by https://github.com/CanonicalLtd/ua-contracts/blob/develop/docs/contracts.yaml#L2187 )
```
GET     /advantage/renewals/{renewal_id}
```

Basically the flow that we need:
- setup an payment method linked to an account (via modal or whatever) -> so one POST to the first api  (`/advantage/payment-method`)
- Open a renewal that needs to be processed
- Accept the renewal (or press pay) (`/advantage/accept-renewal`)
- Those first steps could be done when approving the payment (2 calls in a row)
- Go to a view that display the status of the payment. To do that request every 5 seconds ` /advantage/renewal`

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -40,7 +40,7 @@ from webapp.views import (
     blog_press_centre,
     download_thank_you,
     get_renewal,
-    post_stripe_method_id,
+    put_stripe_method_id,
     releasenotes_redirect,
 )
 from webapp.login import login_handler, logout
@@ -117,8 +117,8 @@ def utility_processor():
 app.add_url_rule("/advantage", view_func=advantage_view)
 app.add_url_rule(
     "/advantage/payment-method",
-    view_func=post_stripe_method_id,
-    methods=["POST"],
+    view_func=put_stripe_method_id,
+    methods=["PUT"],
 )
 app.add_url_rule(
     "/advantage/renewal", view_func=get_renewal, methods=["POST"],
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -40,7 +40,7 @@ from webapp.views import (
     blog_press_centre,
     download_thank_you,
     get_renewal,
-    put_stripe_method_id,
+    post_stripe_method_id,
     releasenotes_redirect,
 )
 from webapp.login import login_handler, logout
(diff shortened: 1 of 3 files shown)


webapp/app.py
```diff
@@ -112,6 +115,18 @@ def utility_processor():
 
 # Simple routes
 app.add_url_rule("/advantage", view_func=advantage_view)
+app.add_url_rule(
+    "/advantage/payment-method",
+    view_func=put_stripe_method_id,
+    methods=["PUT"],
+)
+app.add_url_rule(
+    "/advantage/renewal", view_func=get_renewal, methods=["POST"],
```
tbille, line 124: "/advantage/renewal/{id}", view_func=get_renewal, methods=["GET"],

webapp/app.py
```diff
@@ -112,6 +115,18 @@ def utility_processor():
...
+)
+
+app.add_url_rule(
+    "/advantage/accept-renewal", view_func=accept_renewal, methods=["POST"],
```
tbille, line 128: /advantage/checkout

webapp/app.py
```diff
@@ -112,6 +115,18 @@ def utility_processor():
 
 # Simple routes
 app.add_url_rule("/advantage", view_func=advantage_view)
+app.add_url_rule(
+    "/advantage/payment-method",
+    view_func=put_stripe_method_id,
+    methods=["PUT"],
```
tbille, line 121: post

webapp/app.py
tbille: /advantage/renewals/{id}/process-payment

webapp/views.py
```diff
@@ -225,6 +225,46 @@ def advantage_view():
...
+            return flask.jsonify({"error": "account_id required"}), 400
+
+        return advantage.put_method_id(
+            flask.session, account_id, payment_method_id
+        )
+    else:
+        return flask.jsonify({"error": "authentication required"}), 401
+
+
+def get_renewal(renewal_id=""):
```
jpmartinspt, line 248: `renewal_id` is always going to be set right? otherwise the url rule would not have matched?
same for the view bellow.

PR #7135, opened 2020-04-03
Redirect ubuntu release notes

### LINKED ISSUE #7129: www.ubuntu.com redirection to release notes does not work
## Summary

All Ubuntu ISOs include a file (/media/cdrom/.disk/release_notes_url) which as a link to the release notes. Its contents look like this:

http://www.ubuntu.com/getubuntu/releasenotes?os=ubuntu&ver=18.04.4&lang=${LANG}

This now redirects to https://wiki.ubuntu.com/Releases?os=ubuntu&ver=18.04.4&lang=en which is not the release notes for that release.

## Process

Load this hyperlink http://www.ubuntu.com/getubuntu/releasenotes?os=ubuntu&ver=18.04.4&lang=en

Additionally, it doesn't matter which release I use for ver.

I expect the redirect to the release notes to work.

pmahnke:
## Done

- Redirect ubuntu release notes - mostly
- **Note there are some special cases which might need to be added to this**

## QA

    - http://0.0.0.0:8001/GetUbuntu/releasenotes?os=ubuntukylin&ver=15.04&lang=zh
    - http://0.0.0.0:8001/GetUbuntu/releasenotes?os=ubuntu&ver=16.04
- See that you mostly get to the right place

## Issue / Card

Fixes #7129


webapp/views.py
```diff
@@ -81,11 +81,45 @@ def releasenotes_redirect():
...
+        "13.10": "SaucySalamander",
+        "14.04": "TrustyTahr",
+        "14.10": "UtopicUnicorn",
+        "15.04": "VividVervet",
+        "15.10": "WilyWerewolf",
+        "16.04": "XenialXerus",
+        "16.10": "YakketyYak",
+        "17.04": "ZestyZapus",
+        "17.10": "BionicBeaver",
+        "18.04": "ArtfulAardvark",
```
murraybd, line 97: 17.10 is Artful and 18.04 is Bionic.

webapp/views.py
```diff
@@ -81,11 +81,45 @@ def releasenotes_redirect():
...
+        "xubuntu": "Xubuntu",
+    }
+
+    ver = releaseName[flask.request.args.get("ver")[:5]]
+    os = flavourName[flask.request.args.get("os")]
 
     if ver:
-        return flask.redirect(f"https://wiki.ubuntu.com/{ver}/ReleaseNotes")
+        return flask.redirect(
+            f"https://wiki.ubuntu.com/{ver}/ReleaseNotes/{os}"
```
nottrobin, line 121: i think these pages don't necessarily exist do they?

https://wiki.ubuntu.com/BionicBeaver/ReleaseNotes/Edubuntu -> 404
https://wiki.ubuntu.com/BionicBeaver/ReleaseNotes/Lubuntu -> 404

nottrobin: If that's the case, it does seem like sending everyone to https://wiki.ubuntu.com/Releases to find their own release is probably a more bullet-proof solution.

webapp/views.py
```diff
@@ -81,11 +81,45 @@ def releasenotes_redirect():
 
     Old apache redirects: https://pastebin.canonical.com/p/3TXyyNkWkg/
     """
-
-    ver = flask.request.args.get("ver")
+    releaseName = {
```
nottrobin, line 84: The old solution relied on redirects at e.g.:

https://wiki.ubuntu.com/19.04/ReleaseNotes

This seems to be working fine. Why do we need to do it here? This will just mean we need to maintain another list of release codenames when hopefully it's already handled wiki side?

pmahnke:
@nottrobin how about this for a compromise.  It uses the old format for pre-16.04 releases and the new format going forward.  This way we don't have to maintain the list and the default works for flavours.

nottrobin:
@pmahnke I like this compromise. However if you want to cover everything you have to include `16.10`, `17.04` and `17.10` in the mapping, as they don't have redirects.

webapp/views.py
```diff
@@ -82,7 +82,21 @@ def releasenotes_redirect():
...
+        "13.10": "SaucySalamander",
+        "14.04": "TrustyTahr",
+        "14.10": "UtopicUnicorn",
+        "15.04": "VividVervet",
+        "15.10": "WilyWerewolf",
+    }
+
+    ver = flask.request.args.get("ver")[:5]
+
+    if releaseName[ver]:
```
nottrobin, line 98: This fails if `ver` isn't in the dictionary.

image

What you want is `if ver in releaseName:`

pmahnke:
@nottrobin +1 from me, but I cannot do that on my own PR

PR #7136, opened 2020-04-04
Rename to external links

nottrobin:
Try running "External links" on PR so we can check it runs properly.

It probably won't work on this branch 'cos this is on my fork, but just checking this is still the case.


nottrobin:
Self-merging to fix the check that's gonna run on Monday morning

PR #7149, opened 2020-04-08
Handle UA renewal urls

Lukewh:
## Done

- Added extra conditions for opening the table cells if a subscription is passed
- Added a login redirect if a subscription ID is provided but the user is not authenticated

## QA

- Logout of SSO
  - Visit http://0.0.0.0:8001/advantage?subscription=<id>
  - You should be redirected to login
  - Log in
  - You should be redirected to http://0.0.0.0:8001/advantage?subscription=<id>
- While logged in visit http://0.0.0.0:8001/advantage?subscription=<id>

Try ID: cAM3HMVaKsoTuMVdhGK5lWnQej0GlnuqufQNI0tlz58s

## Issue / Card

Fixes: https://github.com/canonical-web-and-design/web-squad/issues/2582
Fixes: https://github.com/canonical-web-and-design/web-squad/issues/2583

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -100,8 +100,7 @@ def advantage_view():
     open_subscription = flask.request.args.get("subscription", None)
 
     if not auth.is_authenticated(flask.session) and open_subscription:
-        redirect_url = urllib.parse.quote(f"{flask.request.full_path}")
-        return flask.redirect(f"/login?next={redirect_url}")
+        return flask.redirect(f"/login?next={flask.request.full_path}")
 
     if auth.is_authenticated(flask.session):
         try:
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -2,6 +2,7 @@
 import json
 import os
 import re
+import urllib.parse
 
 # Packages
 import dateutil.parser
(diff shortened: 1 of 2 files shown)


matthewpaulthomas:
Thanks @Lukewh.

I see @sowasred2012’s point about scrolling. Without scrolling, it’s “huh, why am I here?” … But with scrolling, and therefore the header scrolled out of sight, it would look less like a real Ubuntu page. Which could be suspicious, if this what you see after clicking a button in an e-mail message.

Since the subscriptions are in no particular order at the moment, perhaps we could cheat — sort the row we’re expanding so that it’s first in the table. Feel free to split that off into a separate issue if you prefer.

webapp/views.py
```diff
@@ -96,6 +97,11 @@ def advantage_view():
     enterprise_contracts = {}
     entitlements = {}
     openid = flask.session.get("openid")
+    open_subscription = flask.request.args.get("subscription", None)
+
+    if not auth.is_authenticated(flask.session) and open_subscription:
+        redirect_url = urllib.parse.quote(f"{flask.request.full_path}")
+        return flask.redirect(f"/login?next={redirect_url}")
```
tbille, line 104: as discussed `return flask.redirect(f"/login?next={flask.request.full_path}")`

PR #7154, opened 2020-04-09
Security API authentication & proof of concept

nottrobin:
This adds an "authorization_required" decorator for securing API endpoints for the security section, and applies it to a POST endpoint at /security/notices.

This will expect to be provided with a valid macaroonbakery macaroon, and if it doesn't find one, will return a macaroonbakery JSON data structure to ask the client to forward the user to api.jujucharms.com/identity to login.

This should be all we need to allow clients to authenticate with our API. All the clients need is the small amount of code you can see in [my gist](https://gist.github.com/nottrobin/8e89667066815a263021a5d3ba248e22), the important part of which is:

``` python3
client = httpbakery.Client(cookies=MozillaCookieJar(".login"))
if os.path.exists(client.cookies.filename):
    client.cookies.load(ignore_discard=True)
response = client.request("POST", "http://ubuntu.com/some/endpoint")
client.cookies.save(ignore_discard=True)
```

Pretty simple.

QA
--

Download [post-cve.py](https://gist.githubusercontent.com/nottrobin/8e89667066815a263021a5d3ba248e22/raw/0fe6ec67595dc732b8bd01035f8ca9ffa1528f00/post-cve.py), which is a script that makes use of macaroonbakery.httpbakery to POST to `https://ubuntu-com-canonical-web-and-design-pr-7154.run.demo.haus/security/cve` on the demo of this PR.

Now run it:

```
python3 -m venv .venv
source .venv/bin/activate
pip3 install macaroonbakery
python3 post-cve.py
```

You should be prompted to open a browser to login, and once you have, you should see:

```
<Response [400]> {
  "message": "No payload received"
}
```

Which means success. ('cos I'm not actually submitting anything with the POST here)

### DIFF
--- a/.env
+++ b/.env
@@ -3,6 +3,7 @@ FLASK_DEBUG=true
 DEVEL=true
 ADVANTAGE_API=https://contracts.staging.canonical.com/
 DATABASE_URL=sqlite:///usn.sqlite3
+SECRET_KEY=insecure_dev_key
 
 # This is a test api key https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
 CAPTCHA_TESTING_API_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,6 +22,5 @@ mistune==0.8.4
 psycopg2-binary==2.8.4
 pyyaml==5.3.1
 ubuntu-release-info==20.2
-
-# This should be removed once we update to canonicalwebteam.http 1.0.2
-httpretty==1.0.2
+launchpadlib==1.10.10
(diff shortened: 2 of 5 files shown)


nottrobin:
@jpmartinspt, @carkod check it out - this should be everything we need to authenticate from client scripts, usn-tool and the script in the CVE repo.

jpmartinspt:
Code looks good.  It's fine for us to use `api.jujucharms/identity`?
All do some QA after standup, could you please just rebase?

nottrobin:
@jpmartinspt rebased. Just have to wait for the demo to be ready again.

nottrobin:
Oh re: api.jujucharms.com, I'm not crazy about it but apparently it's the only way to use candid at this time, creating a new identity provider would be too much effort. And it's only an internal tool, so as long as it works I think it's fine for now.

PR #7173, opened 2020-04-09
Re-model packages

carkod:
## Done

- Revert to previous packages tables and package status release tables

## QA

- load data with `dotrun exec python3 webapp/security/fixtures/load_sample_cve.py`
- Search now by package e.g. http://0.0.0.0:8001/security/cve?package=gitlab (gitlab is the only package in the dummy data)


webapp/security/fixtures/load_sample_cve.py
```diff
@@ -5,41 +5,17 @@
...
+    CVEReference,
+    Bug,
+    Package,
+    PackageReleaseStatus,
+    Release,
+)  # noqa
 
 
 def load_sample_cve():
-
```
carkod, line 12: Since models have changed, all data insertion structure needs to change

webapp/security/models.py
```diff
@@ -107,6 +137,9 @@ class Release(Base):
     version = Column(String, unique=True)
     codename = Column(String, unique=True)
     lts = Column(Boolean)
+    esm = Column(Boolean)
```
carkod, line 140: I believe this should remain as boolean, and when the upload script posts to the API, it should get the value of ESM date and return true or false

nottrobin: How is that going to work for records already in the database? Consider this scenario:

- Security upload a CVE for the Ubuntu Bionic LTS (LTS=True, ESM=False, presumably)
- Time passes, it's now 2023. ESM now needs to be considered "True", but nothing in the database has changed...
- Time passes. It's now 2029. ESM is no longer considered "True", but nothing in the database has changed

nottrobin: Note that LTS and ESM are not equivalent properties, they are not both time dependent in the same way. An LTS is always described as an LTS, even if the support has expired. As in, we still refer to "14.04 LTS" or "12.04 LTS", they are simply descriptions of the type of release. On the other hand, ESM is not a description of a type of release, it is a description of a state that a release might be in. As in, we offer extended support for all LTS releases for those who pay. So ESM is time-dependent: "14.04 LTS is in ESM until April 2022".

nottrobin:
@carkod I think this need rebasing on the CVE branch, before the diff will display accurately. ping me tomorrow when you've had a chance to do that and I'll continue my review.

webapp/security/models.py
carkod: Yes, but don't you need some type of operation to update an entry in the database? I suppose you need to do an `update` query to change this value anyway

carkod:
@nottrobin rebased and adapted to conversation here https://github.com/canonical-web-and-design/ubuntu.com/pull/7221

steverydz:
@nottrobin is there any reason we want to deviate from the pagination pattern used on tutorials and blog? e.g. `?page=4`

We use a shared component for pagination which just works as long as you can pass it `total_pages` and `current_page` values.

webapp/security/models.py
```diff
@@ -50,6 +50,35 @@
...
+    Column("release_id", Integer, ForeignKey("release.id")),
+    Column(
+        "package_release_status_id",
+        Integer,
+        ForeignKey("package_release_status.id"),
+    ),
+)
+
 
 class CVE(Base):
```
jpmartinspt, line 83: Is there a reason we decided to store all dates as strings?

nottrobin:
@steverydz aye there is. A `page` based API doesn't allow flexibility in a bunch of ways. If you ever want pages of a different size it gets very confusing. @jkfran (I think) experienced this issue in the snap store API. This will become relevant if we change in future to pages of a different size, or start doing lazy-loading or whatnot.

This is why the [web API design ebook](https://www.google.com/url?q=https://pages.apigee.com/rs/apigee/images/api-design-ebook-2012-03.pdf&sa=U&ved=2ahUKEwjvpr3ys-joAhUNM8AKHerkBYoQFjAAegQIDRAB&usg=AOvVaw2sYj4JgOsS7mMfIUT2gRTc) (which I love) recommends `limit` and `offset`. Google's search API uses a similar concept of `start` and `size`, and so we follow that interface in https://ubuntu.com/search?q=Ubuntu&start=21, but I feel the `limit` and `offset` standard is stronger.

I like to try to start from the web API design way of doing things when designing URLs.

I'd prefer we revisit the URL structure for those other pages, but happy to open up the discussion wider.

jkfran:
@nottrobin Yeah, I remember that issue, because of the design we were showing the results of the first page in 3 columns and 4 columns for the second page and the following ones. Just with `size` and `page` is very difficult to do something like this.

steverydz:
@nottrobin OK, well my concern was simply that we use the `page` pattern everywhere else and already have a reusable component for that, but if we want to move in this direction I guess this would be a good place to start.

webapp/security/models.py
carkod: Updated. I initially did it because it seems the query changes the format of the date, but we want to show it in ISO format. But I guess there is a `strftime('%Y-%m-%d')`, so updated.

nottrobin:
@steverydz yes, I wasn't closely involved with the Tutorials or Blog implementations I think, but I have already been using the `limit` and `offset` pattern before they came along, which I think you will find exists elsewhere (thought I can't remember where). I've also tried to advocate for the use of Web API Design-based URLs, again prior to those being built, but I clearly didn't do a good enough job of it. I've not heard significant arguments against the "Web API Design" arguments, I just think it clearly hasn't been shared widely enough.

carkod:
@jpmartinspt @nottrobin I believe this discussion is clear now. I updated the models and created a new migration. So I am merging this to cve. Feel free to open again if you see any issues.

PR #7214, opened 2020-04-13
external-links: fix threads argument

OddBloke:
(Completely untested, but seems like the problem.)


nottrobin:
Thanks so much @OddBloke, sorry for the delay.

OddBloke:
No worries, it wasn't a particular blocker, it was just quicker to submit a
PR than to think about how to notify you. :p

PR #7221, opened 2020-04-14
New fields in release table

carkod:
## Done

- Propose changes to the release table for CVEs


webapp/security/models.py
```diff
@@ -77,3 +77,6 @@ class Release(Base):
     version = Column(String, unique=True)
     codename = Column(String, unique=True)
     lts = Column(Boolean)
+    development = Column(Boolean)
+    release_date = Column(DateTime)
+    esm_expire_date = Column(DateTime)
```
jpmartinspt, line 82: If we have the `release_date` can't we work out when ESM is expired?
What is `development`?

carkod: - Well, I guess I did it this way for people who don't know that ESM expires in 5 years?
-`development ` is for unreleased Ubuntu versions that also show in CVEs: https://people.canonical.com/~ubuntu-security/cve/main.html

jpmartinspt: What I mean is you can work out in python if the releases is `esm` if you have a release date.

jpmartinspt:
@carkod needs a migration now.

webapp/security/alembic/versions/a7d1a25813b3_cve_migration.py
```diff
@@ -0,0 +1,34 @@
+"""“cve_migration”
```
jpmartinspt, line 1: might want to rename this `release_update`, same with the filename.

nottrobin:
@jpmartinspt, @carkod I'm concerned by this `release_date`-based implementation for calculating ESM. I understand that, at the moment, free support for an LTS release extends for 5 years after the release date, and then Extended Support Maintenance extends for a further 5 years after free support has expired. And so we can simply say:

``` python3
free_support_expires = release_date + timedelta(years=5)
esm_expires = free_support_expires + timedelta(years=5)
```

My concern is that this relies on those values, `years=5` and `years=5` respectively, never changing. However, I really think it's dangerous to expect the ESM delta to remain constant as time goes on, as it was different as recently as 16.04:

image

Peter confirmed the time delta for ESM support has changed in the past and likely will change again in the future.

And in fact, it's not just ESM, LTS has also changed:

> A new LTS version is released every two years. In previous releases, a Long Term Support (LTS) version had three years support on Ubuntu (Desktop) and five years on Ubuntu Server. Starting with Ubuntu 12.04 LTS, both versions received five years support.
(https://wiki.ubuntu.com/LTS)

Given this, it does seem like it would be silly to hard-code the deltas for either LTS or ESM, and so we should really explicitly require there to be dates provided for every OS version. It seems sensible to keep `release_date` as well, as you can easily imagine that could be useful.

So I think we really need all of:

```
op.add_column("release", sa.Column("release_date", sa.DateTime()))
op.add_column("release", sa.Column("support_expires", sa.DateTime()))
op.add_column("release", sa.Column("esm_expires", sa.DateTime(), nullable=True))
```

(I would also make it so `release_date` and `support_expires` are required fields, 'cos why would they need to be empty?)

PR #7229, opened 2020-04-15
Add script to help load releases

jpmartinspt:
## Done
- Add support and esm expiry dates to the release table.
- Add a little script to help us have initial releases data for ubuntu security features.

## QA

- dotrun clean
- dotrun exec alembic upgrade head
- dotrun exec python3 webapp/security/fixtures/releases.py
- dotrun
- check `/security/notices` and see if releases appear in the dropdown.
- They should appear from newest to oldest.


webapp/security/fixtures/releases.py
```diff
@@ -0,0 +1,340 @@
+import sys
+from datetime import datetime
+
+sys.path.insert(0, "")
```
carkod, line 4: Have you tried running it without this? I think using `dotrun` will avoid this problem.

PR #7231, opened 2020-04-15
Add support and esm expiry dates to release table

jpmartinspt:
## Done
Added `support` and `esm` expiry dates to the release table

## QA

- dotrun clean
- dotrun exec alembic upgrade head
- all migrations should apply correctly


jpmartinspt:
Sorting as part of #7229

PR #7233, opened 2020-04-15
CVEs API endpoints

### LINKED ISSUE #7227: Build API endpoints
## Summary

- Create CVEs
- Delete CVEs
- Update CVEs
- Patch CVEs

carkod:
## Done

Basic API endpoints for Database operations:
At this point I have not been able to test with bulk, but the code is ready to do it. 

- Create a CVE
- Delete a CVE
- Update a CVE
- Authentication

## Quick QA

- `docker-compose up -d` to run the db (postgres)
- Create a CVE e.g. `curl -H "Content-Type: application/json" -X POST -d '{"id": "CVE-2020-11501", "public_date": "2020-04-03 13:15:00 UTC"}' http://localhost:8001/security/CVE-2020-11501`
- Update a CVE e.g. `curl -H "Content-Type: application/json" -X POST -d '{"id": "CVE-2020-11501", "public_date": "2020-10-03 13:15:00 UTC"}' http://localhost:8001/security/CVE-2020-11501` (updated `public_date`)
- Delete a CVE `curl -X DELETE http://localhost:8001/security/CVE-2020-11501`

## Full QA (recommended)

- `docker-compose up -d` to run the db (postgres)
- Clone/pull https://github.com/nottrobin/ubuntu-cve-tracker/pull/2
- Choose a CVE in the `active` folder. We are going to call this CVE-ID
- To create a CVE, run `python scripts/upload-cve.py active/{CVE-ID}`
- You can check if the CVE has been creating by going to the views http://localhost:8001/security/cve or go into the db and `TABLE cve;`
- To update a CVE, run `python scripts/update-cve.py active/{CVE-ID}`
- Repeat check by going to views or db.
- To delete a CVE `curl -X DELETE http://localhost:8001/security/{CVE-ID}' (it accepts a single CVE, so no need to use a file)
- Repeat check by going to views or db, and make sure it's been deleted.
- If all requests return HTTP `200` you are done. If error please report as I haven't tested all CVEs (I am expecting to have some strange ones that do not parse well)

## Additional QA (that I have done):
- Change a CVE package name and `python scripts/update-cve.py active/{CVE-ID}`
- To check in the db the query `TABLE package;` should return the changed package name that you have changed among other packages.
- Change a CVE package release status and `python scripts/update-cve.py active/{CVE-ID}`
- To check in the db the query `TABLE package;` should return the changed package name that you have changed among other packages.
- Pass a string in the public_date (to mess with the date format) and `python scripts/update-cve.py active/{CVE-ID}` should return a SQL error in json with 400.
- Tested on both Sqlite3 and Postgres

## Optionally 
Do previous steps again but this time choose a CVE in the ignored folder, although if it fails it will probably be related to the script, not really to the API endpoints, as these perform the exact same actions:
- To create a CVE, run `python scripts/upload-cve.py ignored/{CVE-ID}`
- You can check if the CVE has been created by going to the views http://localhost:8001/security/cve or go into the db and `TABLE cve;`
- To update a CVE, run `python scripts/update-cve.py ignored/{CVE-ID}`
- Repeat check by going to views or db.
- To delete a CVE `curl -X DELETE http://localhost:8001/security/{CVE-ID}' (it accepts a single CVE, so no need to use a file)
- Also you may try again with `not-for-us.txt` file. For this file the command would be `python scripts/upload-cve.py ignored/not-for-us.txt` and it will POST the first `CVE-2003-1086`

I know it is a lot, so THANKS! 

## Issue / Card

Fixes #7227


webapp/security/alembic/versions/a88c140f4ed1_cve_migration.py
```diff
@@ -99,7 +99,7 @@ def upgrade():
         "cve", sa.Column("last_updated_date", sa.DateTime(), nullable=True)
     )
     op.add_column("cve", sa.Column("mitigation", sa.String(), nullable=True))
-    op.add_column("cve", sa.Column("notes", sa.String(), nullable=True))
+    op.add_column("cve", sa.Column("notes", sa.JSON(), nullable=True))
```
nottrobin, line 102: Why JSON? 'Cos you want to store them line-by-line?

webapp/security/models.py
```diff
@@ -75,7 +75,7 @@
     Column(
         "package_release_status_id",
         Integer,
-        ForeignKey("package_release_status.id"),
+        ForeignKey("package_release_status.id", ondelete="CASCADE"),
```
nottrobin, line 78: Does this mean deleting a CVE will delete a package? Or that deleting a package will delete a CVE? Are we sure we want either?

What happens if a package is referenced in multiple CVEs and then a CVE is deleted?

webapp/security/alembic/versions/a88c140f4ed1_cve_migration.py
carkod: So the current ubuntu tracker is storing them as a pair of author/comment. I think it's useful to keep it with this data structure for display, but since we don't need to query them by author or comment, we don't need to create tables, so I stored them as a JSON.

nottrobin: Ah yes. Nice.

webapp/security/models.py
carkod: So, when we perform a `Delete` operation on the db, this should delete all `package release status` entries associated with that specific package, because if they are orphaned, this data is useless. Unlike e.g. `releases` data, which are useful on their own and used by others. As far as I know this one is applied on a package release status, not a CVE, therefore, if we delete a CVE nothing will happen, if we delete a package, all these release status will be deleted too. But I can confirm it doing an db operation see if that happens correctly.

webapp/security/views.py
```diff
@@ -271,9 +280,9 @@ def cve_index():
 
     # Query parameters
     order_by = flask.request.args.get("order-by", default="oldest")
-    query = flask.request.args.get("q", default="")
-    priority = flask.request.args.get("priority", default="")
-    package = flask.request.args.get("package", default="")
+    query = flask.request.args.get("q")
```
nottrobin, line 283: :+1: thanks for simplifying this.

carkod: Yeah, After your refactoring, having default values was not relevant anymore

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+
+    data = flask.request.get_json()
+    response = flask.jsonify({"message": "Unable to get body"}), 400
+    packages = []
+    references = []
+    bugs = []
+
+    # Check if CVE exists by candidate
+    if db_session.query(CVE).filter(CVE.id == data["id"]).count() > 0:
+        response = flask.jsonify({"message": "CVE already exists"}), 400
```
nottrobin, line 360: Presumably if a CVE already exists then `db_session.commit()` will fail below. Couldn't you roll this response into the `except` statement down there?

That way you avoid doing this extra query, and you simplify the code.

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+        return response
+
+    # Packages
+    # Check if there are packages before mapping
+    if "packages" in data and len(data["packages"]) > 0:
+        for pkg in data["packages"]:
```
nottrobin, line 366: How about `package_data`? Until we start writing Go, I'd prefer we avoid truncated variable names if we can.

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            releases = []
+            for rel in pkg["releases"]:
```
nottrobin, line 368: Simillarly, how about `release`?

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+                release = PackageReleaseStatus(
+                    name=rel["name"],
+                    status=rel["status"],
+                    status_description=rel["status_description"],
+                )
+            releases.append(release)
```
nottrobin, line 374: Pretty sure this needs to be indented 4 more spaces.

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+    # Check if there are packages before mapping
+    if "packages" in data and len(data["packages"]) > 0:
+        for pkg in data["packages"]:
+            releases = []
+            for rel in pkg["releases"]:
+                release = PackageReleaseStatus(
+                    name=rel["name"],
+                    status=rel["status"],
+                    status_description=rel["status_description"],
+                )
```
nottrobin, line 373: Single user variables can often be avoided. How about:

```suggestion
                releases.append(
                    PackageReleaseStatus(
                        name=rel["name"],
                        status=rel["status"],
                        status_description=rel["status_description"],
                    )
                )
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            releases.append(release)
+            package = Package(
+                name=pkg["name"],
+                source=pkg["source"],
+                ubuntu=pkg["ubuntu"],
+                debian=pkg["debian"],
+                releases=releases,
+            )
```
nottrobin, line 381: Similarly, how about:

```suggestion
            packages.append(
                Package(
                    name=pkg["name"],
                    source=pkg["source"],
                    ubuntu=pkg["ubuntu"],
                    debian=pkg["debian"],
                    releases=releases,
                )
            )
```

webapp/security/views.py
nottrobin, line 374: Single user variables can often be avoided. How about:

```suggestion
                releases.append(
                    PackageReleaseStatus(
                        name=rel["name"],
                        status=rel["status"],
                        status_description=rel["status_description"],
                    )
                )
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            packages.append(package)
```
nottrobin, line 382: Similarly, how about:

```suggestion
            packages.append(
                Package(
                    name=pkg["name"],
                    source=pkg["source"],
                    ubuntu=pkg["ubuntu"],
                    debian=pkg["debian"],
                    releases=releases,
                )
            )
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+
+    if "references" in data and len(data["references"]) > 0:
+        for ref in data["references"]:
+            reference = CVEReference(uri=ref)
+            references.append(reference)
```
nottrobin, line 387:
```suggestion
            references.append(CVEReference(uri=ref))
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+
+    if "bugs" in data and len(data["bugs"]) > 0:
+        for b in data["bugs"]:
+            bug = Bug(uri=b)
+            bugs.append(bug)
```
nottrobin, line 392:
```suggestion
            bugs.append(Bug(uri=b))
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+
+    cves = [
+        CVE(
```
nottrobin, line 395: You're only adding one CVE? In which case, why make it a list?

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            id=data["id"],
+            status=data["status"] if "status" in data else "",
+            last_updated_date=data["last_updated_date"]
+            if "last_updated_date" in data
+            else None,
```
nottrobin, line 400:
```suggestion
            last_updated_date=data.get("last_updated_date")
```

nottrobin: In this case this can be shortened anyway. But in general, I don't think it's very nice to put these ternary if statements on multiple lines. In future if you need to do this please make it a proper if statement afterwards:

``` python3
if "last_updated_date" in data:
    cve.last_updated_date = data["last_updated_date"]
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            public_date_usn=data["public_date_usn"]
+            if "public_date_usn" in data
+            else None,
```
nottrobin, line 403:
```suggestion
            public_date_usn=data.get("public_date_usn")
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            priority=data["priority"] if "priority" in data else None,
+            cvss=data["cvss"] if "cvss" in data else "",
+            assigned_to=data["assigned_to"] if "assigned_to" in data else "",
+            discovered_by=data["discovered_by"]
+            if "discovered_by" in data
+            else "",
```
nottrobin, line 409:
```suggestion
            discovered_by=data.get("discovered_by", "")
```

(Or `discovered_by=data.get("discovered_by") or ""`)

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            approved_by=data["approved_by"] if "approved_by" in data else "",
+            description=data["description"] if "description" in data else "",
+            ubuntu_description=data["ubuntu_description"]
+            if "ubuntu_description" in data
+            else "",
```
nottrobin, line 414:
```suggestion
            ubuntu_description=data,get("ubuntu_description", "")
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            if "public_date_usn" in data
+            else None,
+            priority=data["priority"] if "priority" in data else None,
+            cvss=data["cvss"] if "cvss" in data else "",
+            assigned_to=data["assigned_to"] if "assigned_to" in data else "",
+            discovered_by=data["discovered_by"]
+            if "discovered_by" in data
+            else "",
+            approved_by=data["approved_by"] if "approved_by" in data else "",
+            description=data["description"] if "description" in data else "",
```
nottrobin, line 411:
```suggestion
            approved_by=data.get("approved_by", ""),
            description=data.get("description", ""),
```

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
-def api_create_cve():
-    pass
+def create_cve():
+    """
+    Receives a POST request from load_cve.py
+    Parses the object and bulk inserts with add_all()
+    @params json: the body of the request
+    """
+
+    data = flask.request.get_json()
```
nottrobin, line 352: I *believe* this can simply be `flask.request.json`...

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            assigned_to=data["assigned_to"] if "assigned_to" in data else "",
+            discovered_by=data["discovered_by"]
+            if "discovered_by" in data
+            else "",
+            approved_by=data["approved_by"] if "approved_by" in data else "",
+            description=data["description"] if "description" in data else "",
+            ubuntu_description=data["ubuntu_description"]
+            if "ubuntu_description" in data
+            else "",
+            notes=data["notes"] if "notes" in data else "[]",
```
nottrobin, line 415: Don't you actually want this to be submitted as a proper list, rather than a "[]" string? 'cos presumably `data["Notes"]` is in fact going to be a Python data structure rather than a string.

```suggestion
            notes=data.get("notes", []),
```

^ Notice I removed the quotation marks from `"[]")`

webapp/security/views.py
carkod: Because the SQLAlchemy method `add_all` at the end only accepts a list

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+            packages=packages,
+            references=references,
+            bugs=bugs,
+        )
+    ]
+
+    try:
+        db_session.add_all(cves)
+        db_session.commit()
+    except exc.SQLAlchemyError as e:
```
nottrobin, line 425: This is nit-picky, but how about `error`?

webapp/security/views.py
carkod: But you can't do that inside of a class instance

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+        response = flask.jsonify({"message": e}), 400
+        return response
+
+    response = flask.jsonify({"message": "CVE created succesfully"}), 200
```
nottrobin, line 429: Please return a `201` here ("Created") - https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success

webapp/security/views.py
nottrobin: By the way, I think this stackoverflow answer makes a convincing case for a `409` status in this case:

> My feeling is 409 Conflict is the most appropriate, however, seldom seen in the wild of course:
> 
> > The request could not be completed due to a conflict with the current state of the resource. This code is only allowed in situations where it is expected that the user might be able to resolve the conflict and resubmit the request. The response body SHOULD include enough information for the user to recognize the source of the conflict. Ideally, the response entity would include enough information for the user or user agent to fix the problem; however, that might not be possible and is not required.
> > 
> > Conflicts are most likely to occur in response to a PUT request. For example, if versioning were being used and the entity being PUT included changes to a resource which conflict with those made by an earlier (third-party) request, the server might use the 409 response to indicate that it can't complete the request. In this case, the response entity would likely contain a list of the differences between the two versions in a format defined by the response Content-Type.

https://stackoverflow.com/a/3826024/613540

webapp/security/views.py
```diff
@@ -333,5 +342,208 @@ def cve(cve_id):
...
+    return response
```
nottrobin, line 430: Again, you don't need the extra variable:

```suggestion
    return  flask.jsonify({"message": "CVE created succesfully"}), 201
```

nottrobin:
Thanks for this @carkod. I've added a number of suggestions to `create_cve`, and my comments basically apply all the way down the code, so would you mind just going through your other functions and considering my points for all of them?

I'll QA it when you feel it's ready again. Let me know.

webapp/security/models.py
nottrobin: cool. Yeah just check it works as you expect.

webapp/security/views.py
nottrobin: Umm... You use `add_all` if you have a list, you use [`add`](https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.Session.add) if you have a single item. Right?

webapp/app.py
```diff
@@ -170,6 +173,13 @@ def utility_processor():
 app.add_url_rule(
     "/security/<regex('(cve-|CVE-)\\d{4}-\\d{4,7}'):cve_id>", view_func=cve
 )
+app.add_url_rule("/security/cve", view_func=create_cve, methods=["POST"])
+app.add_url_rule("/security/cve", view_func=update_cve, methods=["PUT"])
```
nottrobin, line 177: We agreed that we could support updating multiple CVEs at `/security/cve`, but what you seem to be offering here is to update a single CVE. This is not the right interface for that. If you want to offer updating a single CVE it should look like this:

``` python3
app.add_url_rule(
    "/security/<regex('(cve-|CVE-)\\d{4}-\\d{4,7}'):cve_id>",
    view_func=update_cve,
    methods=["PUT"]
)
```

webapp/app.py
```diff
@@ -170,6 +173,13 @@ def utility_processor():
 app.add_url_rule(
     "/security/<regex('(cve-|CVE-)\\d{4}-\\d{4,7}'):cve_id>", view_func=cve
 )
+app.add_url_rule("/security/cve", view_func=create_cve, methods=["POST"])
```
nottrobin, line 176: You were also talking of supporting adding multiple CVEs here. Did you decide against that? Or were you gonna do it in a follow-up PR instead?

webapp/security/views.py
```diff
@@ -333,5 +342,185 @@ def cve(cve_id):
...
+def update_cve():
+    data = flask.request.json
+    cve = db_session.query(CVE).filter(CVE.id == data["id"]).first()
+
+    # Check if CVE exists by candidate
+    if len(db_session.query(CVE).all()) == 0:
+        response = flask.jsonify({"message": "CVE does not exist"}), 400
+        return response
+
+    cve.status = data.get("status")
```
nottrobin, line 428: A question - what happens if you set an existing property on an existing item to `None`? Does it set the property to `NULL`? Or does it leave the property as it was?

webapp/security/views.py
```diff
@@ -333,5 +342,185 @@ def cve(cve_id):
...
+    cve.crd = data.get("crd")
+    cve.cvss = data.get("cvss")
+    cve.assigned_to = data.get("assigned_to")
+    cve.discovered_by = data.get("discovered_by")
+    cve.approved_by = data.get("approved_by")
+    cve.description = data.get("description")
+    cve.ubuntu_description = data.get("ubuntu_description")
+    cve.notes = data.get("notes")
+
+    if "references" in data and len(data["references"]) > 0:
```
nottrobin, line 442: Assuming `data["references"]` will contain a list, you don't need to test `len(data["references"]) > 0` because `bool([]) == False`. An empty list is Falsey.

webapp/security/views.py
```diff
@@ -333,5 +342,185 @@ def cve(cve_id):
...
+            reference = (
+                db_session.query(CVEReference)
+                .filter(CVEReference.uri == uri)
+                .first()
+            )
+            if not reference:
+                reference = CVEReference(uri=uri)
+            cve.references.append(reference)
+
+    if "bugs" in data and len(data["bugs"]) > 0:
```
nottrobin, line 454:
```suggestion
    if data.get("bugs"):
```

webapp/security/views.py
```diff
@@ -333,5 +342,185 @@ def cve(cve_id):
...
+        cve.bugs.clear()
+        for b in data["bugs"]:
+            bug = db_session.query(Bug).filter(Bug.uri == uri).first()
+            if not bug:
+                bug = Bug(uri=b)
+            cve.bugs.append(bug)
+
+    # Packages
+    # Check if there are packages before mapping
+    if "packages" in data and len(data["packages"]) > 0:
```
nottrobin, line 464:
```suggestion
    if data.get("packages"):
```

webapp/security/views.py
```diff
@@ -333,5 +342,185 @@ def cve(cve_id):
...
+    Delete a CVE from db
+    @params string: query string with the CVE id
+    """
+    cve_query = db_session.query(CVE)
+    cve = cve_query.filter(CVE.id == cve_id).first()
+
+    try:
+        db_session.delete(cve)
+        db_session.commit()
+    except exc.IntegrityError as error:
```
nottrobin, line 520: Could you explain what's going on here? When does an integrity error happen and why is it a special case?

carkod: Well, the only way I could catch the duplicate key error was through this integrity error, and this is still a wrapper of another SQL error.

webapp/app.py
carkod: Well until I don't have a script that sends me a bulk of CVEs (or a list of CVEs) I cannot test that code, so it will not happen in this PR. Or if you prefer I can separate them.

webapp/security/views.py
carkod: Yes. If I set `cve.status`it will set it to `Null` although I think this is SQL alchemy doing its thing

webapp/security/models.py
```diff
@@ -99,7 +99,7 @@ class CVE(Base):
     cvss = Column(String)  # CVSS vector to convert into Base score
     references = relationship("CVEReference", secondary=cve_references)
     bugs = relationship("Bug", secondary=cve_bugs)
-    packages = relationship("Package", secondary=cve_packages)
+    packages = relationship("Package", secondary=cve_packages,)
```
nottrobin, line 102: why the extra comma?

webapp/security/views.py
```diff
@@ -333,5 +342,182 @@ def cve(cve_id):
...
+
+    if data.get("references"):
+        for ref in data["references"]:
+            references.append(CVEReference(uri=ref))
+
+    if data.get("bugs"):
+        for b in data["bugs"]:
+            bugs.append(Bug(uri=b))
+
+    cves = CVE(
```
nottrobin, line 387: please call this `cve`, singular. It is only one.

webapp/security/views.py
```diff
@@ -333,5 +342,182 @@ def cve(cve_id):
...
+    except exc.IntegrityError as error:
+        return flask.jsonify({"message": error.orig.args[0]}), 409
+    # Payload input errors
+    except exc.DataError as error:
+        return (
+            flask.jsonify({"message": error.orig.args[0]}),
+            400,
+        )
+    except exc.SQLAlchemyError as error:
+        return flask.jsonify({"message": error}), 400
```
nottrobin, line 418: What cases couldn't be successfully handled by a `DBAPIError`?

```suggestion
    except exc.DBAPIError as error:
        return flask.jsonify({"message": error.orig.args[0]}), 400
```

webapp/security/views.py
```diff
@@ -333,5 +342,182 @@ def cve(cve_id):
...
+    except exc.DataError as error:
+        return (
+            flask.jsonify({"message": error.orig.args[0]}),
+            400,
+        )
+    except exc.SQLAlchemyError as error:
+        return (
+            flask.jsonify({"message": error.args[0]}),
+            400,
+        )
```
nottrobin, line 502: Similarly, in what cases are you expecting a `SQLAlchemyError` that isn't a `DataError`?

```suggestion
    except exc.DataError as error:
        return flask.jsonify({"message": error.orig.args[0]}), 400
```

nottrobin:
Just a couple of comments, but this is basically code +1 from me now. I just need to QA tomorrow.

webapp/security/views.py
carkod: Yes, these catch duplicate key error (CVE id that already exists) and incorrect input format (in this case only datetime)

webapp/security/views.py
```diff
@@ -333,5 +342,167 @@ def cve(cve_id):
...
+                package.releases.append(package_release_status)
+            cve.packages.append(package)
+
+    try:
+        db_session.add(cve)
+        db_session.commit()
+    # Payload input errors
+    except exc.DataError as error:
+        return (
+            flask.jsonify({"message": error.orig.args[0]}),
```
nottrobin, line 487: Not sure this needs to be wrapped onto multiple lines

carkod: Yes it does. It's over 79 lines, so lint will complain.

webapp/security/views.py
carkod: Data error is nearly always going to be about validation of the format of dates in this API. It may catch some other payload errors, but this is the only thing we have in the db that enforces a certain format.

nottrobin:
OK QAing this I ran into multiple issues.

Firstly, I don't know why you suggested using `python2` to run the script, it doesn't work at all with `python2` (it uses f-strings - 3.6+ - all over the place, and `os.makedirs(exist_ok=True)` - 3 only).

So my QA steps look something like this:

``` bash
git clone -b load-cve git@github.com:carkod/ubuntu-cve-tracker.git
cd ubuntu-cve-tracker
python3 -m venv .venv
source .venv/bin/activate
pip3 install requests gitpython cvss
TRACKER_PORT=8001 scripts/upload-cve.py active/CVE=2020-2727
```

This worked - I saw `{'message': 'CVE created succesfully'}`, and the CVE was indeed created successfully.

However, when I tried `update-cve.py` I wasn't so successful:

``` bash
$ TRACKER_PORT=8001 python scripts/update-cve.py active/CVE-2020-2727
Traceback (most recent call last):
  File "scripts/update-cve.py", line 117, in <module>
    print(response.json())
  File "/home/robin/Projects/ubuntu-cve-tracker/.venv/lib/python3.6/site-packages/requests/models.py", line 898, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

And the error I see in the server is:

```
  File "/home/robin/Projects/ubuntu.com/.venv/lib/python3.6/site-packages/sqlalchemy/sql/elements.py", line 293, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/home/robin/Projects/ubuntu.com/.venv/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1103, in _execute_clauseelement
    distilled_params,
  File "/home/robin/Projects/ubuntu.com/.venv/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1178, in _execute_context
    e, util.text_type(statement), parameters, None, None
  File "/home/robin/Projects/ubuntu.com/.venv/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1482, in _handle_dbapi_exception
    sqlalchemy_exception, with_traceback=exc_info[2], from_=e
  File "/home/robin/Projects/ubuntu.com/.venv/lib/python3.6/site-packages/sqlalchemy/util/compat.py", line 178, in raise_
    raise exception
  File "/home/robin/Projects/ubuntu.com/.venv/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1175, in _execute_context
    context = constructor(dialect, self, conn, *args)
  File "/home/robin/Projects/ubuntu.com/.venv/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 808, in _init_compiled
    param.append(processors[key](compiled_params[key]))
  File "/home/robin/Projects/ubuntu.com/.venv/lib/python3.6/site-packages/sqlalchemy/dialects/sqlite/base.py", line 759, in process
    "SQLite DateTime type only accepts Python "
sqlalchemy.exc.StatementError: (builtins.TypeError) SQLite DateTime type only accepts Python datetime and date objects as input.
[SQL: UPDATE cve SET public_date=?, crd=?, cvss=? WHERE cve.id = ?]
[parameters: [{'cvss': 6.0, 'crd': 'CRD', 'public_date': '2020-01-15 17:15:00 UTC', 'cve_id': 'CVE-2020-2727'}]]
```

This also then corrupts the database session so now nothing works:

image

This was fixed by Joao in his USN work, so either rebasing or copying his fix should help with the last problem.

webapp/security/views.py
```diff
@@ -333,5 +342,167 @@ def cve(cve_id):
 # ===
 
 
-def api_create_cve():
-    pass
+def create_cve(cve_id):
```
nottrobin, line 345: Could we secure these with `authentication_required` please? As we've done with `create_notice` in master.

carkod: Can we do this after we rebase with master? because I don't want to be copying more code from master

nottrobin: That feels pretty risky. It would be easy for this to accidentally go live with no auth.

carkod: Ok, so after I implemented this, I cannot QA until we can generate a discharge in the script?

carkod: Done. Although the script is not working now due to lack of macaroon discharge. I am working on this see if I can come up with one

carkod:
Once It's QAed I will squash. Way too many commits

webapp/security/views.py
nottrobin: I can help fix up the script

carkod: Yeah, but I think it's working for the purposes of this API now. I am expecting that the sec guys will introduce some changes, so once this API is tested, I can focus on search and the script

nottrobin:
@carkod when I first visit `/security/cve` I see:

image

With "None"s in the fields. Is that this branch?

carkod:
> @carkod when I first visit `/security/cve` I see:
> 
> image
> 
> With "None"s in the fields. Is that this branch?

I haven't touched the templates, but yes this is happening because the templates are using the query parameters and they have no defaults set (remember we removed them?) but I can modify the templates to empty string.

nottrobin:
Great work @carkod all works apart from the "None" in the form fields. Once that's fixed, +1

carkod:
@nottrobin thanks, it took me a long time to QA!! I did it about 7 times this week with 3 different types of CVEs

nottrobin:
@carkod ok merge away =)

PR #7261, opened 2020-04-16
add renewals invoice endpoint

sowasred2012:
When a payment fails, the user can attempt to pay with updated payment method details, in which case we need to post to this endpoint in order for the ua-contracts service to use the new payment method on the current payment.

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -41,6 +41,7 @@ from webapp.views import (
     download_thank_you,
     get_renewal,
     post_stripe_method_id,
+    post_stripe_invoice_id,
     releasenotes_redirect,
 )
 from webapp.login import login_handler, logout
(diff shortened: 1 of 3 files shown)


webapp/app.py
```diff
@@ -121,7 +122,12 @@ def utility_processor():
     methods=["POST"],
 )
 app.add_url_rule(
-    "/advantage/renewals/{renewal_id}", view_func=get_renewal, methods=["GET"]
+    "/advantage/renewals/invoice",
```
tbille, line 125: `"/advantage/renewals/<renewal_id>/invoice"`

webapp/views.py
```diff
@@ -258,6 +258,26 @@ def post_stripe_method_id():
         return flask.jsonify({"error": "authentication required"}), 401
 
 
+def post_stripe_invoice_id():
```
tbille, line 261:
```suggestion
def post_stripe_invoice_id(invoice):
```

webapp/views.py
```diff
@@ -258,6 +258,26 @@ def post_stripe_method_id():
...
+    if auth.is_authenticated(flask.session):
+        if not flask.request.is_json:
+            return flask.jsonify({"error": "JSON required"}), 400
+
+        invoice_id = flask.request.json.get("invoice_id")
```
tbille, line 266: sam  it can be removed

webapp/views.py
```diff
@@ -258,6 +258,26 @@ def post_stripe_method_id():
         return flask.jsonify({"error": "authentication required"}), 401
 
 
+def post_stripe_invoice_id():
+    if auth.is_authenticated(flask.session):
+        if not flask.request.is_json:
```
tbille, line 263: no json anymore

PR #7281, opened 2020-04-20
Search based on the selected architecture

anthonydillon:
## Done
- Upgraded the store-api module to the latest.
- Map the board and os to derive the arch and filter snap searches based on it.
- If the selection changes which results in a different arch the snap list is clearer for safety.

## QA
- Sign in and select a board and OS
- You should see a sentence above the search with lets you know the arch you have selected and warn that searches are filtered on this.
- Search for something and add a few snaps
- Then make a different selection, one resulting in different architecture and see that snap lists clear.

## Issue / Card
Fixes https://github.com/canonical-web-and-design/web-squad/issues/2395

### DIFF
--- a/static/js/src/imageBuilder.js
+++ b/static/js/src/imageBuilder.js
@@ -1,4 +1,37 @@
-(function() {
+(function () {
+  const board_architectures = {
+    raspberrypi2: {
+      core16: { arch: "armhf", subarch: "raspi2" },
+      core18: { arch: "armhf", subarch: "raspi3" },
+      "classic16.04": { arch: "armhf", subarch: "raspi3" },
+      "classic18.04": { arch: "armhf", subarch: "raspi3" },
+    },
+    raspberrypi3: {
+      core16: { arch: "armhf", subarch: "raspi3" },
+      core18: { arch: "armhf", subarch: "raspi3" },
+      "classic16.04": { arch: "armhf", subarch: "raspi3" },
+      "classic18.04": { arch: "armhf", subarch: "raspi3" },
+      "classic6418.04": { arch: "arm64", subarch: "raspi3" },
+    },
+    raspberrypi4: {
+      core18: { arch: "armhf", subarch: "raspi3" },
+      "classic18.04": { arch: "armhf", subarch: "raspi3" },
+      "classic6418.04": { arch: "arm64", subarch: "raspi3" },
+    },
+    intelnuc: {
+      core16: { arch: "amd64", subarch: "" },
+      core18: { arch: "amd64", subarch: "" },
+    },
+    snapdragon: {
+      core16: { arch: "arm64", subarch: "snapdragon" },
+      core18: { arch: "arm64", subarch: "snapdragon" },
+    },
+    cm3: {
+      core16: { arch: "armhf", subarch: "cm3" },
+      core18: { arch: "armhf", subarch: "raspi3" },
+    },
+  };
+
   // State management
   class StateArray extends Array {
     push(item) {
@@ -16,6 +49,12 @@
       render();
       return item;
     }
+    reset() {
+      while (this.length > 0) {
+        this.pop();
+      }
+      render();
+    }
   }
 
   class StateManager {
@@ -23,7 +62,7 @@
       this.states = {
         board: new StateArray(),
         os: new StateArray(),
-        snaps: new StateArray()
+        snaps: new StateArray(),
       };
     }
 
@@ -44,6 +83,7 @@
   const osSelection = document.querySelectorAll(".js-os .js-selection");
   const snapSearch = document.querySelector(".js-snap-search");
   const snapResults = document.querySelector(".js-snap-results");
+  const archOutput = document.querySelector(".js-architecture");
   const preinstallResults = document.querySelector(
     ".js-preinstalled-snaps-list"
   );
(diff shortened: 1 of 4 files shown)


static/js/src/imageBuilder.js
```diff
@@ -1,4 +1,37 @@
-(function() {
+(function () {
+  const board_architectures = {
```
nottrobin, line 2: This is not what we discussed. Duplicating this dictionary into the frontend is horrible.

static/js/src/imageBuilder.js
```diff
@@ -78,17 +118,20 @@
...
+    const os = parseSystemValues(state.get("os")[0]);
+    const architecture = board_architectures[board][os]["arch"];
     snapResults.innerHTML =
       '<p><i class="p-icon--spinner u-animation--spin"></i></p>';
     const searchInput = snapSearch.querySelector(".p-search-box__input");
     if (searchInput) {
       const searchValue = encodeURI(searchInput.value);
-      fetch(`/snaps?q=${searchValue}&size=12`)
-        .then(response => {
+      fetch(`/snaps?q=${searchValue}&size=12&architecture=${architecture}`)
```
nottrobin, line 130: As we discussed, let's change this interface to:

```
/snaps?q=${searchValue}&size=12&board=${board}&system=${system}
```

anthonydillon:
@nottrobin would you mind taking a look at this one please?

clagom:
I think when users select a board and go down to step 3 then when they pick a different board/architecture the UI should reset the flow to step 1 again and grey out the reset, instead of keeping step 2 and 3 active. In this way it would be clearer that they changed an important part of the flow that affects step 2 and 3.

Also, the search input field should also empty again, not the previous search stays there:
Screenshot 2020-04-22 at 10 20 42

A nice touch would also be to communicate to users what these two empty boxes are there for, maybe including a call to action 'search for snaps to include' or something instead of this:
Screenshot 2020-04-22 at 10 26 46

One last point, I think "Your selected archtecture is arm64. If you are missing a snap its probably not supported by this archtecture." is not prominent enough, visually, and 'archtecture' is misspelled twice, should be 'architecture'.

static/js/src/imageBuilder.js
```diff
@@ -310,10 +343,22 @@
   function checkDisabled() {
     step2.classList.add("u-disable");
     step3.classList.add("u-disable");
-    if (state.get("board") && state.get("board")[0] != "") {
+    if (state.get("board") && state.get("board")[0]) {
```
barrymcgee, line 346: Can simplify with optional chaining - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Array_item_access_with_optional_chaining

```suggestion
    if (state.get("board")?.[0]) {
```

static/js/src/imageBuilder.js
```diff
@@ -310,10 +343,22 @@
...
       step2.classList.remove("u-disable");
     }
-    if (state.get("os") && state.get("os")[0] != "") {
+    if (state.get("os") && state.get("os")[0]) {
```
barrymcgee, line 349:
```suggestion
    if (state.get("os")?.[0]) {
```

templates/core/build.html
```diff
@@ -199,6 +199,7 @@ <h2><span class="u-number-circle">2</span> What version of Ubuntu do you want?</
     <div class="col-10 u-sv3">
       <h2><span class="u-number-circle">3</span> What packages do you want preinstalled?</h2>
       <p>Select the applications that will be pre-installed on the device. For Ubuntu Core this will be snaps, for Ubuntu Classic it will be snaps or packages.</p>
+      <p>Your selected archtecture is <strong class="js-architecture"></strong>. If you are missing a snap its probably not supported by this archtecture.</p>
```
barrymcgee, line 202: Can this have a fallback value in case JS fails?

static/js/src/imageBuilder.js
anthonydillon: I think that is much harder to read. Isn't it better to keep the code readable over shorter?

templates/core/build.html
anthonydillon: Yeah sure, but the whole page is greyed out if no JS loads anyway so though the masking would cover it.

anthonydillon:
@clagom thanks for the review, working through them. Some of the changes revert suggestions from @matthewpaulthomas but will implement and see if we think it works.

static/js/src/imageBuilder.js
barrymcgee: Optional chaining has been introduced to the spec specifically to replace that hack of truthy short circuit evaluation. As such, it will become _the_ correct way to check if something exists before using it.

It looks weird the first couple of times you see it but then you see duplication everywhere else that can be eradicated by using it.

anthonydillon: I tried to add it but eslint is failing with:
`360:28  error  Parsing error: Unexpected token .`

templates/core/build.html
anthonydillon: Done, made the notification hidden until there is data.

anthonydillon:
@clagom updated with your required changes. Can you review again please.

clagom:
Thanks @anthonydillon I think it works much better now, also the box style for the architecture bit.
Is this supposed to be inactive until users check the agreement?
Screenshot 2020-04-22 at 16 32 36

anthonydillon:
Thanks, no this is just so we can use them to send marketing material too.

PR #7286, opened 2020-04-21
handle payment success response from ua-contract service

sowasred2012:
## Done

A successful payment returns a 200 response with no body from the ua-contracts service, which was causing a 500 on our side when we tried to parse the body.

### DIFF
--- a/webapp/api/advantage.py
+++ b/webapp/api/advantage.py
@@ -144,7 +144,7 @@ def accept_renewal(session, renewal_id):
         raise_http_errors=False,
     )
 
-    if response.status_code == 200:
+    if response.ok:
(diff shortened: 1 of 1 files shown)


webapp/api/advantage.py
```diff
@@ -144,4 +144,7 @@ def accept_renewal(session, renewal_id):
         raise_http_errors=False,
     )
 
-    return response.json()
+    if response.status_code == 200:
```
tbille, line 147:
```suggestion
    if response.ok:
```
In case they start introducing redirects or else.
Doc: https://2.python-requests.org/en/master/api/#requests.Response.ok

PR #7297, opened 2020-04-21
Handle db migration errors

nottrobin:
DB migration errors have caused a lot of difficulties with
QAing. This should improve things by prompting the user
to erase the database when migration errors are encountered.

QA
--

```
git checkout cve
rm usn.sqlite3
git checkout handle-db-migration-errors  # back to this branch
dotrun serve
```

^ this should now error on DB migration and prompt you to delete your database. If you choose "y", it should successfully run the site.

### DIFF
--- a/entrypoint
+++ b/entrypoint
@@ -3,7 +3,17 @@
 set -e
 
 # Run migrations
-alembic upgrade head
+{
+    alembic upgrade head
+} || {
+    read -p "Database migration failed. Erase database and try again? " -n 1 -r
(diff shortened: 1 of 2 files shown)


entrypoint
```diff
@@ -3,7 +3,17 @@
 set -e
 
 # Run migrations
-alembic upgrade head
+{
+    alembic upgrade head
+} || {
+    read -p "Database migration failed. Erase database and try again? " -n 1 -r
```
anthonydillon, line 9: I think you should indicate the options to the user? `(y/n)` for example.

PR #7300, opened 2020-04-22
Add clear filter button to CVE search

### LINKED ISSUE #7268: CVE pages - add button to clear search

steverydz:
## Done

- Added a "Clear filters" button to the CVE search
- Commented out links to /security/cve for soft launch

## QA

- Run the following commands
- `docker-compose up -d`
- If you get an error saying something is already running on that port: `sudo kill -9 `sudo lsof -t -i :5432``
- `dotrun --env DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres`
- `export DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres`
- `alembic upgrade head`
- `dotrun exec python3 webapp/security/fixtures/load_sample_cve.py`
- If you have an error here you may need to drop the database schema. To do that follow the steps in this comment: https://github.com/canonical-web-and-design/ubuntu.com/pull/7242#issuecomment-616435249
- `dotrun`
- Use the search form and submit it
- Click the "Clear filters" button and check that the form fields have been cleared

## Issue / Card

Fixes #7268

### DIFF
--- a/navigation.yaml
+++ b/navigation.yaml
@@ -811,8 +811,9 @@ security:
     - title: Notices
       path: https://usn.ubuntu.com
       persist: True
-    - title: CVEs
-      path: /security/cve
+    # Hidden for soft launch
+    # - title: CVEs
+    # path: /security/cve
 
 licensing:
   title: Licensing
--- a/templates/security/cve/index.html
+++ b/templates/security/cve/index.html
@@ -43,9 +43,14 @@
           </option>
         </select>
       </div>
-      <div class="col-3">
-        <label>&nbsp;</label> <!-- required for layout -->
-        <button type="submit">Search</button>
+      <div class="col-3 u-align--right">
+        <div class="p-form__group">
+          <label class="p-form__label">&nbsp;</label> <!-- required for layout -->
+          <div class="p-form__control">
+            <a href="/security/cve" class="p-button--neutral">Clear filters</a>
+            <button type="submit" class="p-button--positive">Search</button>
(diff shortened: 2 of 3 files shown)


nottrobin:
The QA steps for these branches are getting really onerous. Using postgres looks like it's introducing loads of issues here.

It's my understanding that everything should actually work fine with sqlite, and if it doesn't, it should be fixed so it does. Is that correct @carkod, @steverydz? Or did I miss something?

As long as that's true, in future please just suggest QA steps without postgres. It should all work fine with sqlite.

This should clean up postgres:

``` bash
docker-compose down
unset DATABASE_URL  # And remove any mentions from .env.local or ~/.bashrc
```

And this should be the new, simpler, much less likely to break, steps:

``` bash
# This is the steps without postgres
dotrun exec python3 webapp/security/fixtures/load_sample_cve.py
dotrun
```

If it does break, simply `rm usn.sqlite3` should fix it. Or merging my PR from master: https://github.com/canonical-web-and-design/ubuntu.com/pull/7297

nottrobin:
Oh my bad, it does break with sqlite. That's not good. Let's fix that. But this does need sqlite :(

steverydz:
@nottrobin As far as I was aware the only way to populate the table with data was to use those steps. It seems that's not the case so going forwards I'll use what you've recommended.

carkod:
@nottrobin @steverydz Well, my API PR should fix all the sqlite3 issues

PR #7316, opened 2020-04-23
DON'T MERGE Revert https releases for mirrors

tbille:
# Done 

reverts https restriction for mirrors https://github.com/canonical-web-and-design/ubuntu.com/pull/7264
This is in case reverting is necessary to unload some mirrors etc.


xnox:
Release announcement email & release notes will use `https://` however is that ok?

pmahnke:
@xnox yes, that's fine, this is if releases.u.c cannot handle the https traffic of downloads

tbille:
@xnox it's only for the mirrors. In case the mirrors can't handle the traffic. We'll allow also non https mirrors.

It's just a `in case` and ready to be merged :smile:

PR #7330, opened 2020-04-23
Bulk api

### LINKED ISSUE #7302: Bulk API endpoints
## Summary

- Add API endpoints for bulk inserts and updates.
- Refactor code to accept lists (arrays)
- Update upload script to use these

carkod:
## Done

- One endpoint to bulk create and update
- Refactor (old endpoints and error handling)
- Fixed issues with main table.

## QA

- Delete the `usn.sqlite3` file
- Run `python webapp/security/fixtures/releases.py`
- Download/clone/rebase current ubuntu [tracker upload script](https://github.com/nottrobin/ubuntu-cve-tracker/pull/2) 
- Run  `python scripts/upload-cve.py .../active` in ubuntu tracker, it will upload the last 100 CVEs, but this order could be random depending on filesystem order, so you could not be grabbing the latest CVE. Print the `list_cves` in the upload script if you want to get an accurate CVE list you are sending
- You can check `python scripts/upload-cve.py --help` to remind you of the possible parameters.
- Check `http:localhost:8001/security/cve`, if you can't find it (because 100s of CVEs have been uploaded) check - `> sqlite3 SELECT id FROM cve WHERE id={CVE-id}` where CVE-id is a CVE you uploaded with the script. 

## Issue / Card

Fixes #7302


webapp/security/views.py
```diff
@@ -536,3 +537,55 @@ def delete_cve(cve_id):
...
+        return (
+            flask.jsonify(
+                {"message": "Please, submit a list (array) of CVEs"}
+            ),
+            400,
+        )
+
+    for data in cves_data:
+        cve = db_session.query(CVE).get(data["id"])
+        if cve:
```
carkod, line 565: This was returning `None`, so I couldn't find a way to catch this with `exc.NoResultsFound` (SQLAlchemy's error handler) so had to it this way.

webapp/security/views.py
```diff
@@ -536,3 +513,57 @@ def delete_cve(cve_id):
...
+
+    except DataError as error:
+        return (
+            flask.jsonify(
+                {
+                    "message": "Failed bulk upserting session",
+                    "error": error.orig.args[0],
+                }
+            ),
+            200,
```
nottrobin, line 561: This should not be `200`. This should be either `400 Bad Request` or `409 Conflict`. Up to you ([see the list here](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_errors)).

webapp/security/views.py
```diff
@@ -352,171 +352,148 @@ def cve(cve_id):
...
+def update_cve(cve_data):
 
+    cve_object = []
 
-@authorization_required
-def update_cve(cve_id):
-    data = flask.request.json
-    cve = db_session.query(CVE).get(cve_id)
+    for data in cve_data:
+        cve = db_session.query(CVE).get(data["id"])
```
nottrobin, line 426: You're doing this query twice - you already did it on line 539. You should just pass through that `cve` object rather than getting it again.

webapp/security/views.py
```diff
@@ -536,3 +513,57 @@ def delete_cve(cve_id):
...
+@authorization_required
+def bulk_upsert_cve():
+    """
+    Receives a PUT request from load_cve.py
+    Parses the object and bulk inserts or updates
+    @returns 3 lists of CVEs, created CVEs, updated CVEs and failed CVEs
+    """
+
+    cves_data = flask.request.json
+    create = []
```
nottrobin, line 527: could we call this `cves_to_create` or something to make it clear that it's a collection of CVEs?

webapp/security/views.py
```diff
@@ -352,171 +352,148 @@ def cve(cve_id):
...
 
-    try:
-        db_session.add(cve)
-        db_session.commit()
-    # Duplicate key errors and Payload input errors
-    except (IntegrityError, DataError) as error:
-        return flask.jsonify({"message": error.orig.args[0]}), 409
 
-    return flask.jsonify({"message": "CVE created succesfully"}), 201
+def update_cve(cve_data):
```
nottrobin, line 421: This should now be `update_cves` right? It expects to operate on a list. Same with `create_cves`. Although really it's more like `prepare_cves_to_update` or something, as it doesn't commit anything - using the language `update` or `create` makes it sound like it's going to actually do the action.

Please try to be careful with your singulars and plurals - they are logically extremely important and using the wrong one could easily lead to mistakes in development.

webapp/security/views.py
```diff
@@ -536,3 +513,57 @@ def delete_cve(cve_id):
...
+
+    for data in cves_data:
+        cve = db_session.query(CVE).get(data["id"])
+        if cve:
+            update.append(data)
+        else:
+            create.append(data)
+
+    updated = update_cve(update)
+    created = create_cve(create)
```
nottrobin, line 546: Out of curiosity, what happens if you run `create_cve` for a CVE that already exists? Is there any chance that sqlalchemy simply takes care of it?

carkod: Do you mean `add_all` on existent CVEs? My code will never run `create_cve` on existent CVEs, because it's been filtered by line 539, this is done on purpose, because SQLAlchemy's `add_all` needs the original object to update. If you try to create the way I did it in the `create_cve`. it will complain with `Duplicate key error` so you have to feed it with an existent object, with all the fields correctly mapped. Which is why I separated it in 2 `add_all`.

webapp/security/views.py
```diff
@@ -358,168 +362,223 @@ def cve(cve_id):
...
-            for rel in package_data["releases"]:
-                releases.append(
-                    PackageReleaseStatus(
-                        name=rel["name"],
-                        status=rel["status"],
-                        status_description=rel["status_description"],
+                    # Change and match our releases table
+                    if release_data["name"] == "trusty/esm":
+                        release_data["name"] = "trusty"
+
```
carkod, line 395: We may need to revisit this, once we get from sec team how they are going to update the releases, because we don't always know if ESM has reached end of life or not

nottrobin: I don't quite understand this point. We *do* know whether a release is in the ESM phase or not, right? Because we're asking them to submit expiry dates for ESM expiry for releases, right?

It doesn't seem to me that our API should accept "trusty/esm" as a release name, as it's meaningless. "trusty" is the name, "esm" is extra data they use in their system which is irrelevant. We either know "trusty" is in ESM or we don't, regardless of what they submit here.

I think the removing of "esm" should be something done their side, so all they submit is "trusty".

carkod: No we don't know. Joe sent me this page which indicates some releases are still ESM for certain clients https://wiki.canonical.com/UbuntuEngineering/Security/ESM

webapp/security/views.py
```diff
@@ -539,3 +594,54 @@ def delete_cve(cve_id):
...
+            flask.jsonify(
+                {"message": "Please, submit a list (array) of CVEs"}
+            ),
+            400,
+        )
+
+    for data in cves_data:
+        cve = db_session.query(CVE).get(data["id"])
+        if cve:
+            cves_to_update.append(data)
```
nottrobin, line 622: Is adding this stuff to an array before calling `db_session.add_all` on it significantly more performant than simply calling `db_session.add()` within this for loop? It seems clear to me that the way SQLAlchemy deliberately separates the `add` step from the `commit` step is presumably specifically so you can build up your model piece by piece before you commit.

There's also significant duplication between the `update` and `create` functions.

Wouldn't this code be rather more simple if it were something like:

``` python3
for data in cves_data:
    cve = db_session.query(CVE).get(data["id"])

    if not cve:
        cve = CVE(data["id"])
    
    cve = update_cve_object(cve, data)

    db_session.add(cve)

try:
    db_session.commit()
```

carkod: No, the `add` does not add bulk (list) of objects, it only adds one.

webapp/security/views.py
nottrobin: > Well, we don't know yet. Joe sent me this page which indicates some releases are still ESM for certain clients https://wiki.canonical.com/UbuntuEngineering/Security/ESM, despite the fact they have ended ESM support for quite some time in the standard 5 year period. So these are really the fixes added, so I had to update the releases table. Therefore, we might need a releases API if they are willing to update the ESM dates?

@carkod did Joe tell you that packages in CVEs for expired releases are explicitly considered supported, in the sense of publicly communicating this on the CVE page, while those contracts exists? Yes we have special commercial contracts with some people which means we actually do some support beyond what we publicly promise, but this doesn't seem like it should effect our public data should it?

If we do consider these special contracts in our public information, maybe it could simply be handled by the existing expiry field, in that if we are in practice supporting security for packages in 14.04 until 2023 because of our Ericsson contract, then we could simply set the expiry date as 2023. Or if we support explicit packages in 14.04 because Ericsson asked as to, and we want to expose this publicly, then we would need a "under_special_support" field or something in our packages table.

I guess we need to check with Security which one it is.

> About the last paragraph. I hope that this is the final outcome, but for now the only way to fix the cve/index.html main table, is to fix it our side. I discussed this with @jpmartinspt seems like the most straight-forward way to solve this unpredictable releases they have. If we stick to e.g. trusty, this will cause the package updates to be fail to be recored on our db, because they are updating trusty/esm, not trusty anymore.

You wrote the script you're using to submit CVEs to our API right? It looks to me like that's [here](https://github.com/nottrobin/ubuntu-cve-tracker/pull/2/files#diff-9c88d40408d375390e8bbde08e0ef765R73). You should change this to remove the "ESM", and then change this API so it doesn't accept the "ESM". Any special wrangling of their data should happen in that script, not API side. Our API should be a clean representation of our data model.

carkod: Ok, that makes sense. I'll make the changes

webapp/security/views.py
nottrobin: > No, the add does not add bulk (list) of objects, it only adds one, it expects a single object, it will fail with this condition.

OK, so simply change my code above to say `add_all` instead of `add`.

>  It's not really about performance but about transaction, we would be doing 1 transaction per CVE if we use add().

I don't think this is true, I think the transaction is what happens when you do `db_session.commit()`. You `add` as many times as you like, I would assume including as many `add_all`s as you like, then you `commit`.

> And as mentioned before, to update we require the original object, while to create we do not need the original object, so there will be some degree of duplication (that's why I did 2 add_all).

My point here is that, if you have a new CVE, you can simply create a blank one with `CVE(data["id"])` and then update it exactly the same way you would if it already existed (see my code above). That way you don't need to duplicate any code at all.

templates/security/cve/index.html
```diff
@@ -74,7 +74,7 @@ <h2>Recent CVEs affecting Ubuntu</h2>
           <th>CVE</th>
           <th>Package</th>
           {% for release in releases %}
-          <th>{{ release.name }}</th>
+          <th>{{ "Ubuntu " + release.name + " " + release.version + " " + release.support_tag }}</th>
```
nottrobin, line 77:
```suggestion
          <th>Ubuntu {{ release.name }} {{ release.version }} {{ release.support_tag }}</th>
```

carkod:
QAed:

- Create:
Endpoint entry 1589406820.1440902
Start commit 1589406931.66946
Finished commit 1589407065.5054011

- Update (PUT request over existent CVEs):
Endpoint entry : 1589360.959093452
start commit: 1589405544.907006 
finished commit: 1589406267.5228338

with both ignored and active CVEs

PR #7348, opened 2020-04-27
CVE search form

steverydz:
## Done

Made CVE search form work as intended
Drive by: formatted the date strings on a CVE page

## QA

``` bash
rm usn.sqlite3
dotrun exec python3 webapp/security/fixtures/releases.py
dotrun exec python3 webapp/security/fixtures/load_sample_cve.py
dotrun
```

- Go to `/security/cve` and put a valid CVE ID (e.g. CVE-2020-10535) in the search field
- Check that the package and priority fields are disabled
- Check that the search button has turned into a button which takes you to the CVE page
- Enter a term from the CVE description into the search field e.g. any part of the following string assuming you used the example CVE ID above:

> GitLab 12.8.x before 12.8.6, when sign-up is enabled, allows remote attackers to bypass email domain

- Check that the CVE containing that description is returned as a result

### Check that the tests pass
```bash
dotrun exec yarn run test-js static/js/src/cve/cve-search.test.js
```

### DIFF
--- a/static/js/src/cve/cve-search.js
+++ b/static/js/src/cve/cve-search.js
@@ -1,0 +1,86 @@
+/**
+ * isValidCveId
+ *
+ * Checks if the supplied value is a valid CVE ID in the format
+ *
+ * A CVE ID consists consists of `cve-` (case-insensitive)
+ * followed by 4 integers, follwed by `-`, followed by 4 to 7 integers
+ *
+ * @param {string} value
+ * @returns {boolean}
+ */
+function isValidCveId(value) {
+  if (!value) {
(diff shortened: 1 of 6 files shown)


carkod:
@steverydz So I couple a things:
- The CVE id field serves for both searching for CVE-id and description. You can match the CVE id with the regex we use in the app.py routes, if it matches redirect to `security/cve/{cve-id}`. Else, you would always be searching description.
- When you select a priority, it immediately searches, but our search is commutative, so shouldn't you give the the user the possibility to search package, description and priority?

PR #7350, opened 2020-04-27
Fix bug with blogpost thumbnails not being displayed

steverydz:
## Done

Fix bug where blog post thumbnails aren't displayed

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/robotics/community
- Check that the posts in the "Recent news" section have thumbnails

### DIFF
--- a/static/js/src/latest-news.js
+++ b/static/js/src/latest-news.js
@@ -43,7 +43,9 @@
       article._embedded &&
       article._embedded["wp:featuredmedia"] &&
       article._embedded["wp:featuredmedia"][0] &&
-      article._embedded["wp:featuredmedia"][0].width
+      article._embedded["wp:featuredmedia"][0].media_details &&
+      article._embedded["wp:featuredmedia"][0].media_details.width &&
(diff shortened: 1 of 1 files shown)


static/js/src/latest-news.js
```diff
@@ -43,7 +43,8 @@
       article._embedded &&
       article._embedded["wp:featuredmedia"] &&
       article._embedded["wp:featuredmedia"][0] &&
-      article._embedded["wp:featuredmedia"][0].width
+      article._embedded["wp:featuredmedia"][0].media_details &&
+      article._embedded["wp:featuredmedia"][0].media_details.width
```
hatched, line 47: This check was updated here, but below when defining the `articleImage` you're still referencing `      article._embedded["wp:featuredmedia"][0].width` and       article._embedded["wp:featuredmedia"][0].height` was this intentional?

steverydz: There should be a height if there's a width but you're right, we don't know for sure

PR #7360, opened 2020-04-28
Remove autocomplete and reset from the search field

anthonydillon:
## Done
Remove autocomplete and reset from the search field

## QA
- View the site locally in your web browser at: http://0.0.0.0:8001/build
- Sign in
- Select and board and OS
- Search for something and see you do not see a autocomplete popover
- See that the reset or clear is not there. 

## Issue / Card
Fixes the comments here: https://github.com/canonical-web-and-design/ubuntu.com/pull/6695#issuecomment-620590768

### DIFF
--- a/templates/build/index.html
+++ b/templates/build/index.html
@@ -207,8 +207,7 @@
     <div class="col-6">
       <h4>Available snaps:</h4>
       <form class="p-search-box js-snap-search">
-        <input type="search" class="p-search-box__input" name="search">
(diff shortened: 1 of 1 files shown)


matthewpaulthomas:
I didn’t mean get rid of the clear button all the time, just when the field is empty. But I guess, better to never have it than always have it.

anthonydillon:
I know, but Vanillas clear functionality is provided by a valid input which requires `required` on the input.

PR #7364, opened 2020-04-28
Fix builds by using launchpad v0.5.1

nottrobin:
Introduces image_format=ubuntu-image when calling build API. Also requires
auth_consumer separately from username.

QA
--

`dotrun`, go to `/build`, submit a build, check it works fine.


jkfran:
@nottrobin I think you meant to merge this with image-builder instead of master :)

nottrobin:
@jkfran sorry thanks for QAing, that's a relief. =)

PR #7378, opened 2020-04-29
Move /build back to /core/build

nottrobin:
Because you can actually only build core images

QA
--

`dotrun`, go to `/build/`, get redirected to `/core/build`, check it works.


anthonydillon:
LGTM, do you need to change the webhook in launchpad?

PR #7388, opened 2020-04-30
Update stylelint config plugins to match vanilla

steverydz:
## Done

- Update the stylelint config to use the same plugins as Vanilla

## QA

- Check that the CSS linting passes: `dotrun exec yarn run lint-scss`
- Check that the CSS builds: `dotrun exec yarn run build-css`

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,8 +1,6 @@
 {
-  "extends": "stylelint-config-standard",
-  "plugins": [
-    "stylelint-order"
-  ],
+  "extends": ["stylelint-config-standard", "stylelint-prettier/recommended"],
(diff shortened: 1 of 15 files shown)


.stylelintrc
```diff
@@ -1,8 +1,6 @@
 {
-  "extends": "stylelint-config-standard",
-  "plugins": [
-    "stylelint-order"
-  ],
+  "extends": ["stylelint-config-standard", "stylelint-prettier/recommended"],
+  "plugins": ["stylelint-order", "stylelint-scss"],
```
bartaz, line 3: You should also disable "at-rule-no-unknown" and enable "scss/at-rule-no-unknown".

steverydz: @bartaz done

PR #7484, opened 2020-05-13
Renewals ux tweaks

sowasred2012:
## Done

- Changed date format in the renewal modal to match what was already in the subscription tables
- Added "(if required)" to VAT field
- Changed "Starts:" to "Will continue from:"

## QA
- Visit /advantage in the demo
- Login
- If you don't have test contracts up for renewal, ask Scott to set you up with one
- Open the renewal modal, see that the points [here](https://github.com/canonical-web-and-design/web-squad/issues/2597#issuecomment-627841020) have been addressed

### DIFF
--- a/static/js/src/renewals/set-modal-info.js
+++ b/static/js/src/renewals/set-modal-info.js
@@ -53,12 +53,15 @@ export function getPaymentInformation(paymentMethod) {
   const billingInfo = paymentMethod.billing_details;
   const cardInfo = paymentMethod.card;
 
+  const formattedExpiryMonth = cardInfo.exp_month.toString().padStart(2, "0");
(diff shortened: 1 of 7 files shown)


matthewpaulthomas:
- Not part of these changes, but I’m surprised I didn’t notice this before: Can we please switch from "" to “” for the subscription name in the dialog title?

- [ ] In Pay mode, the expiry date is now showing as M/YYYY, rather than MM/YY like in the Stripe field. I don’t think this was happening before.

  image

- Not part of these changes, but when I use [Stripe’s “This card requires authentication on all transactions” details](https://stripe.com/docs/testing#regulatory-cards), I don’t get the 3D Secure form — I get “Sorry, there was an unknown error with the payment”.

matthewpaulthomas:
Looks good, thanks

PR #7508, opened 2020-05-15
Make cookie message last 10 seconds

pmahnke:
## Done

- Make cookie message last 10 seconds, it was mistakenly 1 second

## QA

- View the site locally in an *incognito* web browser at: http://0.0.0.0:8001/
- See that the cookie message lasts longer

### DIFF
--- a/static/js/src/core.js
+++ b/static/js/src/core.js
@@ -7,9 +7,7 @@ createNav({ showLogins: false });
 // Initalise the cookie policy notification.
 
 var options = {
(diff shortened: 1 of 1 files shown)


nottrobin:
@pmahnke looks good, but I thought we were going to make it persistent / not enable cookies until they accept? Or do you think we haven't decided what we're doing yet so we need to fix this in the interim?

pmahnke:
@nottrobin There is a meeting on cookies tomorrow I think, it would be good to get this in now.

PR #7512, opened 2020-05-15
Update data model; switch to postgres by default

nottrobin:
This speeds up bulk PUTs to /security/cve.

However, it would still take about 4.5 minutes to add 6500ish records, and 1.5 minutes to update them with minimal changes. For this reason, I'm here limiting the number of records you can insert at once to 300, which should always complete in less than 10 seconds.

I've updated https://github.com/nottrobin/ubuntu-cve-tracker/pull/2 to submit the CVEs in chunks of 300, which should still complete in under 5 minutes.

- Remove `reference` and `bug` tables, in favor of simply making them Arrays on `cve`s and `notice`s
- Change the primary key on tables for faster `get`s
- Make `package`s proper unique objects by extending `package_status` to include the `cve_id`, instead of the `package` table itself
- Update CVE and notice pages accordingly

QA
--

Run DB:

```
docker-compose rm -fs
docker-compose up
```

In another terminal:

```
dotrun --env DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres exec alembic upgrade head
dotrun --env DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres exec python3 scripts/create-releases.py
dotrun --env DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres serve
```

In yet another terminal:

```
git clone -b load-cve git@github.com:carkod/ubuntu-cve-tracker
cd ubuntu-cve-tracker/scripts
python3 -m venv .venv
source .venv/bin/activate
pip3 install requests cvss
time python3 upload-cve.py ../active
```

Authenticate and then see the CVEs import. See the time is hopefully less than 5 minutes depending on how quickly you got through the auth step).

Also test importing USNs:

```
wget https://people.canonical.com/~ubuntu-security/usn/database.json
dotrun exec scripts/create-usns.py database.json
```

Go and browse through USNs and CVEs, check they look okay.


carkod:
I don't think that USNs are affected by these changes in data models? Because the only table we share as of now is the `release` table, the references table used by CVEs is called `cve_reference`

nottrobin:
@carkod, @jpmartinspt:

The tests fail because the migrations I wrote don't currently work with sqlite, but apart from that this is ready to test.

Could you both please look over it and give me any comments.

webapp/security/views.py
```diff
@@ -220,48 +211,26 @@ def create_notice():
...
-                .one()
-            )
-        except NoResultFound:
+        release = db_session.query(Release).get(release_codename)
+
+        if not release:
             message = f"No release with codename: {release_codename}."
             return (flask.jsonify({"message": message}), 400)
 
-    # Link CVEs, creating them if they don't exist
```
jpmartinspt, line 244: We need to keep this block otherwise no cves or references will be created.

webapp/security/views.py
```diff
@@ -1,6 +1,6 @@
 # Standard library
```
jpmartinspt, line 1: On line `278` we can remove.
```
    if "isummary" in data:
        notice.isummary = data["isummary"]
```

webapp/security/views.py
```diff
@@ -1,6 +1,6 @@
...
 import re
```
jpmartinspt, line 2: On line `284`  `notice.references.clear()` might not work anymore since it's not a m2m anymore.

webapp/security/views.py
nottrobin: didn't I? I'm so confused. i thought I did.

jpmartinspt: You removed it on one of the endpoints.

carkod:
QA'ed:
- Tested with chunks of 300 - result: 0.34 seconds on average each chunk of 300
- Tested with ignored
- Removed chunk in the script and tested with directly 299 - result: 0.24 seconds
- Tested with not-for-us (299) CVEs
- Tested with both update and create all items above.

Left to test:
- HTTP requests on microk9s cluster

Minor issues to be addressed:
- Use only CVSS 3
- Default sort order `newest`

@steverydz you might want to take a look at the `cve.html` page as it doesn't have much styling in the packages table

webapp/security/views.py
carkod: But we are already creating CVEs with the API?

scripts/create-usns.py
```diff
@@ -0,0 +1,81 @@
...
+                "release_packages": release_packages,
+                "title": notice["title"],
+                "published": datetime.fromtimestamp(
+                    notice["timestamp"]
+                ).isoformat(),
+                "summary": notice["summary"],
+                "instructions": notice.get("action"),
+            },
+        )
+        print(response, response.text)
```
carkod, line 81: Remove print?

carkod:
When I try to run `python scripts/create-usns.py` it asks me `file_path` argument, what file is it referring to?

webapp/security/schemas.py
```diff
@@ -1,33 +1,97 @@
...
-    approved_by = fields.Str()
-    cvss = fields.Str()  # CVSS 3 and Base Score
-    packages = fields.List(fields.Str(), data_key="package")
-    references = fields.List(fields.Str(), data_key="cve_references")
-    bugs = fields.List(fields.Str(), data_key="cve_bugs")
+    id = Str(required=True)
+    published = ParsedDateTime()
+    description = Str()
+    ubuntu_description = Str()
+    notes = List(Nested(Note))
```
carkod, line 91: If this has been changed into a list, do we not need to update the `webapp/security/models.py` in line 40 instead of Json, this would be a list?

webapp/security/models.py
```diff
@@ -24,64 +26,49 @@
...
-    packages = relationship("Package", secondary=cve_packages)
+    priority = Column(
+        Enum(
+            "unknown",
+            "negligible",
+            "low",
+            "medium",
+            "high",
+            "critical",
+            name="priorities",
```
carkod, line 53: So another missing priority here is `untriaged` which is one more reason why I didn't want to make this an Enum. So the payload will be invalid

webapp/security/models.py
```diff
@@ -135,39 +119,40 @@ def support_tag(self):
...
-    __tablename__ = "package_status"
-
-    id = Column(Integer, primary_key=True)
-    status = Column(String)
-    status_description = Column(String)
-
-    package_id = Column(Integer, ForeignKey("package.id"))
-    package = relationship("Package", back_populates="package_statuses")
+    cve_id = Column(String, ForeignKey("cve.id"), primary_key=True)
+    package_name = Column(String, ForeignKey("package.name"), primary_key=True)
```
carkod, line 126: So this is something that is failing but seems a bit contradictory. Status needs a `package_name` but when I give it status.package_name in the bulk insert data, it says `Unknown field` I suppose it's because it's not define in the schema? It's being returned as an error.message after trying the schema_load

webapp/security/schemas.py
nottrobin: A list can be put in a JSON object perfectly fine.

I was using the ARRAY type, but this is PostgreSQL specific, and so less portable. JSON is more portable as it's supported in SQLite and MySQL too.

webapp/security/models.py
nottrobin: Oh I forgot to push the changes to the `ubuntu-cve-tracker` script.

`untriaged` is an internal priority, only relevant to the security team. So I updated the script to turn these into `unknown` instead.

webapp/security/models.py
nottrobin: I don't understand... this works for me, this is probably another issue with me not having pushed the latest script, which I've now done. Please try again with the changes to the script.

nottrobin:
@carkod 

> When I try to run python scripts/create-usns.py it asks me file_path argument, what file is it referring to?

You need to download the database first:

```
wget https://people.canonical.com/~ubuntu-security/usn/database.json
dotrun exec scripts/create-usns.py database.json
```

scripts/create-usns.py
nottrobin: This is there deliberately isn't it? It's the only way we can see whether we got a successful response from the server.

.gitignore
```diff
@@ -69,3 +69,6 @@ webapp/snapcraft_api/
 .venv/
 
 .coverage
+scripts/regenerate-cve-migrations-postgres.sh
```
jpmartinspt, line 72: I think this can go.

webapp/security/schemas.py
```diff
@@ -1,33 +1,97 @@
...
+    codenames = set(rel.codename for rel in db_session.query(Release).all())
+    unrecognised_codenames = set(release_packages.keys()) - codenames
+
+    if unrecognised_codenames:
+        raise ValidationError(
+            f"Unrecognised release codenames: {unrecognised_codenames}"
+        )
+
+
+class Package(Schema):
```
jpmartinspt, line 43: There is another class named `Package` bellow.

jpmartinspt: If we are going to do this I think we need to know if the package is a binary or a source and we might as well ask for them to send a `source_link` and `version_link` for packages that have them.

webapp/security/views.py
```diff
@@ -41,41 +31,30 @@ def notice(notice_id):
...
-            description = package.get("description")
-            package_name = f"{name} - {description}" if description else name
-            notice_packages.add(package_name)
-
-    # Guarantee release order
-    releases_packages = OrderedDict(
-        sorted(releases_packages.items(), reverse=True)
-    )
+        release_packages[release_version] = pkgs
+        packages.update([pkg["name"] for pkg in pkgs])
```
jpmartinspt, line 45: I was previously appending the description, if we are not doing that here we need to pass the entire package so we can do it in the template instead.

.gitignore
nottrobin: yep

webapp/security/schemas.py
nottrobin: I didn't see anywhere in the code that looked for binary packages - did I miss it?

webapp/security/views.py
nottrobin: :+1:

webapp/security/schemas.py
jpmartinspt: You didn't miss it. I was changin it on the PR we discussed yesterday to fix some package list differences with `usn.ubuntu.com`

jpmartinspt: You can just update the class name, and I can fix the rest in a follow up PR

.gitignore
nottrobin: Moving this [to `.git/info/exclude`](https://help.github.com/en/github/using-git/ignoring-files#excluding-local-files-without-creating-a-gitignore-file).

carkod:
@nottrobin syntax error in the filter. If you visit the /security/cve page:
```
syntax error at or near "ASC"
LINE 5: ...'needs-triage', 'pending')) ORDER BY cve.published DESC ASC 
```

carkod:
QA:

- python3 upload-cve.py ../active  0.26s user 0.10s system 79% cpu 0.448 total
- python3 upload-cve.py ../active  0.41s user 0.11s system 1% cpu 33.735 total (updated 1 CVE with changed status)
- python3 upload-cve.py ../ignored  0.35s user 0.12s system 1% cpu 34.327 total
- python3 upload-cve.py ../ignored/not-for-us.txt  0.43s user 0.12s system 7% cpu 7.512 total

Some pending issues:

Release status shows many empty releases, as discussed the filter will solve this. Once this is solved (syntax error currently) - thumbs up on my part.

Bits to review once issues have been solved:
- Ribbon thing is showing up correctly for "not in ubuntu" CVEs
- QA that package filter is not affected by the filter changes

nottrobin:
@carkod I've fixed up the packages search, and package rows with only "not-affected" or "DNE" statuses should also no longer be shown.

Please re-review and let me know of any issues at your earliest convenience.

PR #7513, opened 2020-05-15
Fix error in order_by so it works in PostgreSQL

jpmartinspt:
## Done
Fix the order by so it does not error when the backend is PostgreSQL

## QA

- Make sure the page renders


carkod:
@jpmartinspt might want to add `docker-compose up -d` to the QA steps?

PR #7548, opened 2020-05-21
Add build step for cypress testing

tbille:
## Done

- Add build step for cypress test
- This will make faster the start of the server handled by the cypress action that was taking too long to start (because of building and installing deps)


steverydz:
@tbille what are the QA steps for this? I tried running `dotrun exec yarn cypress:run` and got an error saying there no Cypress executable

tbille:
@steverydz no steps . It's part of the github action. You can try to run `./run build` and make sure it works.

PR #7554, opened 2020-05-21
Hide renewals when not actionable or current

frankban:
## Done

- Make renewals renewable only when actionable and current.

## QA

- Create non-actionable renewals, or renewals whose start date is in the future, or end date in the past: the button to renew should not be shown.
- It should be still possible to renew renewals that are both actionable and with start <= now <= end.

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -1,0 +1,223 @@
+from datetime import (
+    datetime,
+    timedelta,
+    timezone,
+)
+import unittest
+
+from webapp import views
+
+
+class TestMakeRenewal(unittest.TestCase):
+    def test_no_renewals(self):
+        """None is returned if there are no renewals."""
+        advantage = make_advantage()
+        contract_info = {}
+        renewal = views.make_renewal(advantage, contract_info)
+        self.assertIsNone(renewal)
+
+    def test_processing(self):
+        """Procesisng renewals are re-fetched."""
+        advantage = make_advantage(renewal={"id": "1", "actionable": False})
+        contract_info = {
+            "renewals": [{"id": "1", "status": "processing"}],
+        }
+        got = views.make_renewal(advantage, contract_info)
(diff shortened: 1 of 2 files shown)


webapp/views.py
```diff
@@ -511,6 +487,53 @@ def advantage_view():
...
+    # Renewals not pending or processing are never renewable.
+    if renewal["status"] != "processing":
+        return renewal
+
+    invoices = renewal.get("stripeInvoices")
+    if invoices:
+        invoice = invoices[-1]
+        renewal["renewable"] = (
+            invoice["pi_status"] == "requires_payment_method"
+            and invoice["subscription_status"] == "incomplete_expired"
```
sowasred2012, line 531: As I understand it, `incomplete_expired` on an invoice subscription status means it's unrenewable (or at least requires intervention from someone on the sales team). 

This check is about presenting the "Renew..." button to the user if they've attempted to pay, failed, left the page and come back later to make another attempt - in that case, the `subscription_status` will be `incomplete`, and is renewable.

```suggestion
            and invoice["subscription_status"] != "incomplete_expired"
```

tests/test_views.py
```diff
@@ -0,0 +1,223 @@
...
+        want = {
+            "id": "1",
+            "status": "processing",
+            "actionable": True,
+            "start": start,
+            "end": end,
+            "stripeInvoices": [
+                {
+                    "pi_status": "requires_payment_method",
+                    "subscription_status": "incomplete_expired",
```
sowasred2012, line 188: A follow on from my previous comment - attempting to renew when the invoice has `incomplete_expired` status has always failed for me when testing against the contracts service:

```suggestion
                    "subscription_status": "incomplete",
```

frankban: Done.

anthonydillon:
@sowasred2012 you probably have to merge this.

PR #7565, opened 2020-05-22
Remove underscores from exposed files

nottrobin:
The template files for nav and contact-us that we need to expose to the
client for javascript to consume can no longer start with underscores
(because of templatefinder v1.0.0).

So change them to not have underscores.

QA
--

```
dotrun
```

Test out "talk to us" or "ask us" links on pages.

Test out the mega nav.


pmahnke:
It all seems fine and I grepped around a bit and couldn't find anything.

PR #7620, opened 2020-06-02
renewals: add terms checkbox before user can complete payment

### LINKED ISSUE #7612: UA renewal terms and conditions
According to Uros, each renewal of a UA subscription is currently treated as a separate purchase. Therefore, you need to agree to [the UA service terms and conditions](https://ubuntu.com/legal/ubuntu-advantage-service-terms) each time. So, this needs adding to the renewal dialog.

sowasred2012:
## Done

- added a t&c checkbox the user needs to check before payment can be processed
- set equal widths on buttons in the modal footer

## QA

- Login with SSO
- If you don't have any subscriptions, ask Scott to set you up with one
- On the subscription, click "Renew...", fill out the payment method details and click "Continue"
- See that the "Pay" button is disabled
- Click the t&c checkbox

## Issue / Card

Fixes #7612

### DIFF
--- a/static/js/src/renewal-modal.js
+++ b/static/js/src/renewal-modal.js
@@ -19,6 +19,7 @@ const form = document.getElementById("details-form");
 const errorDialog = document.getElementById("payment-error-dialog");
 const progressIndicator = document.getElementById("js-progress-indicator");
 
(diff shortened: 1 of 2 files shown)


anthonydillon:
@matthewpaulthomas if it helps:

image

image

PR #7634, opened 2020-06-03
Show contact numbers if renewal is not actionable

### LINKED ISSUE #7605: UA renewals: We send people to /advantage for subscriptions not renewable there
From Francesco:
> Marketing will send an email for all expiring contracts (tech touch) pointing to the web page. In there, [n]ot all renewals will be actionable. Non-actionable renewals require manually contacting sales. Right now, non-actionable renewals are completely hidden. We should instead suggest how to renew.

sowasred2012:
## Done

- Added a list of contact numbers for cases where a renewal is present but not actionable

## QA

- Sign in with SSO
- Ask Scott to set you up with test data
- Find the subscription, expand the row and see that you are presented with a notification and a list of contact numbers

## Issue / Card

Fixes #7605

## Screenshots
Screenshot from 2020-06-03 12-50-48

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -353,12 +353,10 @@ summary {
 // Advantage section
 .p-table--advantage-contract {
   margin: 0 auto;
-  table-layout: fixed;
   width: auto;
 
   td {
     display: table-cell;
-    width: 50%;
   }
 
   tr {
@@ -405,6 +403,17 @@ summary {
   @extend %vf-has-round-corners;
 
   margin-bottom: 0;
+}
+
+.p-table--advantage-contact {
+  @media only screen and (min-width: $breakpoint-large) {
+    width: 66%;
+  }
+
+  th,
+  td {
+    flex: 1 !important;
+  }
 }
 
 .p-table--advantage {
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -18,8 +18,8 @@
   font-family: "UbtuOverride";
   font-style: normal;
   font-weight: 300;
-  src:
-    url("https://assets.ubuntu.com/v1/e8c07df6-Ubuntu-L_W.woff2") format("woff2"),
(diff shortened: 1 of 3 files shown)


sowasred2012:
@meralonso @cassiocassio - would you mind taking a look at this?

templates/advantage/index.html
```diff
@@ -212,6 +212,57 @@ <h2>Your paid subscriptions</h2>
                                       <span>Renewal in progress...</span>
                                     </td>
                                   </tr>
+                                {% elif contract["renewal"]["actionable"] == false %}
+                                  </table>
+                                  
+                                  <table class="u-no-margin--bottom">
```
anthonydillon, line 218: Why is this notification in a table?

static/sass/styles.scss
```diff
@@ -405,6 +405,17 @@ summary {
   margin-bottom: 0;
 }
 
+.p-table--advantage-contact {
+  @media only screen and (min-width: $breakpoint-large) {
+    width: 66%;
```
anthonydillon, line 410: Can you use row and gird?

static/sass/styles.scss
```diff
@@ -405,6 +405,17 @@ summary {
...
+  }
+
+  th,
+  td {
+    flex: 1 !important;
```
anthonydillon, line 415: Are tables flex?

anthonydillon:
Also it would be good to follow up with a test for this

templates/advantage/index.html
sowasred2012: I think because I was tired and surrounded by other tables... I'll update this.

static/sass/styles.scss
sowasred2012: Good point!

static/sass/styles.scss
sowasred2012: `p-table--expanding` are flex, and `td`s under the `p-table--advantage ` class have been given various flex values - the parent table uses both of these classes.

PR #7639, opened 2020-06-03
Use Talisker session for docs calls

nottrobin:
Instead of the CachedSession from canonicalwebteam.http.

This is because we're getting timeout errors in Sentry,
because the CachedSession has a timeout of (.5, 3) by default.

We should no longer explicitly time out API calls, as
the combination of the content-cache in front of our sites
and the fact that we use gsync gevent workers in production
should mean that there's no problem with request waiting for
the API calls to complete.

Also, using a Talisker session allows us to get more information
about these API requests in Graylog.

This should fix Sentry errors like this one: https://sentry.is.canonical.com/canonical/ubuntu-com/issues/9582/

## QA

In the demo, or running locally, go to `/tutorials` and `/server/docs` and check they both still work fine.

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -80,7 +80,11 @@ app.config["CANONICAL_LOGIN_URL"] = os.getenv(
     "CANONICAL_LOGIN_URL", "https://login.ubuntu.com"
 ).rstrip("/")
 
+session = talisker.requests.get_session()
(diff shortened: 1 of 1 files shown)


anthonydillon:
@nottrobin just a lint failure

PR #7651, opened 2020-06-04
Update renewed tag logic

sowasred2012:
## Done

- Changed the logic around how we present a "renewed" label when a renewal has completed
  - previously it was based on a query param given by JS, which you wouldn't get if the renewal took > 30 seconds to complete.

**Note:** I've updated contract API and Stripe key variables to use staging values so this can be QAd by anybody - I'll drop that commit before merging.

## QA

- Ask Scott to set you up with a test contract and renewal
- Complete a test purchase using 4242 4242 4242 4242 as the card number, any future expiry date and any zip code
- When the page reloads, see that the "Renewed" tag is there in the expanded row, and that "renewed=true" isn't present in the url
- If you have older contracts with renewals completed in previous tests, expand those rows and see that the "renewed" tag is not there

### DIFF
--- a/.env
+++ b/.env
@@ -3,7 +3,7 @@ FLASK_DEBUG=true
 DEVEL=true
 DATABASE_URL=sqlite:///usn.sqlite3
 SECRET_KEY=insecure_dev_key
-CONTRACTS_API_URL=https://contracts.canonical.com/
+CONTRACTS_API_URL=https://contracts.staging.canonical.com/
 
 # This is a test api key https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
 CAPTCHA_TESTING_API_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -5,10 +5,7 @@ import json
 import math
 import os
 import re
-from datetime import (
-    datetime,
-    timezone,
-)
+from datetime import datetime, timezone, timedelta
 
 # Packages
 import dateutil.parser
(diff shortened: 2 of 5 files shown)


.env
```diff
@@ -3,7 +3,7 @@ FLASK_DEBUG=true
 DEVEL=true
 DATABASE_URL=sqlite:///usn.sqlite3
 SECRET_KEY=insecure_dev_key
-CONTRACTS_API_URL=https://contracts.canonical.com/
+CONTRACTS_API_URL=https://contracts.staging.canonical.com/
```
frankban, line 6: Did you intend to change this?

webapp/views.py
```diff
@@ -5,10 +5,7 @@
 import math
 import os
 import re
-from datetime import (
-    datetime,
-    timezone,
-)
+from datetime import datetime, timezone, timedelta
```
frankban, line 8: Alphabetical order?

webapp/views.py
```diff
@@ -506,6 +503,13 @@ def make_renewal(advantage, contract_info):
 
     renewal["renewable"] = False
 
+    if renewal["status"] == "done":
```
frankban, line 506: Do we need a test for this?

webapp/views.py
```diff
@@ -506,6 +503,13 @@ def make_renewal(advantage, contract_info):
...
+        renewalModifiedDate = dateutil.parser.parse(renewal["lastModified"])
```
frankban, line 507: Why using mixed case in here, change them to last_modified and one_hour_ago to be PEP8 compliant?

webapp/views.py
```diff
@@ -506,6 +503,13 @@ def make_renewal(advantage, contract_info):
...
+        oneHourAgo = datetime.now(timezone.utc) - timedelta(hours=1)
+
+        if oneHourAgo < renewalModifiedDate:
```
frankban, line 510: Perhaps better: `renewal["recently_renewed"] = oneHourAgo < renewalModifiedDate`, so that the key is present in the false case as well.

webapp/views.py
alnvdl, line 507: I'd recommend catching errors here (wrapping this in a `try`) and handling it somehow (i.e. not marking it as recently renewd).
We try to keep our APIs sane in ua-contracts, but you never know when things can go wrong :)

.env
sowasred2012: Yes, this is just for anyone on the webteam who QAs it - I'm going to revert this change before merging.

webapp/views.py
sowasred2012: Good point!

sowasred2012:
@frankban @alnvdl - updated per your comments, could you take another look?

PR #7675, opened 2020-06-07
Remove reference to the ubuntu-support-status command

### LINKED ISSUE #7663: ubuntu-support-status command is no longer available on 20.04 LTS
"Maintenance and security updates" section in https://ubuntu.com/about/release-cycle has `ubuntu-support-status` command to check the support status of the system. However, the command has been removed from 20.04 LTS so it's no longer available.
https://bugs.launchpad.net/ubuntu/+source/update-manager/+bug/1873362

There is no direct alternative so I'm personally not sure what to put instead.

Screenshot from 2020-06-05 23-03-43

---
*Reported from: https://ubuntu.com/about/release-cycle*

fitojb:
## Done

- Remove reference to the `ubuntu-support-status` command. This has been removed from Ubuntu 20.04 LTS, per [LP: #1873362](https://pad.lv/1873362)

## Issue / Card

Fixes #7663

### DIFF
--- a/templates/about/release-cycle.html
+++ b/templates/about/release-cycle.html
@@ -235,11 +235,6 @@
     <div class="row">
       <div class="col-8">
         <p>Ubuntu LTS releases transition into Extended Security Maintenance (ESM) phase as the standard, five-year public support window comes to a close. It is recommended for users and organisations to upgrade to the latest LTS release or <a href="/esm">subscribe to ESM</a> for continued security coverage.</p>
-        <p>This command will print the exact status of your system:</p>
-        <div class="p-code-copyable" style="margin-top: .5rem;">
-          <input class="p-code-copyable__input" value="ubuntu-support-status" readonly="readonly">
-          <button class="p-code-copyable__action">Copy to clipboard</button>
-        </div>
         <p>The <a class="p-link--external" href="https://wiki.ubuntu.com/Releases">Ubuntu Releases wiki</a> has current information on previous and upcoming versions.</p>
       </div>
     </div>


templates/about/release-cycle.html
```diff
@@ -235,11 +235,6 @@ <h2>Maintenance and security updates</h2>
     <div class="row">
       <div class="col-8">
         <p>Ubuntu LTS releases transition into Extended Security Maintenance (ESM) phase as the standard, five-year public support window comes to a close. It is recommended for users and organisations to upgrade to the latest LTS release or <a href="/esm">subscribe to ESM</a> for continued security coverage.</p>
-        <p>This command will print the exact status of your system:</p>
-        <div class="p-code-copyable" style="margin-top: .5rem;">
-          <input class="p-code-copyable__input" value="ubuntu-support-status" readonly="readonly">
```
nobuto-m, line 240: Looks like `ubuntu-security-status` is a successor of `ubuntu-support-status` at least on Ubuntu 20.04 LTS as of today. Instead of deleting the section, we could have conditions to use `ubuntu-security-status` > 20.04, then the previous command as `ubuntu-support-status` for older releases.

fitojb: I don’t know how to implement conditions, and I don’t think doing it with UA sniffing is a good idea, privacy-wise. Maybe tabs?

pmahnke:
@nobuto-m you were correct about the other command.

@fitojb thank you for helping.  I got confirmation from the head of the server team and put in a fix in PR #7746 so I am closing this.

PR #7738, opened 2020-06-11
Speed up CVE schema checking

nottrobin:


carkod:
Looks like we need to add a postgres service to the github test-python?

carkod:
Ok... so this new error in test-python is because we are not running using .entrypoint, so it is not executing the releases script... So there are a few options, should we change it to run .entrypoint? or should we simply add the releases script?

PR #7836, opened 2020-06-26
Fix the endpoint to update notices

jpmartinspt:
## Done
Adds `release_codenames` to `NoticeSchema` context on the update endpoint.

## QA
- `docker-compose up -d`
- `jq '."4147-1" | { (.id): .}' database.json > 4147-1.json`
- `dotrun exec ./scripts/create-usns.py 4147-1.json` should return a `200`
- Browse http://localhost:8001/securtity/notices/USN-4147-1
- Change the `title` value in `4147-1.json`
- `dotrun exec ./scripts/create-usns.py --update 4147-1.json` should return a `200`
- Browse http://localhost:8001/security/notices/USN-4147-1 and see the updated title.

### DIFF
--- a/scripts/create-usns.py
+++ b/scripts/create-usns.py
@@ -123,7 +123,7 @@ with open(args.file_path) as usn_json:
         # Build endpoint
         endpoint = notice_endpoint
         if http_method == "PUT":
-            endpoint = f"{notice_endpoint}/{notice['id']}"
+            endpoint = f"{notice_endpoint}/USN-{notice['id']}"
 
         response = client.request(
             http_method,
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -253,25 +253,27 @@ def update_notice(notice_id):
     """
     PUT method to update a single notice
     """
+    notice = db_session.query(Notice).get(notice_id)
+
+    if not notice:
+        return (
+            flask.jsonify({"message": f"Notice {notice_id} doesn't exist"}),
+            404,
+        )
+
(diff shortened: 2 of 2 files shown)


webapp/security/views.py
```diff
@@ -253,25 +253,29 @@ def update_notice(notice_id):
...
             flask.jsonify(
-                {"message": "Invalid payload", "errors": error.messages}
+                {"message": f"Notice {notice_id} doesn't exist"}
             ),
-            400,
+            404,
         )
 
-    notice = db_session.query(Notice).get(notice_id)
+    notice_schema = NoticeSchema()
```
nottrobin, line 266: This line seems redundant

scripts/create-usns.py
```diff
@@ -123,7 +123,7 @@ def guess_binary_links(binary, version, sources):
         # Build endpoint
         endpoint = notice_endpoint
         if http_method == "PUT":
-            endpoint = f"{notice_endpoint}/{notice['id']}"
+            endpoint = f"{notice_endpoint}/USN-{notice['id']}"
```
nottrobin, line 126: I thought the `USN` would be part of the ID. But we can discuss this down the line, no need to block this PR.

jpmartinspt: It is, but this script is used to import data from the `database.json` file, and all the ID's in there are not prefixed.

nottrobin: Oooooooh, I see. Great :+1:

PR #7844, opened 2020-06-29
Advantage shop checkout markup

### LINKED ISSUE #7573: UA purchase: Build checkout
@matthewpaulthomas commented on [Tue May 26 2020](https://github.com/canonical-web-and-design/master-epics/issues/155)

sowasred2012:
## Done

- Adapted the renewal modal to accommodate multiple products and display information about each during checkout.

**Note:** some of this code is purely for demo purposes and will be removed when we have real data coming in from the contracts API.

## QA

- Select your products and their quantities
- Click "Buy"
- See that the modal matches the [brief](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.mz73r7dju49w)
- Add payment method details and click "Continue", see that the second page looks correct
- Click "Cancel" (the "Pay" button does not work for this demo)
- Select a different combination of products, click "Buy", see that the summary has been updated.

## Issue / Card

Fixes #7573


package.json
```diff
@@ -74,6 +74,7 @@
     "@canonical/cookie-policy": "2.1.0",
     "@canonical/global-nav": "2.4.1",
     "@canonical/latest-news": "1.0.3",
+    "moment": "^2.27.0",
```
steverydz, line 77: Can you pin the dependency? Although I think https://date-fns.org/ would be better for this as it's more modular. It's got a similar API to moment but you can just import the functions you need rather than the whole library.

PR #7923, opened 2020-07-13
Advantage shop product selection

sowasred2012:
## Done

- Added product selection markup and "waterfall" form logic to /advantage/subscribe

**Note:** clicking "Buy" won't trigger the payment modal, that will be covered in another PR that implements the shop payment endpoints

## QA

- Use the form to make a product selection, see that at the end of it you are able to add a product to the cart
- Add the product to the cart
- Change quantities on the added product, see that the price and totals updates accordingly
- See that the header has been updated to "Any more to add?"
- Run through the process again, this time making different selections
  - Clicking AWS or Azure should disable the rest of the form
  - Clicking "Other" in the Ubuntu version should disable the rest of the form
  - Adding a second, different product should add a new line item to the cart
  - Adding multiples of the same product should update the existing line item with the combined quantity of both selections
- Remove products from the cart, see that totals are updated
- Remove all products from the cart, see that the cart disappears and the page header "Subscribe to Ubuntu Advantage" reappears, and the heading on the first section reads "What are you setting up?"


jpmartinspt:
@sowasred2012 re-created the demo in the new demoservice

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,165 @@
...
+    steps.forEach((step) => {
+      const wrapper = form.querySelector(`${stepClassPrefix}${step}`);
+
+      wrapper.classList.remove("u-disable");
+    });
+  }
+
+  function handleStepSpecificAction(inputElement) {
+    const step = inputElement.name;
+    const validQuantity = step === "quantity" && inputElement.value > 0;
```
anthonydillon, line 57: You dont need the `step === "quantity"` check here as your filtering it in the switch case below

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,165 @@
...
+    const quantityTypeEl = document.querySelector(".js-type-name");
+
+    switch (step) {
+      case "type":
+        quantityTypeEl.innerHTML = `${inputElement.dataset.productName}s`;
+        updateProductState(inputElement);
+        break;
+      case "quantity":
+        if (validQuantity) {
```
anthonydillon, line 66: As we can reduce the conditional above we can just move the logic here, so:
```
if (inputElement.value > 0) {
```

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,165 @@
...
+      const wrapper = form.querySelector(`${stepClassPrefix}${step}`);
+
+      wrapper.classList.remove("u-disable");
+    });
+  }
+
+  function handleStepSpecificAction(inputElement) {
+    const step = inputElement.name;
+    const validQuantity = step === "quantity" && inputElement.value > 0;
+    const quantityTypeEl = document.querySelector(".js-type-name");
```
anthonydillon, line 58: Can you cache this query in the init function? At this point, it is querying the DOM on every input event.

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,165 @@
...
+      cartStep.classList.remove("u-hide");
+    } else {
+      cartStep.classList.add("u-hide");
+    }
+  }
+
+  function updateProductState(inputElement) {
+    productState.set(inputElement.name, [inputElement.value]);
+
+    updateCartLineItem();
```
anthonydillon, line 161: This should be run by the render function. Which, is called when state is mutated.

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,165 @@
...
+      } else {
+        tab.setAttribute("aria-selected", false);
+        target.classList.add("u-hide");
+      }
+    });
+  }
+
+  function updateCartLineItem() {
+    const cartName = form.querySelector(".js-shop-product-id");
+    const cartStep = form.querySelector(`${stepClassPrefix}add`);
```
anthonydillon, line 142: Can these queries be cached above so you are not performing expensive DOM lookups on each render?

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,165 @@
...
+    const quantity = productState.get("quantity")[0];
```
anthonydillon, line 143: It's a shame about the `[0]` to access the value. I did not have time to find a more elegant method.

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,165 @@
...
+    const support = productState.get("support")[0];
+    const type = productState.get("type")[0];
+
+    const productString = `uai-${support}-${type} x ${quantity}`;
+
+    cartName.innerHTML = `Your selected product id is ${productString}`;
+
+    if (type && quantity && support) {
```
anthonydillon, line 151: Do you want to print the product id before checking these values exist?

static/js/src/utils/tests/state.test.js
```diff
@@ -0,0 +1,109 @@
...
+    const items = ["foo", "bar"];
+    const key = "four";
+
+    it("should remove a given item from a given key's array of values", () => {
+      state.set(key, items);
+      state.remove(key, "foo");
+      expect(state.get(key)).toEqual(["bar"]);
+    });
+
+    it("should return an error if no value is passed", () => {
```
anthonydillon, line 88: Can you add a test to check the callback is called on each action that requires it. It's kind of the main thing the stateManager does.

templates/advantage/subscribe.html
```diff
@@ -6,29 +6,132 @@
...
+      <div class="row">
+        <div class="col-12">
+          <!-- <a
             href="#ua-payment-modal"
             id="showModal"
             class="js-ua-payment-cta p-button--positive u-no-margin--bottom"
             data-transaction-type="purchase"
-          >Buy</a>
-        </form>
+          >Add</a> -->
```
anthonydillon, line 128: Can this be removed or easier for now?

wgx:
👍  LGTM

meralonso:
Thanks @sowasred2012 this is looking nice. The radio buttons are not centered when selected - could you please fix that?
Screenshot 2020-07-21 at 14 57 17

matthewpaulthomas:
I’m really sorry I dropped the ball on reviewing this after returning from leave. It looks like an accurate implementation of the mockup. However, I predict it would get Review: CEO–1 for one or more of these reasons:

- Radio buttons are (a) awkward for a set of options where the default is intentionally none, (b) highly unusual for a set of options that wraps to multiple lines (without being an actual 2-D grid), and (c) least importantly, inconsistent with the other places we have this interaction, ubuntu.com/core/build and snapcraft.io/first-snap. (And it’s not as if those other pages could be changed to match — 13 radio buttons on /first-snap would look daft.) Any one of these problems by itself might be tolerable, but the combination is a bit much.

- “Physical”, “Physical”, “Public cloud”, “Public cloud”. Martin’s heading layout avoided this repetition.

- The options you don’t buy on this page (AWS and Azure) are more colourful than the ones you do. Either all the icons should be colourful, or none of them should be.

- We should be proudest of the latest LTS, so the Ubuntu versions should be in descending order.

@meralonso has kindly made a mockup that addresses three of these issues. (Disregard the “Which services do you want?” section — that’s for [web-squad#2981](https://github.com/canonical-web-and-design/web-squad/issues/2981).)

build-v2

Separate from that:

- The contents of the “What version of Ubuntu…” tab is inaccurate (probably not implemented yet). Refer to [the coverage spreadsheet](https://docs.google.com/spreadsheets/d/1krXNNq1BkZCDSJeTARteLxZpylCep6jG6KmUBMA8RwA/edit?usp=sharing).
- “No, thanks” shouldn’t have a comma.
- [I’ve tweaked the wording](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.vlvlt0gikzjy) to accommodate the future addition of UA Apps. (For example, “What version of Ubuntu are they running?” is now “What Ubuntu version are you running?” because that section no longer immediately follows the quantity field.) This future addition will also make it every more important to keep the page really compact — for example by using plain text labels rather than headings.

wgx:
Should "Infra" be a user-facing term, or should it instead be "Infrastructure"  @matthewpaulthomas

lechsandecki:
Hey team. I like it! A few comments:
1. The desktop option is almost a border case for us. It shouldn't go first and it also shouldn't be the default one.
2. If we ask which version of Ubuntu they are running before we give UA Infra/ UA Apps choice, then we will be able to list all features of UA Infra/ Apps precisely for a selected Ubuntu version.
3. Do we say Ubuntu version or Ubuntu release?
4. we can say AWS and Azure to avoid 2 lines of text.
5. I think that we have 4 distinct cases here: Physical, VM, Public Cloud and Desktop. Virtual Machine does not equal to Cloud.

please let me know if there is anything else that I should review

cheers
Lech

matthewpaulthomas:
> Should "Infra" be a user-facing term, or should it instead be "Infrastructure"

@wgx “Ubuntu Advantage for Infrastructure” is unfortunately long enough that people will inevitably abbreviate it, as demonstrated by [Canonical staff already abbreviating it several ways](https://ubuntu.com/search?q=%22ua+for+infrastructure%22+OR+%22UA+Infrastructure%22+OR+%22UA+Infra%22+OR+%22UA-I%22). And sometimes abbreviating is vital, for example [in CVE status cells](https://github.com/canonical-web-and-design/web-squad/issues/2982#issuecomment-661691993). So to minimise brand dilution, in April I proposed [standard short names](https://docs.google.com/document/d/1AX-kSNztuAmShEoohe8L3LNLRnSKF7I0qkZGNeoGOok/edit#heading=h.ervb8eo2tyy), which I’m using here.

> 1. The desktop option is almost a border case for us. It shouldn't go first

@lechsandecki Ok, we’ve moved it to the end.

> 2. … UA Infra/ UA Apps choice …

Let’s discuss that in [web-squad#2981](https://github.com/canonical-web-and-design/web-squad/issues/2981) since it isn’t part of this.

> 3. Do we say Ubuntu version or Ubuntu release?

Tiny details have tiny rationales: “Other” here includes not just non-LTS releases, but also the development version (currently Groovy), which is not yet a release. For example, you could/should not have bought UA for Ubuntu Focal until it became the 20.04 LTS release. (I made [the same distinction for CVE search](https://docs.google.com/document/d/1VuAElUpFay7MBFNmE3NhSyY9TkXiqHcA1ZHQ11AYO-Q/edit#bookmark=id.jzmf6qyiy6f3).)

> 4. we can say AWS and Azure to avoid 2 lines of text.

That would slimplify the choice a bit. But I’d rather keep them separate, because (a) combining them would mean using neither logo, harder to scan, and (b) the links will go to different places (their respective Marketplaces). We can easily avoid two lines without combining them.

> 5. I think that we have 4 distinct cases here: Physical, VM, Public Cloud and Desktop. Virtual Machine does not equal to Cloud.

Fair enough. That, in turn, means that the category headings are no longer useful, and that “Other VMs” should not have a cloud icon.

Other tweaks:
- As discussed today, the first question no longer uses the word “support”: it’s now “What kind of machines are you setting up?” or “What kind of machines are you adding?”, depending on first/subsequent subscription.
- “See the details of our response times” becomes “Details of our coverage and response times”.

Wireframe | Mockup
-|-
builder | …

meralonso:
@matthewpaulthomas I updated the design and put it on [zeplin](https://app.zeplin.io/project/5f19ae878554826b64d30775), I kept the [old version](https://app.zeplin.io/project/5efa107ccb7ffa15fe136cab) just in case.

matthewpaulthomas:
@wgx Are you happy with the latest changes?

matthewpaulthomas:
Looking really good! Just a few glitches remaining:

- [ ] When you are signed in, the sign-in blurb disappears correctly, but the “Sign out” button shouldn’t be there.
- [ ] The Usabilla popup appears after 20 seconds, which is (a) far too quick (I thought we’d increased this to 60?) and (b) not appropriate for this page since it’s a transitory task. Can we suppress it altogether on this page?
- [ ] Until you answer the tech-support question, the “Your chosen plan” is not present, which obscures how long the form is going to take. It should be [present but greyed out, with “…” plan name and blank price](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#bookmark=id.yplpcljyvivg).
- [ ] The contents of the “What version of Ubuntu…” tab isn’t implemented yet.
- [ ] “Get details on our tech support response times” above the buttons should be [“Details of our coverage and response times” below the buttons](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#bookmark=id.h3gfubad2pr7). Please also make it open in a new window/tab, so that you don’t lose all the choices you’ve just made (paging @cassiocassio).
- [ ] “No thanks” really really shouldn’t have a comma, srs.
- [ ] Once you’ve added a subscription, [“What are you setting up?” should be “Any more to add?”](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#bookmark=id.ruwuzr3po8xx)
- [ ] Once you’ve added a subscription, the “What are you setting up?”/“Any more to add?” section is closer to the grey section above than it is to the rest of the form below. It should be at least as far away.
  image

meralonso:
@sowasred2012 this is looking great, thank you - just some pixel perfect tweaks to match my design

Could you make the 3 logos the original svg size, please?
[Server](https://assets.ubuntu.com/v1/fdf83d49-Server.svg)
[VMs](https://assets.ubuntu.com/v1/9ed50294-Virtual+machine.svg)
[Desktop](https://assets.ubuntu.com/v1/4b732966-Laptop.svg)

Screenshot 2020-08-05 at 10 55 07

For consistency, could you give 24px between heading (questions) and block underneath on every section, please?
Screenshot 2020-08-05 at 11 15 31

Could you give 16px from text top and bottom on these buttons? Unless Matthew wants 24px, which I'm fine with but they just seem a bit too big
Screenshot 2020-08-05 at 11 16 03

Could the icon be 32px high on this section? I can export these for you - let me know
Screenshot 2020-08-05 at 11 24 03

wgx:
- [ ]  I proposed a text change to:

> We recommend buying for an hourly rate on the AWS Marketplace:

Which is in the google doc spec.

meralonso:
@matthewpaulthomas Checkout mobile design below
UA - Build your subscription - 4
UA - Build your subscription - Checkout

matthewpaulthomas:
Very close now…
- [ ] Now that the AWS and Azure sections have only one option each, and it doesn’t have a bullet (good), putting it on a separate line looks like it’s being squirted into a template. Instead, can we say: “You can buy [Ubuntu Pro on the AWS Marketplace](https://aws.amazon.com/marketplace/search/results?page=1&filters=VendorId&VendorId=e6a5002c-6dd0-4d1e-8196-0a1d1857229b&searchTerms=ubuntu+pro) at an hourly, per-machine rate, with all UA software features included.”. And equivalent for Azure.
- [ ] When I choose the “Other” tab, the remainder of the form is correctly greyed out. But if I choose a numbered version, then make a tech-support choice, _then_ go back and choose “Other”, the “Your chosen plan” selection is not greyed out and should be.
- [ ] I’d prefer if the whole of the “Your chosen plan” section, rather than just the heading, was present (and greyed out) until you complete your selections, so that you can see how the task will end. (I guess you need to implement this anyway, to address the previous point.)

In the version tabs:
- [ ] “All UA plans include:” > “For Ubuntu {version}, all UA plans include:”
- [ ] “Extended Security Maintenance (ESM) LTS” > “Extended Security Maintenance (ESM)”
- [ ] ESM should be listed only for 14.04 ESM, since it isn’t applicable for any other versions.
- [ ] “CIS” > “CIS benchmark” (as on [/pricing/infra](https://ubuntu.com/pricing/infra))
- [ ] “STIG” > “DISA STIG” (just for a little more explanation)
- [ ] “Landscape” > “Landscape on-prem systems management” (as on [/pricing/infra](https://ubuntu.com/pricing/infra))
- [ ] “Support for Infra (24x5, 24x7)” _must not_ be present (it isn’t included in all plans!)
- [ ] The “Other” > “You can:” bulleted list should not be ticked — it’s not a feature checklist.

meralonso:
@matthewpaulthomas @sowasred2012 Can this text go on the same line?

Screenshot 2020-08-06 at 11 31 47

sowasred2012:
> It concerns me slightly that the list of features per version is raw, mostly-repeated HTML elements. I don’t know whether that contributed to this being left unfixed…
> 
> > “Support for Infra (24x5, 24x7)” must not be present (it isn’t included in all plans!)
> 
> …but it will invite errors in future when adding, removing, changing supported versions, or changing the description of any UA features. I’ll leave it to code review whether that should be tackled at all as part of this.

That being unfixed was a case of mistaken, over-eager box checking on my part, I've updated with the suggestions.

I think the repeated features are fine - we could reduce repetition and introduce more JS to toggle the visibility of each per version based on, for example, data attributes, but I think that would be unnecessarily complicating things. As it is, you only need to glance at the code to see which features a particular version has.

anthonydillon:
## QA
- You should limit quantities in the cart to [0-9] and not minus.
image

- If you use [Vanilla's list `is-split`](https://vanillaframework.io/docs/patterns/lists#split) instead of two separate lists in the version features list it looks better on small screens:

| Before | After |
| --- | --- |
| image | image |

sowasred2012:
@anthonydillon good spot - I've updated with a change that prevents that value being less than zero, and if the subtotal of the cart is zero, the "Buy" button is disabled.

meralonso:
> @matthewpaulthomas @sowasred2012 Can this text go on the same line?
> 
> Screenshot 2020-08-06 at 11 31 47

Same for this text - Can 'ESM. You can:' go up on the previous line?
Screenshot 2020-08-06 at 14 32 42

anthonydillon:
@sowasred2012 nice work! :+1:

meralonso:
> > @matthewpaulthomas @sowasred2012 Can this text go on the same line?
> > Screenshot 2020-08-06 at 11 31 47
> 
> Same for this text - Can 'ESM. You can:' go up on the previous line?
> Screenshot 2020-08-06 at 14 32 42

Sorry @sowasred2012 my bad - could it look like this?
Screenshot 2020-08-06 at 14 32 42

wgx:
👍

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,492 @@
...
+    </div>
+    `;
+  }
+
+  function buildLineItemHTML(productId, quantity, imageURL, action) {
+    let product;
+    let rawTotal;
+    let cost = "$0";
+    let productString = "&mldr;";
+    let quantityHTML = `x${quantity}`;
```
matthewpaulthomas, line 110:
```suggestion
    let quantityHTML = `× ${quantity}`;
```

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,492 @@
...
+        }">Buy now</button>
+      </div>
+    </div>
+    `;
+  }
+
+  function buildLineItemHTML(productId, quantity, imageURL, action) {
+    let product;
+    let rawTotal;
+    let cost = "$0";
```
matthewpaulthomas, line 108:
```suggestion
    let cost = "";
```

static/js/src/ua-product-selector.js
```diff
@@ -0,0 +1,492 @@
...
+        lineItem.get("productId")[0],
+        lineItem.get("quantity")[0],
+        lineItem.get("imageURL")[0],
+        "remove"
+      )}</li>`;
+    });
+
+    return `<div class="row">
+      <div class="col-12">
+        <h2>Your subscription so far:</h2>
```
matthewpaulthomas, line 74: Labels end in colons, headings don’t.
```suggestion
        <h2>Your subscription so far</h2>
```

PR #7940, opened 2020-07-16
Replace Wordpress class with BlogAPI class

### LINKED ISSUE #2995: update copy download/server/provisioning/ - build
copy doc: https://docs.google.com/document/d/12GXYcP9WrOgoso6Sf3bQql6nALRUzpdlcRW3Gbu-npM/edit

albertkol:
Replaces `admin.insights.ubuntu.com/wp-content/uploads` with `ubuntu.com/wp-content/uploads`. 
Additionally it enables the cloudinary service for our images again.

## QA

Check out the image url, it will be converted to a cloudinary url, and instead of `admin.insights.ubuntu.com/wp-content` the img url will show `ubuntu.com/wp-content`.

Browse to the blog to check the thumbnails:
https://ubuntu-com-7940.demos.haus/blog

Browse to an article:
https://ubuntu-com-7940.demos.haus/blog/dell-xps-13-developer-edition-with-ubuntu-20-04-lts-pre-installed-is-now-available

Browse to another article:
https://ubuntu-com-7940.demos.haus/blog/design-and-web-team-summary-20th-july-2020

Everything should look exactly like on prod.

## Issue / Card

Fixes #2995

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,10 @@
 alembic==1.4.2
 canonicalwebteam.flask_base==0.6.3
 canonicalwebteam.http==1.0.3
-canonicalwebteam.blog==6.0.0
+canonicalwebteam.blog==6.2.1
 canonicalwebteam.search==0.2.1
 canonicalwebteam.templatefinder==1.0.0
-canonicalwebteam.image-template==1.1.0
+canonicalwebteam.image-template==1.2.0
 canonicalwebteam.discourse-docs==1.0.0
 canonicalwebteam.launchpad==0.7.6
 python-dateutil==2.8.1
--- a/templates/blog/blog-card.html
+++ b/templates/blog/blog-card.html
@@ -5,7 +5,7 @@
     {% if article.image and article.image.source_url %}
     <div class="u-crop--16-9">
       <a href="/blog/{{ article.slug }}" aria-hidden="true" tabindex="-1">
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
 alembic==1.4.2
 canonicalwebteam.flask_base==0.6.3
 canonicalwebteam.http==1.0.3
-canonicalwebteam.blog==6.0.0
+git+git://github.com/albertkol/canonicalwebteam.blog.git@re-enable-cloudinary#egg=canonicalwebteam.blog
```
jpmartinspt, line 4: can you try `git+https` instead?

nottrobin: why? is there a problem with using SSH?

jpmartinspt: The image was not building, because it did not have `git` as a dependency in the python step. Just wanted to try this to check if it avoided `pip` attempting to git clone. It does not so we added `git` in the dockerfile.

Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Build stage: Install python dependencies
 # ===
 FROM ubuntu:focal AS python-dependencies
-RUN apt-get update && apt-get install --no-install-recommends --yes python3-pip python3-setuptools python3-wheel build-essential
+RUN apt-get update && apt-get install --no-install-recommends --yes python3-pip python3-setuptools python3-wheel build-essential git
```
nottrobin, line 6: Is this just so the demo will spin up with the `git` line in the `requirements.txt`?

That's fine, but please make sure we remove this again if we don't need it for any other reason.

PR #7990, opened 2020-07-24
Fix page jump as global nav loads

steverydz:
## Done

- Fixed the page jump as the page loads
- Removed an SCSS comment that was causing a linting error

## QA

- Load the page and compare the experience to ubuntu.com
- The page should no longer jump when the global nav loads


steverydz:
@anthonydillon thanks, I've mad the change

Issue #7926, opened 2020-07-14
Add search parameters for CVE component search


steverydz:
There is a requirement for the CVE search form to have a search field for components (Any, Main, Partners, Universe). For this to work there needs to be a URL parameter for this, i.e. `?component=main`

[Project brief](https://docs.google.com/document/d/1ns0OxQVnICxgq2xGsvWz6YFMUJgQI35Vsqy1yOokhko/edit)
[UX spec](https://docs.google.com/document/d/1VuAElUpFay7MBFNmE3NhSyY9TkXiqHcA1ZHQ11AYO-Q/edit#)

steverydz:
@albertkol [This PR](https://github.com/canonical-web-and-design/ubuntu.com/pull/7945) adds the `component` query parameter which you should be able to use to return the results based on this filter.

PR #7994, opened 2020-07-26
WIP: Search exact package name

nottrobin:
Rather than fuzzy match, which is more confusing. (may fix https://github.com/canonical-web-and-design/ubuntu.com/issues/7883)


albertkol:
On top of this we also need to filter out the cve packages that do not match the search package name.

Since it's very similar to the new component search that we plan to implement (#7926), once that is done, we will be a big step closer to filter for package name too.

PR #8004, opened 2020-07-28
Fix CVE package filter

### LINKED ISSUE #7883: CVEs: Package name filter returns unwanted packages
The CVEs on https://staging.ubuntu.com/security/cve can be filtered by package, however, the current filtering is not specific enough. For example, if I want to see all of the CVEs that affect the package "git", I also get results for gitlab, golang-github-containers-buildah, golang-code.gitea-git, gitaly, etc. I also get results for some packages which do not contain the string "git", such as hhvm.

I would expect a quick and easy way (perhaps via regex) that would allow me to see only the CVEs affecting a single package.

albertkol:
Make package filter by exact name instead of using `LIKE`.

Display only packages of the cve that match the search criteria.
Before it was displaying all cve packages as long as one package matches the criteria.

## QA

1. Prepare local database (optional; if you have no CVE data on local):
```bash
docker-compose down # reset db
docker-compose up -d 
dotrun # to run the migrations
dotrun exec python3 scripts/create-releases.py # to populate release table
dotrun # serve the website
```
In a new terminal; Run the importer:
```bash
git clone -b test-status-component-field git@github.com:albertkol/ubuntu-cve-tracker.git # fetch testing branch

cd ubuntu-cve-tracker/scripts
python3 -m venv .venv
source .venv/bin/activate
pip3 install requests cvss macaroonbakery
time python3 upload-cve.py ../active
```

2. Browse to `/security/cve`
Browse for packages with the name fx: `php` you will only see the packages listed with that name.

Before it would list all the packages of the cve as long as one is matching the search criteria.

## Issue / Card

Fixes #7883

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -16,6 +16,8 @@ from sqlalchemy import asc, desc, or_
 from sqlalchemy.exc import IntegrityError, DataError
 
 # Local
+from sqlalchemy.orm import contains_eager
+
 from webapp.security.database import db_session
 from webapp.security.models import CVE, Notice, Package, Status, Release
 from webapp.security.schemas import CVESchema, NoticeSchema
@@ -333,13 +335,9 @@ def cve_index():
         db_session.query(CVE)
         .filter(CVE.statuses.any(Status.status.in_(Status.active_statuses)))
         .filter(CVE.status == "active")
+        .join(CVE.statuses)
+        .options(contains_eager(CVE.statuses))
     )
-
-    # Apply search filters
-    if package:
-        cves_query = cves_query.filter(
-            CVE.statuses.any(Status.package_name.ilike(f"%{package}%"))
-        )
 
     if priority:
         cves_query = cves_query.filter(CVE.priority == priority)
@@ -347,15 +345,16 @@ def cve_index():
     if query:
         cves_query = cves_query.filter(CVE.description.ilike(f"%{query}%"))
 
-    cves = (
-        cves_query.order_by(desc(CVE.published))
-        .offset(offset)
-        .limit(limit)
-        .all()
-    )
+    # Apply search filters
+    if package:
+        cves_query = cves_query.filter(Status.package_name == package)
+
+    cves = cves_query.order_by(desc(CVE.published)).all()
 
     # Pagination
-    total_results = cves_query.count()
+    total_results = len(cves)
+    last_item = offset + limit
+
     releases = (
         db_session.query(Release)
         .order_by(desc(Release.release_date))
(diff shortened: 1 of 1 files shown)


webapp/security/views.py
```diff
@@ -354,6 +354,15 @@ def cve_index():
...
     )
 
+    cve_tree = defaultdict(dict)
+    for cve in cves:
+        packages = cve.status_tree
+
+        if package:
+            packages = {k: v for k, v in packages.items() if k == package}
+
+        cve_tree[cve.id] = packages
```
carkod, line 364: Ideally we should do this in the original cves_query, it could be expensive to loop through thousands of packages, @jpmartinspt any ideas?

jpmartinspt: I might be misreading the purpose of this.. But if is to only show packages on the template that match this filters we could just add a couple of `if` statmentes on the template itself since we already loop through the packages there to render them?

jpmartinspt: In the appropriate place in `_cve-table.html` -> `{% if request.args.get('package') and package_name = request.args.get('package' ) %}` so we only render packages with the matching name.

albertkol: Filtering in the packages in the view would be a nice way to avoid some extra for loops. The only problem is that there are more filters to be applied fx: filtering by component in my second PR: https://github.com/canonical-web-and-design/ubuntu.com/pull/8006 The view would end up containing too much logic. 

I do agree that it would be the cleanest to make `cves_query` do the filtering, but right now, I am not sure how..

jpmartinspt: Two other oprions:
1) Transform each `cve` into a dict instead of just passing down the sqlalchemy object to the template. That way we can change and filter the packages field.(can be tricky if we are using sqlalchemy specific stuff in the template logic)

2) Make `cve.status_tree` a `hybrid_method` that takes arguments so we can filter down the tree to whatever we want.

albertkol: 1. Would this not ruin the pagination which is implemented with `sqlalchemy`'s help (using `offset` and `limit`)?
2. This is interesting. Would you pass `request.args.get('package')` as a parameter inside the `hybrid_method`?

jpmartinspt: For the second option, yes you would call the method on the template with arguments.

albertkol: I found a way to make it work using `cves_query`. @carkod could you please check the code one more time. It's much cleaner now.

nottrobin:
This solution looks perfect! Thanks @albertkol. Code +1 from me. I haven't QAed it, but I guess @carkod has.

carkod:
Yes I QAed again just in case and searched `busybox`

PR #8006, opened 2020-07-29
Search cves by component

### LINKED ISSUE #7926: Add search parameters for CVE component search

albertkol:
Enable component filter.

Display only packages of the cve that match the search criteria.

## QA

1. Have the latest CVEs imported in your local database. `CVE-2020-8834` should be the most recent cve on `/security/cve`
2. All the packages should have the component `main` right now. We have to create a `universe` component to test the filter.
```bash
docker exec -it db psql -U postgres # go to the databse
```
```SQL
UPDATE status SET component='universe' WHERE cve_id='CVE-2020-8834' AND release_codename='focal' AND package_name='linux' 
```
3. Fetch my branch changes: 
```bash 
git fetch albert-fork
git checkout search-cves-by-component
```

4. Browse to `/security/cve`. Filter by component, you should only see one row.

## Issue / Card

Fixes #7926

### DIFF
--- a/templates/security/cve/_cve-table.html
+++ b/templates/security/cve/_cve-table.html
@@ -17,10 +17,10 @@
   </thead>
   <tbody>
     {% for cve in cves %}
-      {% for package_name, statuses in cve.status_tree.items() %}
+      {% for package_name, statuses in cve.packages.items() %}
         <tr>
           {% if loop.index == 1 %}
-            <td rowspan="{{ cve.status_tree | length }}"><a href="/security/{{ cve.id }}">{{ cve.id }}</a></td>
+            <td rowspan="{{ cve.packages | length }}"><a href="/security/{{ cve.id }}">{{ cve.id }}</a></td>
           {% endif %}
           <td>{{ package_name }}</td>
           {% for release in releases %}
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -14,8 +14,6 @@ from mistune import Markdown
 from sortedcontainers import SortedDict
 from sqlalchemy import asc, desc, or_
 from sqlalchemy.exc import IntegrityError, DataError
-from sqlalchemy.orm import contains_eager
-from sqlalchemy.sql.expression import nullslast
 
 # Local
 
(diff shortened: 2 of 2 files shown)


carkod:
But have you guys been able to import the CVEs from the other components? e.g. partners, universal ...

albertkol:
> But have you guys been able to import the CVEs from the other components? e.g. partners, universal ...

I think everything that comes in right now is for main. There is some work they need to do to start sending data for other components, but as long as it comes in the expected format, the filter should be working correctly.

webapp/security/views.py
```diff
@@ -354,6 +359,22 @@ def cve_index():
...
+            packages = {k: v for k, v in packages.items() if k == package}
+
+        if component:
+            packages = {
+                k: v
+                for k, v in packages.items()
+                if any(component == a.component for a in v)
+            }
+
+        cve_tree[cve.id] = packages
```
carkod, line 376: Ideally we should do this in the original `cves_query`, it could be expensive to loop through thousands of packages, @jpmartinspt any ideas?

konf/site.yaml
```diff
@@ -13,7 +13,7 @@ env:
       key: google-custom-search-key
       name: google-api
 
-  - name: LAUNCHPAD_TOKEN
+  - name: LAUNCHPAD_TOKEpp/security/views.pyN
```
nottrobin, line 16: A mistake, I assume?

templates/security/cve/_cve-table.html
```diff
@@ -17,10 +17,10 @@
   </thead>
   <tbody>
     {% for cve in cves %}
-      {% for package_name, statuses in cve.status_tree.items() %}
+      {% for package_name, statuses in cve.packages.items() %}
```
nottrobin, line 20: Love this change. This is much more straightforward. Thanks for massively improving on my bad `status_tree` name.

I'd in fact love it if we could rename this all the way down to the model at https://github.com/albertkol/ubuntu.com/blob/search-cves-by-component/webapp/security/models.py#L68. Do you think we could do that? You don't have to do it in this PR if you think it'll delay things.

konf/site.yaml
albertkol: Yes. Fixed it..

templates/security/cve/_cve-table.html
albertkol: I will rename the `status_tree` method to `packages` as part of my second PR (https://github.com/canonical-web-and-design/ubuntu.com/pull/8090) to make my life easier.

PR #8015, opened 2020-07-30
Fix single CVE page

albertkol:
Fix single CVE page so that if status is missing for a package, the table does not get messed up.

## QA

```bash
git remote add albert-fork git@github.com:albertkol/ubuntu.com.git  # get my fork if you don't have it
git fetch albert-fork  # fetch the branches
git checkout fix-cve-page-for-missing-status  # checkout the branch
```

- Run website
```bash
dotrun
```

- browse to: `/security/CVE-2020-8834` check the table

Fixes #

## Screenshots

This is what it looks now:
image

### DIFF
--- a/templates/security/cve/cve.html
+++ b/templates/security/cve/cve.html
@@ -135,8 +135,8 @@
           <tbody>
             {% for package_name, statuses in cve.status_tree.items() %}
               {% for release in releases %}
-                {% if release.codename in statuses %}
-                  <tr>
+                <tr>
+                  {% if release.codename in statuses %}
                     {% if loop.index == 1 %}
                       <td rowspan="{{ releases | length }}">{{ package_name }}</td>
                     {% endif %}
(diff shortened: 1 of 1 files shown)


templates/security/cve/cve.html
```diff
@@ -188,6 +188,8 @@ <h2>Status</h2>
                       {% endif %}
                     </td>
                   </tr>
+                {% else %}
```
steverydz, line 191: I think the better fix would be to move the opening `tr` on line 139 above `{% if release.codename in statuses %}` and the closing `tr` on line 190 beneath the closing `{% endif %}`

albertkol: That will avoid the `else`. Fine.

Also I will remove a number of extra, redundant `{% if release.codename in statuses %}` inside the main `{% if release.codename in statuses %}`.

PR #8048, opened 2020-08-07
Advantage shop payments

sowasred2012:
# Done

Added ability to purchase UA products when logged in via SSO

## notes
- we don't have the ability to checkout as a guest yet - CS team are working on it
- when you're redirected to /advantage, you won't be able to see how many machines you have available to attach, that's coming in https://github.com/canonical-web-and-design/ubuntu.com/pull/8140

# QA

- Go to /advantage/subscribe
- If you're not already logged in via SSO, there'll be a prompt below the header asking you to do so - do so
  - (if you _are_ logged in, there won't be a prompt)
- Use the form to select a product
  - AWS and Azure should prevent you from being able to progress through the form, as should selecting "Other" in the version section
- Add the product to the basket, and click "Buy now"
- Using test cards from https://stripe.com/docs/testing#cards, try to complete your purchase:
  - card number 4242 4242 4242 4242 should result in the modal showing a success message before taking you to /advantage, where you will see your new subscription
  - Test card numbers designed to trigger a 3DS check should trigger a 3DS modal when you click "Pay" on the second page in the modal.
  - Test card numbers designed to fail various checks should return an appropriate error message when clicking "Continue" on the first page of the modal
    - The exception to this is 4000000000000341, which will not return an error on the first page, but should return a "Card declined" message when you click "Pay"
  - Ensure that, after any failed payment attempts, you are able to input a new payment method and have it succeed during payment (using the 4242 4242 4242 4242 test card)

### DIFF
--- a/.env
+++ b/.env
@@ -4,7 +4,7 @@ DEVEL=true
 DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres
 SECRET_KEY=insecure_dev_key
 CONTRACTS_API_URL=https://contracts.develop.canonical.com/
-STRIPE_PUBLISHABLE_KEY=pk_test_yndN9H0GcJffPe0W58Nm64cM00riYG4N46
+STRIPE_PUBLISHABLE_KEY=pk_test_LbxZRQZdP7xsZenWT1TAhbkX00VioMBflp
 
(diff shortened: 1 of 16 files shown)


steverydz (approved):
LGTM :+1:

PR #8050, opened 2020-08-07
Enable Cypress

sowasred2012:
## Done

Fixed Cypress GitHub action and added Cypress to the project

## QA

See that the GitHub action runs and that there is output from the test runner

### DIFF
--- a/.github/workflows/cypress.yaml
+++ b/.github/workflows/cypress.yaml
@@ -12,21 +12,16 @@ on:
 jobs:
   run-cypress:
     services:
-      # Label used to access the service container
       postgres:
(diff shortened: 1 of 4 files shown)


.github/workflows/cypress.yaml
```diff
@@ -1,27 +1,54 @@
...
+  CONTRACTS_API_URL: contracts.staging.canonical.com
 
 on:
   schedule:
     - cron: "0 7 * * *"
 
 jobs:
   run-cypress:
+    services:
+      # Label used to access the service container
```
tbille, line 15: Maybe those comments can be removed

PR #8057, opened 2020-08-10
Remove built JS files

steverydz:
## Done

In [this commit](https://github.com/canonical-web-and-design/ubuntu.com/commit/96d9ba90f9fe361fe7901e8e5b2a0e64ad285806) the built JS files were added to the repository. This PR removes them from git tracking.

## QA

- Remove your local built JS files: `rm -rf static/js/dist`
- Run the site: `dotrun`
- Check that the build JS files have been built: `ls --ltr static/js/dist`
- Run `git status` to make sure no changes are flagged on the built JS files


pmahnke:
do people need to update the .gitignore?

steverydz:
> do people need to update the .gitignore?

@pmahnke  As long as you're working from the latest upstream then this shouldn't be an issue. The `.gitignore` is as it should be. If you are having an issue locally then you may want to remove the cached files:

```
git rm -r --cached static/js/dist
```

PR #8090, opened 2020-08-14
Search cves by version status

### LINKED ISSUE #7928: Add search parameters for CVE Ubuntu version/status search
There is a requirement to have search fields for Ubuntu version/status pairs. It needs to be possible for multiple fields of version/status pairs.

See the **Search function** section of the [UX spec](https://docs.google.com/document/d/1VuAElUpFay7MBFNmE3NhSyY9TkXiqHcA1ZHQ11AYO-Q/edit#)

[Project brief](https://docs.google.com/document/d/1ns0OxQVnICxgq2xGsvWz6YFMUJgQI35Vsqy1yOokhko/edit)

albertkol:
Enable version-status filter.

Display only packages of the cve that match the search all the version-status criteria.

## QA

1. Have CVEs imported in your local database.
2. Fetch my changes:
```bash
git fetch albert-fork
git checkout search-cves-by-component
```
3. Browse to `/security/cve`. Play with the version-status filter.

## Issue / Card

Fixes #7928

### DIFF
--- a/webapp/security/models.py
+++ b/webapp/security/models.py
@@ -67,18 +67,18 @@ class CVE(Base):
     )
 
     @hybrid_property
-    def status_tree(self):
-        status_tree = defaultdict(dict)
+    def packages(self):
+        packages = defaultdict(dict)
         for status in self.statuses:
-            status_tree[status.package_name][status.release_codename] = status
+            packages[status.package_name][status.release_codename] = status
 
-        return status_tree
+        return packages
 
     @hybrid_property
     def active_status_tree(self):
         active_package_statuses = {}
 
-        for package_name, release_statuses in self.status_tree.items():
+        for package_name, release_statuses in self.packages.items():
             for status in release_statuses.values():
                 if (
                     status.status in Status.active_statuses
--- a/templates/security/cve/_status-fields.html
+++ b/templates/security/cve/_status-fields.html
@@ -5,7 +5,7 @@
 <option value="needs-triage" {% if statuses[index] == "needs-triage" %}selected{% endif %}>
   Needs triage
 </option>
-<option value="not-vulnerable" {% if statuses[index] == "not-vulnerable" %}selected{% endif %}>
+<option value="not-affected" {% if statuses[index] == "not-affected" %}selected{% endif %}>
   Not vulnerable
 </option>
 <option value="needed" {% if statuses[index] == "needed" %}selected{% endif %}>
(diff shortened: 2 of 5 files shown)


webapp/security/models.py
```diff
@@ -147,6 +147,17 @@ class Status(Base):
...
+    all_statuses = [
+        "released",
+        "DNE",
+        "needed",
+        "not-affected",
+        "deferred",
+        "needs-triage",
+        "ignored",
+        "pending",
+    ]
```
carkod, line 159: Where are we using this? the variable `all_statuses` is not being mentioned anywhere else in `models.py`.

As an aside, I was wondering if we are interested in converting `active_statuses` as an `Enum`?

carkod:
## QA
Do we want to show `Upstream None` in the Ubuntu version list?

Screenshot from 2020-08-25 15-57-34

webapp/security/models.py
albertkol: `all_statuses` is used in view.py:375 to default to `all_statuses` if `Any` is selected in the status drop-down. I chose to put it in the model, since it belongs to the `status` entity.

Not sure what you mean by converting `active statuses` as an `Enum`.

albertkol:
> ## QA
> * Do we want to show `Upstream None` in the Ubuntu version list?
> * Also if the [UX spec](https://docs.google.com/document/d/1VuAElUpFay7MBFNmE3NhSyY9TkXiqHcA1ZHQ11AYO-Q/edit#heading=h.aoek8nxke7k9) hasn't changed, I am aware that we need to order this list by date (latest version first, oldest version last)
> 
> Screenshot from 2020-08-25 15-57-34

Good catch. Should be fixed now.

webapp/security/models.py
carkod: Well, you don't have to do this, but I was suggesting to get `all_statuses` list from the table (as query) as I believe these are stored in a table line 166 of this `models.py`?

albertkol: Ok.. I found out how to fetch the enum of statuses from the database. In a way I am not keen on adding an extra database query, but it is probably nice to have only one place that stores all the statuses. I will make the change shortly.

albertkol: Done.

carkod: > Ok.. I found out how to fetch the enum of statuses from the database. In a way I am not keen on adding an extra database query, but it is probably nice to have only one place that stores all the statuses. I will make the change shortly.

Yeah, I would not have asked if I didn't see the other `Enum`. Because it is true that initially we did it without `Enums` but if we are using them in one place, and then using static statuses somewhere else, this could lead us to different versions of these statuses.

Nice!

carkod (approved):
LGTM

PR #8190, opened 2020-08-28
Refactor how machines usage is calculated

frankban:
## Done

- Fixed a type error: `"0/" in x` where x was an int.
- Changed how allowance is calculated: now the contract is used rather than listings.
- Changed how machine usage is returned: a more structured named tuple rather than either a string or an int.
- Some linting I was forced to commit by commit hooks. 

## QA

- Check that the number of machines is correctly displayed both when having an allowance and when not.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ database.json
 .login
 database.json.1
 static/js/build/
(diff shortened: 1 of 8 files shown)


sowasred2012 (approved):
LGTM

PR #8195, opened 2020-08-31
use staging contracts api

sowasred2012:
For the sake of having a demo pointing to staging contract service - do not merge


sowasred2012:
redundant, closing

PR #8211, opened 2020-09-02
Optimise CVE filters

albertkol:
Reduce the number of queries done on the CVE page. From 25 down to 4, with an over 50% speed improvement. (I did not take before and after screenshots so you will have to take my word for it.) 

Two big changes: 
1) `total_results = cves_query.count()` -> removed

`cve_query` is already a massive query which used to be executed twice: 
 - One time to get the total results found 
 - One time to apply the `limit` and the `offset` on the results

Now we do it all in one, selecting the CVEs and the `count(*)`.

2) Load `CVE.statuses` one time forever, so every time we do `cve.statuses`, the values are already there. This reduced the number of queries by 20. 

## QA

1. Have the CVEs imported

2. Fetch changes
```bash
git fetch albert-fork
git checkout optimize-filter
```
3. Play with the filters on `/security/cve`. It should work.

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -12,11 +12,12 @@ from marshmallow import EXCLUDE
 from marshmallow.exceptions import ValidationError
 from mistune import Markdown
 from sortedcontainers import SortedDict
-from sqlalchemy import asc, desc, or_, and_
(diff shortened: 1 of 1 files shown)


carkod (approved):
Nice. LGTM

PR #8217, opened 2020-09-03
Advantage shop polish

### LINKED ISSUE #8182: Advantage subscribe has some keyboard navigation funkiness
- I can't tab through the different server options
- I can tab to disabled areas and interact with input fields

Screenshot 2020-08-27 at 15 58 48

sowasred2012:
## Done

- Added a check to make sure a user is an admin on one of their accounts
- The product selection form now scrolls between steps
- If the user is not logged in, or doesn't have an account with admin permissions, prevents the user being able to select a product
  - Temporary, until guest checkout can be implemented
- Prevents tabbing to inputs within `u-disable` sections

## QA

- Assuming you are not signed in, the form should be disabled
  - You should also not be able to tab to any of the inputs
- Follow the link to login with SSO, and do so
- When you are returned to the page, you should see the first section of the form is enabled (if not, you don't have an account with admin permissions, Scott can help you)
- Try to tab to inputs in sections that are disabled, you should not be able to

## Issue / Card

Fixes #8182

### DIFF
--- a/.env
+++ b/.env
@@ -3,7 +3,7 @@ FLASK_DEBUG=true
 DEVEL=true
 DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres
 SECRET_KEY=insecure_dev_key
-CONTRACTS_API_URL=https://contracts.develop.canonical.com/
+CONTRACTS_API_URL=https://contracts.staging.canonical.com/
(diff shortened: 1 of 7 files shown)


steverydz (approved):
LGTM :+1:

PR #8226, opened 2020-09-04
Check Stripe env when attempting to send GA events on /advantage

sowasred2012:
## Done

For shop & renewals, added a check to see whether we're using the Stripe prod or test environment (Stripe key in prod begins with "pk_live", and in test it's "pk_test"), and prevent GA events from being fired outside of production.

Also updated the event action, so we know whether the transaction was a renewal or a purchase.

## QA

- Install and activate the [GA debugger](https://addons.mozilla.org/en-GB/firefox/addon/ga-debugger/)
- Open your browser's console
- Select a product, add it to cart, and open the payment modal
- The console should show no GA activity when you open the modal

### DIFF
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -668,7 +668,7 @@ function sendGAEvent(label) {
     dataLayer.push({
       event: "GAEvent",
       eventCategory: "advantage",
-      eventAction: "renewal",
+      eventAction: currentTransaction.type,
       eventLabel: label,
       eventValue: undefined,
     });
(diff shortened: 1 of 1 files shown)


static/js/src/ua-payment-modal.js
```diff
@@ -605,15 +605,17 @@ function processStripePayment() {
       currentTransaction.previousPurchaseId
     )
       .then((data) => {
-        if (data.id) {
-          currentTransaction.transactionId = data.id;
-        }
-        handlePaymentAttemptResponse(data);
+        console.log(data);
```
steverydz, line 608: Please remove `console.log()` and commented out code

steverydz (approved):
LGTM :+1: with a small change

Issue #8235, opened 2020-09-07
Stripe payment method needs to be set before we can show VAT


sowasred2012:
Encountered a wrinkle whilst implementing the VAT/purchase preview endpoint:

- we want to show the VAT to the user on the payment method screen (the first in the payment modal)
- the contracts API requires us to have created a payment method in Stripe, including address, so that they can make their VAT calculations based on what is in Stripe
- we don't create the payment method until the user has clicked "Continue" i.e. they are moving to the second screen of the modal
- this means, at the moment the contracts service doesn't have an address to work with and just assumes VAT applies regardless

Possible solutions to be investigated/validated:
- can we send an address to the VAT/purchase preview endpoint to get the figures we need without creating a payment method?
- can we only show VAT and Total figures on the second screen of the modal, when a payment method has been created?
- can we somehow create a temporary payment method when the user has filled out the required information, whose only purpose is to be used to calculate VAT and Total, then automatically discarded?

#### Tangentially related question:
- it looks like VAT is being returned in USD. I'm not certain what the rules are in other VAT-applicable countries, but the UK requires that VAT totals are shown in GBP, can we have that total returned in the appropriate currency?

Tagging @alnvdl too

matthewpaulthomas:
> - can we send an address to the VAT/purchase preview endpoint to get the figures we need without creating a payment method?

A variation of this:
- Is there some non-Stripe mechanism we can use to determine VAT existence+rate before creating the payment method?

> - can we only show VAT and Total figures on the second screen of the modal, when a payment method has been created?

Well, we can, but this is the last resort. The amount of tax added might tip you over a threshold where you can’t:
- pay with this particular card; or
- pay by yourself, so you need to get approval/payment from someone else in your organisation; or
- afford the service at all.

So I’d rather avoid a situation where we’re keeping around a payment method that you aren’t going to use.

> at the moment the contracts service doesn't have an address to work with and just assumes VAT applies regardless

Since tax rates are different in different countries, how does it “assume VAT applies”?

sowasred2012:
> Since tax rates are different in different countries, how does it “assume VAT applies”?

[...]

PR #8244, opened 2020-09-09
Advantage shop VAT

### LINKED ISSUE #7647: UA purchase: Implement VAT validation and total
Currently when you renew a UA subscription, in a country that has VAT, the dialog just says “Excluding VAT”.

Now that we have more time, we should make this more informative by showing:
- the VAT rate in your country
- the resulting VAT amount
- the total to pay including VAT.

[This is the sequel to #7619.]

sowasred2012:
## Done

Implemented a VAT/purchase preview endpoint that returns a VAT total (when applicable) and actual total and presents it to the user during the payment modal journey.

## QA

- Whilst logged in, make a product selection, add it to the cart, and click "Buy Now" and QA the following two flows:

#### VAT
- In the payment modal, select United Kingdom (VAT is always applicable)
- After a short wait, VAT and total should appear below the form
- Change the country to United States (VAT is not applicable)
- After a short wait, VAT should disappear and total should update
- Change the country to France (VAT is applicable without a VAT number)
- After a short wait, VAT and total should appear below the form
- Fill out the VAT number field with "FR77784608416" (Credit Agricole's VAT number)
- After a short wait, VAT should disappear and total should update

#### Prorating
- In the payment modal, allow a moment for the subtotal to appear
- If you have an existing subscription that ends in less than a year, the end date of each plan should change to reflect that, and the first plan will contain a note explaining the change.
  - If you don't have such a subscription, ask Scott to set you up with one.

## Issue / Card

Fixes #7647 and #8235

### DIFF
--- a/.env
+++ b/.env
@@ -4,7 +4,7 @@ DEVEL=true
 DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres
 SECRET_KEY=insecure_dev_key
 CONTRACTS_API_URL=https://contracts.develop.canonical.com/
-STRIPE_PUBLISHABLE_KEY=pk_test_yndN9H0GcJffPe0W58Nm64cM00riYG4N46
+STRIPE_PUBLISHABLE_KEY=pk_test_LbxZRQZdP7xsZenWT1TAhbkX00VioMBflp
 
 # This is a test api key https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
 CAPTCHA_TESTING_API_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
--- a/.env
+++ b/.env
@@ -4,7 +4,7 @@ DEVEL=true
 DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres
 SECRET_KEY=insecure_dev_key
 CONTRACTS_API_URL=https://contracts.develop.canonical.com/
-STRIPE_PUBLISHABLE_KEY=pk_test_LbxZRQZdP7xsZenWT1TAhbkX00VioMBflp
+STRIPE_PUBLISHABLE_KEY=pk_test_yndN9H0GcJffPe0W58Nm64cM00riYG4N46
 
 # This is a test api key https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
 CAPTCHA_TESTING_API_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
--- a/.env
+++ b/.env
@@ -3,7 +3,7 @@ FLASK_DEBUG=true
 DEVEL=true
 DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres
 SECRET_KEY=insecure_dev_key
-CONTRACTS_API_URL=https://contracts.develop.canonical.com/
+CONTRACTS_API_URL=https://contracts.staging.canonical.com/
 STRIPE_PUBLISHABLE_KEY=pk_test_LbxZRQZdP7xsZenWT1TAhbkX00VioMBflp
 
 # This is a test api key https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do
--- a/.env
+++ b/.env
@@ -3,8 +3,8 @@ FLASK_DEBUG=true
 DEVEL=true
 DATABASE_URL=postgres://postgres:pw@localhost:5432/postgres
 SECRET_KEY=insecure_dev_key
-CONTRACTS_API_URL=https://contracts.staging.canonical.com/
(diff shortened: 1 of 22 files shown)


matthewpaulthomas:
> - After a short wait, VAT and total should appear below the form

It never appears for me, for either UK or France.

Meanwhile I reported #8247 and #8248, but I doubt they’re specific to this branch.

sowasred2012:
> It never appears for me, for either UK or France.

Fixed

matthewpaulthomas:
From our discussion just now:

- With a large subscription, switching repeatedly between UK and France, the VAT amount fluctuates by a few cents each time. Why is this? Is Stripe calculating it in the original currency, then converting to USD with real-time exchange rates? (Perhaps this doesn’t matter, if it’s not going to be noticeable for the kinds of amounts people use a credit card for.)

- When you choose a country, the “VAT:” line isn’t shown — or, showing a previous value, isn’t updated — until the API returns. If the server is bogged down or your Internet connection is slow, you might submit the form while an incorrect total is being shown. Better to show “VAT: …” and “Total: …” instantly, replaced by the new values when available.

matthewpaulthomas:
From our discussion today, here’s what I think should happen: 

Condition | “VAT:” line | “Total:” line
-|-|-
Initial | not present | “Total: $XXXX”
Immediately after choosing a country/region that doesn’t have VAT | not present | “Total: $NNNN”
Immediately after choosing a country/region that has VAT | “VAT: …” | “Total: …”
Once VAT amount is calculated | “VAT: $YYY” | “Total: $ZZZZ”
If you’ve entered a valid VAT number | “VAT: $0” | “Total: $XXXX”

Does that make sense?

matthewpaulthomas:
One minor glitch:
- [ ] When changing country, the Subtotal changes to “Subtotal: …” despite the fact that it’s not going to change.

Apart from that, looks good to me.

matthewpaulthomas:
- [ ] If I enter an invalid VAT number and choose “Continue”, after a couple of seconds, “Continue” re-enables and … nothing else visible happens. An error has appeared above the form, but it’s not in view. Let’s move that error text immediately above the VAT field, and focus the field.

matthewpaulthomas:
- [ ] Enter card details and billing address, and choose country/region UK; “Continue” is enabled. Now focus the VAT number field; “Continue” becomes disabled and shouldn’t be. (Typing a VAT number doesn’t fix the problem, but tabbing away then back to the field does fix it.)

- [ ] Change country/region, then click “Continue” before the total calculates. The button becomes disabled, but [the rest of the form does not](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.jwi4rns7fr68), and nothing further happens.

- [ ] When there’s no VAT, in Pay mode, there’s too much gap between the Subtotal and the Total.

  image

matthewpaulthomas:
- [ ] Every time I enter a character in the “How many ______?” field, the page scrolls up so that I can’t see what I just typed. That doesn’t seem like it should be part of this branch, but I’m ~90% sure it wasn’t happening yesterday.

- When a subscription is prorated, the “Subtotal” near the top of the page is different from the “Subtotal” inside the dialog. As discussed, you’ve fixed this temporarily by changing the former to “Yearly cost”. (So we should have a better fix before introducing monthly subscriptions.)

- [ ] In the dialog, at the moment you first read the subtotal, it’s fairly likely to say “Subtotal: …”, causing bewilderment. And if the subscription is prorated, “Ends:” gets taller, pushing the rest of the dialog down, making it look unstable. For those two reasons, I think we’d better keep the whole dialog body area blank (or, after ≥2 seconds, show a centered spinner) until we know whether prorating is happening or not. (But don’t spend a lot of time on this, since it will be moot once the parent page knows the subtotal.)

>  - Enter card details and billing address, and choose country/region UK; “Continue” is enabled. Now focus the VAT number field; “Continue” becomes disabled and shouldn’t be.

This looks fixed to me now.

> - Change country/region, then click “Continue” before the total is updated. The button becomes disabled, but [the rest of the form does not](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.jwi4rns7fr68), and nothing further happens.

You’ve prevented this by disabling “Continue” until the total is updated. I don’t think that’s the right approach — someone’s unlikely to understand why the button is disabled, especially if the VAT and Total aren’t in view, for example if they Tabbed to the country/region menu. Instead, the whole form — including the Continue button — is supposed to disable when you choose “Continue”, regardless of whether the total has updated or not.

Other checked issues look fixed now.

matthewpaulthomas:
> Let’s move that error text immediately above the VAT field, and focus the field.

The error text now appears below the VAT number field, which is fine. (That risks being covered by an autocomplete menu — but if it was above, the field would jump up while you were retyping. I haven’t yet figured out an elegant way to avoid both.)

However, it’s the payment card field that gets focused, when it should be the VAT number field.

> In the dialog, at the moment you first read the subtotal, it’s fairly likely to say “Subtotal: …”

This is fixed now, nice work.

One final tweak: Can we make the total bold? It’s a little lost below all the other numbers.

steverydz (approved):
LGTM :+1:

PR #8265, opened 2020-09-10
Fix CVE order

### LINKED ISSUE #8263: Check why old CVEs are at the top
If you go to https://ubuntu.com/security/cve CVEs from 2001-2003 are at the top. It is suspected that these are missing a `published` date, so as a consequence they show up at the top.

Check to see if that's the case. If it is make those CVEs not be displayed at the top.

albertkol:
CVEs are ordered desc by `published` date. Some old CVEs do not have a
published date, so null gets inserted in the database. When these get
ordered NULL values are listed at the top.

This change will list the NULL values at the end.

## QA

0. Do not fetch changes.
1. Have the CVEs imported. 
2. Update some CVEs to have published date NULL
```bash
docker exec -it db psql -U postgres
```
And run:
```sql
UPDATE cve SET published=NULL WHERE id='CVE-2012-0937'; 
UPDATE cve SET published=NULL WHERE id='CVE-2012-1191'; 
UPDATE cve SET published=NULL WHERE id='CVE-2012-1148'; 
```
3. Browse to `/security/cve`. You will see these cves at the top.
4. Fetch changes:
```bash
git fetch albert-fork
git checkout cve-fix-order-if-published-null
```
5. Browse to `/security/cve`. You will not see these cves at the top.

## Issue / Card

Fixes #8263

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -12,7 +12,7 @@ from marshmallow import EXCLUDE
 from marshmallow.exceptions import ValidationError
 from mistune import Markdown
 from sortedcontainers import SortedDict
-from sqlalchemy import asc, desc, or_, and_, func
+from sqlalchemy import asc, desc, or_, and_, func, case
 from sqlalchemy.exc import IntegrityError, DataError
 from sqlalchemy.orm import contains_eager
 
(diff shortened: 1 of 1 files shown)


steverydz (approved):
LGTM :+1:

Issue #8201, opened 2020-09-01
Print formatting is bad at www.ubuntu.com/uasd


andrewjamessinclair:
When printing the page at www.ubuntu.com/uasd, there are some messy page breaks at the top of the document and the spacing between list items is very tight (perhaps that's intentional, but it's not consistent with the browser view of that page).

barrymcgee:
## List of items to address

- [ ] The spacing between line items in a ordered list is very tight
- [ ] Spacing above and below tables is very tight
- [ ] Sidebar headings are printed
- [ ] Large blank whitespaces at top of page

barrymcgee:
I dug into this a little to see why the content was being pushed onto the next page in print preview, what I found;

- This only happens on pages where the content is generated from a markdown file
- There's no much rhyme nor reason as to when the content is pushed onto the next page

You can resolve it replicate the bug by removing all content bar the first header and a the first table. You can then fix it by removing the table. However, if you add all content back, bar the table, the bug reappears.

This is as far as I have got so far.

anthonydillon:
This has been improved in the last 6 months.

PR #8268, opened 2020-09-11
Fix print page layout

bethcollins92:
## Done

- Fix line height
- Remove contents sidebar
- Fix margin on table in ol

## Blocker

- In print preview can't remove first page break, so first page is blank

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/legal/ubuntu-advantage-service-description
- View print preview to see changes. First page should appear blank with only Canonical link. 

## Issue / Card

Issue #8201 

## Screenshots

image
image

### DIFF
--- a/static/sass/print.scss
+++ b/static/sass/print.scss
@@ -23,7 +23,7 @@ body {
   color: $color-x-dark;
   font-family: "Ubuntu", "Segoe UI", "Roboto", "Oxygen", "Cantarell",
     "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
-  line-height: 1.3;
+  line-height: 1.5em;
 }
 
 a,
@@ -58,9 +58,9 @@ dl {
   page-break-before: avoid;
 }
 
+a,
 a:link,
-a:visited,
-a {
+a:visited {
   background: transparent;
   color: $color-link;
   font-weight: bold;
--- a/static/sass/print.scss
+++ b/static/sass/print.scss
@@ -28,11 +28,14 @@ body {
 
 a,
 blockquote,
-table,
 pre,
 [class^="p-strip"],
 [class*=" p-strip"] {
   page-break-inside: avoid;
+}
+
+table {
+  margin: 1.5em 0;
 }
 
 th,
@@ -91,7 +94,8 @@ footer,
 .global-nav__header-list,
 .dropdown-window,
 .dropdown-window-overlay,
-.p-navigation--secondary {
+.p-navigation--secondary,
+.p-table-of-contents {
   display: none;
 }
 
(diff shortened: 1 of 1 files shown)


static/sass/print.scss
```diff
@@ -23,18 +23,21 @@ body {
   color: $color-x-dark;
   font-family: "Ubuntu", "Segoe UI", "Roboto", "Oxygen", "Cantarell",
     "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif;
-  line-height: 1.3;
+  line-height: 1.5em;
```
barrymcgee, line 26: The line-height value in CSS is unitless - see: https://allthingssmitty.com/2017/01/30/nope-nope-nope-line-height-is-unitless/

static/sass/print.scss
```diff
@@ -91,7 +95,8 @@ footer,
 .global-nav__header-list,
 .dropdown-window,
 .dropdown-window-overlay,
-.p-navigation--secondary {
+.p-navigation--secondary,
+.p-table-of-contents {
```
barrymcgee, line 99: We can remove this whole sidebar so let's target `.p-sticky-toc` instead of `.p-table-of-contents`

static/sass/print.scss
```diff
@@ -23,18 +23,21 @@ body {
...
 blockquote,
-table,
 pre,
 [class^="p-strip"],
 [class*=" p-strip"] {
   page-break-inside: avoid;
 }
 
+table {
+  margin: 1.5em 0;
```
barrymcgee, line 38: We use `rem` instead of `em` across all of our codebases to keep our spacing consistent. You can read more about the difference between them here: https://webdesign.tutsplus.com/tutorials/comprehensive-guide-when-to-use-em-vs-rem--cms-23984

barrymcgee (changes requested):
Print preview looks much better! I have left some comment in the code.

static/sass/print.scss
```diff
@@ -91,7 +95,12 @@ footer,
 .global-nav__header-list,
 .dropdown-window,
 .dropdown-window-overlay,
-.p-navigation--secondary {
+.p-navigation--secondary,
+.p-sticky-toc {
+  display: none;
+}
+
+#canonical-global-nav {
```
barrymcgee, line 103: I asked you to add this but the class `.global-nav` in the selector above already applies this rule so this is not needed.

barrymcgee (changes requested):

static/sass/print.scss
```diff
@@ -91,7 +95,12 @@ footer,
 .global-nav__header-list,
 .dropdown-window,
 .dropdown-window-overlay,
-.p-navigation--secondary {
+.p-navigation--secondary,
+.p-sticky-toc {
+  display: none;
```
barrymcgee, line 100:
```suggestion
  display: none !important;
```

Let's make this trump all other display rules as we know for sure we don't need to override it anywhere else down the cascade.

barrymcgee (approved):

barrymcgee:
@bethcollins92 Looks great, thanks! 🙌🏻

PR #8289, opened 2020-09-15
Remove /desktop/activedirectory redirect

### LINKED ISSUE #8288: Update redirect for active directory
## Summary

This is a follow up to previous PR #8282 
mpt's recommendation on [MP 390662](https://code.launchpad.net/~jibel/ubiquity/+git/ubiquity-1/+merge/390662) is:

""""
Please drop the added “/desktop” here. The point is for the URL to be short enough that people can memorize or write it down before continuing. (And besides, it goes to a server page, not a desktop one!)

That means the redirect will need changing, but it needs fixing anyway — currently it points at the https://discourse.ubuntu.com/t/service-sssd/11579 staging version, when it should point at the https://ubuntu.com/server/docs/service-sssd user-facing version instead.
""""

Can you modify the redirection to be
ubuntu.com/activedirectory
and point to
https://ubuntu.com/server/docs/service-sssd 

Thank you.

pmahnke:
## Done

- Remove /desktop/activedirectory redirect
- Redirect wasn't published, so removed, plan to add it to help.ubuntu.com instead

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/desktop/activedirectory
- See that http://0.0.0.0:8001/desktop/activedirectory no longer redirects

## Issue / Card

Fixes #8288


redirects.yaml
```diff
@@ -53,6 +53,7 @@ aboutus/online-accounts-legal-notice/?: "/legal/online-account-terms"
 aboutus/privacypolicy/?: "/legal/data-privacy"
 aboutus/privacypolicy/thirdparties/?: "/legal/data-privacy"
 aboutus/trademarkpolicy.*/?: "/legal/intellectual-property-policy"
+activedirectory/?: "/server/docs/service-sssd"
```
tbille, line 56: the reason why I moved it too `/desktop/activedirectory` instead of a top level domain was mostly so we don't get used to this pattern of reserving toplevel domains.

If you think this won't be an issue and that we have hard requirements I don't mind putting this at this level.

matthewpaulthomas: There are two issues with including `/desktop`:
- We’re displaying the URL in an installer error dialog, for you to write down / photograph and use later (because visiting it now is premature). So the shorter, the better.
 - The redirect goes (both before and after this change) to a server-focused page, not a desktop one.

However, I just realised that in the two other cases where the installer needs to do this, the URL is help.ubuntu.com/topic, not ubuntu.com/topic. So I guess it would make more sense to put this redirect on help.ubuntu.com instead.

matthewpaulthomas: @jibel has kindly updated the MP for the dialog to display “help.ubuntu.com/activedirectory” instead. So the ubuntu.com redirect can be dropped altogether. Sorry for taking your time on this.

tbille (approved):

PR #8297, opened 2020-09-15
Show machine allowances on /renewals when possible

sowasred2012:
**⛔ revert API environment back to prod before merging ⛔**

## Done

- if a contract has info on its machine allowance, present it to the user on /advantage as "X/Y" where X is the number of attached machines, and Y is the number of machines they have paid for.

## QA

- Login
- If you don't have contract staging data already, talk to Scott and he'll set you up with what you need
- See that under "attached", you can see how many machines you have, and how many have been attached.

and/or: 

- see that the tests pass, and that the screenshot looks good

## Screenshots

Screenshot from 2020-09-15 15-30-00

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -1,4 +1,5 @@
 # Standard library
+from collections import namedtuple
 import hashlib
 import hmac
 import json
@@ -30,6 +31,9 @@ from webapp.advantage import AdvantageContracts
 ip_reader = geolite2.reader()
 session = talisker.requests.get_session()
 store_api = SnapStore(session=talisker.requests.get_session())
+
+# Define the metric name for the number of active machines.
+ALLOWANCE_METRIC_ACTIVE_MACHINES = "active-machines"
 
 
 def download_thank_you(category):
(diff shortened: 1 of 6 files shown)


frankban (approved):
Looks good!

matthewpaulthomas:
Working as planned:
image

webapp/views.py
```diff
@@ -499,6 +498,32 @@ def advantage_view():
...
+def get_machine_usage(advantage, contract):
+    """Return machine usage for the given contract as a MachineUsage object."""
+    allowances = contract.get("contractInfo", {}).get("allowances", [])
+    allowed = sum(
+        a["value"]
+        for a in allowances
+        if a["metric"] == ALLOWANCE_METRIC_ACTIVE_MACHINES
+    )
+    attached_machines = advantage.get_contract_machines(contract).get(
+        "machines"
```
jpmartinspt, line 510: If you you provide `[]` as a default in this function, bellow you can just have `attached = len(attached_machines)`.

jpmartinspt (approved):

PR #8302, opened 2020-09-15
Update bulk cve update

albertkol:
## Done

- Releases no longer needed at all, since we do not create releases through this endpoint
- Only update cves/statuses/packages if needed. We use a bool for each to find out if there are any changes coming from the data, if not do not update anything
- Delete only `cve.statuses` that we do not find in the cve data we receive; before we were deleting everything the inserting again which was too heavy and results in time outs. 

From my tests:
```
Before:
1st run - Insert round - Time: 7m0.526s
2nd run - Update round - Time: 7m32.567s

After:
1st run - Insert round - Time: 6m36.694s
2nd run - Update round - Time: 5m50.945s
```

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -571,40 +571,62 @@ def cve(cve_id):
 
 # CVE API
 # ===
-def update_statuses(cve, data, packages, releases):
-    updated_statuses = []
+def update_statuses(cve, data, packages):
+    statuses = cve.packages
+
+    statuses_query = db_session.query(Status).filter(Status.cve_id == cve.id)
+    statuses_to_delete = {
+        f"{v.package_name}||{v.release_codename}": v
+        for v in statuses_query.all()
+    }
 
     for package_data in data.get("packages", []):
         name = package_data["name"]
-        package = packages.get(name) or Package(name=name)
-        package.source = package_data["source"]
-        package.ubuntu = package_data["ubuntu"]
-        package.debian = package_data["debian"]
-        packages[name] = package
 
-        statuses = cve.packages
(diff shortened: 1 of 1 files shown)


carkod:
QA? is this an endpoint?

webapp/security/views.py
```diff
@@ -607,6 +607,36 @@ def update_statuses(cve, data, packages, releases):
...
+def delete_statuses(cve, data):
+    statuses_to_delete = (
+        db_session.query(Status).filter(Status.cve_id == cve.id).all()
+    )
+    statuses = cve.packages
+
+    for package_data in data.get("packages", []):
+        name = package_data["name"]
+
+        if not statuses[name]:
```
jpmartinspt, line 619: Won't this result in a `KeyError` if the key is not there? I think you want `if name in statuses`?

albertkol: You are correct. My mistake.

webapp/security/views.py
```diff
@@ -571,40 +571,67 @@ def cve(cve_id):
...
 
-            status.status = status_data["status"]
-            status.description = status_data["description"]
+            if "component" not in status_data:
+                status_data["component"] = None
+            if "pocket" not in status_data:
+                status_data["pocket"] = None
 
-            if "component" in status_data:
+            if status.component != status_data["component"]:
```
jpmartinspt, line 621: You could remove the if statments above by doing `status_data.get("component")`, right?

albertkol: Done.

jpmartinspt (approved):

5325833510:
Fatma

PR #8326, opened 2020-09-21
WIP: Add component on CVE page

### LINKED ISSUE #8155: Design presentation of package component on CVE pages
The CVEs main page lets you filter CVEs by component — that is, by whether they affect any packages  that are in a given component in any release.

However, when you get to the page for an individual CVE, you do not see which package(s) for which release(s) match that filter. This makes a CVE’s presence in those search results mysterious.

This can be fixed by adding the component for each package-release combination to the individual CVE page.

albertkol:
Add component on CVE page.

List component as a separate column.

Make sure the coloured status strips have same spacing on both sides.

## Issue / Card

Fixes #8155

## Screenshots

image

### DIFF
--- a/templates/security/cve/cve.html
+++ b/templates/security/cve/cve.html
@@ -130,6 +130,7 @@
               <th>Package</th>
               <th>Release</th>
               <th>Status</th>
+              <th>Component</th>
             </tr>
           </thead>
           <tbody>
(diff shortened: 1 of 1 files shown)


anthonydillon:
@albertkol why is this blocked?

albertkol:
@anthonydillon this is adding an extra column showing the type of component for package-release. But for now, we still do not have more than one component type. We decided waiting on it until the other component types exist.

PR #8339, opened 2020-09-22
Add endpoint to read notices

### LINKED ISSUE #7916: [usn website] please provide a json version of items under /security/notices/
## Summary

Please provide an endpoint that provides JSON versions of the notices under /security/notices/

This is useful for the security team to ensure that a given security notice has been published. It would also be useful for users that wish to track security notices programmatically; the security team provides a JSON of all our USNs, but working with all of them to extract the information from one is overkill.

Thank you.

albertkol:
Get notice:
`/security/api/notices/<notice_id>`
- returns notice with id `<notice_id>`
- `400` if no notice found with the id

`/security/api/notices`
- accepts two get parameters: `limit`, `offset`
These will paginate the results returned.
Offset defaults to 0.
Limit defaults to 20.

## QA

0) Have notices imported
1) Fetch changes
2) Browse to routes and see that you get 200

## Issue / Card

Fixes #7916

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -226,6 +226,50 @@ def _update_notice_object(notice, data):
     return notice
 
 
+def read_notice(notice_id):
+    """
+    GET method to get notice by id
+    """
+
+    notice = db_session.query(Notice).get(notice_id)
+
+    if not notice:
+        return (
+            flask.jsonify({"message": f"Notice {notice_id} does not exist"}),
+            404,
+        )
+
+    return flask.jsonify({"data": notice.as_dict()}), 200
+
+
+def read_notices():
+    """
+    GET method to get notices
+    """
+
+    limit = flask.request.args.get("limit", default=20, type=int)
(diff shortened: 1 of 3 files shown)


jpmartinspt:
Code is looking neat. I would split this into two views and were you are returning a 400 I think it should be a 404.

webapp/security/views.py
```diff
@@ -226,6 +226,47 @@ def _update_notice_object(notice, data):
...
+        .order_by(Notice.published)
+        .offset(offset)
+        .limit(limit)
+        .all()
+    )
+
+    return (
+        flask.jsonify(
+            {
+                "data": [n.as_dict() for n in notices],
```
jpmartinspt, line 261: We tend to avoid one letter variable names, can we go with `notice` please?

albertkol: Changed. Also split the endpoint in two.

jpmartinspt (approved):

PR #8349, opened 2020-09-23
Add information tool-tip for CVE statuses

### LINKED ISSUE #7763: CVEs: Implement priority+status explainer popup
[From the CVEs spec](https://docs.google.com/document/d/1VuAElUpFay7MBFNmE3NhSyY9TkXiqHcA1ZHQ11AYO-Q/edit#heading=h.32o4oau13zr0):
> image
>
> When you focus or mouse over a table cell, an “ℹ️” icon should appear to indicate that you can click it or press Enter to reveal info on the current priority and status. (The icon should not be there all the time, to reduce noise in a table with hundreds of status cells.)

albertkol:
Add tool-tip to give details on CVE status and priority.

We have no texts for the following statuses:
needs triage, needed, deferred, ignored, released, DNE

Right now cells with those statuses will show no status explanation in the tool-tip.

`Not vulnerable` and `Does not exist` do not show any priority explanation in the tool-tip because it does not show a priority icon in the cell.

`Does not exist` does not display the information icon since it's missing status explanation and has no priority icon

## QA

- Have CVEs imported
- `dotrun`
- Browse to `/security/cve` hover over statuses

## Issue / Card

Fixes #7763

### DIFF
--- a/static/js/src/cve/cve.js
+++ b/static/js/src/cve/cve.js
@@ -71,3 +71,71 @@ searchInput.addEventListener("keyup", handleSearchInput);
 attachEvents();
 handleButtons();
 disableSelectedVersions();
+
+const priorities = {
+  critical:
+    "<strong>Critical:</strong> A world-burning problem that is exploitable for most Ubuntu users. Examples include remote root privilege escalations or remote data theft.",
+  high:
+    "<strong>High:</strong> Exploitable for many users in the default configuration of the affected software. Examples include serious remote denial of service of the system, local root privilege escalations or local data theft.",
+  medium:
+    "<strong>Medium:</strong> Exploitable for many users of the affected software. Examples include network daemon denial of service, cross-site scripting and gaining user privileges.",
+  low:
+    "<strong>Low:</strong> Does very little damage or is otherwise hard to exploit, due to small user base or other factors such as requiring specific environment, uncommon configuration, or user assistance. These tend to be included in security updates only when higher priority issues require an update or if many low-priority issues have built up.",
+  negligible:
+    "<strong>Negligible:</strong> May be a problem, but does not impose a security risk due to various factors. Examples include when the vulnerability is only theoretical in nature, requires a very special situation, has almost no install base or does no real damage. These typically will not receive security updates unless there is an easy fix and some other issue causes an update.",
+  unknown:
+    "<strong>Unknown:</strong> Open vulnerability where the priority is currently unknown and needs to be triaged.",
+};
+
+const statuses = {
+  "not-affected":
+    "<strong>Not vulnerable:</strong> Packages which do not exist in the archive, are not affected by the vulnerability or have a fix applied in the archive.",
+  pending:
+    "<strong>Pending:</strong> A fix has been applied and updated packages are awaiting arrival into the archive. For example, this might be used when wider testing is requested for the updated package.",
+};
+
+const tooltipIconList = document.querySelectorAll(".cve-tooltip-icon");
+
+tooltipIconList.forEach(function (tooltipIcon) {
+  tooltipIcon.addEventListener(
+    "mouseover",
+    function (event) {
+      if (tooltipIcon.parentElement.querySelector(".cve-tooltip") == null) {
+        const priority = this.dataset.priority;
+        const status = this.dataset.status;
+
+        const tooltip = document.createElement("div");
+        tooltip.classList.add("cve-tooltip");
+
+        tooltip.innerHTML +=
+          "<div class='arrow-up back'></div><div class='arrow-up front'></div>";
+
+        if (priority in priorities) {
+          tooltip.innerHTML += priorities[priority];
+        }
+
+        if (priority in priorities && status in statuses) {
+          tooltip.innerHTML += "<br><br>";
+        }
+
+        if (status in statuses) {
+          tooltip.innerHTML += statuses[status];
+        }
+
+        tooltipIcon.parentElement.append(tooltip);
+      }
+    },
+    false
+  );
+  tooltipIcon.addEventListener(
+    "mouseout",
+    function (event) {
+      const tooltip = tooltipIcon.parentElement.querySelector(".cve-tooltip");
+
+      if (tooltip != null) {
+        tooltipIcon.parentElement.removeChild(tooltip);
+      }
+    },
+    false
+  );
+});
--- a/static/sass/_pattern_cve.scss
+++ b/static/sass/_pattern_cve.scss
@@ -12,6 +12,15 @@
     color: $color-x-light;
   }
 
+  .u-overflow-inherit {
+    overflow-x: inherit !important;
+  }
+
+  .cve-table {
+    font-size: 0.8rem;
+    line-height: 1rem;
+  }
+
   .cve-table thead tr {
     border-bottom: 0;
   }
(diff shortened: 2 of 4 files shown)


templates/security/cve/index.html
```diff
@@ -81,7 +81,7 @@ <h2>Search</h2>
   </form>
 </section>
 
-<section class="p-strip is-shallow">
+<section class="p-strip is-shallow" style="overflow-x: inherit">
```
carkod, line 84: I think this is not good practice. it's harder to maintain inline styles. Maybe adding a class to `cve-pattern` or maybe it would be even better to have a utility class e.g. `u-overflow-inherit` in Vanilla.

albertkol: Made the change.

templates/security/cve/_cve-table.html
```diff
@@ -98,6 +98,40 @@
...
+              <div class="icon-container__icon icon-cve-info">
+                {% if statuses[release.codename].status != 'DNE'%}
+                  <i class="p-icon--information"></i>
+                {% endif %}
+                <small class="cve-tooltip">
+                  <div class="arrow-up back"></div>
+                  <div class="arrow-up front"></div>
+                  {% if statuses[release.codename].status != 'DNE' and statuses[release.codename].status != 'not-affected' %}
+                    {%if cve.priority == "critical" %}
+                      <b>Critical:</b> Open vulnerability that is a world-burning problem and is exploitable for most Ubuntu users. Examples include remote root privilege escalations or remote data theft.
```
matthewpaulthomas, line 110:
```suggestion
                      <b>Critical:</b> A world-burning problem that is exploitable for most Ubuntu users. Examples include remote root privilege escalations or remote data theft.
```
Now that we’re showing status and priority separately, it’s no longer true that CVEs with a priority are always “open”. (Sorry I didn’t notice this until now.)

templates/security/cve/_cve-table.html
```diff
@@ -98,6 +98,40 @@
...
+                    {% elif cve.priority == "high" %}
+                      <b>High:</b> Open vulnerability that is a real problem and is exploitable for many users in the default configuration of the affected software. Examples include serious remote denial of service of the system, local root privilege escalations or local data theft.
```
matthewpaulthomas, line 112:
```suggestion
                      <b>High:</b> Exploitable for many users in the default configuration of the affected software. Examples include serious remote denial of service of the system, local root privilege escalations or local data theft.
```

templates/security/cve/_cve-table.html
```diff
@@ -98,6 +98,40 @@
...
+                    {% elif cve.priority == "medium" %}
+                      <b>Medium:</b> Open vulnerability that is a real problem and is exploitable for many users of the affected software. Examples include network daemon denial of service, cross-site scripting and gaining user privileges.
```
matthewpaulthomas, line 114:
```suggestion
                      <b>Medium:</b> Exploitable for many users of the affected software. Examples include network daemon denial of service, cross-site scripting and gaining user privileges.
```

templates/security/cve/_cve-table.html
```diff
@@ -98,6 +98,40 @@
...
+                    {% elif cve.priority == "low" %}
+                      <b>Low:</b> Open vulnerability that is a problem but does very little damage or is otherwise hard to exploit due to small user base or other factors such as requiring specific environment, uncommon configuration, user assistance, etc. These tend to be included in security updates only when higher priority issues require an update or if many low priority issues have built up.
```
matthewpaulthomas, line 116:
```suggestion
                      <b>Low:</b> Does very little damage or is otherwise hard to exploit, due to small user base or other factors such as requiring specific environment, uncommon configuration, or user assistance. These tend to be included in security updates only when higher priority issues require an update or if many low-priority issues have built up.
```

templates/security/cve/_cve-table.html
```diff
@@ -98,6 +98,40 @@
...
+                    {% elif cve.priority == "negligible" %}
+                      <b>Negligible:</b> Open vulnerability that may be a problem but otherwise does not impose a security risk due to various factors. Examples include when the vulnerability is only theoretical in nature, requires a very special situation, has almost no install base or does no real damage. These typically will not receive security updates unless there is an easy fix and some other issue causes an update.
```
matthewpaulthomas, line 118:
```suggestion
                      <b>Negligible:</b> May be a problem, but does not impose a security risk due to various factors. Examples include when the vulnerability is only theoretical in nature, requires a very special situation, has almost no install base or does no real damage. These typically will not receive security updates unless there is an easy fix and some other issue causes an update.
```

templates/security/cve/_cve-table.html
```diff
@@ -98,6 +98,40 @@
                   {% endif %}
                 </small>
               </div>
+              <div class="icon-container__icon icon-cve-info">
+                {% if statuses[release.codename].status != 'DNE'%}
+                  <i class="p-icon--information"></i>
+                {% endif %}
+                <small class="cve-tooltip">
```
matthewpaulthomas, line 105: If I read this correctly, it includes the popup text inside every cell. That seems like it would multiply the weight of the page, especially long search results pages (and be tiresome for text-mode browsers).

Have you considered instead generating the popup with script when the icon is clicked? Perhaps something like:
- the icon has `<a href="#key">`
- `<dl id="key">`, after the table, contains `<dt>` `<dd>` with the text for each status and priority
- if you have JavaScript, you get (a) `#key {display: none}`, because instead you have (b) a script that overrides all the `href="#key"` with popup-creating functions.

matthewpaulthomas (changes requested):

templates/security/cve/_cve-table.html
albertkol: The pop-up is indeed generated in every cell and with the help of CSS it get's displayed on hover. 
This way I avoid using javascript which I thought was the cleaner/optimal way to get this working. Also the tool-top is just plain text element. 

However I do not mind changing it if dynamically generating the content of the tool-tip on hover of the icon is the better solution. What do you think, @steverydz?

templates/security/cve/_cve-table.html
albertkol: Done.

templates/security/cve/_cve-table.html
albertkol: Done.

templates/security/cve/_cve-table.html
albertkol: Done.

templates/security/cve/_cve-table.html
albertkol: Done.

templates/security/cve/_cve-table.html
albertkol: Done.

templates/security/cve/_cve-table.html
steverydz: I think what @matthewpaulthomas is saying is right. It would be nice to do this without JS but having the text in every cell could cause problems with search, accessibility etc, so I think doing it dynamically might be the better solution. It does mean there will be no tooltip without JS but that's probably not a major concern in this case.

static/sass/_pattern_cve.scss
```diff
@@ -31,6 +40,70 @@
...
+  .cve-table .icon-container__icon {
+    position: relative;
+  }
+
+  .cve-tooltip {
+    z-index: 1;
+    background: #fff;
+    position: absolute;
+    top: 30px;
+    left: -125px;
```
steverydz, line 80: Instead of this you could do: `transform: translateX(-125px)`. It's better for in-page performance.

steverydz (approved):
Looks good. Just one suggestion in the code.

static/js/src/cve/cve.js
```diff
@@ -71,3 +71,54 @@ searchInput.addEventListener("keyup", handleSearchInput);
 attachEvents();
 handleButtons();
 disableSelectedVersions();
+
+const priorities = {
+  "critical": "<b>Critical:</b> A world-burning problem that is exploitable for most Ubuntu users. Examples include remote root privilege escalations or remote data theft.",
```
steverydz, line 76: can you use `<strong>` instead of `<b>`

static/js/src/cve/cve.js
```diff
@@ -71,3 +71,54 @@ searchInput.addEventListener("keyup", handleSearchInput);
...
+  "not-affected": "<b>Not vulnerable:</b> Packages which do not exist in the archive, are not affected by the vulnerability or have a fix applied in the archive.",
+  "pending": "<b>Pending:</b> A fix has been applied and updated packages are awaiting arrival into the archive. For example, this might be used when wider testing is requested for the updated package.",
+}
+
+const tooltipIconList = document.querySelectorAll('.cve-tooltip-icon')
+
+tooltipIconList.forEach(function(tooltipIcon) {
+  tooltipIcon.addEventListener("mouseover", function( event ) {
+    if (tooltipIcon.parentElement.querySelector(".cve-tooltip") == null) {
+      const priority = this.getAttribute('data-priority')
```
steverydz, line 94: instead of `this.getAttribute()` you can do `this.dataset.priority`

steverydz (commented):
A couple of small comments in the code

static/js/src/cve/cve.js
albertkol: Done.

static/js/src/cve/cve.js
albertkol: Done.

matthewpaulthomas (approved):

PR #8350, opened 2020-09-23
Fix bulk CVE update when published date is None

albertkol:
## Done

- When CVE's data has no published date it errors out. This fixes it.


carkod (approved):
LGTM

PR #8373, opened 2020-09-28
Fall back to shopify if the shop is not enabled

frankban:
## Done

- Fall back to shopify if the shop is not enabled, meaning there are no product listings.
- Also made associating products and listings slightly more efficient.

## QA

- Check that you are redirected to the subscribe page as usual on staging, and everything works, meaning you are able to select and buy a product.
- Then connect the app to a ua-contracts service without listings. When clicking to purchase a subscriptions you should be redirected to buy.ubuntu.com.


sowasred2012 (approved):
Works!

PR #8428, opened 2020-10-06
Advantage shop guest checkout

sowasred2012:
## Done

- Adds ability for non logged in users to make purchases
- Also removes the need for logged in users who aren't already an admin on an enterprise account to have to be manually setup as one.

## QA

- Make sure you're logged out by visiting https://login.ubuntu.com/+logout
- Run through the following [QA steps](https://canonical-web-and-design.github.io/practices/workflow/qa-steps.html
- See that the page header informs you that you may want to sign in if you have an existing account, but _do not do that_.
- Use the product selection form to select a product, add it to the cart, and click "Buy Now"
- Fill out the form with a plus variant of an email address you have access to e.g. scott.mason.nash<b>+guest</b>@canonical.com
  - You can also attempt to use your Ubuntu SSO email address. When clicking "Continue" in that case, you should see an error notification prompting you to log in.
- Use 4242 4242 4242 4242 as the card number, 04/24 as expiry, 242 as CVC, and 42424 as ZIP
- Proceed through the form, to the second stage, accept terms and conditions, then click Pay.
- You should be directed to /advantage/subscribe/thanks, and prompted to create an SSO account with the email address you filled out.
- Follow the link to Ubuntu SSO and create your account (you may need to verify your email).
- When the account has been created, visit /advantage and see that the product you purchased is there.


static/js/src/advantage/contracts-api.js
```diff
@@ -7,6 +7,25 @@ export async function getPurchase(purchaseID) {
...
+  let response = await fetch(`/advantage/purchase-account`, {
+    method: "POST",
+    cache: "no-store",
+    credentials: "include",
+    headers: {
+      Accept: "application/json",
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      email: email,
```
alnvdl, line 20: I said we didn't take fields other than this (and that is still true), but there were discussions today about accepting the customer name in this endpoint as well for guest purchases. We're waiting for feedback on whether this is really needed. Just something to have in mind as a probable future modification.

static/js/src/ua-payment-modal.js
```diff
@@ -128,12 +130,17 @@ function attachCTAevents() {
...
       const cartItems = JSON.parse(data.cart);
+      currentTransaction.type = "purchase";
 
       // make sure the product array is empty
       // before we start adding to it
       currentTransaction.products = [];
 
-      currentTransaction.type = "purchase";
+      // for guest checkout, we'll show the annual
+      // subtotal until they've added a payment method
```
alnvdl, line 140: Technically, it's the tax ID that determines the taxes.

static/js/src/ua-payment-modal.js
```diff
@@ -316,35 +323,49 @@ function applyTotals() {
...
-          }
-        );
-      }
-    })
-    .catch((error) => {
-      console.error(error);
-    });
+          });
+        } else if (currentTransaction.type === "renewal") {
+          postRenewalPreviewData(currentTransaction.transactionId).then(
```
alnvdl, line 344: The renewal preview is only valid if you are accepting renewals as purchases, which I think is not being done yet, right? In that case, I'd recommend against including it for now.

static/js/src/ua-payment-modal.js
```diff
@@ -316,35 +323,49 @@ function applyTotals() {
...
+        console.error(error);
+      });
+  } else {
+    const purchaseTotals = {
+      total: currentTransaction.subtotal,
+    };
+    guestPurchase = true;
+    modal.classList.remove("is-processing");
+
+    // set the "Country" and vatApplicable parameters
```
alnvdl, line 363: In which scenario would this happen (not having an account ID)?

alnvdl (commented):
Generally LGTM, with a few questions/opinions.

static/js/src/ua-payment-modal.js
sowasred2012: Yeah, this is a guest purchase thing only - because at this point I don't have enough information to send to the purchase preview endpoint and get an actual VAT amount back, we just use the unit price * number of machines from the form to give a subtotal.

static/js/src/ua-payment-modal.js
sowasred2012: Ah OK - I'll update

webapp/views.py
```diff
@@ -590,24 +594,27 @@ def post_advantage_subscriptions(preview):
...
-            extra={"payload": payload}
-        )
-        return (
-            flask.jsonify(
-                {"error": "could not retrieve account subscriptions"}
-            ),
-            500,
-        )
-    except KeyError:
+    if not guest_token:
```
frankban, line 597: What do you think about structuting this as follows
```python
last_subscription = {}
if not guest_token:
    try:
        subscriptions = ...
    except HTTPError:
        ...
    if subscriptions:
        last_subscription = subscriptions[0]
```
and dropping the else block?

webapp/views.py
```diff
@@ -762,24 +763,21 @@ def advantage_shop_view():
...
         stripe_publishable_key=stripe_publishable_key,
-        account=account,
+        purchase_account_id=purchase_account_id,
         previous_purchase_id=previous_purchase_id,
     )
 
 
-def get_admin_accounts(accounts):
-    admin_accounts = []
+def advantage_thanks_view():
```
frankban, line 774: Add a docstring?

webapp/views.py
```diff
@@ -934,6 +944,38 @@ def get_purchase(purchase_id):
...
+    if user_info(flask.session):
+        auth_token = flask.session["authentication_token"]
+
+    advantage = AdvantageContracts(
+        session,
+        auth_token,
+        api_url=flask.current_app.config["CONTRACTS_API_URL"],
+    )
+
+    if not flask.request.is_json:
```
frankban, line 959: Should we do this check at the beginning of the function?

webapp/views.py
```diff
@@ -934,6 +944,38 @@ def get_purchase(purchase_id):
...
+    email = flask.request.json.get("email")
+    payment_method_id = flask.request.json.get("payment_method_id")
+
+    try:
+        account = advantage.get_purchase_account(email, payment_method_id)
+
+        flask.session["guest_authentication_token"] = account["token"]
+    except HTTPError as http_error:
+        flask.current_app.extensions["sentry"].captureException()
+        return (
```
frankban, line 971: As mentioned, this could not be always an actual 500. SOmetimes we should just suggest users to log in instead, because you can only purchase as a guest if the email that you claim is not known by ua-contracts.

frankban (approved):
Looks good with some changes, thanks!

static/js/src/ua-payment-modal.js
```diff
@@ -316,35 +323,49 @@ function applyTotals() {
...
+              setOrderTotals(country, vatApplicable, purchasePreview, modal);
+            }
+          );
+        }
+      })
+      .catch((error) => {
+        console.error(error);
+      });
+  } else {
+    const purchaseTotals = {
```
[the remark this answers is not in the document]
anthonydillon: Could the contents of each conditional be split into separate functions to be easier to read?

webapp/app.py
```diff
@@ -150,6 +152,9 @@ def utility_processor():
 # Simple routes
 app.add_url_rule("/advantage", view_func=advantage_view)
 app.add_url_rule("/advantage/subscribe", view_func=advantage_shop_view)
+app.add_url_rule(
+    "/advantage/subscribe/thanks", view_func=advantage_thanks_view
```
anthonydillon, line 156: We normally use thank-you instead of thanks. Or would confirmation be a better URL?

sowasred2012: @matthewpaulthomas what do you think?

static/js/src/ua-product-selector.js
```diff
@@ -372,20 +372,16 @@ function productSelector() {
...
-    if (window.accountId) {
-      steps.forEach((step) => {
-        if (stepsToEnable === undefined) {
-          if (!state.get(step)[0]) {
-            stepsToEnable = steps.slice(0, i + 1);
-            stepsToDisable = steps.slice(i + 1);
-          } else if (i < steps.length) {
-            i++;
-          }
+    steps.forEach((step) => {
```
steverydz, line 375: can you not use the index from the `forEach` loop? `steps.forEach((step, i) => {`

steverydz (approved):
LGTM :+1: 

Just one small comment in the code

static/js/src/ua-product-selector.js
sowasred2012: yes I can!

webapp/app.py
sowasred2012: Updated to /thank-you

templates/advantage/subscribe/thank-you.html
```diff
@@ -0,0 +1,25 @@
...
+{% block head_extra %}<meta name="robots" content="noindex" />{% endblock %}
+
+{% block content %}
+  <section class="p-strip--suru-topped">
+    <div class="row">
+      <div class="col-12 u-sv3">
+        <h1>Thanks for your purchase</h1>  
+
+        {% if email %}
+          <p class="p-heading--four">We've sent your invoice to {{ email }}</p>
```
matthewpaulthomas, line 15:
```suggestion
          <p class="p-heading--four">We’ve sent your invoice to {{ email }}</p>
```
Is there a reason for this paragraph, and the next one, to be `p-heading--four`?

matthewpaulthomas (changes requested):
- It’s a bit weird (though not a blocker) that if I do guest checkout from a country with VAT, “VAT: …” and “Total: …” rows appear after I’ve submitted the form, only to disappear a moment later.

- [x] On paying, I was sent to https://ubuntu-com-8428.demos.haus/advantage/subscribe/thanks?email=mpt%2Btest20201009%40canonical.com but this returned a 404 error.

matthewpaulthomas:
- [ ] I am not returned to /advantage, to see my new subscription, after signing in _or_ registering.
  - If I register a new account with the address I provided, I’m marooned at https://login.ubuntu.com/.
  - If I sign in to my existing account instead, I’m marooned at https://login.ubuntu.com/?return_to=https://ubuntu-com-8428.demos.haus/advantage&_ga=… (same page, fluffier URL).

sowasred2012:
@matthewpaulthomas that's fixed now, you should be returned in both cases to /advantage.

wgx:
Some UX issues / bugs:

1. The spinner that appears in the the lower left is too subtle - the spinner should be larger, in the center and should 'block' input to prevent the user interacting with the view while payment details are submitted to Stripe.

2. If the use cancels the modal and returns to the 'basket' - all their details are cleared, I would expect payment details to clear (PCI DSS compliance!) but not the email, address, etc - could we retain these?

3. Guest checkout: If the user submits payment and email details, THEN cancels and re-tries, the user cannot progress a second time because their email is in use (they get a red error notification) - could we *not* create their account until the end of the process, as this would block a guest user from completing checkout

wgx:
> So this is a facet of presenting the user with a VAT total. In order to show a guest user a VAT total, we need to make a "simulated" purchase via the contracts API. The endpoint for this requires us to send along an account ID and a snapshot of their cart, so before making this call, we call another endpoint to create an account for the guest user, which returns the account ID.

Would it be possible for the VAT endpoint to not require an account ID? 🤔

sowasred2012:
> > So this is a facet of presenting the user with a VAT total. In order to show a guest user a VAT total, we need to make a "simulated" purchase via the contracts API. The endpoint for this requires us to send along an account ID and a snapshot of their cart, so before making this call, we call another endpoint to create an account for the guest user, which returns the account ID.
> 
> Would it be possible for the VAT endpoint to not require an account ID? thinking

@frankban @alnvdl ^ - any thoughts?

wgx:
Fixed! 👍🏼

PR #8462, opened 2020-10-13
Remove typos from the base index template

m-czernek:
When randomly browsing the [W3C validator](https://validator.w3.org/nu/?doc=https%3A%2F%2Fubuntu.com%2F), I noticed there were a few typos on the Ubuntu website. This PR fixes the most obvious `clas` typo, and based on other image rows, I removed the redundant div.

WDYT?

### DIFF
--- a/templates/base_index.html
+++ b/templates/base_index.html
@@ -1348,7 +1348,7 @@
     </div>
   </div>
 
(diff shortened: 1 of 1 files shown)


tbille (approved):
Good catch :smile:

PR #8496, opened 2020-10-15
Rework MicroStack docs to use Secure Clustering

dshcherb:
As of revision 215 in the store, MicroStack contains the Secure
Clustering feature:
https://review.opendev.org/#/c/757658/

The workflow no longer includes a cluster password and instead relies on
a connection string (more details in the opendev review).

It is also mandatory to specify the node role via command line when
using the --auto mode too.

Passwords are now randomly generated to avoid insecure practices which
is also reflected in the documentation:
https://review.opendev.org/#/c/756567/

### DIFF
--- a/templates/openstack/install.html
+++ b/templates/openstack/install.html
@@ -155,7 +155,7 @@
         <div class="p-stepped-list__content">
           <p>Before using your OpenStack installation, it has to be initialised, so that networks and databases get configured. In order to do so, run:</p>
 
-<pre><code>$ sudo microstack.init --auto
+<pre><code>$ sudo microstack init --auto --control
 ...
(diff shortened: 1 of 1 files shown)


SirSamTumless:
@pmatulis Can you confirm this work is needed if we're moving to a Discourse docs model?

dshcherb:
@SirSamTumless This is quite urgent - not sure what the timeline is to move to docs on Discourse.

pmahnke:
@dshcherb and @pmatulis this might be superceeded by #8512 ? can you confirm?

pmatulis:
@pmahnke Yes, this PR has been superseded.

Issue #8394, opened 2020-09-30
Small screens is unusable 


anthonydillon:
On small screens the table of subscriptions is unusable:
![image](https://user-images.githubusercontent.com/1413534/94692745-ce532f80-032a-11eb-8aab-8bc80b16087c.png)

sowasred2012:
@wgx @matthewpaulthomas @meralonso - wanted to get opinions on how we can solve this; there's so much data, and content - do we have an existing pattern in any of our projects that would suit?

meralonso:
This table has always been a bit of a pain @sowasred2012 - we should redesign it at some point @matthewpaulthomas @wgx 

As a temporary solution I'd suggest to not compress the table on mobile, keep the desktop size and scroll horizontally. I've seen this practice on other websites i.e https://stamped.io/pricing - please check desktop and mobile versions

matthewpaulthomas:
A couple of small possible improvements:
- Make the subscription name and plan type cells wrap their text, rather than ellipsizing.
- [Make the table headers diagonal](https://css-tricks.com/rotated-table-column-headers-now-with-fewer-magic-numbers/) at small widths.

> keep the desktop size and scroll horizontally

Could we keep the table’s first column fixed, while all the other columns scroll horizontally? (Like a “frozen” column in a spreadsheet.) Maybe that’s beyond CSS’s power at the moment.

anthonydillon:
You could try the Vanilla responsive solution? https://vanillaframework.io/docs/base/tables#responsive

anthonydillon:
This has been merged.

Issue #7718, opened 2020-06-11
/advantage: List of UA services doesn’t include Landscape


matthewpaulthomas:
All Ubuntu Advantage subscribers get access to Landscape. When they subscribe, they receive a message saying “If you are a new customer, you will receive emails from noreply@canonical.com with an invitation to become an administrator of your Landscape account…”

This makes sense only if you know what Landscape is. Currently buy.ubuntu.com mentions Landscape, but /advantage does not. If this is not changed, once buy.ubuntu.com is retired, it will be possible for someone to buy UA and then be confused by the mention of Landscape.

---
*Reported from: https://ubuntu.com/advantage*

pmahnke:
@matthewpaulthomas is this a reminder for you to do this work?

SirSamTumless:
@lechsandecki Can you check to see if you agree with this and if you do, update the copy doc.

matthewpaulthomas:
@pmahnke I’ve expanded the report to explain the problem further.

lechsandecki:
yes, I agree.  thanks for pointing that out
Looking at the copy doc - it seems that this job would be more complex than
just a doc update:
https://docs.google.com/document/d/1AW8wQu1Uvq0xFtd_c2JwNdic075MPOTx7k2KHqFPPeo/edit#heading=h.2p4jyl34ww3l

wgx:
Easy part: adding Landscape to the logged-out comparison table: https://docs.google.com/document/d/1AW8wQu1Uvq0xFtd_c2JwNdic075MPOTx7k2KHqFPPeo/edit?disco=AAAAGqmgBe8

wgx:
Less easy part: add Landscape to the table of subscriptions, and, if not installed, show this help panel:

![image](https://user-images.githubusercontent.com/930099/85699740-af158800-b6d3-11ea-8373-6ad511cccfd2.png)

matthewpaulthomas:
This is a good start. A small issue is that it’s quite different from the layout of the other expanded cells. Probably it could be more similar.

A much bigger issue is that currently, from /advantage, we have no way of telling whether you have Landscape installed. (I imagine it would be pretty hard to implement, given that the contracts service doesn’t manage Landscape.) And even if we did know, you might have it installed on some machines and not others. So we probably shouldn’t provide an imperative instruction as if you don’t have it installed.

I think there are three things we should help you do here:
[...]

Issue #7137, opened 2020-04-05
/advantage: Tech support isn’t indicated/offered anywhere


matthewpaulthomas:
When you have an Ubuntu Advantage subscription, nothing on [ubuntu.com/advantage](https://ubuntu.com/advantage) indicates whether that subscription includes tech support.

And if it does include tech support, there is no indication of how to access it.

[Reported by Maryan Dahir.]

carkod:
The new advantage page has a support portal.

PR #8532, opened 2020-10-20
Improve the small screen view of the subscriptions

anthonydillon:
## Done
Provide an entirely new UI for small screens.

## QA
- Login
- Shrink the screen to a small screen or better yet use your phone
- Is there that information is accessible

## Issue / Card
Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/8394


pmahnke (approved):
+1

matthewpaulthomas:
This is an impressive start, thanks @anthonydillon.

One big regression, though: there’s no longer any way to access the details for ESM, Livepatch, or Certifications, because there’s no equivalent to the collapsed/expanded cells. Chevrons to reveal the details could appear to the right of the data values, just as they do in the tabular view.

anthonydillon:
@matthewpaulthomas thanks, I was trying to think of a nice way to display the info and commands but have been torn between displaying the command in each subscription in this viewport. There is a lot of information and commands under each affordance. Therefore I was thinking of providing a link to the manpage where all this information is and explained.

matthewpaulthomas:
How about this? Retaining the chevrons for expanding/collapsing, just with rows rather than cells.

mobile

anthonydillon:
@matthewpaulthomas I have used the accordion pattern to hide and show the information for each affordance.

@sowasred2012 they are sharing partials to help to maintain the panel info.

**Closed**
image

**Open**
image

sowasred2012:
@anthonydillon oh nice - I was just going through the Vanilla docs looking for a solution for this! Reviewing.

templates/advantage/index.html
```diff
@@ -401,6 +517,72 @@ <h2>Plans for enterprise use</h2>
   }
 </script>
 
+
+<script>
```
sowasred2012, line 521: This script is in two places now (the other is /openstack/compare) - I think it'd be worth adding it to /js/utils and including a link to it on both pages.

sowasred2012 (changes requested):
LGTM - tested it with an active renewal too, it works. Just one comment about reducing some repetition.

templates/advantage/index.html
anthonydillon: Good point, done

sowasred2012 (approved):
LGTM

templates/advantage/index.html
```diff
@@ -58,7 +58,170 @@ <h2>Your paid subscriptions</h2>
...
+              <p>To detach a machine: <code>sudo ua detach</code></p>
+              <p>
+                Once a machine is attached, you can configure its services. Visit the <a href="http://manpages.ubuntu.com/manpages/focal/en/man1/ua.1.html" class="p-link--external">UA manpages</a> for details.
+              </p>
+              {% else %}
+              <div class="col-12 u-align--center u-sv3">
+                To attach a machine: <br /><code>sudo ua attach {{ contract['token'] }}</code>
+              </div>
+              {% endif %}
+            </div>
```
matthewpaulthomas, line 92: In the large-screen view:
- the attach instruction is hidden by default (except for a new subscription)
- the detach instruction is hidden by default (ditto).

But in this mobile view:
- the attach instruction is _always_ shown, with no ability to hide it
- the detach instruction is _never_ shown, with no ability to show it!

matthewpaulthomas (commented):
UX +1 because this is a massive improvement, but it could be improved further.

> @matthewpaulthomas I have used the accordion pattern to hide and show the information for each affordance.

Well, it does have the advantage that it’s a Vanilla pattern. However:
- It looks quite odd to have a column of “+” symbols that are — at least on a narrow tablet or a phone in landscape mode — far away from the text of the things they’re indicating the expandability of. (Whereas with chevrons, we’re used to that kind of distance, from wide radio menus.) I suspect this will get Design -1 for that.
- It makes the interaction difference bigger than it needs to be if chevrons were used for both.
- It introduces inconsistency in the text itself…

templates/advantage/index.html
```diff
@@ -58,7 +58,170 @@ <h2>Your paid subscriptions</h2>
...
+            </h5>
+            <hr class="u-sv1">
+            <div class="row p-card__content">
+              {% if new_subscription_id == contract['contractInfo']['id'] %}
+              <h4>Welcome to your new subscription.</h4>
+              <p>To get started, first attach each machine:</p>
+              <p>To attach a machine: <code>sudo ua attach {{ contract['token'] }}</code></p>
+              <p>To detach a machine: <code>sudo ua detach</code></p>
+              <p>
+                Once a machine is attached, you can configure its services. Visit the <a href="http://manpages.ubuntu.com/manpages/focal/en/man1/ua.1.html" class="p-link--external">UA manpages</a> for details.
```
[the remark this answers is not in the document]
anthonydillon: Now that we have the information in the hidden sections I have removed the man page link and added the Where indicators to small screens too.

templates/advantage/index.html
anthonydillon: Would you prefer all info to be hidden by default as the large-screen view does?

matthewpaulthomas: By default, yes, with the ability to reveal it. This would involve “Name” switching from a header to an expandable row, and the “Attached” row becoming expandable too (both shown in the wireframe).

long-chung:
Alignment issues of the titles on my iPhone see screenshot attached
Advantage

anthonydillon:
Oh strange I don't see that on Android chrome. Are you using Safari?

long-chung:
Yep. iPhone Safari

matthewpaulthomas:
From Product Design Review this morning:
- @danielmutis reminded me that on mobile, you wouldn’t get the grey background on mouseover, so there’d be even less indication that the “+” and the label were a single clickable area. He suggested maybe lines drawn between the rows instead.
- @cassiocassio wondered if needing a different presentation for small screens (plus the need to cram more into the table with #7137 and #7718) should nudge us towards using a different presentation regardless of screen size.
- @danielmutis went a step further than my request to hide the attach instructions by default, suggesting hiding everything about a subscription by default except for its name.
- @spencerbygraves mentioned that we now have [new icons for expand and collapse](https://vanillaframework.io/docs/patterns/icons). He wondered whether (a) the accordion pattern should use those icons, and (b) `p-icon--plus` and `p-icon--minus` should be renamed to `p-icon--add` and `p-icon--remove`. I guess neither involves changes here — we’d pick up any changes from Vanilla automatically.

Meanwhile, we now have the “Where?” button and animations in the small-screen view — thanks! — but:
- the animations are no longer working in the large-screen view
- there’s no animation for the Certifications row
- the animations are highlighting nothing in particular, because the chevrons aren’t there.

nowhere

anthonydillon:
In light of these review points, I believe the best course of action would be to close this PR and address small screens design sometime in the future?

pmahnke:
@matthewpaulthomas @spencerbygraves why can't we go with this and improve later?  obviously fix the safari bug?

long-chung:
Agree with @pmahnke at least its usable now on mobile and we can revisit this at a later date as @anthonydillon mentioned.

matthewpaulthomas:
Given how bad the existing UI is on small screens, I understand having a lower standard for what counts as a blocker for this PR. Personally I think only two of the current issues qualify:
- [ ] the large-screen “Where?” button no longer working at all
- [ ] the small-screen “Where?” button highlighting nothing in particular.

The rest can certainly wait, given that #8394 is arguably not a shop release blocker anyway.

spencerbygraves:
>@danielmutis reminded me that on mobile, you wouldn’t get the grey background on mouseover, so there’d be even less indication that the “+” and the label were a single clickable area. He suggested maybe lines drawn between the rows instead.

@anthonydillon [the accordion pattern](https://vanillaframework.io/docs/examples/patterns/accordion/default) has dividing lines, they seem to be missing here.

>@matthewpaulthomas @spencerbygraves why can't we go with this and improve later? obviously fix the safari bug?

I'm happy to go with this including the bug fix and improve later.

anthonydillon:
@long-chung safari issue resolved
image

anthonydillon:
@spencerbygraves Good spot, added the divider lines:
image

anthonydillon:
@matthewpaulthomas Do you still think this "Where" on small screens is a blocker? What would you do differently? Highlight the "+"
Peek 2020-10-30 12-22

matthewpaulthomas:
> @matthewpaulthomas Do you still think this "Where" on small screens is a blocker?

Unfortunately yes, highlighting empty space just looks broken.

> What would you do differently? Highlight the "+"

Exactly. Alternatively, temporarily drop the button to get this merged.

anthonydillon:
@matthewpaulthomas done and it has fixed the large screen Where functionality

matthewpaulthomas:
Retested, works well, thanks for your hard work on this @anthonydillon

PR #8549, opened 2020-10-21
Fix path for download mirror list

nottrobin:
Download mirrors are broken because the path to the mirrors file is incorrect.

This bug was introduced in https://github.com/canonical-web-and-design/ubuntu.com/pull/8073, and brought to our attention in https://rt.ubuntu.com/Ticket/Display.html?id=34579.

## QA

On the demo, go to /download/desktop, download a version of Ubuntu, see that you download from a mirror rather than releases.ubuntu.com.

### DIFF
--- a/scripts/create-usns.py
+++ b/scripts/create-usns.py
@@ -12,13 +12,13 @@ from macaroonbakery import httpbakery
 
 
 def guess_binary_links(binary, version, sources):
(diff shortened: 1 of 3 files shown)


pmahnke (approved):
+1, however mirros.melbourn.co.uk's cert is old.

PR #8550, opened 2020-10-21
advantage guest purchases: provide account/comany name

frankban:
## Done
- When purchasing without an account, allow specifying an account/company name.
- When already having an account, prevent the user from providing a different one.
- Implemented an API client method for getting an account for purchasing if it exists, without creating it.
- Improved test coverage around UA API client.

## QA
- Purchase as a guest, as a new user (with no UA accounts) and as an existing UA user.
- When not having an account, you should be able to set an account name.
- When returning later, the account name should be there.
- Now try a renewal, just to check that the stripe form is still working for that use case as well.
- Bonus: when purchasing again in the same day, the price is not prorated by the second anymore (not a fix in this branch, but nice to double check that's working).

## Issue / Card
Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/8522

### DIFF
--- a/templates/advantage/_stripe-modal.html
+++ b/templates/advantage/_stripe-modal.html
@@ -48,7 +48,7 @@
           {% if account is defined %}
             <div class="p-form__group p-form-validation row u-no-padding u-vertically-center">
               <div class="col-3">
-                <label class="u-no-padding--top" for="name">Account name:</label>
+                <label class="u-no-padding--top" for="name">{% if account %}Account name{% else %}Choose account name{% endif %}:</label>
               </div>
 
               <div class="col-9">
@@ -57,7 +57,7 @@
 
               {% if not account %}
                 <div class="col-9 col-start-large-4">
-                  <small>For example, your organization ot team name</small>
+                  <small>Your name, or the name of your organisation.</small>
                 </div>
               {% endif %}
 
--- a/templates/advantage/_stripe-modal.html
+++ b/templates/advantage/_stripe-modal.html
@@ -16,19 +16,6 @@
         </div>
 
         <form class="p-form p-form--stacked u-sv3" id="details-form">
-          <div class="p-form__group p-form-validation row u-no-padding u-vertically-center">
-            <div class="col-3">
-              <label class="u-no-padding--top" for="email">Email my receipt to:</label>
-            </div>
-
-            <div class="col-9">
-              <input class="p-form-validation__input" name="email" type="email" required value="{{ user_info.email }}" />
-            </div>
-
-            <div class="col-9 col-start-large-4">
-              <span class="p-form-validation__message u-hide"></span>
-            </div>
-          </div>
 
           <div class="p-form__group row u-no-padding u-vertically-center">
             <div class="col-3">
@@ -46,17 +33,39 @@
 
           <div class="p-form__group p-form-validation row u-no-padding u-vertically-center">
             <div class="col-3">
-              <label class="u-no-padding--top" for="name">Name on card:</label>
+              <label class="u-no-padding--top" for="email">Email my receipt to:</label>
             </div>
 
             <div class="col-9">
-              <input class="p-form-validation__input" name="name" type="text" required value="{{ user_info.fullname }}" />
+              <input class="p-form-validation__input" name="email" type="email" required value="{{ user_info.email }}" />
             </div>
 
             <div class="col-9 col-start-large-4">
               <span class="p-form-validation__message u-hide"></span>
             </div>
           </div>
+
+          {% if account is defined %}
+            <div class="p-form__group p-form-validation row u-no-padding u-vertically-center">
+              <div class="col-3">
+                <label class="u-no-padding--top" for="name">Account name:</label>
+              </div>
+
+              <div class="col-9">
+                <input class="p-form-validation__input" name="name" type="text" required value="{{ account.name }}"{% if account %}disabled{% endif %} />
+              </div>
+
+              {% if not account %}
+                <div class="col-9 col-start-large-4">
+                  <small>For example, your organization ot team name</small>
+                </div>
+              {% endif %}
+
+              <div class="col-9 col-start-large-4">
+                <span class="p-form-validation__message u-hide"></span>
+              </div>
+            </div>
+          {% endif %}
 
           <div class="p-form__group p-form-validation row u-no-padding">
             <div class="col-3">
@@ -109,6 +118,7 @@
               <small>e.g. GB101111, CY99999999L</small>
             </div>
           </div>
+
         </form>
 
         <div id="order-totals" class="u-hide u-sv1">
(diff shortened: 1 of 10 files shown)


matthewpaulthomas:
@anthonydillon Could you launch a demo for this PR please? I tried but [failed](https://github.com/canonical-web-and-design/demos.haus/issues/19).

anthonydillon:
Sure, here you go: https://ubuntu-com-8550.demos.haus/

tests/test_advantage.py
```diff
@@ -0,0 +1,173 @@
...
+    def __init__(self, resp: Response):
+        self._resp = resp
+        self.request_kwargs = None
+
+    def request(self, **kwargs):
+        self.request_kwargs = kwargs
+        return self._resp
+
+
+def make_client(session: Session):
```
nottrobin, line 35: The mock tooling you've written here is nice and elegant.

After trying a few different things, our most common attitude to mocking API calls like this is now to use real API URLs, and then use [VCR](https://pypi.org/project/vcrpy/1.5.2/) to save the responses, so future calls are mocked versions of the real API response from the time of the snapshot. We find that this gives us the benefit of quick mock responses while also giving us the opportunity to catch errors relating to subtle changes in the API responses.

Do you think that could work here? Or do you think there are clear reasons why your approach here is better?

frankban: Never used vcrpy and it seems neat.
There are pros and cons of using a tool like that. On the bright side, once in a while you can actually execute a real API call, therefore checking that the client code works, and that the API is stable enough.
But some problems come to mind:
- how simple is to simulate error cases? how do you make the API server return an error, to check that the client deals with that correctly?
- how simple is to log in interactively in case the API is authenticated, for instance against USSO?
- did we review the code in that library? Can we trust that it does not have bugs that lead to always making HTTP requests, or to leaking secrets, in case auth is required?
- is it worth introducing a dependency when the actual mock code is more or less 30 lines, and the actual benefit is partially overruled by QA exercises, that needs to be done anyway?

templates/advantage/_stripe-modal.html
```diff
@@ -46,18 +33,40 @@ <h2 class="p-modal__title js-contract-name u-no-margin--bottom" id="modal-title"
...
 
             <div class="col-9 col-start-large-4">
               <span class="p-form-validation__message u-hide"></span>
             </div>
           </div>
 
+          {% if account is defined %}
+            <div class="p-form__group p-form-validation row u-no-padding u-vertically-center">
+              <div class="col-3">
+                <label class="u-no-padding--top" for="name">Account name:</label>
```
matthewpaulthomas, line 51: Regardless of any future layout changes, I think “Account name:” here needs to be “Choose an account name:”. Otherwise, no matter what caption we used, it would look like something you’re supposed to know already when you don’t.

Unfortunately, in the Ubuntu font, “Choose an account name:” doesn’t fit inside               `<div class="col-3">`. I don’t know why we don’t use `<table>` for this kind of thing, but from quick experimenting I think the dialog would look fine with all the `col-3` changed to `col-4` and all the `col-9` changed to `col-8`.

templates/advantage/_stripe-modal.html
```diff
@@ -46,18 +33,40 @@ <h2 class="p-modal__title js-contract-name u-no-margin--bottom" id="modal-title"
...
+              </div>
+
+              <div class="col-9">
+                <input class="p-form-validation__input" name="name" type="text" required value="{{ account.name }}"{% if account %}disabled{% endif %} />
+              </div>
+
+              {% if not account %}
+                <div class="col-9 col-start-large-4">
+                  <small>For example, your organization ot team name</small>
```
matthewpaulthomas, line 60:
```suggestion
                  <small>Your name, or the name of your organisation.</small>
```
(Canonical uses UK spelling, so we need to say “organisation”.)

matthewpaulthomas (changes requested):

templates/advantage/_stripe-modal.html
frankban: Done, "Choose account name" fits even without the column changes. "Account name" is shown if the account is already set.

templates/advantage/_stripe-modal.html
frankban: Done.

matthewpaulthomas:
From discussion with @cassiocassio:

It’s not clear why we need an “account name” at all. Granted that “The email is not enough and not correct for naming accounts”. But why would a user-provided, non-unique string be any better? If J. Random Customer can claim to be, say, “Pan American World Airways” — whether or not we already have an account with that name — then what purpose does it serve?

To put it another way, what would be the drawbacks of using an auto-generated number from 10000 to 99999 instead of a user-provided string?

matthewpaulthomas (approved):

frankban:
@matthewpaulthomas, the question can be answered from two different perspectives:
- users: this is the name of their account in our system, shown as part of the ua cli experience, and in the future as part of account management (to be implemented). A sensible name is better for organizing company accounts, compared to autogenerated identifiers (that are there anyway).
- salesops: they explicitly requested this feature, possibly because they don't want to deal with practically anonymous accounts in Salesforce.

tests/test_advantage.py
nottrobin: I can respond to some of those:

> how simple is to simulate error cases? how do you make the API server return an error, to check that the client deals with that correctly?

I don't think it provides anything along these lines, so if you're not able to come up with a real API call that produces the error you want to test, then you should probably bypass vcrpy in that case and manually create a test error case.

> how simple is to log in interactively in case the API is authenticated, for instance against USSO?

Not trivial. I think we've come up with workaround for this in the past... @jkfran might know more?

> did we review the code in that library? Can we trust that it does not have bugs that lead to always making HTTP requests, or to leaking secrets, in case auth is required?

No we haven't specifically reviewed the code (although we've read some of it as needed), just the same as for all the other dependencies of ubuntu.com. We know that it doesn't always make HTTP calls, as that would have been pretty obvious. It does nothing automatic to prevent leaking of secrets, but it lets you strip specific headers, so that's what we've done when necessary. But the onus is on the developer.

> is it worth introducing a dependency when the actual mock code is more or less 30 lines, and the actual benefit is partially overruled by QA exercises, that needs to be done anyway?

I think the dependency concern only has much weight if it's literally the only dependency needed for our test suite. Even if this is true now, I doubt it's going to be for much longer. The QA point is a general argument against testing, but the general wisdom is that the existence of QA does not excuse your tests from being thorough.

frankban: Yeah I was not using the QA argument against testing in general, of course.
That's why this branch introduces unit tests for the ua contracts API client.
I am just not sure that vcrpy would make our tests much more thorough in this specific case, in which we want to check both happy and unhappy cases, and all requests are authenticated.
I am also not sure about the example you linked: when recording, and real HTTP requests are made, those test cases check that the target service is up and running and all pages return 200. But what is actually tested when not doing real requests?

anthonydillon:
@sowasred2012 assigning Scott now he is back

sowasred2012 (approved):
Code and QA +1 - tested on guest and logged in purchases, and renewals.

Issue #8392, opened 2020-09-30
Pricing seem wrong


anthonydillon:
Even though I am sure there is a just reason for the difference it is jarring and confusing for users.
![image](https://user-images.githubusercontent.com/1413534/94692099-0dcd4c00-032a-11eb-8484-0bab78aa464d.png)

10 × US$25/year = $250
Subtotal: $235.53

matthewpaulthomas:
Thanks @anthonydillon. “The same date as your existing annual subscription” is meant to convey that you aren’t buying a full year. Instead, you’re being charged for the period from the moment you purchased, on 30 September 2020, to the moment your existing subscription ends on 9 September 2021.

9 September 2021 – 30 September 2020 = 344 days
$235.53 / (10 × $25) × 365 days = 343.1 days

(The difference between 344 and 343.1 is because you’re being charged by the second rather than by the day, which is [a separate problem](https://docs.google.com/document/d/1kRqawNqi3ISXPqGohaAyu-QQEJlrGZiVv-Bn8bQfago/edit?disco=AAAAKWWt1v4).)

One way to make this clearer might be changing
> Subtotal: $235.53

to

> For this period: $235.53

What do you think?

anthonydillon:
I think the real issue here is that I believed I was purchasing a brand new subscription and in that case would expect the support to cover the full year from today. 

I assume this has something to do with an existing contract but I am not expecting this subscription to be add additional quantity to an existing contract but be entirely separate.

pmahnke:
Perhaps the message could be more like 'Prorated to end on the same date as
your existing subscription'

However, what if there was 1 more day on the existing sub?  I that would be
odd, pay £0.68...

matthewpaulthomas:
> I am not expecting this subscription to be add additional quantity to an existing contract but be entirely separate.

Agreed. However, it’s a limitation of the contracts service that each customer can have only one subscription per level×machinetype combination. For example, you can have a UA Infra Standard Desktop subscription, and a UA Infra Advanced Desktop subscription, and a UA Infra Virtual Server Advanced subscription, happily separate … but if you then try to buy another UA Infra Advanced Desktop subscription, instead of becoming your fourth subscription, it’s added to your second one. (I can see how this would be useful for slowly growing a fleet over time, but it’s weird for subscriptions that might be for different teams in a large organisation. Hopefully we can make enlarging a separate feature eventually.)

[...]

PR #8567, opened 2020-10-22
Use “For this period:” for subtotal when purchase is prorated

matthewpaulthomas:
## Done

- Changed “Subtotal:” to “For this period:” when a purchase is prorated.

## QA

- With at least one existing subscription, select another and choose “Buy now”

## Disclaimer

- I have not tested this, and there are probably better ways to implement it.
- It seems brittle to me that the way to select the label is with the `u-text-light` class. I was expecting it to be something more semantic, like a `<th>`.

## Issue / Card

Addresses #8392

### DIFF
--- a/static/js/src/advantage/set-modal-info.js
+++ b/static/js/src/advantage/set-modal-info.js
@@ -225,7 +225,7 @@ export function setOrderInformation(listings, modal) {
 export function setOrderTotals(country, vatApplicable, purchasePreview, modal) {
   const currency = "USD";
   const totalsContainer = modal.querySelector("#order-totals");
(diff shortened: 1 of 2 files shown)


anthonydillon (approved):
This looks good to me :+1:

anthonydillon:
This just doesn't seem right. The dates are a year apart but the math doesn't add up:
image

sowasred2012:
I've updated this with a bugfix for https://github.com/canonical-web-and-design/ubuntu.com/issues/8533, which also addresses your comment @anthonydillon

anthonydillon (approved):
Code and QA :+1:

PR #8608, opened 2020-10-27
Allow missing form-id in engage pages

### LINKED ISSUE #8607: 20.10 engage page - change whitepaper form to webinar
## Summary

https://ubuntu.com/engage/ubuntu-20.10 page which is being linked from the takeover has a form to download a whitepaper and is sending to a 404 page instead of being a link to this BrightTalk webinar https://www.brighttalk.com/webcast/6793/448578

pmahnke:
## Done

- For engage pages, you had to have a blank form-id to not present a form. It would print a broken form if you skipped the value, so I added a check for the existence as well.

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/engage
- See that the form works for a webinar and for a whitepaper, see that the logic makes sense

Would have fixed #8607

### DIFF
--- a/templates/engage/base.html
+++ b/templates/engage/base.html
@@ -82,7 +82,7 @@
       {{ text | safe }}
       {% endfor %}
     </div>
-    {% if document["metadata"]["form_id"] != "" %}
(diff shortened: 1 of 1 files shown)


carkod (approved):
Thanks. LGTM

Issue #8626, opened 2020-10-28
/advantage 500 - bad key lookup on renewal


ajbeattie:
This may be the result of a weird account state on the back end, but nevertheless, when logging into /advantage with my sso account, I get:

```
Traceback (most recent call last):
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/app.py", line 2450, in wsgi_app
    response = self.handle_exception(e)
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/app.py", line 1867, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/app.py", line 2447, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/app.py", line 1952, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/app.py", line 1821, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/_compat.py", line 39, in reraise
    raise value
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/multipass/code/ubuntu.com/.venv/lib/python3.6/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/multipass/code/ubuntu.com/webapp/views.py", line 535, in advantage_view
    advantage, contract["contractInfo"]
  File "/home/multipass/code/ubuntu.com/webapp/views.py", line 818, in make_renewal
    if renewal["status"] == "processing":
KeyError: 'status'
```

The `make_renewal` method in question ([here](https://github.com/canonical-web-and-design/ubuntu.com/blob/advantage-shop/webapp/views.py#L790)) accesses several fields on renewals which are not guaranteed to exist.

[...]

PR #8627, opened 2020-10-28
Don't 500 on invalid renewal

ajbeattie:
See #8626 

This will handle invalid renewal state (or invalid parsing) gracefully and send an exception sentry instead of displaying a 500 page on `/advantage` in this case.

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -531,9 +531,15 @@ def advantage_view():
                             "daysTillExpiry"
                         ] = date_difference.days
 
-                    contract["renewal"] = make_renewal(
-                        advantage, contract["contractInfo"]
-                    )
+                    try:
+                        contract["renewal"] = make_renewal(
+                            advantage, contract["contractInfo"]
+                        )
+                    except KeyError:
+                        flask.current_app.extensions[
+                            "sentry"
+                        ].captureException()
+                        contract["renewal"] = None
 
                     enterprise_contract = enterprise_contracts.setdefault(
                         contract["accountInfo"]["name"], []


jaycee (approved):
LGTM

webapp/views.py
```diff
@@ -795,10 +795,22 @@ def make_renewal(advantage, contract_info):
...
+    # Check for the status key before returning a renewal. It is possible
+    # that it is not present, which would indicate an invalid renewal.
+    renewal = None
+    for r in sorted_renewals:
+        # Return first renewal with a status present.
+        if "status" in r:
+            renewal = r
+            break
+
+    if not renewal:
```
anthonydillon, line 812: Why not just return renewal? If it's None it will return None.

anthonydillon (commented):
This seems like strange business logic to have in the frontend. Is this not something the backend should handle before exposing over the API?

webapp/views.py
ajbeattie: The dict lookups below will error if renewal is None. 

In any case, I'm going to remove this and just wrap the whole thing in a KeyError check in a few minutes when I get back to my keyboard

anthonydillon: Oh sorry of course, the diff cut off the rest of the function.

ajbeattie:
@anthonydillon this is simplified now to just catch the error and log it.

ajbeattie:
@jaycee @alnvdl ptal again :)

alnvdl (approved):

sowasred2012 (approved):
LGTM

webapp/views.py
```diff
@@ -531,9 +531,15 @@ def advantage_view():
...
                         ] = date_difference.days
 
-                    contract["renewal"] = make_renewal(
-                        advantage, contract["contractInfo"]
-                    )
+                    try:
+                        contract["renewal"] = make_renewal(
+                            advantage, contract["contractInfo"]
+                        )
+                    except KeyError:
```
frankban, line 538: A key error is a bit accidental here. I guess either make_renewal should catch the KeyError and return a more explicit ValueError or we just have a comment here explaining why make_renewal would fail with a KeyError.

frankban (commented):
Looks good with a possible fix for a follow up branch. Good catch!

webapp/views.py
ajbeattie: @frankban I agree - I think `make_renewal` should just be generally refactored in a followup, to remove direct dict access and use consts for the several string values in it.

PR #8632, opened 2020-10-30
add version specific esm and livepatch instructions to /advantage

sowasred2012:
## Done
Updated the instructions in the ESM and Livepatch cells on /advantage, according to the changes made [here](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.dqabkhv1pde3)

## QA

- Login with SSO
- If you don't have already have any enterprise subscriptions, go to /advantage/subscribe and buy one with test card details
  - Card number: 4242 4242 4242 4242, expiry: 04/24, cvc: 242, zip: 42424
- back on /advantage, in your subscription, expand the ESM and Livepatch cells, and see that they match the copy in [the doc](https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.dqabkhv1pde3)

## Issue / Card

Fixes #

## Screenshots


sowasred2012:
@matthewpaulthomas @spencerbygraves could I get some design input on this? The cells I've updated are ESM and Livepatch, and I'm not sure about the positioning. As you'll see below, different (existing) cells have different alignment and positioning from each other - I went with left aligned for the two updated cells, because I didn't think centered looked particularly good, but will defer to a designer's opinion! Could we make the various alignments uniform across each cell?

cells

matthewpaulthomas:
@sowasred2012 Left alignment is Canonical’s usual style. There were two reasons I went with centering instead for these expanded cells:

- At typical screen sizes, left alignment would put the info far away from where you clicked to access it.

 image

- Left-aligning the whole block would require left-aligning the command labels too (as you’ve done here), which results in unaligned colons and unaligned identical text, making it harder to notice how the commands differ.

  image

I realise adding headings makes this more difficult. I suggest using Vanilla columns to make the cell contents as a whole more centered.

matthewpaulthomas:
Until @spencerbygraves gets here … The spacing could do with a bit of work. All the bits of text are more closely related to each other than they are to the cell border. Therefore, it should be true that _a_ ≥ _c_ and _a_ ≥ _d_.

image

(It should also be true that _b_ ≥ _c_ and _b_ ≥ _d_. At large window widths, centering the block as a whole accidentally achieves this.)

spencerbygraves:
@sowasred2012 could we have left aligned and use the key value layout where needed, as below? 

image

This may help with alignment of titles.

sowasred2012:
@matthewpaulthomas @spencerbygraves have updated with the feedback, let me know what you think.

matthewpaulthomas:
@sowasred2012 As above I think the left alignment looks lopsided, but I’ll defer to @spencerbygraves on that.

Sticking with left alignment does mean we need to fix [the spacing problem](#issuecomment-719598897), though. A simple fix is to add `padding: 1.5rem` to the existing `.p-table-expanding th.p-table-expanding__panel, .p-table-expanding td.p-table-expanding__panel {`…`}` rule.

Before | After
-|-
image | image
image | image

spencerbygraves:
@matthewpaulthomas @sowasred2012 I think we should look at alignment when using the key value pattern as part of the [Vanilla issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/2848). For now I think this is fine.

A bit of extra padding would be good as @matthewpaulthomas says. The vanilla card uses a bit more.

Thanks.

bethcollins92 (approved):
LGTM apart from no new line at the end of a couple of files

templates/advantage/shared/_livepatch-info.html
```diff
@@ -1,5 +1,38 @@
...
+          <td>To deactivate:</td>
+          <td><code>sudo canonical-livepatch disable</code></td>
+        </tr>
+      </tbody>
+    </table>
+
+    <h4>Other Ubuntu versions</h4>
+    <p>Livepatch is available only on LTS versions. Consider upgrading to Ubuntu 20.04 LTS.</p>
   </div>
 </div>
```
bethcollins92, line 38: no new line here

PR #8660, opened 2020-11-04
Fix wrong parameter name and remove unused locale

carkod:
## Done

- Fix wrong parameter name and remove unused locale

## QA

- Check that the Spanish takeovers are not showing when your browser has English language


anthonydillon (approved):
LGTM thanks :+1:

PR #8674, opened 2020-11-05
Remove secrets for discourse

tbille:
## Done

We keep making authenticated API calls even if the endpoints are public
This might cause an outage due to the limit of requests from discourse

## QA

- Make sure site deploys well to demo


carkod (approved):
LGTM

PR #8696, opened 2020-11-09
Redirect buy.ubuntu.com to /advantage

nottrobin:
Set up the ubuntu.com ingress to listen for buy.ubuntu.com.

This domain will be pointed at the ubuntu.com cluster once
https://portal.admin.canonical.com/C128440
is completed, so this is ready for when that happens.

Visitors coming from buy.ubuntu.com will be redirected to
/advantage.

## QA

Not sure how to QA. You could do it using microk8s, but I can't remember exactly how the tooling for this works right now.

### DIFF
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -54,6 +54,7 @@ extraHosts:
   - domain: webapps.ubuntu.com
   - domain: ubuntuserver.org
   - domain: developer.ubuntu.com
+  - domain: buy.ubuntu.com
 
 # Overrides for production
 production:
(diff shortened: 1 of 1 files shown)


konf/site.yaml
```diff
@@ -104,6 +105,9 @@ production:
     if ($host = 'apps.ubuntu.com' ) {
       rewrite ^ https://snapcraft.io/ permanent;
     }
+    if ($host = 'buy.ubuntu.com' ) {
+      rewrite ^ https://ubuntu.com/advantage$request_uri? permanent;
```
jkfran, line 109: @nottrobin should it be without the `$request_uri?` in this case?

eg: https://buy.ubuntu.com/collections/ubuntu-advantage-virtual will 404 in https://ubuntu.com/advantage/collections/ubuntu-advantage-virtual

jkfran (approved):

PR #8703, opened 2020-11-09
UA shop: add marketo form

sowasred2012:
## Done

Added a marketo form to the payment modal, that triggers on successful payment.

## QA

- Select a product, add it to cart, checkout and use these card details with any address:
  - 4242 4242 4242 4242, expiry: 04/24, CVC: 242, ZIP: 42424
- Open the network tab in dev tools
- Click "Continue", then check the Ts & Cs box, then click "Pay"
- In the network tab, you should see a post request to "https://app-sjg.marketo.com/index.php/leadCapture/save2"
- for Marketo gurus: verify that the form is sending the necessary data

### DIFF
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -514,6 +514,23 @@ function analyticsFriendlyProducts() {
   return products;
 }
 
+function getSessionData(key) {
+  const keyValue = localStorage.getItem(key);
+
+  if (keyValue) {
+    if (key === "gclid") {
+      const gclid = JSON.parse(keyValue);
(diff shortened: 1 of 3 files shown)


caldav:
@sowasred2012  Thanks! I'm noticing a couple things during testing:

It's missing some form of email capture:
`<input type="hidden" aria-hidden="true" aria-label="hidden field" name="Email" class="mktoField mktoEmailField" id="Email" value="[email value]">`

The POST request seems to be canceled every time, and inspecting it doesn't show any response data. I'm assuming it's because of the page change happening and the request not being able to be fully sent?

caldav (approved):
Works like a charm! :+1: I'll set Marketing review to green when I have feedback from a couple more testers.

PR #8714, opened 2020-11-11
Keep test-redirects script in scripts dir

nottrobin:
Just keeping files tidy. No real QA to do here, unless you're @pmahnke who presumably wrote this script and knows how to use it, in which case you could just check it still works in its new location.


pmahnke (approved):
+1

PR #8752, opened 2020-11-17
filter out "closed" status renewals

sowasred2012:
## Done

If a user has multiple renewals, filter out any with "closed" status so we can surface any that are open.

## QA

- See that the Python tests pass.

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -244,7 +244,7 @@ class TestMakeRenewal(unittest.TestCase):
         self.assertEqual(got, want)
 
     def test_closed(self):
-        """Closed renewals are not renewable."""
+        """Closed renewals are not returned to the view."""
         advantage = make_advantage()
         now = datetime.now(timezone.utc)
         start = (now - timedelta(days=1)).isoformat()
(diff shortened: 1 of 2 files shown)


anthonydillon (approved):
LGTM

webapp/views.py
```diff
@@ -816,7 +816,14 @@ def make_renewal(advantage, contract_info):
         key=lambda renewal: dateutil.parser.parse(renewal["start"]),
     )
 
-    renewal = sorted_renewals[0]
+    open_renewals = [
+        renewal for renewal in sorted_renewals if renewal["status"] != "closed"
```
frankban, line 820: We could target just pending renewals rather than "not closed", and also do that while sorting, in a single pass:
```python
sorted_renewals = sorted(
    r for r in renewals if r["status"] == "pending", 
    key=lambda r: dateutil.parser.parse(r["start"]),
)
```
What do you think?

tests/test_views.py
```diff
@@ -244,30 +244,33 @@ def test_multiple_pending(self):
...
         contract_info = {
             "renewals": [
                 {
                     "id": "1",
                     "status": "closed",
                     "actionable": True,
                     "start": start,
-                    "end": end,
+                },
+                {
```
frankban, line 259: Let's add a renewal in error state here to check that this is not taken?

frankban (approved):
Looks good with a couple of suggestions. Thanks for the quick fix!

webapp/views.py
sowasred2012: We have a case in there where a renewal that is "done" can be returned, so that we can show that contract as having been recently renewed (it will have a "Renewed" label on it), and this would stop that from happening - that said, that's another case where if there are multiple renewals to be renewed, the user could renew the first, it would be marked as "done", and then they wouldn't be able to see their second renewal...

sowasred2012: Agreed that this is out of this PR's scope, and will be addressed in https://github.com/canonical-web-and-design/ubuntu.com/issues/8755

tests/test_views.py
sowasred2012: Will be addressed by https://github.com/canonical-web-and-design/ubuntu.com/issues/8755

PR #8760, opened 2020-11-18
Sum all contract allowances regardless of metric

### LINKED ISSUE #8759: UA Shop: machine usage display may display the wrong count
The current code for calculating allowances seems to be mixing active machines with attached machines:

https://github.com/canonical-web-and-design/ubuntu.com/blob/13e6bf7f2028a0d882133d71d93e43f0b4a4a918/webapp/views.py#L569-L581

At the moment in ua-contracts, we're in a situation where:
- we use two allowance metrics: `attached-machines` and `active-machines`;
- a contract may have allowances with these two types of metrics, at the same time.

In the long term, neither metric should be used for this purpose anymore. We're still discussing how we want to model this going forward.

Regardless of what we decide, in order for the web interface to be accurate now and most likely in the future, it'd be best if you just summed up all allowance values when displaying, regardless of the metric name.

alnvdl:
Fixes #8759.

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -355,13 +355,13 @@ class TestGetMachineUsage(unittest.TestCase):
         contract = {
             "contractInfo": {
                 "allowances": [
-                    {"metric": "discared-machines", "value": 42},
+                    {"metric": "discarded-machines", "value": 42},
                     {"metric": "joined-machines", "value": 47},
                 ]
             }
         }
         got = views.get_machine_usage(advantage, contract)
-        self.assertEqual(got, views.MachineUsage(attached=3, allowed=0))
+        self.assertEqual(got, views.MachineUsage(attached=3, allowed=89))
 
     def test_both_attached_and_allowed(self):
         """Both attached and allowed counts are returned."""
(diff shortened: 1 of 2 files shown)


frankban (approved):

sowasred2012:
@alnvdl looks good, but wanted to check with you the test we have for invalid allowances - to make it pass, we have to sum them all, but if they're invalid, should we be summing them?

```
def test_invalid_allowance_metric(self):
        """Machines allowance is defined in a specific metric."""
        advantage = make_advantage(machines={"machines": [1, 2, 3]})
        contract = {
            "contractInfo": {
                "allowances": [
                    {"metric": "discarded-machines", "value": 42},
                    {"metric": "joined-machines", "value": 47},
                ]
            }
        }
        got = views.get_machine_usage(advantage, contract)
        self.assertEqual(got, views.MachineUsage(attached=3, allowed=89))
```

alnvdl:
@sowasred2012 Please sum them all. ua-contracts should not return invalid allowances. If an allowance is returned, you can consider it valid. Thanks for fixing the failing tests by the way!

sowasred2012 (approved):
LGTM

PR #8777, opened 2020-11-20
Changes CVE table from 0.8rem to 0.875rem

matthewpaulthomas:
To match the Vanilla font size scale.

## Done

- Changed CVE table font size from 0.8rem to 0.875rem.

## QA

- Check that the table still looks okay

## Issue / Card

No issue, but raised by @lyubomir-popov in [vanilla-framework#3399](https://github.com/canonical-web-and-design/vanilla-framework/issues/3399).

### DIFF
--- a/static/sass/_pattern_cve.scss
+++ b/static/sass/_pattern_cve.scss
@@ -17,7 +17,7 @@
   }
 
   .cve-table {
-    font-size: 0.8rem;
+    font-size: 0.875rem;
     line-height: 1rem;
   }
 
--- a/static/sass/_pattern_cve.scss
+++ b/static/sass/_pattern_cve.scss
@@ -17,8 +17,7 @@
   }
 
   .cve-table {
-    font-size: 0.875rem;
-    line-height: 1rem;
+    @extend %small-text;
   }
 
   .cve-table thead tr {


lyubomir-popov:
@matthewpaulthomas how can I make the table appear? I see a search form but no table.

static/sass/_pattern_cve.scss
```diff
@@ -17,7 +17,7 @@
   }
 
   .cve-table {
-    font-size: 0.8rem;
+    font-size: 0.875rem;
```
lyubomir-popov, line 20: Ideally this should refer to the variable so if we change something it gets updated here automatically, e.g.:
```
    font-size: #{map-get($font-sizes, small)}rem;
```

Small text also has smaller line-height, and some other adjusted properties, so something like this may be better:

```
@extend %small-text;
```

lyubomir-popov (approved):
Approving as this change will fix the issue, but I've added some optional improvement suggestions.

matthewpaulthomas:
Thanks @lyubomir-popov . Adopted your suggestion of using `@extend %small-text;`.

> @matthewpaulthomas how can I make the table appear? I see a search form but no table.

You can see it below the “Recent CVEs affecting Ubuntu” heading [on production](https://ubuntu.com/security/cve).

PR #8804, opened 2020-11-23
Only move chars with versions right

pmahnke:
## Done

- Only move chars with versions right

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/about/release-cycle#ubuntu-openstack-release-cycle
- See that all charts align properly


steverydz (changes requested):
I would do something like:

```
var chartTranslateX = margin.left;

if (taskVersions) {
  chartTranslateX = margin.left * 1.6;
}
```

steverydz (approved):
LGTM :+1:

PR #8805, opened 2020-11-23
Allow LSNs

albertkol:
## Done

- Change NoticeSchema to allow LSNs too.

### DIFF
--- a/webapp/security/schemas.py
+++ b/webapp/security/schemas.py
@@ -84,7 +84,7 @@ class NoticePackage(Schema):
 
 
 class NoticeSchema(Schema):
-    id = String(required=True, validate=Regexp(r"USN-\d{1,5}-\d{1,2}"))
+    id = String(required=True, validate=Regexp(r"(LSN|USN)-\d{1,5}-\d{1,2}"))
     title = String(required=True)
     summary = String(required=True)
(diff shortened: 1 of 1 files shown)


webapp/security/schemas.py
```diff
@@ -84,7 +84,7 @@ class NoticePackage(Schema):
 
 
 class NoticeSchema(Schema):
-    id = String(required=True, validate=Regexp(r"USN-\d{1,5}-\d{1,2}"))
+    id = String(required=True, validate=Regexp(r"(LSN-|USN-)\d{1,5}-\d{1,2}"))
```
nottrobin, line 87: why not

```suggestion
    id = String(required=True, validate=Regexp(r"(LSN|USN)-\d{1,5}-\d{1,2}"))
```

?

I'm gonna approve this anyway, this is just a suggestion.

nottrobin (approved):
Great work thanks!

webapp/security/schemas.py
albertkol: @nottrobin no good reason.. :D  I used as an example the regex we used for cves where the - is included. :P

albertkol: Done.

PR #8812, opened 2020-11-24
3448 build tutorials search

ClementChaumel:
## Done

- Added support for searching tutorials using google API in `views.py`
- Uncommented the search bar and added an event listener so it would update the query parameters on change
- Added a "No results found" state

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/tutorials
- Check [design](https://app.zeplin.io/project/5df26dadead180a5c29b569d/screen/5dfa10d6d30a77171fd01cbe)

## Issue / Card

Fixes https://github.com/canonical-web-and-design/web-squad/issues/3448

### DIFF
--- a/templates/tutorials/index.html
+++ b/templates/tutorials/index.html
@@ -1,8 +1,8 @@
 {% extends "tutorials/base_tutorials.html" %}
 
-{% block title %}Tutorials{% endblock %}	
+{% block title %}Tutorials{% endblock %}
 
-{% block content %}	
+{% block content %}
 
 {% if page == 1 %}
 <section class="p-strip--suru-blog-header is-dark">
@@ -115,12 +115,12 @@
     {% endif %} {% endfor %}
   </div>
 </section>
-{% endif %} 
+{% endif %}
 
 {% with %}
-  {% set total_pages = total_pages %}	
-  {% set current_page = page %}	
-  {% include "shared/_pagination.html" %}	
-{% endwith %}	
+  {% set total_pages = total_pages %}
+  {% set current_page = page %}
+  {% include "shared/_pagination.html" %}
+{% endwith %}
 
 {% endblock %}
--- a/templates/tutorials/index.html
+++ b/templates/tutorials/index.html
@@ -64,6 +64,39 @@ endblock %} {% block content %} {% if page == 1 %}
   </div>
 </section>
 
+{% if query and result_number == 0 %}
+
+<div class="p-strip">
+  <div class="row">
+    <div class="col-6">
+      <h3>Why not try widening your search?</h3>
+      <p>You can do this by:</p>
+      <ul class="p-list">
+        <li class="p-list__item is-ticked">
+          Adding alternative words or phrases
+        </li>
+        <li class="p-list__item is-ticked">
+          Using individual words instead of phrases
+        </li>
+        <li class="p-list__item is-ticked">Trying a different spelling</li>
+      </ul>
+    </div>
+    <div class="col-6">
+      <h3>Still no luck?</h3>
+      <ul class="p-list">
+        <li class="p-list__item is-ticked">
+          <a href="/">Visit the Ubuntu homepage</a>
+        </li>
+        <li class="p-list__item is-ticked">
+          <a href="/desktop/contact-us?product=search-page">Contact us</a>
+        </li>
+      </ul>
+    </div>
+  </div>
+</div>
+
+{% else %}
+
 <section class="p-strip has-shadow">
   <div class="row u-equal-height">
     {% if query %}
@@ -94,6 +127,5 @@ endblock %} {% block content %} {% if page == 1 %}
     {% endif %} {% endfor %}
   </div>
 </section>
-
-{% with %} {% set total_pages = total_pages %} {% set current_page = page %} {%
-include "shared/_pagination.html" %} {% endwith %} {% endblock %}
+{% endif %} {% with %} {% set total_pages = total_pages %} {% set current_page =
+page %} {% include "shared/_pagination.html" %} {% endwith %} {% endblock %}
--- a/templates/tutorials/index.html
+++ b/templates/tutorials/index.html
@@ -1,16 +1,20 @@
-{% extends "tutorials/base_tutorials.html" %}
-
-{% block title %}Tutorials{% endblock %}
-
-{% block content %}
-
-{% if page == 1 %}
+{% extends "tutorials/base_tutorials.html" %} {% block title %} Tutorials {%
+endblock %} {% block content %} {% if page == 1 %}
 <section class="p-strip--suru-blog-header is-dark">
   <div class="row">
     <div class="col-6">
       <h1>Tutorials</h1>
-      <p>These tutorials provide a step-by-step process to doing development and dev-ops activities on Ubuntu machines, servers or devices.</p>
-      <p><a class="p-link--inverted" href="https://discourse.ubuntu.com/c/tutorials">You can write your own&nbsp;&rsaquo;</a></p>
+      <p>
+        These tutorials provide a step-by-step process to doing development and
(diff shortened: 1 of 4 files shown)


anasereijo:
Hi @ClementChaumel this is looking great! Just a few things:

- [ ] I've been testing searching for something that doesn't exist I get 2 things:

_500: Server error_
Screenshot 2020-11-24 at 17 24 26

or

_no results text_
Screenshot 2020-11-24 at 17 27 09

-----------------------------

- [ ] when I want to clean up the search results by using the `X` on the search field I have to click outside to make it work:

Screen Recording 2020-11-24 at 17 32 23


- [ ] when I clean up the search results the Topic is still selected - I think it should go back to _Any_?

Screen Recording 2020-11-24 at 17 53 25

-----------------------------

@wgx at the moment we are displaying the same no results text as we use on ubuntu.com. Any thoughts about this? Might need some copy changes to fit the tutorials search better.
 
Screenshot 2020-11-24 at 17 27 09

wgx:
@anasereijo @ClementChaumel Yes, the re-search should occurr on mouseUp of the X removal, rather than de-focus of the input field. 

Second issue.. not sure. We could say something like "View all Tutorials" so they have a handy shortcut back to the main list? Not sure where else we could send them...

templates/tutorials/index.html
```diff
@@ -1,33 +1,41 @@
-{% extends "tutorials/base_tutorials.html" %}
-
-{% block title %}Tutorials{% endblock %}
-
-{% block content %}
-
-{% if page == 1 %}
+{% extends "tutorials/base_tutorials.html" %} {% block title %} Tutorials {%
```
nottrobin, line 1: These shouldn't be collapsed into 1 line, it's much harder to read.

This will be an auto formatter of some kind. Could you disable it for HTML pages? They never do a good job with template tags.

ClementChaumel: Yeah I noticed that, I'll try to figure out which one of my plugins is misbehaving

webapp/views.py
```diff
@@ -28,6 +29,10 @@
 from webapp.decorators import store_maintenance
 
 
+class NoAPIKeyError(Exception):
```
nottrobin, line 32: why not:

```python3
from canonicalwebteam.search.views import NoAPIKeyError
```

?

webapp/views.py
```diff
@@ -1037,7 +1042,32 @@ def tutorials_index():
...
+        """
+
+        # The webteam's default custom search ID
+        search_engine_id = "009048213575199080868:i3zoqdwqk8o"
+
+        # API key should always be provided as an environment variable
+        search_api_key = os.getenv("SEARCH_API_KEY")
+
+        if not search_api_key:
+            raise NoAPIKeyError("Unable to search: No API key provided")
```
nottrobin, line 1059: It's actually a shame you have to do this again here. It would be nice if the code inside the search module was encapsulated such that it contained all this logic but allowed you to get results for your own view.

I'll file an issue. But this is fine for now.

webapp/views.py
ClementChaumel: Cause I didn't think of it! Nice catch, I've changed it

templates/tutorials/index.html
ClementChaumel: I've reverted it!

ClementChaumel:
@nottrobin I'm not sure why the python tests are failing. They pass locally and I don't really know what the issue is (I'm thinking possibly the missing API key is causing it to throw an error and return a `500` but it if was the case the search page would also fail so :shrug: )

webapp/views.py
nottrobin: Filed https://github.com/canonical-web-and-design/canonicalwebteam.search/issues/5

webapp/views.py
```diff
@@ -1058,7 +1098,8 @@ def tutorials_index():
                 metadata, key=lambda k: k["difficulty"], reverse=False
             )
 
-        total_pages = math.ceil(len(metadata) / posts_per_page)
+        result_number = len(metadata)
```
nottrobin, line 1101: `number_of_results` maybe?

ClementChaumel: Done!

nottrobin:
@ClementChaumel your problem with the tests is that the `/tutorials` view now *requires* the API key to load. The tests don't provide that key.

Could you please make it so the tutorials homepage itself loads fine without the key, and the key is only needed when you actually perform a search?

nottrobin (approved):
This works and code looks good to me. Really great work @ClementChaumel, not a simple task, you accomplished it brilliantly.

I have one UX comment, but I don't think it should be a blocker - this is fine for an MVP:

To me, it's a bit weird that you have to *submit* the search box, but then when you do, the page doesn't look any different. In fact it might jump you back up to the top of what looks like the original tutorials homepage.

I would expect that it would work in one of the following 2 ways:

1. The search box filters the result-set live, when you pause typing for a short time; or:
2. When you submit the search box, the layout changes to collapse the header a bit, show some text like "Results matching <your search" or something

I think option 1 is smoother, since it is in keeping with how the rest of the fields in that form work.

Just let me know if you don't want / have time to fix this in this PR, and I'll file an issue about it.

wgx:
Yep, agree with @nottrobin  - hitting return or clicking to submit is fine for an MVP - but generally, if I'm just **filtering** an on-page resource, I'd expect that to filter "live" in the view rather than waiting on a submit. Agree this is ok for first version though. 👍🏼

ClementChaumel:
Thanks!!

> I would expect that it would work in one of the following 2 ways:
> 
> 1. The search box filters the result-set live, when you pause typing for a short time; or:
> 2. When you submit the search box, the layout changes to collapse the header a bit, show some text like "Results matching <your search" or something
> 
> I think option 1 is smoother, since it is in keeping with how the rest of the fields in that table work.

So yeah at first I imagined it as the first one as well and I've done it before in react just filtering elements from the state on (debounced) field change. In here I tried to adhere to the "On submit" logic which I assumed was built in the vanilla search field https://vanillaframework.io/docs/patterns/search-box 
Also I'm not sure how "Real time" flask can be. Wouldn't changing the query param cause a refresh anyway?

nottrobin:
> @ClementChaumel  Also I'm not sure how "Real time" flask can be. Wouldn't changing the query param cause a refresh anyway?

Oh I see, the "Topic" drop-down actually reloads the page. This isn't ideal either.

Flask is not any sort of limitation here, it's just a server-side framework. The solution needs a server-side and client-side piece.:

- JavaScript (can be React, doesn't matter) will listen for "keyup", with some debounce delay
- Then use JS to show a spinner of some sort and trigger an XHR request to the server
- The server has an endpoint that returns JSON data about tutorials
- JS then updates the page with the new data, and updates the URL with the query parameter (it can do this without a page reload no problem)

This would be true in any framework, even if they tried to sweep the detail away from you (which in my opinion is usually a bad thing, as things will end up breaking in confusing ways that are difficult to debug).

I'm happy to have a meeting about this, but also might be worth pinging `@webteam` in [~web--design](https://chat.canonical.com/canonical/channels/web--design) to see if other frontenders have suggestions about where this is done in other places that you could copy.

However, this does seem quite involved, so I've filed [an issue](https://github.com/canonical-web-and-design/ubuntu.com/issues/8834) for this to be solved as an enhancement down the line, and we can merge this PR now.

nottrobin:
> I'm happy to have a meeting about this, but also might be worth pinging `@webteam` in [~web--design](https://chat.canonical.com/canonical/channels/web--design) to see if other frontenders have suggestions about where this is done in other places that you could copy.

Actually I think this pattern in followed quite heavily on in snapcraft.io on e.g. [snapcraft.io snap details pages](https://snapcraft.io/lxd) where it's used to dynamically pull in blog articles. @tbille would probably know more.

ClementChaumel:
Ok cool I'll merge this for now then, thanks!

PR #8819, opened 2020-11-25
UA shop: increase the amount of time we wait for a payment to process

### LINKED ISSUE #8811: [shop] extend timeout time for purchase confirmation
extend to 60 sec

sowasred2012:
## Done

Increased maximum waiting time during payment processing from 30 to 60 seconds.

There are three messages presented to the user over time while a payment is processing, one after two seconds, another after a further two seconds, and a third after a further 11 seconds. We then waited another 15 seconds before sending the user to /advantage. This PR increases that final wait to 45 seconds, rather than 15, for a total of 60.

## QA

- Check the code

## Issue / Card

Fixes #8811


ClementChaumel (approved):
:+1:

PR #8824, opened 2020-11-26
8756 make account name optional

### LINKED ISSUE #8756: [UA shop handover] advantage shop: purchase form: make account name optional
We can allow Customers that are not purchasing on behalf of a company to avoid providing the account/cmpany name, as their full name is sufficient for creating an account.
Logic would be like:
```
if acocunt_name.strip() == "":
    account_name = full_name
```

ClementChaumel:
⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️
### BEFORE MERGING / AFTER QA
Contract API url and Stripe public key must be updated to production values. They can be found [here](https://docs.google.com/document/d/1BG2mRWUF9pHAiwLPgm-7gelrZdstjlkQit84kXBum4A), and also here:
- contract API url: https://contracts.canonical.com
- Stripe public key: pk_live_68aXqowUeX574aGsVck8eiIE

⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️

## Done

- Made account name optional in the payment modal
- Default to user name if there's no account name provided

## QA

- try to fill out the address form without an account name

## Issue / Card

Fixes #8756

### DIFF
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -968,7 +968,9 @@ function setCustomerInfo() {
   customerInfo = {
     email: formData.get("email"),
     name: formData.get("name"),
-    accountName: formData.get("account_name").trim() || formData.get("name"),
+    accountName:
+      (formData.get("account_name") && formData.get("account_name").trim()) ||
+      formData.get("name"),
     address: {
       city: formData.get("city"),
       country: formData.get("Country"),
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -56,6 +56,7 @@ const forOrganisationRadio = document.getElementById(
   "buying_for_an_organisation"
 );
 const accountNameField = document.getElementById("account_name");
+const accountNameLabel = document.getElementById("account_name_label");
 
 // initialise Stripe
 const stripe = window.Stripe(window.stripePublishableKey);
@@ -249,8 +250,6 @@ function attachFormEvents() {
     checkVAT();
   });
 
-  accountNameField.disabled = true;
-
   forMyselfRadio.addEventListener("change", handleNameFieldRadio);
   forOrganisationRadio.addEventListener("change", handleNameFieldRadio);
 
@@ -267,8 +266,10 @@ function handleNameFieldRadio() {
   if (forMyselfRadio.checked) {
     accountNameField.value = "";
     accountNameField.disabled = true;
+    accountNameLabel.classList.add("u-text--muted");
   } else {
     accountNameField.disabled = false;
+    accountNameLabel.classList.remove("u-text--muted");
   }
 }
 
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -51,6 +51,12 @@ const closeModalButton = modal.querySelector(".js-close-modal");
 const cardErrorElement = document.getElementById("card-errors");
 const paymentErrorElement = document.getElementById("payment-errors");
 
+const forMyselfRadio = document.getElementById("buying_for_myself");
+const forOrganisationRadio = document.getElementById(
+  "buying_for_an_organisation"
+);
+const accountNameField = document.getElementById("account_name");
+
 // initialise Stripe
 const stripe = window.Stripe(window.stripePublishableKey);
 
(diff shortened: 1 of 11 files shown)


static/js/src/ua-payment-modal.js
```diff
@@ -942,7 +942,8 @@ function setCustomerInfo() {
   customerInfo.email = formData.get("email");
   customerInfo.country = formData.get("Country");
   customerInfo.name = formData.get("name");
-  customerInfo.accountName = formData.get("account_name");
+  customerInfo.accountName =
+    formData.get("account_name") || formData.get("name");
```
frankban, line 946: Should we trim the account name before doing the check?

frankban (commented):
Looks good. I'd also change the description for the account name field: "This can be your organisation, or your name again". We should make it clearer that the field is not required, nor is required to repeat your name again.

Also, in my QA, the stripe field seems broken, as it does not limit the number of digits on the zip code.

static/js/src/ua-payment-modal.js
ClementChaumel: > Looks good. I'd also change the description for the account name field: "This can be your organisation, or your name again". We should make it clearer that the field is not required, nor is required to repeat your name again.

Do you know who I should ask to get a new copy?

anthonydillon: You could ask @matthewpaulthomas

matthewpaulthomas:
Hi @ClementChaumel 👋

All else being equal, the fewer fields a form has, the better. So I’m confused as to why we were directed to add an account name field to this form in the first place, if it wasn’t actually required.

Separately, the later someone has to make a decision in a UI, the easier it is to make. By that rule, it would be better for organisation name to be something you can (but don’t have to) choose _after_ the purchase, rather than trying to decide whether to use one when you’re in the middle of making a payment. (Your organisation might later change its name, or get bought/merged, so eventually we should allow setting account name after payment anyway.)

_If_ it really does have to be in the payment form, I suggest this:

image

That would avoid any possibility of people thinking that they’re being expected to type their name twice.

anthonydillon:
QA issue when filling out the form. It looks like each section of the form is expecting a `p-form-validation__message` element to write validation messages too. This missing and so gett console errors.

Peek 2020-12-01 11-36

anthonydillon (changes requested):
QA issue

ClementChaumel:
> QA issue

I've added a hidden validation message to remove the errors

anthonydillon (approved):
LGTM :+1: Thanks

webapp/views.py
```diff
@@ -398,7 +398,7 @@ def advantage_view():
     new_subscription_id = None
     open_subscription = flask.request.args.get("subscription", None)
     stripe_publishable_key = os.getenv(
-        "STRIPE_PUBLISHABLE_KEY", "pk_live_68aXqowUeX574aGsVck8eiIE"
+        "STRIPE_PUBLISHABLE_KEY", "pk_test_yndN9H0GcJffPe0W58Nm64cM00riYG4N46"
```
squidsoup, line 401: It seems weird to be hardcoding the default from `os.getenv` here?

matthewpaulthomas:
It would be less distracting/confusing if the “Choose account name:” label was greyed out whenever the field is. But that problem already existed, so +1 for this.

webapp/views.py
nottrobin: I'm not too fussed. Yeah it could be in `.env` and then also written into Kubernetes config, but that way it is a tiny bit of a pain 'cos it's kept in at least 2 places. I think whether we hard-code it depends on whether it needs to change between environments, and how we're gonna want to update it. If it needs to be different between dev / staging / prod then it can't be hard coded. Or if we might want to change the publishable key without doing a release, then k8s configs would allow us to do that.

If neither of those are the case - it's easiest for it to be the same between all environments, and we are perfectly happy to update it through PR + release, then I'm actually wondering if there's any point using `os.getenv` at all. Maybe we should just do `stripe_publishable_key = "pk_test_yndN9H0GcJffPe0W58Nm64cM00riYG4N46"` for now, as the most straightforward, declarative solution, and only add in the ability to read it from an environment variable as and when we actually need it.

PR #8843, opened 2020-12-01
Lazy load the default takeover image

pmahnke:
## Done

- To reduce the chance you see a flash of the gorilla while the homepage loads, I have turn on lazy loading

## QA

- Turn off JS and see that the 20.10 takeover loads
- Turn on JS and see less flash than on live.


anthonydillon (approved):
LGTM nice idea. 👍

PR #8854, opened 2020-12-02
Prefill name and address

### LINKED ISSUE #8757: [UA shop handover] advantage form: purchase form: pre-fill name and address with stripe info
The full name and all address fields could be pre-filled for returning customers by making a call to the ua-contracs GetCustomerInfo API endpoint and using the returned information.
We still allow to change that information, but at least we don't require customers to provide it every time.

ClementChaumel:
⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️
### BEFORE MERGING / AFTER QA
Contract API url and Stripe public key must be updated to production values. They can be found [here](https://docs.google.com/document/d/1BG2mRWUF9pHAiwLPgm-7gelrZdstjlkQit84kXBum4A), and also here:
- contract API url: https://contracts.canonical.com
- Stripe public key: pk_live_68aXqowUeX574aGsVck8eiIE

⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️

## Done

- Added a call to `/v1/accounts/{account}/customer-info/stripe` to get the name and address from Stripe
- Refactored the way changing your payment methods fills the form to use the same function for both

## QA

- If logged out, log in
- if logged in, see that your address should be pre filled to the last one you used to make a purchase
- make a new purchase with a new address, refresh the page and see that it should be pre filled next time

## Issue / Card

Fixes #8757 

## Side note
I've noticed that in the response from `/v1/accounts/{account}/customer-info/stripe` we also have the paymentId and the 4 last digits of the card. Have we thought of pre-filling the card number input with something like 'Card ending in 1234 [Use another card button]'? 
That would turn the purchase process in a single click (granted there's no ATV number required)

### DIFF
--- a/static/js/src/advantage/contracts-api.js
+++ b/static/js/src/advantage/contracts-api.js
@@ -36,6 +36,15 @@ export async function getRenewal(renewalID) {
   return data;
 }
 
+export async function getCustormerInfo(accountId) {
+  let response = await fetch(`/advantage/customer-info/${accountId}`, {
+    cache: "no-store",
+  });
+
+  let data = await response.json();
+  return data;
+}
+
 export async function postInvoiceID(transactionType, transactionID, invoiceID) {
   let response = await fetch(
     `/advantage/${transactionType}/${transactionID}/invoices/${invoiceID}`,
--- a/static/js/src/advantage/contracts-api.js
+++ b/static/js/src/advantage/contracts-api.js
@@ -36,7 +36,7 @@ export async function getRenewal(renewalID) {
   return data;
 }
 
-export async function getCustormerInfo(accountId) {
+export async function getCustomerInfo(accountId) {
   let response = await fetch(`/advantage/customer-info/${accountId}`, {
     cache: "no-store",
   });
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -4,7 +4,7 @@ import {
   ensurePurchaseAccount,
   getPurchase,
   getRenewal,
-  getCustormerInfo,
+  getCustomerInfo,
   postInvoiceID,
   postCustomerInfoToStripeAccount,
   postCustomerInfoForPurchasePreview,
@@ -402,7 +402,7 @@ function applyGuestPurchaseTotals() {
 }
 
 function fetchCustomerInfo(accountId) {
-  getCustormerInfo(accountId)
+  getCustomerInfo(accountId)
     .then((res) => {
       const { name, address } = res.customerInfo;
       customerInfo = { ...customerInfo, name, address };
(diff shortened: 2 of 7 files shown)


static/js/src/advantage/contracts-api.js
```diff
@@ -36,6 +36,15 @@ export async function getRenewal(renewalID) {
   return data;
 }
 
+export async function getCustormerInfo(accountId) {
```
squidsoup, line 39: typo: `getCustomerInfo` (without the `r`)

static/js/src/ua-payment-modal.js
```diff
@@ -431,6 +401,43 @@ function applyGuestPurchaseTotals() {
...
 
+function fetchCustomerInfo(accountId) {
+  getCustormerInfo(accountId)
+    .then((res) => {
+      const { name, address } = res.customerInfo;
+      customerInfo = { ...customerInfo, name, address };
+      setFormElements();
+      isCustomerInfoSet = true;
+    })
+    .catch((e) => console.error(e));
```
squidsoup, line 412: Do we have sentry configured on ubuntu.com? Might nice to send an error here.

static/js/src/advantage/contracts-api.js
ClementChaumel: Whoops! Good catch

sowasred2012:
> I've noticed that in the response from /v1/accounts/{account}/customer-info/stripe we also have the paymentId and the 4 last digits of the card. Have we thought of pre-filling the card number input with something like 'Card ending in 1234 [Use another card button]'?
That would turn the purchase process in a single click (granted there's no ATV number required)

I'm pretty sure we can't manipulate the contents of the input itself, so we'd need to hide that element and show something else in its place (plus make sure we handle toggling those elements when the user clicks "Change" on the second page of the modal), but that's a good idea - worth running by @matthewpaulthomas to see what he thinks, and how it should look.

ClementChaumel:
> I'm pretty sure we can't manipulate the contents of the input itself, so we'd need to hide that element and show something else in its place (plus make sure we handle toggling those elements when the user clicks "Change" on the second page of the modal)

Yeah we don't have the full card number anyway just the last 4 digits I believe so it would need to be a different component

static/js/src/ua-payment-modal.js
sowasred2012: Yeah, Sentry is setup on the backend at least - this call hits that, which then calls the contracts API and raises any errors at that point, to be captured by Sentry.

sowasred2012 (approved):
LGTM, nice work!

PR #8888, opened 2020-12-07
Switch API environment depending on query param

### LINKED ISSUE #8828: [UA shop] add feature flag to easily switch between staging and live environments
From a meeting on 27/11/20 with @nottrobin, @ClementChaumel and @sowasred2012, and is a follow up to the conversation that happened here: https://github.com/canonical-web-and-design/meeting-notes/issues/597

When developing UA renewals and shop, 99% of the time you need to be working with the staging variants of the contracts API and Stripe implementation, but the outcome of the conversation linked to above was that any external services should be configured to use their production variants by default.

The solution I used when working on the initial launch of the shop was to have a feature branch that I'd committed staging values for the API and Stripe to, and any shop work would be merged into that, so I only actually had to change those values back to production once, when the shop was finally launched.

Now that the shop is live, and Clement is working on bug fixes and feature requests that will be merged straight into master, having to update those values and then remember to change them back is going to get old pretty quickly, so in a meeting with Robin we discussed potential solutions. 

One that we thought would work well with the current setup is to have a feature flag that would tell the site to use staging versions of the API and Stripe, so if a user was to append `?ua_staging=true` to the URL, you'd get the shop staging environment.

ClementChaumel:
## Done

- Added checks for the query params. if the url contains `?ua_staging=true` we use staging values for both Stripe and Contracts API if not we default to live ones.

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/advantage/subscribe?test_backend=true
- check that the purchase works with a test card (4242424242...)
- Go to http://0.0.0.0:8001/advantage/subscribe without the query param
- Try to complete a purchase using a test card and see that it fails with an error

## Issue / Card

Fixes #8828


static/js/src/advantage/contracts-api.js
```diff
@@ -37,17 +46,24 @@ export async function getRenewal(renewalID) {
...
+      cache: "no-store",
+    }
+  );
 
   let data = await response.json();
   return data;
 }
 
 export async function postInvoiceID(transactionType, transactionID, invoiceID) {
+  const queryString = window && window.location ? window.location.search : "";
```
nottrobin, line 63: Is it really feasible that `window` or `window.location` are not available?

webapp/views.py
```diff
@@ -974,6 +1035,20 @@ def ensure_purchase_account():
     contains an auth token required for subsequent calls to the
     contract API.
     """
+    is_staging = flask.request.args.get("ua_staging", False)
+
+    api_url = flask.current_app.config["CONTRACTS_LIVE_API_URL"]
+
+    if is_staging:
+        print("Bah alors?????")
```
albertkol, line 1043: I don't think these were meant to be left in the code. There are a few more below.

webapp/views.py
```diff
@@ -399,15 +399,26 @@ def advantage_view():
     new_subscription_start_date = None
     new_subscription_id = None
     open_subscription = flask.request.args.get("subscription", None)
+    is_staging = flask.request.args.get("ua_staging", False)
```
nottrobin, line 402: You very smartly use the wording "live" and "test" in the env var names, which I like. Because you could switch backend on any environment, so it's more about testing and live than it is about staging and production.

But here you use the word "staging". Maybe this should be "test_backend" or something?

albertkol (changes requested):
Also we should make sure we are not indexing the page if we have the flag up. (@nottrobin)

Add a check and if the staging flag is there add/make `<meta name="robots" content="noindex">`

webapp/views.py
```diff
@@ -399,15 +399,26 @@ def advantage_view():
...
+
     stripe_publishable_key = os.getenv(
-        "STRIPE_PUBLISHABLE_KEY", "pk_live_68aXqowUeX574aGsVck8eiIE"
+        "STRIPE_LIVE_PUBLISHABLE_KEY", "pk_live_68aXqowUeX574aGsVck8eiIE"
     )
 
+    api_url = flask.current_app.config["CONTRACTS_LIVE_API_URL"]
+
+    if is_staging:
```
nottrobin, line 410: If we are serving a page built with staging data, we absolutely have to have to have to include a `x-robots-tag: noindex` header (that's from memory, please check the syntax). These pages *must not* be indexed by Google.

nottrobin (dismissed):
a few suggestions inline

webapp/views.py
nottrobin: Ah Albert suggested putting it in the html. I prefer the http header solution I think, but either will work.

webapp/views.py
ClementChaumel: :see_no_evil: I removed the console.log but forgot about these

static/js/src/advantage/contracts-api.js
ClementChaumel: Yeah I don't think so now that I think on it. Just a habit from my Node days

webapp/views.py
ClementChaumel: I've added it in the head as you suggested.

albertkol (approved):
:+1:

PR #8891, opened 2020-12-07
Security API ingress rule

jkfran:
**This PR can only be merged after https://github.com/canonical-web-and-design/konf/pull/4**

## Done

- Ingress rule for security API

## QA

- Follow the QA steps in the Konf repo: https://github.com/canonical-web-and-design/konf/pull/4

### DIFF
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -106,6 +106,9 @@ production:
         - name: SENTRY_DSN
           value: https://0293bb7fc3104e56bafd2422e155790c@sentry.is.canonical.com//13
 
+    - paths: [/security/cves.json, /security/cves/*.json, /security/notices.json, /security/notices/*.json, /security/spec.json, /security/docs]
+      service_name: ubuntu-com-security-api
+
   nginxConfigurationSnippet: |
     if ($host = 'apps.ubuntu.com' ) {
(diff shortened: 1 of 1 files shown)


konf/site.yaml
```diff
@@ -106,6 +106,9 @@ production:
         - name: SENTRY_DSN
           value: https://0293bb7fc3104e56bafd2422e155790c@sentry.is.canonical.com//13
 
+    - paths: [/security-api]
```
albertkol, line 109: Instead of `/security-api` we were planning to have `[/security/cves.json, /security/cves/*.json, /security/notices.json, /security/notices/*.json, /security/spec.json, /security/docs]`

Not sure if comma separated is the syntax to enumerate multiple paths. Also not sure if * is the way to wildcard the `cve_id` in `/security/cves/*.json` or `notice_id` for the `/security/notices/*.json`.

jkfran: Done, the wildcard should be valid.

jkfran:
@albertkol before merging this, it will be nice to deploy the `ubuntu-com-security-api` service first and then QA all these endpoints. I think it will be better to leave this PR the last one.

albertkol (approved):

PR #8892, opened 2020-12-07
Integrate smartquotes package

anthonydillon:
## Done
- Integrate smartquotes package
- Change one quote on the homepage for testing

## QA
- Find on the page "Ubuntu supports the fastest" and the That’s at the beginning on the sentence should look nice even though its a straight quote in the markup.
- Insure there is no errors or page slowness


ClementChaumel (approved):

static/js/src/smartquotes.js
```diff
@@ -0,0 +1,3 @@
+var smartquotes = require("smartquotes");
+
+smartquotes(document.getElementById("main-content"));
```
ClementChaumel, line 3: Only focusing the `main-content` means there could be dumb quotes in the navigation and the footer still. Do we take a big performance hit if we apply it to the whole page?

anthonydillon:
My reasoning was that the navigation very rarely changes and it might cause issues in the complex navigation mark up. The main concern was the page content that often contain unencoded quotes. 

Happy to try it out.

pmahnke (approved):
+1

PR #8901, opened 2020-12-09
Add general /security/api to path

albertkol:
## Done

Do not merge until https://github.com/canonical-web-and-design/ubuntu-com-security-api/pull/18 is merged!

Changes the paths we link to the API app.

## Issue
Fixes https://github.com/canonical-web-and-design/ubuntu-com-security-api/issues/6

### DIFF
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -106,7 +106,7 @@ production:
         - name: SENTRY_DSN
           value: https://0293bb7fc3104e56bafd2422e155790c@sentry.is.canonical.com//13
 
-    - paths: [/security/cves\.json, /security/cves/.*\.json, /security/notices\.json, /security/notices/.*\.json, /security/spec\.json, /security/docs]
+    - paths: [/security/cves\.json, /security/cves/.*\.json, /security/notices\.json, /security/notices/.*\.json, /security/api/.*]
       service_name: ubuntu-com-security-api
 
   nginxConfigurationSnippet: |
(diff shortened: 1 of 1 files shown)


konf/site.yaml
```diff
@@ -106,7 +106,7 @@ production:
         - name: SENTRY_DSN
           value: https://0293bb7fc3104e56bafd2422e155790c@sentry.is.canonical.com//13
 
-    - paths: [/security/cves\.json, /security/cves/.*\.json, /security/notices\.json, /security/notices/.*\.json, /security/spec\.json, /security/docs]
+    - paths: [/security/cves\.json, /security/cves/.*\.json, /security/notices\.json, /security/notices/.*\.json, /security/api/.*]
```
jkfran, line 109: @albertkol @nottrobin I have no context about this but why not just everything under `/security/api/`, like `/security/api/cves.json`, `/security/api/cve/xxxx.json`, etc...?

jkfran (approved):
Look good, I added a small question.

konf/site.yaml
albertkol: We can have this discussion separately, as I want to get this merged ASAP.

@nottrobin can go into more details. The idea was that it would look neat to have:
https://ubuntu.com/security/notices - HTML view of the notices from the website
https://ubuntu.com/security/notices.json - Json view of the notice using the API

I think the only potential issue will be if we ever find ourselves releasing a v2 of the api. People tend to add the version number in the url too.

It is a bit annoying having to amend this `site.yaml` every time we will have a new endpoint that does not fall in the current rules set. 

We should maybe discuss this tomorrow during the **Tooling for independent services** meeting.

PR #8904, opened 2020-12-10
Change footer design

### LINKED ISSUE #8877: Visual design: Global vs. contextual footer borders seem reversed
This has been bothering me for a few years. On ubuntu.com, all pages (that I’ve seen) have a global footer:

image

Many, but not all, also have a “contextual footer” above the global footer:

image

The contextual footer’s top border (a 0.1875rem orange strip) is much more vivid than the global footer’s top border (a light grey drop shadow).

**The problem:** On pages that don’t have a contextual footer — like [/kernel](https://ubuntu.com/kernel), or [/advantage/subscribe](https://ubuntu.com/advantage/subscribe) — there is, therefore, much less border after the main page contents than on pages that do. Where the last section happens to have a grey background — as in these two examples — there seems to be no particular border before the sitemap at all, so it could be misinterpreted as part of the page body.

image

**I suggest:** This can be fixed by swapping the two borders — the contextual footer should have the grey shadow border (or maybe something else entirely), while the global footer has the orange strip.

As a bonus, this could lead us to pay more attention to the contents of the contextual footer when it is present, checking that it’s relevant and doesn’t duplicate material from the rest of the page.

---
*Reported from: https://ubuntu.com/kernel*

albertkol:
Switch global and contextual footer borders.

## QA

- Go to https://ubuntu-com-8904.demos.haus/support/community-support scroll down and take a look at the contextual and the global footer the borders should be switched.

## Issue / Card

Fixes #8877

### DIFF
--- a/static/sass/_pattern_contextual-footer.scss
+++ b/static/sass/_pattern_contextual-footer.scss
@@ -1,6 +1,6 @@
 // contextual footer
 @mixin ubuntu-p-contextual-footer {
   .p-contextual-footer {
-    @include vf-highlight-bar($color-brand);
+    box-shadow: 0 4px 4px -4px rgba(0, 0, 0, 0.3) inset;
   }
 }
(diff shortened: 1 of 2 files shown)


bethcollins92 (approved):
LGTM! that looks so much better 👍🏼

PR #8906, opened 2020-12-10
Regex for proxy locations

jkfran:
## Done

- Regex for proxy locations

## QA

- Check staging.ubuntu.com is working as usual, I deployed this config there to check.


tbille (approved):
Works well

PR #8920, opened 2020-12-14
Add RSS metatag

tbille:
## Done

- Add RSS metatag for blog

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/blog
- Make sure that blog index page has the metatag for rss feed


carkod (approved):
LGTM

PR #8952, opened 2020-12-18
Show status of each microcert for user

jpmartinspt:
## Done
Fetch the status for each course from the edx API.

## QA

- dotrun
- go to `cube//microcerts`
- check all courses have the correct status defined.


bethcollins92 (approved):
LGTM 👍🏼

PR #8954, opened 2020-12-19
Enable the card field on the Silo browser

anthonydillon:
## Done
Remove the font src when initialising the card input element.

**This unblocks the US Navy and other institutions that are required to use Silo browser**

## QA
- Open the demo and check the card input is working as live
- Go to https://www.authentic8.com/company/ and click "Try Silo risk free."
- Then enter the demo link into the simulated browser and check the card input works 
- If you want to switch to production in Silo and see that the card input is not rendered.

### DIFF
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -85,8 +85,6 @@ const elements = stripe.elements({
   fonts: [
     {
       family: "Ubuntu",
-      src:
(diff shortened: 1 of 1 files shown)


jaycee:
I am flabbergasted that loading a font makes the browser block the element, but one can't argue with results. QA confirmed, the silo browser works for me w/ the demo link.

jaycee (approved):

nottrobin (approved):

PR #8983, opened 2021-01-06
Get list of assertions for user from Badgr

jpmartinspt:
## Done
Get a list of assertions from badgr to know what badges a certain user has and use those assertions to work out if I user has passed courses or is already CUBE certified.

## QA

- Pair with me to QA


jkfran (approved):
Code looks good

PR #8984, opened 2021-01-06
Cube EDX API proxy settings

jpmartinspt:
## Done
This CUBE edx API will live under a sub-domain of ubuntu.com but requests to it still need proxying, so we configure the session manually to avoid it loading the configurations from environment variables, since those default to not proxy requests for ubuntu.com sub -domains and that is  the intended behaviour for most of our apps.

## QA

- make sure `https://ubuntu-com-8984.demos.haus/cube/microcerts` loads.

### DIFF
--- a/webapp/cube/views.py
+++ b/webapp/cube/views.py
@@ -2,6 +2,7 @@ import os
 import flask
 import talisker.requests
 import yaml
+from requests import Session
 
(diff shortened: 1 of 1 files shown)


bethcollins92:
@jpmartinspt QA LGTM! 👍🏼

jkfran (approved):
Looks good! 👍🏻

PR #8988, opened 2021-01-07
Get user enrollments

jpmartinspt:
## Done
Get the user enrollments and use them to fix the logic that calculates the status of the course for the authenticated user.
When you are unenrolled from a course you can still have a grade in the system for it, so we need an extra check.

## QA

- dotrun
- go to `/cube/microcerts`
- make sure your courses statuses look correct

### DIFF
--- a/webapp/cube/views.py
+++ b/webapp/cube/views.py
@@ -15,7 +15,7 @@ with open("webapp/cube/mappings/courses.yaml", "r") as stream:
 # This API lives under a sub-domain of ubuntu.com but requests to it
 # still need proxying, so we configure the session manually to avoid
 # it loading the configurations from environment variables, since those
-# deafult to not proxy requests for ubuntu.com sub-domains and that is
+# default to not proxy requests for ubuntu.com sub-domains and that is
 # the intended behaviour for most of our apps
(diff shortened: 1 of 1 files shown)


bethcollins92:
QA LGTM 👍🏼

webapp/cube/views.py
```diff
@@ -50,15 +55,17 @@ def cube_microcerts():
...
-                status = "passed" if grade["passed"] else "enrolled"
+            if course_id in enrollments:
+                status = "enrolled"
+                try:
+                    grade = cube_api.get_course_grades(course_id, username)[0]
+                    if grade["passed"] and grade["username"] == username:
+                        status = "passed"
+                except KeyError:
+                    # There is no grade the user is just "enrolled"
+                    pass
```
jkfran, line 66: There is no need to put
```python
grade = cube_api.get_course_grades(course_id, username)[0]
```
inside this try block, right?

Also can we not change the condition to avoid the try/except?

jpmartinspt: Update do `IndexError` so it probably makes more sense now :smile:

jkfran (approved):
LGTM

webapp/cube/views.py
jkfran: I see, okay!

PR #8999, opened 2021-01-08
Revert "Remove the engage takeover on the homepage"

anthonydillon:
Reverts canonical-web-and-design/ubuntu.com#8994

PR #9000, opened 2021-01-08
Handle discourse API errors for homepage takeovers

carkod:
## Done

- This should catch any Http exceptions and return a simple that will be handled by template finder
- 
## QA

- I wasn't able to test it using real 429 error, we instead used a fake endpoint
May need pairing up to QA this

## Issue / Card

Temporarily Fixes #3615

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -4,6 +4,7 @@ A Flask application for ubuntu.com
 
 # Packages
 import os
+import requests
 import talisker.requests
 import flask
 from canonicalwebteam.flask_base.app import FlaskBase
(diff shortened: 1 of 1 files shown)


jkfran:
I don't think this is what we agree to do:

1.
```python
app.add_url_rule("/", view_func=build_takeovers(engage_pages))
```
Should be removed and let templatefinder serve the index page.

2. We need to create a JSON route that serves the takeovers.
3. We need JS to consume this new JSON endpoint and show that data in the index page.

carkod:
Superseeded by #8999

PR #9007, opened 2021-01-11
Switch account name property to the right case

### LINKED ISSUE #8997: Fix UA form data not sending correctly
When purchasing the account name is set to the user email instead of the user name

ClementChaumel:
## Done

- Fix the case of `account_name`

## QA

- View the site https://ubuntu-com-9007.demos.haus/advantage/subscribe?test_backend=true
- Make a purchase using a new email and check in the network tab that the right username is used in the customer-info response

## Issue / Card

Fixes #8997

### DIFF
--- a/static/js/src/advantage/contracts-api.js
+++ b/static/js/src/advantage/contracts-api.js
@@ -29,7 +29,7 @@ export async function ensurePurchaseAccount(
     },
     body: JSON.stringify({
       email: email,
-      accountName: accountName,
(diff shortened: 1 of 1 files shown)


frankban (approved):
Looks good as it is so that it can be deployed as quickly as possible to production.
I'd suggest a follow up branch with a regression test.
Thanks a lot for the quick fix!

anthonydillon (approved):

PR #9012, opened 2021-01-12
Create LSN template

### LINKED ISSUE #9005: Create LSN page template
## Summary

Create LSN template, `lsn.html`. The default `notice.html` will become `usn.html`.  

## Screenshot

https://docs.google.com/document/d/183fLUXOLL17_baS617aX7Sj133n4lLfS6-rLhXXv9I0/edit#heading=h.bbi2iqqwhp35

albertkol:
## Done

- Create LSN template compare to https://docs.google.com/document/d/183fLUXOLL17_baS617aX7Sj133n4lLfS6-rLhXXv9I0/edit#heading=h.bbi2iqqwhp35 to confirm that everything is correct

## QA

1. Have the database with an LSN

2. Go to your ubuntu.com repo and run:
```bash
docker-compose down # this resets your databse
docker-compose up -d # NOTE: do not run `dotrun`

# Now to import the databse
docker exec -i db psql -U postgres -d postgres < ~/Downloads/backup.sql # replace with correct path

dotrun # run site
```

3. Browse to http://0.0.0.0:8001/security/notices/LSN-0000-0 and compare to the image in the docs

## Issue / Card

Fixes #9005

### DIFF
--- a/templates/security/notices/lsn.html
+++ b/templates/security/notices/lsn.html
@@ -1,0 +1,194 @@
+{% extends "security/base_security.html" %}
+
+{% block title %}{{ notice.id }}: {{ notice.title }} | Ubuntu security notices{% endblock %}
+
+{% block content %}
+<section class="p-strip--suru-topped">
+  <div class="row">
+    <div class="col-12">
+	    <h1>{{ notice.id }}: {{ notice.title }}</h1>
+      <p class="p-muted-heading">{{ notice.published.strftime("%d %B %Y") }}</p>
+      <p>{{ notice.summary | safe }}</p>
+      <h2>Releases</h2>
+      <ul class="p-inline-list u-no-margin--bottom">
+        {% for release in notice.releases %}
+          <li class="p-inline-list__item">
+            <a href="/security/notices?release={{ release.codename }}" class="p-button--tag">Ubuntu {{ release.version }} {{ release.support_tag }}</a>
+          </li>
+        {% endfor %}
+      </ul>
+    </div>
+  </div>
+
+  {% if notice.package_descriptions %}
+  <div class="row">
+    <div class="col-12">
+      <h2>Packages</h2>
+      <ul class="p-list">
+        {% for package_name, package_title in notice.package_descriptions.items() %}
+          <li class="p-list__item"><a href="/security/cve?package={{ package_name }}">{{ package_name }}</a> - {{ package_title }}</li>
+        {% endfor %}
+      </ul>
+    </div>
+  {% endif %}
+
+  <div class="row">
+    <div class="col-8">
+      <h2>Details</h2>
+      <p>{{ notice.details|safe }}</p>
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="col-8">
+      <h2>Checking update status</h2>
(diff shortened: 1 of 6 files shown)


templates/security/notices/lsn.html
```diff
@@ -0,0 +1,194 @@
...
+            </tr>
+          {% endfor %}
+        </tbody>
+      </table>
+      <p>To check your kernel type and Livepatch version, enter this command:</p>
+      <div class="p-code-copyable">
+        <input class="p-code-copyable__input" value="canonical-livepatch status" readonly="readonly">
+        <button class="p-code-copyable__action">Copy to clipboard</button>
+      </div>
+    </div>
```
albertkol, line 73: @sowasred2012 I am trying to add the copyable element on this page. I have copy-pasted the js script at the bottom. Was that right, or was I supposed to import the script neatly?

sowasred2012: I don't think you need to add or import anything actually - the necessary JS is already present in the project [here](https://github.com/canonical-web-and-design/ubuntu.com/blob/master/static/js/src/scratch.js#L12), so as long as main.js is on the page, you should have it.

albertkol: I investigated a bit, because `main.js` is already present on my lsn template page, and as soon as I delete the block of javascript, the copy does not work anymore. 

I have detected the usage of the `p-code-copyable` on these pages:
https://ubuntu.com/kubernetes/install
https://ubuntu.com/internet-of-things/edgex
https://ubuntu.com/ai/install-kubeflow
https://ubuntu.com/openstack/install

Out of these, the only one working is `/internet-of-things/edgex` because @anthonydillon  also added the script on the page.

`main.js` has some `p-code-copyable` related js inside `/static/scratch.js`, but it seems to be not working.

So I will take the liberty to to replace the code with the script from the docs: https://vanillaframework.io/docs/base/code#copyable
(also I will delete ant's copyable code from `/internet-of-things/edgex`)

As a result all the above links' copyable elements started to work.

templates/security/notices/lsn.html
```diff
@@ -0,0 +1,139 @@
...
+  </div>
+
+  {% if notice.release_packages %}
+    <div class="row">
+      <div class="col-8">
+        <h2>Support Information</h2>
+        <p>Kernels older than the levels listed below do not receive livepatch updates. If you are running a kernel version earlier than the one listed below, please upgrade your kernel as soon as possible.</p>
+        {% for version in notice.release_packages.__reversed__() %}
+          <h5>Ubuntu {{ version }}</h5>
```
sowasred2012, line 82: Descending headings should step down sequentially, the previous heading was a `h2`, so this should be a `h3` - but I guess you need the `h5` style, so you could use:
```suggestion
          <h3 class="p-heading--five">Ubuntu {{ version }}</h3>
```

sowasred2012 (changes requested):
Looks good, checked it locally and it works, just one semantic suggestion.

templates/security/notices/lsn.html
albertkol: Fixed! @sowasred2012

PR #9013, opened 2021-01-12
CVE results cells: Make “Does not exist” more compact

### LINKED ISSUE #8832: CVE tables: “Does not exist” wraps to 3 lines
CVE results tables are much less information-dense than they could be, because many of the cells contain the text “Does not exist”, and that text wraps to three lines.

image

(These should also be the least prominent cells in the table, but being almost the _only_ cells that wrap to three lines challenges this.)

These cells have plenty of width for “Does not exist” to occupy two lines rather than three:

image

It doesn’t use this space because it is leaving room for an “ℹ️” icon alongside. However:
- Cells generally don’t need to leave room for this icon; it could float over top on hover.
- There’s no point in “Does not exist” cells particularly leaving room for this icon, because they don’t have it anyway!

What should happen: either
- no cells should leave room for the “ℹ️” icon, because it floats over top if necessary; or
- “Does not exist” cells in particular shouldn’t leave room for the “ℹ️” icon, because they doesn’t have it.

---
*Reported from: https://ubuntu.com/security/cve*

matthewpaulthomas:
## Done

- Applied semantic classes to the CVE cells, for easier styling changes later
- Removed the invisible (i) icon from cells where it never appears
- Expanded the cell text to use the space taken by the never-appearing icon

## QA

- Observe that the “Recent CVEs” table is now more compact

## Issue / Card

Fixes #8832.

## Screenshots

I cannot test this or make screenshots because there is no local access to the CVE database.

### DIFF
--- a/static/sass/_pattern_cve.scss
+++ b/static/sass/_pattern_cve.scss
@@ -107,7 +107,7 @@
     border-bottom: 10px solid #b0b0b0;
   }
 
-  .cve-table-cell--muted {
+  .cve-table-cell.status-DNE, .cve-table-cell.status-not-affected {
     @extend .cve-table-cell;
 
     color: $color-mid-dark;
--- a/static/sass/_pattern_cve.scss
+++ b/static/sass/_pattern_cve.scss
@@ -46,7 +46,7 @@
   .cve-table .icon-container__text {
     width: calc(100% - 3rem);
   }
-  
+
   .cve-table .status-DNE .icon-container__text {
     width: calc(100% - 1.5rem);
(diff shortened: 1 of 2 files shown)


matthewpaulthomas:
I don’t know why prettier is still complaining about the SCSS: when I put it into [the Prettier Playground](https://prettier.io/playground) it makes no changes.

albertkol:
This will be done as part of https://github.com/canonical-web-and-design/ubuntu.com/pull/9036

This PR can be closed.

PR #9015, opened 2021-01-12
Update subscription call to use filters

### LINKED ISSUE #9004: [UA-shop] error making purchases with a canceled subscription

## Summary
I am unable to make purchase on staging if my account contains a canceled subscription.

## Process
1. Cancel an active subscription on UA shop
2. try to make a new purchase with the same account as above

With the above steps done, i would be unable to make a new purchase as the shop. This is because the shop chooses the first subscription from the get subscription endpoint(which may not be active), leading to the error during purchase.

https://github.com/canonical-web-and-design/ubuntu.com/blob/499749b55a023778c89f1b4dd7a583c38347cf87/webapp/advantage.py#L153-L164

affected code locations:
https://github.com/canonical-web-and-design/ubuntu.com/blob/76657087f79db6b5956bd2594d450d06ab706fcd/webapp/views.py#L647
https://github.com/canonical-web-and-design/ubuntu.com/blob/76657087f79db6b5956bd2594d450d06ab706fcd/webapp/views.py#L795

this should be updated to use the filters added on the backend as documented below.
https://github.com/CanonicalLtd/ua-contracts/blob/e43c466be115a3b9388f7434db6753d43a3e5bcd/docs/contracts.yaml#L2599-L2614

we should update the call to get the subscription to the one below to ensure we can get the current active subscription for use in the purchase flow.
```
https://<sontracts-server>/v1/accounts/<account-id>/marketplace/<marketplace>/subscriptions?period=yearly&status=active
```

The endpoint above should be hard coded until we have end to end support for alternative subscriptions periods.

## Screenshot

## Browser details

[Optionally - if you can, copy the report generated by [mybrowser.fyi](https://mybrowser.fyi/) - this might help us debug certain types of issues.]

ClementChaumel:
## Done

- Added yearly and active filters to the subscription call to make sure we are not looking at cancelled subscriptions

## QA

Not too sure how to QA this honestly. The ticket mentions cancelling an active subscription on the UA shop but I don't think we can do this yet via the web interface? (@sowasred2012 Is that right?) So I guess just go to the link, log in and try to make a purchase

- https://ubuntu-com-9015.demos.haus/advantage/subscribe?test_backend=true

## Issue / Card

Fixes #9004

### DIFF
--- a/webapp/advantage.py
+++ b/webapp/advantage.py
@@ -157,7 +157,7 @@ class AdvantageContracts:
             method="get",
             path=(
                 f"v1/accounts/{account_id}"
-                f"/marketplace/{marketplace}/subscriptions"
+                f"/marketplace/{marketplace}/subscriptions?period=yearly&status=active"
             ),
         )
 
--- a/webapp/advantage.py
+++ b/webapp/advantage.py
@@ -157,10 +157,10 @@ class AdvantageContracts:
             method="get",
             path=(
                 f"v1/accounts/{account_id}"
-                f"/marketplace/{marketplace}/subscriptions?period=yearly&status=active"
+                f"/marketplace/{marketplace}"
+                f"/subscriptions?period=yearly&status=active"
             ),
         )
-
         return response.json()
 
     def get_account_purchases(self, account_id: str) -> dict:


sowasred2012:
@ClementChaumel that's right, there's no way to cancel them via the dashboard. In terms of QAing, the CS team might be able to create the data for a given user?

ClementChaumel:
Thanks! @brotich can you confirm it works as expected?

brotich:
> Thanks! @brotich can you confirm it works as expected?

thanks @ClementChaumel and @sowasred2012 I will check and confirm if this works as expected. 

btw to cancel a subscription, one needs to make an API call as describe here https://pastebin.canonical.com/p/C5V65jPJSm/

brotich:
Thanks for the change. In the process of testing this, i found a subtle bug from the backend. therefore this may be open until monday/tuesday as i work on a fix on the backend.

brotich:
@ClementChaumel This works as expected. On your end, there should be no change in the purchase flow.

this will become more relevant when the canceled subscriptions are implemented. 

Thanks for working on this

ClementChaumel:
@brotich Great news Thanks for testing

albertkol (approved):
Just needs a rebase to the latest master and should be good to go!

PR #9019, opened 2021-01-13
redirect /guide

nottrobin:
## QA

Check you get a redirect at /server/guide

``` bash
dotrun

# Then in another terminal
curl -sI http://127.0.0.1:8001/server/guide | grep -i location
```

Check you see a location header for `/server/docs`.


anthonydillon (approved):
LGTM

PR #9036, opened 2021-01-15
Optimise CVE table cells

### LINKED ISSUE #8832: CVE tables: “Does not exist” wraps to 3 lines

albertkol:
## Done

- Optimise CVE table cells

## QA

Go to /security/cve look at the table it should have:
- `Does not exist` on 2 lines instead of 3
- Update the UA texts under the status, inside the cell and they should be smaller
- It should look better than https://ubuntu.com/security/cve

## Issue / Card

Fixes #8832
Fixes #9036

### DIFF
--- a/static/sass/_pattern_cve.scss
+++ b/static/sass/_pattern_cve.scss
@@ -47,6 +47,8 @@
   }
 
   .cve-table .icon-container__text {
+    overflow: hidden;
+    text-overflow: ellipsis;
     width: calc(100% - 1.5rem);
   }
 
--- a/static/sass/_pattern_cve.scss
+++ b/static/sass/_pattern_cve.scss
@@ -40,11 +40,14 @@
   }
 
   .cve-table .icon-cve-info {
-    margin: 0 0 0 0.5rem;
+    margin: 0;
+    position: absolute !important;
+    right: 0;
+    top: calc(0.5rem - 1px);
   }
 
   .cve-table .icon-container__text {
-    width: calc(100% - 3rem);
+    width: calc(100% - 1.5rem);
   }
 
   .cve-table .p-icon--information {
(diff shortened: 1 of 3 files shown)


static/sass/_pattern_cve.scss
```diff
@@ -71,6 +74,16 @@
     position: relative;
   }
 
+  .cve-table .esm-info {
+    width: 100%;
```
sowasred2012, line 78: Linting - these should be in alphabetical order

static/sass/_pattern_cve.scss
```diff
@@ -40,11 +40,14 @@
   }
 
   .cve-table .icon-cve-info {
-    margin: 0 0 0 0.5rem;
+    margin: 0;
+    position: absolute !important;
+    top: calc(0.5rem - 1px);
+    right: 0;
```
sowasred2012, line 46: Same here, alphabetical order

sowasred2012 (changes requested):
Couple of linting issues, and there's a strange 1px border in the gap between the last two columns:
Screenshot from 2021-01-18 12-26-11

static/sass/_pattern_cve.scss
albertkol: Fixed!

sowasred2012:
Updates look good, but I just noticed that longer words break out of their cells at smaller browser widths:

Screenshot from 2021-01-19 10-18-05

We could fix this by adding `overflow: hidden; text-overflow: ellipsis;` - unless there are any objections. @matthewpaulthomas ?

Screenshot from 2021-01-19 10-19-12

sowasred2012 (approved):
LGTM!

PR #9037, opened 2021-01-15
Fix JS error on renewal form blocking purchases

### LINKED ISSUE #9018: Handle Stripe failing to initialise
A customer saw this while trying to renew - the odd looking card input suggests Stripe failed to initialise, possibly because of an aggressive adblocker blocking stripe domains. Whatever the cause, it leaves the customer unable to make a purchase - perhaps we could show a message informing the user there was a problem, and point them to a sales contact?

sowasred2012:
## Done

- Check for existence of "I'm purchasing for" fields before attaching events to them

## QA

- View the site in your web browser at: https://ubuntu-com-9037.demos.haus/advantage?test_backend=true
- Log in with your SSO account
- If you don't see a contract up for renewal, ask @alnvdl to create one for you
- When you have that, find it in the table, open the dropdown, and click "Renew"
- The payment modal should launch and the Stripe field should render correctly

This is related to, but doesn't fix #9018 - this fixes the problem the user was seeing, we should still handle Stripe failing to initialise.

## Screenshots

Before:
Screenshot from 2021-01-15 14-13-16

After:
Screenshot from 2021-01-15 14-11-57

### DIFF
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -248,8 +248,13 @@ function attachFormEvents() {
     checkVAT();
   });
 
-  forMyselfRadio.addEventListener("change", handleNameFieldRadio);
-  forOrganisationRadio.addEventListener("change", handleNameFieldRadio);
+  // these elements aren't rendered on the renewal form
+  // or if an account isn't present, so check for their
+  // existence first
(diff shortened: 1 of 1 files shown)


ClementChaumel (approved):
:pray: :pray: :pray:

PR #9041, opened 2021-01-15
JS homepage takeover with endpoint to avoid taking down homepage

carkod:
## Done

- Make sure Homepage doesn't go down if discourse APIs fail.

## QA

- To test robustness change the `/takeovers.json` endpoint to anything else, and you will see the default 20.04 takeover.


carkod:
This PR is not reacting to my force pushes, looks like wrong branch. Replaced by #9042

PR #9042, opened 2021-01-15
JS homepage takeover with endpoint to avoid taking down homepage

carkod:
## Done

- Make sure Homepage doesn't go down if discourse APIs fail.

## QA

- To test robustness change the `/takeovers.json` endpoint to anything else, and you will see the default 20.04 takeover.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/web-squad/issues/3615

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -819,8 +819,8 @@ body {
 .p-takeover-animation {
   animation: FadeIn map-get($animation-duration, "fast")
     map-get($animation-easing, "out");
-  animation-fill-mode: backwards;
   animation-delay: 2.5s;
+  animation-fill-mode: backwards;
 
   &.is-loaded {
     animation-delay: 0s !important;
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -806,7 +806,6 @@ body {
   }
 }
 
-// Homepage takeover animation
 @keyframes FadeIn {
   0% {
     opacity: 0;
@@ -816,14 +815,42 @@ body {
   }
 }
 
+@keyframes SlideInFromRight {
+  0% {
+    transform: translateX(0.5rem);
+  }
+  100% {
+    transform: translateX(0);
+  }
+}
+
+@keyframes SlideInFromLeft {
+  0% {
+    transform: translateX(-0.5rem);
+  }
+  100% {
+    transform: translateX(0);
+  }
+}
+
 .p-takeover-animation {
-  animation: FadeIn map-get($animation-duration, "fast")
-    map-get($animation-easing, "out");
+  animation: FadeIn map-get($animation-duration, "snap")
+    map-get($animation-easing, "in");
   animation-delay: 2.5s;
   animation-fill-mode: backwards;
 
   &.is-loaded {
     animation-delay: 0s !important;
+
+    .col-7 {
+      animation: SlideInFromLeft map-get($animation-duration, "fast")
+        map-get($animation-easing, "out");
+    }
+
+    .col-5 {
+      animation: SlideInFromRight map-get($animation-duration, "fast")
+        map-get($animation-easing, "out");
+    }
   }
 
   &.is-loading {
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -805,3 +805,24 @@ body {
     }
   }
 }
+
+// Homepage takeover animation
+@keyframes FadeIn {
+  0% {
+    opacity: 0;
+  }
+  100% {
+    opacity: 1;
+  }
+}
+
+.p-takeover-animation {
+  animation: FadeIn map-get($animation-duration, "fast")
+    map-get($animation-easing, "out");
+  animation-fill-mode: backwards;
+  animation-delay: 2.5s;
+
+  &.is-loaded {
+    animation-delay: 0s !important;
+  }
+}
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -825,4 +825,8 @@ body {
   &.is-loaded {
     animation-delay: 0s !important;
   }
+
+  &.is-loading {
+    animation-delay: 500s !important;
+  }
 }
--- a/templates/base_index.html
+++ b/templates/base_index.html
@@ -8,28 +8,28 @@
 
 {% block outer_content %}
 {% block content %}
-<template id="takeovers" class="u-hide">
+<template class="u-hide">
   {% block takeover_content %}
   {% endblock takeover_content %}
 </template>
 
 <!-- Default Takeover: Download the latest Ubuntu -->
-<section data-lang="all" id="takeover" class="p-strip--image is-dark is-deep" style="background-color: #5E2750; background-image: linear-gradient(-89deg, #e95420 0%, #772953 42%, #2c001e 94%)">
-  <div class="row">
+<section data-lang="all" id="takeover" class="p-takeover--grad is-dark is-deep">
+  <div class="row p-takeover-animation">
     <div class="col-7 u-vertically-center">
-      <h1>What's new in<br>Ubuntu {{ releases.latest.short_version }}?</h1>
-      <h4>Introducing the Ubuntu Desktop for Raspberry Pi, the latest desktop features and micro clouds.</h4>
-      <div class="u-hide--small">
-        <p><a href="/download" class="p-button--positive">Get Ubuntu {{ releases.latest.short_version }}</a></p>
+      <h1 id="takeover-title">What's new in<br>Ubuntu {{ releases.latest.short_version }}?</h1>
+      <h4 id="takeover-subtitle">Introducing the Ubuntu Desktop for Raspberry Pi, the latest desktop features and micro clouds.</h4>
+      <div id="takeover-ctas">
+        <p><a id="takeover-primary-url" href="/download" class="p-button--positive">Get Ubuntu {{ releases.latest.short_version }}</a></p>
         <p>
-          <a href="/raspberry-pi" class=" p-link--inverted">
+          <a id="takeover-secondary-url" href="/raspberry-pi" class=" p-link--inverted">
             Learn more about Ubuntu on the Raspberry Pi&nbsp;›
           </a>
         </p>
       </div>
     </div>
     <div class="col-5 u-vertically-center u-align--center u-align--left-medium u-hide--small">
-      {{  image(      url="https://assets.ubuntu.com/v1/fe951eda-20.10_Groovy+Gorilla_RPi_Sketch.svg",      alt="",      width="250",      height="388",      hi_def=True,      loading="lazy",      attrs={"style": "width: 250px; height: 388px;"},  ) | safe }}
+      <img id="takeover-image" src="https://assets.ubuntu.com/v1/fe951eda-20.10_Groovy+Gorilla_RPi_Sketch.svg" width="200" alt="" />
     </div>
   </div>
 </section>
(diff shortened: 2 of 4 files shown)


templates/base_index.html
```diff
@@ -1409,56 +1409,83 @@ <h2>
...
+    }
+    });
+  }
 
-      if (! document.body.contains(selectedTakeover)) {
-        // If it's not in the current document (as in, it's in a template) import it
-        // Note: This is for all browser *except* IE, which doesn't support <template>s
-        selectedTakeover = document.importNode(selectedTakeover, true);
-      }
+  function showTakeover(takeovers, index = 0) {
```
bartaz, line 1450: That's quite a lot of JS to keep inline on the page. Ideally it should be in a separate file.

And it seems it could use some transpiling as well. There are a lot of ES6 features that will fail in older browsers.

Most notably:
- default values of function parameters
- template literals ` `${...}` `

Other things that may fail in older IE
- const
- .classList API

If we want to keep it as inline script it should at least be written in most cross-browser way, avoiding unnecessary ES6 features.

templates/base_index.html
```diff
@@ -1409,56 +1409,83 @@ <h2>
...
     * choose one (that matches the client's language), and replace the
     * base template with it.
     */
 
-    var baseTakeover = document.getElementById('takeover');  // The base takeover element
-    var takeovers = document.getElementById('takeovers');  // The list of current takeovers to choose from
-
-    // For browsers that support <template> (all but IE), grab the template's ".content"
-    takeovers = 'content' in takeovers ? takeovers.content: takeovers;
+    fetch("/takeovers.json").then(response => response.json()).then(takeovers => {
```
bartaz, line 1417: Does ubuntu.com have `fetch` polyfill for older browsers or do we rely on native implementation - quite fresh browsers are required to use it natively: https://caniuse.com/?search=fetch

templates/base_index.html
carkod: Rewritten with old JS. Should support IE > 9

templates/base_index.html
carkod: Rewritten with XHR

bartaz: There is a well grounded fetch implementation. If you call APIs a lot probably worth adding as dependency: https://github.com/github/fetch

anthonydillon (changes requested):
The jump here is far too much. Can you try and think of a way to visually improve it? Maybe hide the base takeover content for a min before the new content arrives. Also can the default takeover use the new suru background, which might help.

anthonydillon:
## QA
### Desktop
- **IE11**: Not working
- **Chrome 87**: Pass
- **Chrome 79**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
- **Opera 73**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
- **Opera 63**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
- **Firefox 84**: Pass
- **Firefox 76**: Pass
- **Safari 12.1**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
- **Edge 87**: Pass
- **Edge 80**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)

### Mobile
- **iOS Safari**: Pass
- **iOS Chrome**: Pass
- **OnePlus 8 FF**: Pass

carkod:
> ## QA
> ### Desktop
> * **IE11**: Not working
> * **Chrome 87**: Pass
> * **Chrome 79**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
> * **Opera 73**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
> * **Opera 63**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
> * **Firefox 84**: Pass
> * **Firefox 76**: Pass
> * **Safari 12.1**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
> * **Edge 87**: Pass
> * **Edge 80**: Text is applied but the image is not updated (`cannot set property "src" of undefined`)
> 
> ### Mobile
> * **iOS Safari**: Pass
> * **iOS Chrome**: Pass
> * **OnePlus 8 FF**: Pass

 So it works on all browsers, but keyframe effects don't work on browserstack Opera and Chrome. But then I tried it on native Chrome and Opera and there are no problems.Similar with Safari.

anthonydillon:
@carkod thanks, three things from a quick QA.
- You should show the most common background as the white for such a long time looks broken. Just fade the content to the `.row`.
- When the JS has loaded the content it should immediately add a class which triggers the animation by removing the `animation-delay`.
- 1s animation is far to slow. Could you try `map.get($animation-duration, "fast")`

carkod:
> @carkod thanks, three things from a quick QA.
> 
> * You should show the most common background as the white for such a long time looks broken. Just fade the content to the `.row`.
> * When the JS has loaded the content it should immediately add a class which triggers the animation by removing the `animation-delay`.
> * 1s animation is far to slow. Could you try `map.get($animation-duration, "fast")`

Done, but I am not sure if I understood your second point, but maybe it's best that you take a look, when you can (a line a code is worth a thousand words). Thanks

anthonydillon:
@carkod What I'm suggesting is that once the content has been replaced then add a class like `is-loaded` to `p-takeover-animation`  element. Then something like this should remove the delay and make the content reveal.

```scss
.p-takeover-animation.is-loaded {
  animation-delay: 0s;
}
```

carkod:
> @carkod What I'm suggesting is that once the content has been replaced then add a class like `is-loaded` to `p-takeover-animation` element. Then something like this should remove the delay and make the content reveal.
> 
> ```css-scss
> .p-takeover-animation.is-loaded {
>   animation-delay: 0s;
> }
> ```

Oh well that's exactly what I did, but with a different class name

templates/base_index.html
```diff
@@ -1409,55 +1409,100 @@ <h2>
...
+    primaryUrl.textContent = selectedTakeover.primary_cta;
+    if (selectedTakeover.secondary_url && selectedTakeover.secondary_url !== "") {
+      secondaryUrl.href = selectedTakeover.secondary_url;
+      secondaryUrl.textContent = selectedTakeover.secondary_cta + " ›";
+    } else {
+      secondaryUrl.remove();
     }
+
+    // Remove animation delay
+    takeoverAnimation.className +=  ".is-loaded";
```
anthonydillon, line 1505: You get an array back from `getElementsByClassName` therefore you need:
```js
takeoverAnimation[0].className += " is-loaded";
```
or switch to using an id.

anthonydillon (dismissed):

templates/base_index.html
bartaz: There is also `element.querySelectorAll()` that returns all child elements for given selector or `element.querySelector()` that returns first child element of given selector.

carkod: Done. Additionally, I have added a fixed height (reduced a little the default takeover height), and centered aligned with flexbox. Tested on IE11, Opera 65 and Firefox 76 no major issues.

I'd rather avoid using query selectors, because you don't explicitly name them with a class or id. I would use id or class.

carkod:
Rebased

static/sass/styles.scss
```diff
@@ -805,3 +805,29 @@ body {
...
+    opacity: 0;
+  }
+  33% {
+    opacity: 0.33;
+  }
+  66% {
+    opacity: 0.66;
+  }
+  100% {
+    opacity: 1;
```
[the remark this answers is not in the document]
carkod: I did that way to make it smoother, but it doesn't seem to make a big impact. Done.

static/sass/styles.scss
```diff
@@ -805,3 +805,29 @@ body {
...
+  }
+}
+
+.p-takeover-animation {
+  animation: FadeIn map-get($animation-duration, "fast") linear;
+  animation-fill-mode: both;
```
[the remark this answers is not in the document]
carkod: Done with `backwards`

static/sass/styles.scss
```diff
@@ -805,3 +805,29 @@ body {
...
+  66% {
+    opacity: 0.66;
+  }
+  100% {
+    opacity: 1;
+  }
+}
+
+.p-takeover-animation {
+  animation: FadeIn map-get($animation-duration, "fast") linear;
```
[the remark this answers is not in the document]
carkod: Done

templates/base_index.html
```diff
@@ -8,28 +8,28 @@
...
+<section data-lang="all" id="takeover" class="p-takeover--grad is-dark is-deep">
+  <div class="row p-takeover-animation">
     <div class="col-7 u-vertically-center">
-      <h1>What's new in<br>Ubuntu {{ releases.latest.short_version }}?</h1>
-      <h4>Introducing the Ubuntu Desktop for Raspberry Pi, the latest desktop features and micro clouds.</h4>
-      <div class="u-hide--small">
-        <p><a href="/download" class="p-button--positive">Get Ubuntu {{ releases.latest.short_version }}</a></p>
+      <h1 id="takeover-title">What's new in<br>Ubuntu {{ releases.latest.short_version }}?</h1>
+      <h4 id="takeover-subtitle">Introducing the Ubuntu Desktop for Raspberry Pi, the latest desktop features and micro clouds.</h4>
+      <div id="takeover-ctas" class="u-hide--small">
```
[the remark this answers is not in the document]
carkod: Thanks. Done.

anasereijo:
It's taking a long time to load the images and the text every time I refresh the page.

Also, why are the illustrations smaller now? It doesn't loo right in some of the takeovers.

thank

templates/base_index.html
```diff
@@ -1409,55 +1409,103 @@ <h2>
...
+    // Set language attributes
+    takeover.setAttribute("lang", selectedTakeover.lang);
+    takeover.setAttribute("lang-skip", selectedTakeover.lang_skip);
+
+    // Set takeover content
+    title.textContent = selectedTakeover.title;
+    subtitle.textContent = selectedTakeover.subtitle;
+    image.src = selectedTakeover.image;
+    image.removeAttribute("style");
+    image.style.width = selectedTakeover.image_width;
```
pmahnke, line 1493: I think this might not be overriding the parameter `width=""`, perhaps the default should have `style="width:200px;"` not `width=""`

templates/base_index.html
```diff
@@ -8,28 +8,28 @@
...
-          <a href="/raspberry-pi" class=" p-link--inverted">
+          <a id="takeover-secondary-url" href="/raspberry-pi" class=" p-link--inverted">
             Learn more about Ubuntu on the Raspberry Pi&nbsp;›
           </a>
         </p>
       </div>
     </div>
     <div class="col-5 u-vertically-center u-align--center u-align--left-medium u-hide--small">
-      {{  image(      url="https://assets.ubuntu.com/v1/fe951eda-20.10_Groovy+Gorilla_RPi_Sketch.svg",      alt="",      width="250",      height="388",      hi_def=True,      loading="lazy",      attrs={"style": "width: 250px; height: 388px;"},  ) | safe }}
+      <img id="takeover-image" src="https://assets.ubuntu.com/v1/fe951eda-20.10_Groovy+Gorilla_RPi_Sketch.svg" width="200" alt="" />
```
pmahnke, line 32:
```suggestion
      <img id="takeover-image" src="https://assets.ubuntu.com/v1/fe951eda-20.10_Groovy+Gorilla_RPi_Sketch.svg" style="width: 200px;" alt="" />
```
for the js to work, does it need to use a style not a width parameter?

templates/base_index.html
pmahnke: I don't see how this is changing the markup. Is it correct?

templates/base_index.html
```diff
@@ -1409,55 +1409,103 @@ <h2>
...
+    takeover.setAttribute("lang", selectedTakeover.lang);
+    takeover.setAttribute("lang-skip", selectedTakeover.lang_skip);
+
+    // Set takeover content
+    title.textContent = selectedTakeover.title;
+    subtitle.textContent = selectedTakeover.subtitle;
+    image.src = selectedTakeover.image;
+    image.removeAttribute("style");
+    image.style.width = selectedTakeover.image_width;
+    image.style.height = selectedTakeover.image_height;
```
pmahnke, line 1494:
```suggestion
    image.height = selectedTakeover.image_height;
```

anthonydillon:
@anasereijo @pmahnke I have added an interrupt to the delay when the load is complete. I am awaiting the load of the hero image too so might be able to improve further.

templates/base_index.html
anthonydillon: Done, yes this is correct.

anasereijo:
@anthonydillon looking much better now, the loading and the size of illustrations. thanks for fixing it

pmahnke (approved):
+1 thanks guys!

PR #9055, opened 2021-01-18
Focus quantity field when product is selected

### LINKED ISSUE #8247: /advantage/subscribe: Choosing machine type doesn’t focus quantity field
1\. Open /advantage/subscribe on an [`advantage-shop`](https://github.com/canonical-web-and-design/ubuntu.com/tree/advantage-shop) branch (for example, https://ubuntu-com-8244.demos.haus/advantage)
2\. Choose Servers, Other VMs, or Desktops

What happens:
- The page scrolls to the “How many” field

What should happen:
- The page scrolls to the “How many” field
- The field is focused so that you can type in it

[Discovered while testing #8244.]

steverydz:
## Done

Focus quantity field when option has been selected on the subscribe page

## QA

- Go to https://ubuntu-com-9055.demos.haus/advantage/subscribe
- Select either "Servers", "Other VMs", or "Desktops"
- Check that the page scrolls so the "How many ..." field is visible and that it is focused
- Select either "AWS" or "Microsoft Azure" and check that the page doesn't scroll or focus the "How many..." field

## Issue / Card

Fixes #8247

### DIFF
--- a/static/js/src/ua-product-selector.js
+++ b/static/js/src/ua-product-selector.js
@@ -344,6 +344,7 @@ function productSelector() {
         } else {
           quantityTypeEl.innerHTML = `How many ${inputElement.dataset.productName}s?`;
           state.set(inputElement.name, [inputElement.value]);
+          focusQuantityField();
         }
 
         break;
@@ -439,6 +440,11 @@ function productSelector() {
     });
 
     input.closest(".p-card--radio").classList.add("is-selected");
+  }
+
+  function focusQuantityField() {
+    const quantityField = document.querySelector(".js-product-quantity");
+    quantityField.focus();
   }
 
   function setVersionTabs() {
--- a/templates/advantage/subscribe/index.html
+++ b/templates/advantage/subscribe/index.html
@@ -121,11 +121,11 @@
           </div>
         </div>
 
-        <div id="aws" class="col-12 u-hide js-public-cloud-info" style="background-color: #f7f7f7; padding: 1.5rem; margin-bottom: 1.5rem">
+        <div id="aws" class="col-12 u-hide js-public-cloud-info" style="background-color: #f7f7f7; margin-bottom: 1.5rem; padding: 1.5rem">
           <span>You can buy <a href="https://aws.amazon.com/marketplace/search/results?page=1&filters=VendorId&VendorId=e6a5002c-6dd0-4d1e-8196-0a1d1857229b&searchTerms=ubuntu+pro" class="p-link--external">Ubuntu Pro on the AWS Marketplace</a> at an hourly, per-machine rate, with all UA software features included. If you need tech support as well, <a href="/support/contact-us">contact us</a>.</span>
         </div>
 
-        <div id="azure" class="col-12 u-hide js-public-cloud-info" style="background-color: #f7f7f7; padding: 1.5rem; margin-bottom: 1.5rem">
+        <div id="azure" class="col-12 u-hide js-public-cloud-info" style="background-color: #f7f7f7; margin-bottom: 1.5rem; padding: 1.5rem">
           <span>You can buy <a href="https://azuremarketplace.microsoft.com/en-us/marketplace/apps?search=Ubuntu%20Pro&page=1" class="p-link--external">Ubuntu Pro on the Azure Marketplace</a> at an hourly, per-machine rate, with all UA software features included. If you need tech support as well, <a href="/support/contact-us">contact us</a>.</span>
         </div>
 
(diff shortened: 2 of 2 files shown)


ClementChaumel (approved):
LGTM :+1:

elioqoshi:
@steverydz Thanks!

I just tried it and while technically it works fine, there are 2 small things which could be improved.

1. the scrolling goes abruptly to the relevant section. Would it be possible to have a smoother and slightly slower transition with an ease-in duration? It would help guide the user's eyes and track the site changes.
2. Currently the scrolling leads to the "How many" section as soon as it appears on the view. The expected behaviour should be that the "How many" section ends up at the top of the page. Screenshots for context:

### Currently

image


### Expected behaviour

image

Let me know if these make sense!

elioqoshi:
I just noticed another issue (in 2 places) in terms of keyboard accessibility in navigation. 

1. When selecting Servers, Other VMs or Desktops with the arrow keys, focus goes automatically to the "How many servers" section. Expected behaviour should be that the user needs to confirm with the Tab key to confirm and continue to the next section. I am unsure if this is intended, but this behavior would be worse accessibility-wise than what is already on the Ubuntu.com production site currently (you select the kind of machine you want, press Tab and the "How many servers" section is focused.

2. This also seems to be a bug in the current production website: After adding a subscription to your cart and selecting "Buy now", a modal with the user's payment information shows up. When trying to go back with the keyboard here (Shift+Tab) the selected focus is not the modal, but the background page elements. The background page should not be accessible without closing the modal.

Let me know if these make sense and if you need further context, I'd be happy to further illustrate these interactions.

matthewpaulthomas:
> Would it be possible to have a smoother and slightly slower transition with an ease-in duration?

Agreed.

> Currently the scrolling leads to the "How many" section as soon as it appears on the view. The expected behaviour should be that the "How many" section ends up at the top of the page.

In that detail, Steve has implemented [what I asked for](https://github.com/canonical-web-and-design/ubuntu.com/issues/8247#issuecomment-711905806): “_if_ the quantity field is not already visible, the page scrolls smoothly just enough to reveal it, and then it’s focused. That way you can still see and double-check your machine-type choice if you need to.”

I can see there’s a benefit to showing future steps, but:
- I think it’s less than the reassurance of letting you see and double-check the choice you’ve just made
- the less distance scrolled, the less the surprise
- scrolling “How many” to the top of the screen would, unhelpfully, scroll even if your display was big enough to show the entire form at once.

A subconscious reason for disliking the “scroll just enough” behaviour is that it leaves the bottom of the field scraping the bottom of the window. Perhaps there’s a way to scroll a _little bit_ (say 1 rem) more than that.

> Expected behaviour should be that the user needs to confirm with the Tab key

Yes, or Space or Enter.

elioqoshi:
@matthewpaulthomas That makes sense. If the scrolling would be smoother, having the anchor point of the next step at the bottom of the view could suffice. It's however a pattern I haven't seen often, usually navigating via anchor links on a page ends up with the anchor at the top of the view, not the bottom. If the scrolling effect would go just a tad further down that could be a good compromise though.

ClementChaumel:
If you want to scroll just a bit further down I think you can use something like this
```css
[id] {
  scroll-margin-bottom: 2em;
}
```

steverydz:
I've added the `scroll-margin-bottom` as @ClementChaumel recommended.

With the smooth scroll, the most robust solution would be to use CSS `scroll-behaviour` (https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior). The problem with this approach is that it would have to be applied to the `html` element, meaning every in-page scroll will inherit this behaviour, which may or may not be problematic. The alternative would be to use JS but I'm not keen to add JS but this is less robust and can have performance implications.

@lyubomir-popov do you have any thoughts on the above?

lyubomir-popov:
@steverydz I haven't used smooth scroll before, but it seems to work on the mozilla example page. I'd check if there are any performance issues with large amounts of content, and user scrolling in the opposite direction. I.e. whether the problems plaguing scroll hijacking are resolved here.

steverydz:
> @steverydz I'd check if there are any performance issues with large amounts of content, and user scrolling in the opposite direction. I.e. whether the problems plaguing scroll hijacking are resolved here.

I can't find any cause for concern around performance and scrolljacking. My only concern is as this applies to the entire site, there may be cases where we don't want this?

lyubomir-popov:
@steverydz maybe push a separate demo so whoever proposed it can QA it thoroughly?

steverydz:
@elioqoshi I've made a separate demo pr for the smooth scroll as it would be a change that effects the whole site so would need thorough QA: https://github.com/canonical-web-and-design/ubuntu.com/pull/9080

steverydz:
@elioqoshi I've made a separate issue for your accessibility concerns: https://github.com/canonical-web-and-design/ubuntu.com/issues/9081

matthewpaulthomas (approved):
Since #9081 is tracked separately, and this is an improvement by itself, +1 from me. Thanks @steverydz.

PR #9059, opened 2021-01-19
Update organisation field

### LINKED ISSUE #3623: Bring beta takeover up to date with live site

steverydz:
## Done

Updated the labelling on the UA payment form

## QA

- Go to https://ubuntu-com-9059.demos.haus/advantage/subscribe
- Check that the "I'm buying for" field defaults to "An organisation"
- Check that the following field is labelled "Organisation" rather than "Choose an account name"

## Issue / Card

Fixes #3623

### DIFF
--- a/templates/advantage/_stripe-modal.html
+++ b/templates/advantage/_stripe-modal.html
@@ -42,7 +42,7 @@
             <div class="col-9 col-start-large-4">
               <span class="p-form-validation__message u-hide"></span>
             </div>
-            
+
             {% if type == "purchase" %}
             <div class="col-9 col-start-large-4">
               <small>We'll also send setup instructions to this address.</small>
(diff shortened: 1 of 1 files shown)


matthewpaulthomas:
Thanks @steverydz

templates/advantage/_stripe-modal.html
```diff
@@ -76,7 +76,7 @@ <h2 class="p-modal__title js-contract-name u-no-margin--bottom" id="modal-title"
                   </label>
 
                   <label class="p-radio--inline">
-                    <input type="radio" class="p-radio__input" name="buying_for" id="buying_for_an_organisation" {% if account %}checked=""{% endif %} aria-labelledby="organisation">
+                    <input type="radio" class="p-radio__input" name="buying_for" id="buying_for_an_organisation" {% if account %}checked=""{% endif %} aria-labelledby="organisation" checked="true">
```
ClementChaumel, line 79: The checked attribute is duplicated here I think we can remove the first instance `{% if account %}checked=""{% endif %}`

matthewpaulthomas (approved):

templates/advantage/_stripe-modal.html
steverydz: @ClementChaumel good catch. Fixed!

PR #9066, opened 2021-01-21
Protect /takeovers.json from Discouse errors (a bit)

nottrobin:
If the Discourse requests have errors which lead to /takeovers.json failing and returning a 500 status, the content-cache would stop cache responses. This would potentially lead the the application being inundated by requests to /takeovers.json, and in turn inundating the Discourse API with requests.

Including the `stale-if-error` header will hopefully lead to the content-cache providing us with more cover (although I don't know exactly how the content-cache behaves in this case).

QA
--

``` bash
$ curl -sI https://ubuntu-com-9066.demos.haus/takeovers.json | grep -i cache-control
cache-control: max-age=300, stale-while-revalidate=360, stale-if-error=600
```

### DIFF
--- a/templates/takeovers/index.html
+++ b/templates/takeovers/index.html
@@ -10,16 +10,16 @@
   <div class="row">
     <ul class="p-inline-list u-no-margin--bottom">
       <li class="p-inline-list__item">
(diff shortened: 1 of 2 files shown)


pmahnke (approved):
works +1

Issue #9065, opened 2021-01-21
Landscape is not listed as a feature of UAI on the UAI table


alysonrose88:
Landscape is not listed as a feature of Ubuntu Advantage on the table that appears when you click 'Add another subscription'

---
*Reported from: https://ubuntu.com/advantage*

PR #9068, opened 2021-01-22
Add pricing and what's including headings to all pricing tables and Landscape row on /advantage

### LINKED ISSUE #8932: Not clear that UA prices are annual
[/advantage/subscribe](https://ubuntu.com/advantage/subscribe) makes clear that once it shows you a price, it’s per year.

However, other pages list prices without making this clear:
- https://ubuntu.com/advantage
- https://ubuntu.com/pricing/infra
- https://ubuntu.com/legal/ubuntu-advantage-service-description

This will become more important once monthly prices are introduced.

[Reported by the field sales team via David Callé.]

pmahnke:
## Done

- Add 'Pricing per year' and 'What's included' heading rows to all pricing tables 
- Add Landscape row on /advantage
- Updated /uasd copy doc

## QA

    - http://0.0.0.0:8001/advantage
    - http://0.0.0.0:8001/pricing/infra
    - http://0.0.0.0:8001/legal/ubuntu-advantage-service-description
- See that all tables have ' / year' added to the pricing row per #8932 
- See that /advantage has a row for landscape added

## Issue / Card

Fixes #8932 and #9065

## Screenshots

image

### DIFF
--- a/templates/advantage/_table.html
+++ b/templates/advantage/_table.html
@@ -35,6 +35,12 @@
           <td class="u-align--center"><img src="https://assets.ubuntu.com/v1/f1a7515d-tick-darkaubergine.svg" alt="Yes" width="16" height="16"></td>
           <td class="u-align--center"><img src="https://assets.ubuntu.com/v1/f1a7515d-tick-darkaubergine.svg" alt="Yes" width="16" height="16"></td>
         </tr>
+        <tr>
+          <td>Ubuntu systems management with <a href="https://landscape.canonical.com" class="p-link--external">Landscape</a></td>
+          <td class="u-align--center"><img src="https://assets.ubuntu.com/v1/f1a7515d-tick-darkaubergine.svg" alt="Yes" width="16" height="16"></td>
+          <td class="u-align--center"><img src="https://assets.ubuntu.com/v1/f1a7515d-tick-darkaubergine.svg" alt="Yes" width="16" height="16"></td>
+          <td class="u-align--center"><img src="https://assets.ubuntu.com/v1/f1a7515d-tick-darkaubergine.svg" alt="Yes" width="16" height="16"></td>
+        </tr>
         <tr id="kvm-guests">
           <td>Certified Windows drivers for KVM guests</td>
           <td class="u-align--center"><img src="https://assets.ubuntu.com/v1/f1a7515d-tick-darkaubergine.svg" alt="Yes" width="16" height="16"></td>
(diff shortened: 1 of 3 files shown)


elioqoshi:
@pmahnke Shouldn't /year be included in all of the price cells instead the "What's included" cells? While it would be repetitive it would have better readability to include the timeframe right with the price. Azure does the same for example:
https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux/

image

I think there would be enough space to add it with the price.

matthewpaulthomas:
@elioqoshi I think Azure is doing the wrong thing there: “/month” is repeated 40 times, and almost half of each cell is the same as other cells, making it much harder to compare the actual differences. It’s what Tufte would call a low data-ink ratio. (And the unaligned units in almost every column suggest that the table hasn’t had much visual design attention anyway.)

This case isn’t as extreme, but the reasoning is the same. I put “/year” in the first column, rather than the others, to reduce repetition and make the actual prices easier to compare. It’s not about space, and even if it was, much of that space will soon be taken by a UA Apps column ([web-squad#3378](https://github.com/canonical-web-and-design/web-squad/issues/3378#issuecomment-723086206)).

matthewpaulthomas:
> the "What's included" cells

That does raise a good point, though, that there’s no indication in that column of where “What’s included” ends and “And here are the prices” begins. A new header row there would help, or even just a thicker row border.

elioqoshi:
Fair enough, that makes sense. I will keep the discussion in the issue so it's not duplicated.

pmahnke:
with a header, it could say 'Annual pricing' in the first column.

pmahnke:
image

anasereijo:
@pmahnke this is being discussed [here](https://github.com/canonical-web-and-design/ubuntu.com/issues/8932)

anasereijo:
@pmahnke looks good, just tiny things:

- [ ] the background on rows should be `#f7f7f7` (as per [vanilla's example](https://vanillaframework.io/docs/examples/patterns/tables/table-colored-rows))
- [ ] /pricing/infra: swap the order of light grey backgrounds: "Security maintenance and support" should be white and then the following section should be `p-strip--light` and so on

elioqoshi:
Did we decide that pricing rows come before "What's included"? If yes it looks good to me.

pmahnke:
@anasereijo fixed
@elioqoshi there are actually 2 tables, /advantage and the the one that is on /pricing/infra and /uasd, which I was more hesitant to change the order of.  We can ask Lech in the Ubuntu Advantage meeting Tuesday?

anasereijo:
LGTM 👍

bethcollins92 (approved):
LGTM 👍🏼

PR #9071, opened 2021-01-22
Index all CVEs, regardless of status

nottrobin:
Index all CVEs, regardless of status

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9003

Hard to QA (we'd have to populate the CVEs in the demo) - maybe a code review is good enough?


tbille (approved):
:tada:

PR #9077, opened 2021-01-22
Fix squashed passed badge from badgr issue

bethcollins92:
## Issue

Passed microcert badge received from badger was squashed 

## Done

- Fix size and position of passed badge to line up with not yet passed badges 

## QA

- Check the passed badge is the correct shape and lines up with the not yet passed badges

## Screenshots

image

### DIFF
--- a/templates/cube/microcerts.html
+++ b/templates/cube/microcerts.html
@@ -205,20 +205,14 @@
             </div>
           </td>
           <td aria-label="Badge" class="u-no-padding--left u-no-padding--right">
-            {% if module.status == "passed" %}
               <img
                 class="p-table--cube__badge"
                 src="{{ module.badge.url }}"
                 alt="Badge for {{ module.name }}"
+                {% if module.status == 'passed' %}
                 style="margin-left: -0.2rem;"
+                {% endif %}
               />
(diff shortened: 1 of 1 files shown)


templates/cube/microcerts.html
```diff
@@ -203,12 +204,21 @@ <h2>Microcertifications</h2>
               <p class="u-text--muted">{{ loop.index }}</p>
             </div>
           </td>
-          <td aria-label="Badge">
+          <td aria-label="Badge" class="u-no-padding--left u-no-padding--right">
+            {% if module.status == "passed" %}
```
jpmartinspt, line 208: I think we could make it shorter:
``` html
<td aria-label="Badge" class="u-no-padding--left u-no-padding--right">
<img
  class="p-table--cube__badge"
  src="{{ module.badge.url }}"
  alt="Badge for {{ module.name }}"
  {% if module.status == "passed" %}
  style="margin-left: -0.2rem;"
  {% endif %}
/>
</td>
```

bethcollins92: Good point, done! Thanks

jpmartinspt (approved):

PR #9085, opened 2021-01-25
Fix dependency conflicts

nottrobin:
Upgrade ubuntu-release-info to 21.1, which loosens dependencies
so they no longer conflict.

See https://github.com/powersj/ubuntu-release-info/issues/4

## QA

``` bash
dotrun exec pip3 install --upgrade pip  # Get on pip 21
dotrun install  # See install succeed
```


ClementChaumel (approved):
Works great, thanks! :tada:

PR #9097, opened 2021-01-26
WIP: Spike to show the login functionality in the navigation

anthonydillon:
Just a spike at the moment to investigate the capabilities of the login in the navigation.

Design: https://app.zeplin.io/project/5ffdb21a3714151de9cf750c/screen/5fff03c52885841ced879a74

## QA
- Load the demo
- Click "Sign in" in the navigation
- See that you are returned to the same page
- Check your name is now displayed
- Click your name for a dropdown
- Your number of subscriptions should be displayed in the UA subscriptions section
- Awaiting CUBE to land to pull that data through
- Account settings and Logout should also work

Fixes https://github.com/canonical-web-and-design/web-squad/issues/3660

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -35,6 +35,7 @@ from webapp.context import (
 from webapp.views import (
     accept_renewal,
     advantage_view,
+    advantage_query,
     advantage_shop_view,
     advantage_thanks_view,
     BlogCustomGroup,
@@ -166,6 +167,7 @@ def utility_processor():
 # Simple routes
 app.add_url_rule("/sitemap.xml", view_func=sitemap_index)
 app.add_url_rule("/advantage", view_func=advantage_view)
+app.add_url_rule("/advantage.json", view_func=advantage_query)
 app.add_url_rule("/advantage/subscribe", view_func=advantage_shop_view)
 app.add_url_rule(
     "/advantage/subscribe/thank-you", view_func=advantage_thanks_view
@@ -586,10 +588,10 @@ core_als_autils_docs.init_app(app)
 
 @app.after_request
 def cache_headers(response):
-    """
-    Set cache expiry to 60 seconds for homepage and blog page
-    """
 
+    """
+    Set cache to private for image builder and advantage to prevent caching
+    """
     if flask.request.path in ["/core/build", "/advantage"]:
         response.cache_control.private = True
 
@@ -599,6 +601,9 @@ def cache_headers(response):
         ] = "no-cache, no-store, must-revalidate"
         response.headers["Pragma"] = "no-cache"
 
+    """
+    Set cache expiry to 60 seconds for homepage and blog page
+    """
     if flask.request.path in ["/", "/blog"]:
         response.headers[
             "Cache-Control"
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -34,6 +34,7 @@ from webapp.context import (
 
 from webapp.views import (
     accept_renewal,
+    account_query,
     advantage_view,
     advantage_query,
     advantage_shop_view,
@@ -168,6 +169,7 @@ def utility_processor():
 app.add_url_rule("/sitemap.xml", view_func=sitemap_index)
 app.add_url_rule("/advantage", view_func=advantage_view)
 app.add_url_rule("/advantage.json", view_func=advantage_query)
+app.add_url_rule("/account.json", view_func=account_query)
 app.add_url_rule("/advantage/subscribe", view_func=advantage_shop_view)
 app.add_url_rule(
     "/advantage/subscribe/thank-you", view_func=advantage_thanks_view
--- a/templates/templates/_navigation.html
+++ b/templates/templates/_navigation.html
@@ -50,55 +50,150 @@
           {% include "templates/header_noscript.html" %}
         </ul>
       </noscript>
-      <div class="p-navigation__search">
+      <div class="p-navigation__search u-align--right">
+        <div class="p-navigation__user">
+          {% if user_info %}
+            <span class="p-contextual-menu">
+              <button style="background: transparent !important; color: white !important; margin-bottom: 0;"
+                class="p-button--base p-contextual-menu__toggle has-icon"
+                data-trigger="click"
+                aria-controls="#menu-3"
+                aria-expanded="false"
+                aria-haspopup="true">
+                <span>{{ user_info.fullname }}</span>
+                <i class="p-icon--chevron-down is-light p-contextual-menu__indicator"></i>
+              </button>
+              <span class="p-contextual-menu__dropdown" id="menu-3" aria-hidden="true">
+                <span class="p-contextual-menu__group">
+                  <a href="/advantage" class="p-contextual-menu__link" data-js="ua-subscriptions">UA subscriptions</a>
+                </span>
+                <span class="p-contextual-menu__group">
+                  <a href="/cube" class="p-contextual-menu__link" data-js="ua-subscriptions">CUBE microcertifications</a>
+                </span>
+                <span class="p-contextual-menu__group">
+                  <a href="https://login.ubuntu.com" class="p-contextual-menu__link">Account settings</a>
+                  <a href="/logout" class="p-contextual-menu__link">Logout</a>
+                </span>
+              </span>
+            </span>
+            <script>
+              var uaElement = document.querySelector('[data-js="ua-subscriptions"]');
+              if (uaElement) {
+                fetch('/advantage.json')
+                  .then(response => response.json())
+                  .then(data => {
+                    var contracts = data.enterprise_contracts.enterprise_contracts[""].filter(contract => contract.contractInfo.status != 'expired' );
+                    uaElement.innerHTML = `<div>UA subscriptions</div><div><small>See your <div class="p-label--new">${contracts.length}</div> subscriptions</small></div>`;
+                  });
+              }
+            </script>
+          {% else %}
+            <a href="/login" style="position: relative;top: .75rem;padding-left: 1rem;color: white;">
+              Sign in
+            </a>
+          {% endif %}
+        </div>
+        <div class="p-navigation__user js-account">
+          <span style="position: relative;padding-left: 1rem;color: white;">
+            <i class="p-icon--spinner u-animation--spin is-light">Loading account data</i>
+            <i class="p-icon--chevron-down is-light p-contextual-menu__indicator"></i>
+          </span>
+        </div>
+
+        <script>
+          var accountContainer = document.querySelector('.js-account');
+
+          if (accountContainer) {
+            fetch('/account.json')
+              .then(response => response.json())
+              .then(data => {
+                if (data.account === null) {
+                  accountContainer.innerHTML = '<a href="/login" style="position: relative;top: .75rem;padding-left: 1rem;color: white;">Sign in</a>';
+                } else {
+                  var contracts = data.contracts.enterprise_contracts[""].filter(contract => contract.contractInfo.status != 'expired' );
+                  accountContainer.innerHTML = `<span class="p-contextual-menu">
+                    <button style="background: transparent !important; color: white !important; margin-bottom: 0;"
+                      class="p-button--base p-contextual-menu__toggle has-icon"
+                      data-trigger="click"
+                      aria-controls="#menu-5"
+                      aria-expanded="false"
+                      aria-haspopup="true">
+                      <span>${data.account.fullname}</span>
+                      <i class="p-icon--chevron-down is-light p-contextual-menu__indicator"></i>
+                    </button>
+                    <span class="p-contextual-menu__dropdown" id="menu-5" aria-hidden="true">
+                      <span class="p-contextual-menu__group">
+                        <a href="/advantage" class="p-contextual-menu__link" data-js="ua-subscriptions">
+                          <div>UA subscriptions</div><div><small>See your <div class="p-label--new">${contracts.length}</div> subscriptions</small></div>
+                        </a>
+                      </span>
+                      <span class="p-contextual-menu__group">
+                        <a href="/cube" class="p-contextual-menu__link">CUBE microcertifications</a>
+                      </span>
+                      <span class="p-contextual-menu__group">
+                        <a href="https://login.ubuntu.com" class="p-contextual-menu__link">Account settings</a>
+                        <a href="/logout" class="p-contextual-menu__link">Logout</a>
+                      </span>
+                    </span>
+                  </span>`;
+                  menuListener(accountContainer);
+                }
+              });
+          }
+
+          function menuListener(accountContainer) {
+            var toggle = accountContainer.querySelector(".p-contextual-menu__toggle");
+            toggle.addEventListener("click", (e) => {
+              e.preventDefault();
+              const menuAlreadyOpen = toggle.getAttribute("aria-expanded") === "true";
+
+              const dropdown = document.querySelector(
+                toggle.getAttribute("aria-controls")
+              );
+
+              toggle.setAttribute("aria-expanded", !menuAlreadyOpen);
+              dropdown.setAttribute("aria-hidden", menuAlreadyOpen);
+            });
+          }
+        </script>
+
+        <script>
+          var uaElement = document.querySelector('[data-js="ua-subscriptions"]');
+          if (uaElement) {
+            fetch('/advantage.json')
+              .then(response => response.json())
+              .then(data => {
+                var contracts = data.enterprise_contracts.enterprise_contracts[""].filter(contract => contract.contractInfo.status != 'expired' );
+                uaElement.innerHTML = `<div>UA subscriptions</div><div><small>See your <div class="p-label--new">${contracts.length}</div> subscriptions</small></div>`;
+              });
+          }
+        </script>
+
         <form action="/search" class="p-search-box" id="google-appliance-search-form">
           <input type="search" class="p-search-box__input" name="q" placeholder="Search" required="" aria-label="Search">
           <button type="reset" class="p-search-box__reset u-no-margin--right"><i class="p-icon--close">Clear</i></button>
           <button type="submit" class="p-search-box__button"><i class="p-icon--search">Search</i></button>
         </form>
       </div>
-      <div class="p-navigation__user">
-        {% if user_info %}
-          <span class="p-contextual-menu">
-            <button style="background: transparent !important; color: white !important; margin-bottom: 0; margin-top: .25rem;"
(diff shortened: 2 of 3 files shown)


pmahnke:
@anthonydillon that's an impressive start.  @nottrobin can you take a look at this?  We are a bit worried about performance of this... perhaps we should look at cookies and local storage as well? or have the user cache advantage.json.  Also, should we put all the logged in info in a single json file ultimately?  Also, we need to think about all the states for each product (i.e. do they have a sub, free sub, etc..)

nottrobin:
> Your number of subscriptions should be displayed in the UA subscriptions section

I only have 1 subscription, which is the free one. Should I be seeing this counted in the drop-down?

image

anasereijo:
> > Your number of subscriptions should be displayed in the UA subscriptions section
> 
> I only have 1 subscription, which is the free one. Should I be seeing this counted in the drop-down?
> 
> image

that's a good point @nottrobin. I will update the design to include that.
could you please leave comments regarding ux or visual on [zeplin](https://app.zeplin.io/project/5ffdb21a3714151de9cf750c/screen/5fff03c52885841ced879a74)? that might be easier for me to track the feedback

anthonydillon:
> I only have 1 subscription, which is the free one. Should I be seeing this counted in the drop-down?

@nottrobin sorry this prototype just shows your enterprise subscriptions count. I would show you a message saying you have no contracts but use free account?

anthonydillon:
@anasereijo I made up the subscription display as there was nothing in Zeplin. Its was just to show the info.

anthonydillon:
Will look like this with subscriptions:
image

anthonydillon:
@pmahnke So the performance of loading any pages has no-hit as the authentication data is already available to every route and the data request is performed on the client. I have noticed it takes a while to display the data but that is in the dropdown so probably ok.

> perhaps we should look at cookies and local storage as well?

I don't like the idea of using cookies or local storage as you could logout and the data will show when you walked off.

> have the user cache advantage.json

Do you mean instead of calling /advantage.json we call /ya-bo-ng/advantage.json and then cache that endpoint? 

> should we put all the logged in info in a single json file ultimately

I think keeping the services as separate calls makes sense as they require separate calls which would slow down the render of any data.

nottrobin:
> perhaps we should look at cookies and local storage as well? or have the user cache advantage.json

I guess in this case local storage would be effectively another cache, and so would be subject to all the usual problems with caches - how do you know when it's out of date? What's your refresh mechanism? It could get complicated.

This should already be cached. The response headers on `https://ubuntu-com-9097.demos.haus/advantage.json` include:

``` bash
HTTP/2 200 OK
vary: Accept-Encoding
cache-control: public, max-age=300, stale-while-revalidate=360
vary: Cookie
```

In theory `vary: cookie` should mean that the caches are split by cookie, so one per user. However, I wouldn't risk it, content-cache often does some odd things. It would be much safer to make this `cache-control: private` so content-cache never stores it.

Anyway, as things stand, the browser should be caching `/advantage.json` for 5 minutes. We could increase this, but I'm not sure we need to. The `stale-while-revalidate` of 6 minutes is even [supported by Chrome and Firefox](https://web.dev/stale-while-revalidate/), so if someone is browsing round a lot their `/advantage.json` should even refresh in the background.

I'm also not sure we need to worry *too much* about performance of these items within the drop-down. Can't we just display a spinner until the info arrives? It should be fine for this to happen in the background shouldn't it?

webapp/app.py
```diff
@@ -592,13 +594,16 @@ def takeovers_index():
...
-    Set cache expiry to 60 seconds for homepage and blog page
+    Set cache to private for image builder and advantage to prevent caching
     """
-
     if flask.request.path in ["/core/build", "/advantage"]:
         response.cache_control.private = True
 
+    """
+    Set cache expiry to 60 seconds for homepage and blog page
+    """
```
nottrobin, line 606: Docstrings are only for the top of functions. If you want an inline comment it looks like this:

```suggestion
    # Set cache expiry to 60 seconds for homepage and blog page
```

anasereijo:
@anthonydillon @nottrobin @pmahnke @cassiocassio   FYI @wgx is proposing to [have invoices and payment methods](https://github.com/canonical-web-and-design/ubuntu.com/issues/8540#issuecomment-768396829) in the user profile dropdown.

wgx:
@anthonydillon @nottrobin @pmahnke @cassiocassio @anasereijo  ... but only if the user is signed-in to UA

cassiocassio:
I'm unsure about this.

In terms of hierarchy -

Do Payment methods and invoices  belong to CUBE or/and /advantage ?

They feel like secondary nav options for each of those.

I think in @wgx's prototypes we were at the stage of. "and we will need a
hope for these kind of options somewhere.

cassiocassio:
some ideas on homes for identity and account = money

 https://miro.com/app/board/o9J_lWhzgyg=/

nottrobin:
Here's a bit of a curveball (this is mostly about implementation with only light implications for UX).

If we change the fundamental content of the response for every page on the site based on whether you're logged in, we then won't be able to use content-cache for any pages at all for logged in users. As soon as we release this feature, we load we get from logged-in users will massively increase as their uncached requests will extend to the whole site (This might have been what @pmahnke meant in his comment, although we interpreted him differently).

If we inspect analytics carefully we can probably work out what proportion of traffic is logged-in users, and we would probably find that it's a fraction of 1% (although I haven't checked). So it's *probably* safe to release this feature in its current implementation for now, and we could consider addressing this in follow-up work.

However, a fraction of 1% for ubuntu.com can still be a fair bit of traffic, and if we are successful in our various efforts to engage users, this number could grow quite considerably.

This problem is the motivation behind [jamstack](https://jamstack.org/). What jamstack says is that you should build all your HTML pages to have static content, and then manage any interactivity by pulling it content through APIs with JavaScript (client-side XHR requests).

So the jamstack way to build this would be to build an `/account.json` endpoint (or something) that would return info about the logged-in user (or a "not logged in" 401). When a page loads, JavaScript would call this endpoint and then update the account widget in the top right with the data returned from that endpoint.

This is actually more-or-less how I was planning to achieve global log-in too, across all our sites. Every site we own could call `https://ubuntu.com/account.json`, and then each display the account widget based on the information returned - so each site was displaying whether you were logged into `ubuntu.com`.

We'd then have a choice of whether to include, say, the number of advantage contracts you have in the original `account.json` response, or instead to keep the current structure of making extra calls for extra data like that.

The main UX implication would be that we would need to decide what to display initially while we wait for the logged-in state to be returned by the background call.

anthonydillon:
Thanks @nottrobin, of course you are right. I took a shortcut to gain the logged in in the header. This was to explore pulling in contextual data into the header. Your exactly right this needs to be a full JS implementation. [Partly discussed in the kickoff notes](https://github.com/canonical-web-and-design/master-epics/issues/241#issuecomment-767646824).

We would need to allow other sites to hit ubuntu.com to avoid CORS.  If we were to use ubuntu.com for all authentication requests. 

The issue with using ubuntu.com as the canonical source of user details, is that we could not provide a login option, as it would need to come from ubuntu.com. For example, if you hit login on maas.io it would bounce you to maas.io -> ubuntu.com/login -> login.ubuntu.com -> ubuntu.com. Am I right? Can't remember the implementation of the redirect. 

I was under the impression that each site that wants this functionality would need to implement the basic authentication itself. But that limits what data we can display in the drop down.

nottrobin:
Yeah I was thinking we'd update the ubuntu.com return URL view to then forward the user back to the original domain. It would mean 4 redirects, but we could maybe talk to e.g. roadmr about if there might be ways to cut down the number of redirects.

pmahnke:
@nottrobin the caching was exactly my point and why I wanted all services on a site to use one json endpoint we can't afford to let this feature make the site non-static.

cassiocassio:
- I was always open to this only appearing -short-term- on /cube,
/advantage, >1% of the page if that helped implementation

- but login everywhere just feels cleaner, and was always the goal. It fees
us to ask for all future features - what do we give, extra, to people we
know about

- the extra data surfacing, like how many CUBE certs do you have, how many
subscriptions are nice to haves, but the essential part is do I have
anything or nothing in these services

- when we habitually put our services on many individual domains, jaas,
snapcraft, each with their accounts we got away with u.com being static and
optimised for performance

- assume we'll only add services, and having a plan that scales for many
will be important

nottrobin:
> @pmahnke: @nottrobin the caching was exactly my point and why I wanted all services on a site to use one json endpoint we can't afford to let this feature make the site non-static.

I think we *may* be able to afford it, 'cos as I say I'm pretty sure the logged-in traffic is gonna be a tiny fraction of the total, and any logged-out users will stiff get cached pages. But actually the `/account.json` implementation isn't very hard, so we should probably go for it right off the bat.

Whether we want to split the account drop-down into multiple JSON calls will depend on latency. If we can speed up the initial response, allowing us to display the basic logged-in state sooner, by splitting out extra information about advantage contracts, then we might as well do that and load the advantage stuff (which is hidden under the dropdown) later, as it will probably still be ready before the user needs it. The time saved by splitting it out might be minimal, in which case it's not worth it. But I assume @anthonydillon split it out in this implementation for a reason.

anthonydillon:
@nottrobin @pmahnke @cassiocassio I have now added both methods. 
1. The left name in the navigation is using a server-side rendered name with client-side queries for the personal product information. 
2. The name on the right (by the search) is fully client-side and pulls from a single end-point (`/account.json`).

@anasereijo moved the search and made it expand to see how it feels.

nottrobin:
@anthonydillon Got bugs bro.

For me it stays in this state forever:

image

I can see that `/account.json` succeeded with 200 in the network tab, but I get these errors in the console:

```
Uncaught (in promise) TypeError: data.enterprise_contracts.enterprise_contracts[''] is undefined
    <anonymous> https://ubuntu-com-9097.demos.haus/:177
    promise callback* https://ubuntu-com-9097.demos.haus/:176
ubuntu-com-9097.demos.haus:177:37
```

Clearly you haven't accounted for someone who doesn't have any paid contracts in your code.

anthonydillon:
@nottrobin you are right should be fixed now.

anthonydillon:
@nottrobin @anasereijo I have removed the static auth method leaving a fully client-side solution with a separate call for the services in the background to speed up the initial name display.

nottrobin:
@anthonydillon demo seems broken. I think it's cos it doesn't build, rather than memory or other issues.

anthonydillon:
@nottrobin works now, just needed a rebase

nottrobin:
I'm unsure what state this is in. Is this just a proof of concept to inform our decisions? Or is is supposed to be a production ready solution?

Because if the latter, I have a few gripes:

- The state on the page is often out of date
  - When I first visited it, it showed me as logged in, but when I tried to browse to another page I was logged out and had to log in again
  - Then when I first log in, it returns me to a page that says "sign in" again. I have to refresh before I see my logged-in state
  - (I guess we need to look at the caching of the `/account.json` endpoint?)
- When I first visited the demo, the account drop-down said "see your 0 subscriptions" which was a bit weird messaging. But now I don't see it any more, now it just says "UA subscriptions" without any extra message, which is good. Maybe it was just a cached page
- You now see a spinner initially when ever page loads. This is much more jarring than I anticipated
  - Maybe the spinner is just too noticeable, we should replace it with something subtler,
  - Or maybe any kind of placeholder is going to be jarring. Maybe we should store a boolean for your logged-in state in local storage, just for the initial page load, and then update it once we hear back from `/account.json`?

cassiocassio:
- for a first version we don't have definitie "what we should show (if
anything) for each service".

- being able to fetch the set of services you have login to is the first
step

- we're super-interested in being able to display some state for each
service, so a technical spike that shows this is possible in a secure and
performant way is great

- Number of subs is a good guess for UA

- In order to get CUBE up,  I would be happy to start without any
service-no-status

- Since everyone has theoretical access to the free subscription, that
shouldn't count

- I vaugely remember we investigated the time thresholds below which we
shouldn't show a spinner in MAAS -  HCI *research*
<https://www.nngroup.com/articles/response-times-3-important-limits/> on
this goes way back. Somewhere between 200-300ms is the upper limit of "I
clicked and nothing happened and I think it's broken", displaying the
spinner immediately in a relatively performant system is noise - the web
seems littered with delay
<https://egghead.io/lessons/react-delay-the-appearance-of-a-loading-spinner-with-css-in-react>
(1)
before spinner
<https://www.bennadel.com/blog/3711-delaying-loading-indicators-using-css-animations-in-angular-9-0-0-next-14.htm>
(2)
implementations
<https://ckjaersig.dk/2020/05/showing-a-loading-spinner-delayed-with-rxjs/>
(3).
I'm sure differing network performances make second guessing this hard.

cassiocassio:
Looping in UA people:

@wgx and @elioqoshi and @anasereijo - when we have -Login-in-the-navigation- you are -encouraged- to add this to u.com/advantage :-)

Questions:

 - What kind of status e.g. @anthonydillon 's demo shows number of subscriptions, would it make sense to show?

-  Is an alert-state rather than a count more interesting, e.g. expired payment method? 

- Are alerts another thing, that should be shown -in- the navigation, rather than -inside- the pull-down

for CUBE

- there's nothing operationally critical about CUBE that I would want to show in v1

wgx:
@cassiocassio 

100% We _need_ this in /advantage :) 

> What kind of status e.g. @anthonydillon 's demo shows number of subscriptions, would it make sense to show?

I'd vote, at this stage, for: no status, BUT we only show Ubuntu Advantage or My Subscriptions if the users subscriptions are not zero. So - ubuntu advantage users will see "My Subscriptions" in the drop-down when signed in, while non-users will not. 

> Is an alert-state rather than a count more interesting, e.g. expired payment method?

Yes - having the little orange exclaim icon next to "My Subs" if they're in an alert-state would be ideal. 

> Are alerts another thing, that should be shown -in- the navigation, rather than -inside- the pull-down

No - not for now, i see the notification system handling these things. 

One more thing: I see the 'alert state in the drop down' as a reminder rather than an alert - a user will be coming here because of an email or because they've seen a big orange banner - they are unlikely to _discover_ their alert state from the drop-down, but it does serve as a great reminder.

nottrobin:
> @wgx I'd vote, at this stage, for: no status, BUT we only show Ubuntu Advantage or My Subscriptions if the users subscriptions are not zero. So - ubuntu advantage users will see "My Subscriptions" in the drop-down when signed in, while non-users will not.

I think this is complicated by the fact that every logged-in user has access to a free "personal subscription" token for up to 5 machines. So I guess whether to include it in the drop-down depends on whether we care more about advertising your free UA token, so that you might actually use it, or about eliminating noise in the drop-down.

Or maybe I'm wrong? Maybe your free token is only automatically created when you first visit `/advantage`, so if you've never been there you won't have one and we can use that fact to exclude it from the nav? @anthonydillon ?

anthonydillon:
In the act of requesting the account information the free token is generated so technically they will always have one.

nottrobin:
> @cassiocassio - I vaugely remember we investigated the time thresholds below which we shouldn't show a spinner in MAAS -  [HCI *research*](https://www.nngroup.com/articles/response-times-3-important-limits/) on this goes way back. Somewhere between 200-300ms is the upper limit of "I clicked and nothing happened and I think it's broken", displaying the spinner immediately in a relatively performant system is noise -
> 
> the web seems littered with delay before spinner implementations ([egghead.io](https://egghead.io/lessons/react-delay-the-appearance-of-a-loading-spinner-with-css-in-react), [bennadel.com](https://www.bennadel.com/blog/3711-delaying-loading-indicators-using-css-animations-in-angular-9-0-0-next-14.htm), [ckjaersig.dk](https://ckjaersig.dk/2020/05/showing-a-loading-spinner-delayed-with-rxjs/)). I'm sure differing network performances make second guessing this hard.

It does sound like whenever displaying a spinner we should be doing it with a delay anyway, as in one of these implementations. If this can be a CSS-only solution then the delay could even be centralised through Vanilla, which would be nice.

But you say the upper limit for such delays before showing a loading state would be 200-300ms. The delay on the `/account.json` endpoint is very likely to be longer than this, so we can expect a lot of people to see a spinner anyway. This suggest to me that we should look into ways to update the account state more quickly anyway, that the spinner delay work (that we should do anyway) isn't enough. So I think we do still need some short-lived local storage cache of this state.

cassiocassio:
Agree the spinner dones't absolve you from not trying to improve performance

A spinner that could see the future would know if should appear, or not
bother, because those packets are about to turn up.

It's complicated.

The loading progress bar of macOS compares it's progress to previous boot
speeds, and -fakes- the animation to make the sense of progress seem more
-real-.

The 200-300ms number that flies around is the "I clicked and nothing
happend, it's broken threshold", above which you need a "don't worry, I'm
doing something". If you knew this operation takes an average of 330ms -on
this browser- you might choose to never bother showing a spinner. If you
knew it was likely to take 5s, then always displaying the spinner
-immediately- would feel good.

It's -just- possible we could overthink and overengineer this.

Agree  the current implementation is a bit "look at me! I'm spinning!"

anasereijo:
Hey @anthonydillon ,

great work, thanks for exploring this 👍 

I have a few comments/amendments:

- [ ] spinner: remove chevron
- [ ] pause 250ms before showing spinner
- [ ] when signing in in incognito I'm not returned to a page showing my name
- [ ] change the sequence of login and search (this was discussed [here](https://app.zeplin.io/project/5ffdb21a3714151de9cf750c/screen/5ffdb2ce647d0519ad789d5d) scroll down and you'll see Claudio's and Martin's comments)
- [ ] search should have the text before the icon
- [ ] double check if the search icon is the new one
- [ ] make search icon the same colour as text `#F7F7F7`
- [ ] chevron should be the same size as the ones from the nav (enterprise, developer, etc) as well as spacing between text and icon
- [ ] dropdown: UA settings data isn't needed for this first version - we will come to this later (@wgx @elioqoshi can you start thinking if you want it visible in the dropdown? If so, how it should be displayed?) 
- [ ] dropdown: please match with [design](https://zpl.io/VxNjdOk) 
Screenshot 2021-02-16 at 15 42 46
Screenshot 2021-02-16 at 15 43 04

- [ ] expanded search (input field) please refer to my design for visual amendments and alignment 
Screenshot 2021-02-16 at 15 48 18

 - [ ] expanded search (input field) search icon should be replaced by the close icon
 - [ ] Esc key to close search input field
 - [ ] not sure if we need this message?
Screenshot 2021-02-16 at 15 57 35

Happy to join a call if you want to go through this together.

anthonydillon:
Closing in favour of the follow-up cards to complete this work.
https://github.com/canonical-web-and-design/master-epics/issues/241

Issue #9017, opened 2021-01-12
[/security/cve] cve webpages do not expose mitigation field/information


stevebeattie:
## Summary

The CVE pages do not expose mitigation information where the security team has annotated it in their tracker. This is information intended for customers to be able to mitigate a security issue while awaiting a fix to land in the Ubuntu archive.

## Process

Example: CVE-2020-28374

In the source tracker https://git.launchpad.net/ubuntu-cve-tracker/tree/active/CVE-2020-28374 the mitigation section has been populated; however, that does not show up as a separate section in https://ubuntu.com/security/CVE-2020-28374 ; it only appears on the page waaaaaaaay at the bottom because the information has been copied to to the Notes section. 

The script that exports this information to the website endpoint may need to be fixed to include this section.

## Current and expected result

Mitiagation annotations that the security team makes get surfaced to users. Additionally, because mitigation steps often involve running commands, they need to not lose their formatting, or an agreed upon schem for formattion needs to be agreed upon.
(The Notes section also has some formatting expectations.)

matthewpaulthomas:
I’m sorry I didn’t include this field in the design because I hadn’t encountered it in any of the CVEs I looked at.

For [the same CVE on the old site](https://people.canonical.com/~ubuntu-security/cve/2020/CVE-2020-28374) “Mitigation” is between “References“ and “Notes”.

Coincidentally, though, we plan to add a “Remediation” section with boilerplate for installing relevant updates whenever available. Because it will be both short and important, it’s likely to go high up the page. “Mitigation” is very closely related to “Remediation”, so perhaps they should be consecutive.

The old page also demonstrates that “not lose their formatting” includes “convert any indented line into a `<pre>` element”. @stevebeattie Is there other formatting we should know about?

matthewpaulthomas:
[Design proposed](https://docs.google.com/document/d/1VuAElUpFay7MBFNmE3NhSyY9TkXiqHcA1ZHQ11AYO-Q/edit#heading=h.9a0ml5bppl8c). Very simple: a new “Mitigation” section between the “From the Ubuntu Security Team” section (when present) and the new “Remediation” section (which is then followed by the “Status” section).

matthewpaulthomas:
@elioqoshi Could you take a look at this design please?

stevebeattie:
@matthewpaulthomas re formatting, indented as pre-formatted is all we had with the old site, I think that's fine to continue in the new site.

[...]

PR #9106, opened 2021-01-27
Remove alembic stuff

albertkol:
## Done

- Remove `alembic` stuff. The API will take care of it.

## QA

- Are checks passing?

## Issue / Card

Part of #9017


entrypoint
```diff
@@ -5,24 +5,7 @@ set -e
 RUN_COMMAND="talisker.gunicorn.gevent webapp.app:app --bind $1 --worker-class gevent --name talisker-`hostname`"
 
 if [ "${FLASK_DEBUG}" = true ] || [ "${FLASK_DEBUG}" = 1 ]; then
```
jpmartinspt, line 7: This needs to be an if else.

albertkol: Not sure I understand. Isn't this what we use for most of our websites? Fx:
https://github.com/canonical-web-and-design/jp.ubuntu.com/blob/master/entrypoint
https://github.com/canonical-web-and-design/charmed-kubeflow.io/blob/master/entrypoint

jpmartinspt: NVM, I was looking at the diff and did not realise you swapped the logic inside the if to be an assignment.

jpmartinspt (approved):

jpmartinspt:
Looks good, needs a rebase before merging though.

carkod:
https://ubuntu-com-9106.demos.haus/security/cve seems to be broken

jpmartinspt:
@carkod, since we are removing migrations from ubuntu.com there is no schema created on the demo database. That is why you see the error on `/cve`. The way to QA this would be locally against a database that already has the tables in place.

albertkol:
Closed. For now we will keep the migrations on ubuntu.com. 

We should do this at the very end when ubuntu-com is not reliant on the database anymore. 

That way we avoid any 500s caused by the missing database tables.

PR #9126, opened 2021-01-29
Remove renewal preview endpoint

### LINKED ISSUE #9107: Incorrect shop backend API request during renewal purchase process.

## Summary

Ubuntu Advantage shop appears to be making an incorrect API request during renewal purchase process.

## Process

This appears to be caused by a customer purchasing a renewal. We see an API call to `/v1/renewals/rREDACTED/purchase/preview`.

Purchase preview is only supported by the backend for new purchases. It doesn't work with old-style renewals yet (which pre-date the shop). So this API endpoint shouldn't be called for renewals.

## Screenshot

Not available for data protection reasons.

## Browser details

We don't have this information, but if you ping us in chat, we might be able to get customer success to get this information.

sowasred2012:
## Done

Removed a renewal preview endpoint that would show VAT on a renewal, and added "Excluding VAT" below the subtotal when we're not able to calculate VAT.

For UA shop, we have the ability to make a "preview purchase" in order to get a correct VAT amount and display it to the user before they make the actual purchase. The same was implemented for renewals, but it was causing issues on older renewals, as mentioned in https://github.com/canonical-web-and-design/ubuntu.com/issues/9107#issuecomment-768674581

## QA

- Visit https://ubuntu-com-9126.demos.haus/advantage?test_backend=true
- Login
- Open your dev tools, and the network tab within
- If you don't have a contract and renewal, speak to @alnvdl, who can set you up with one.
- Once you have one, click the "Renew" button on the relevant contract.
- When the modal opens, it should instantly show you a summary of the renewal, including the subtotal and "Excluding VAT" underneath.
- Fill out the form with these card details:
  - 4242 4242 4242 4242 / expiry: 04/24 / cvc: 242 / zip: 42424
  - and valid address details
  - set country to United Kingdom (which would ordinarily trigger a VAT calculation)
- Click "Continue", see that you are taken to the second modal screen, and that the subtotal amount still informs you it is "Excluding VAT"
- In your dev tools network tab, you should _not_ see a request to "preview"

## Issue / Card

Fixes #9107


ClementChaumel (approved):
LGTM!

PR #9132, opened 2021-01-31
Removed external from blog like

anthonydillon:
## Done
Removed the external link class from the blog link on OpenStack features page.

## QA

- Check the code


pmahnke (approved):
good catch!

PR #9143, opened 2021-02-01
Restrict CUBE views to Canonical employees

jpmartinspt:
## Done
Make CUBE views only accessible to users with `@canonical.com` emails.
## QA

- Make sure you can still see `/cube` and `/cube/microcerts` you should be redirected to SSO if you are not logged in.
- Login as a user that does not have a `@canonical.com` email and see you get a 403 page.

### DIFF
--- a/webapp/cube/views.py
+++ b/webapp/cube/views.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from requests import Session
 
 from webapp.cube.api import BadgrAPI, EdxAPI
+from webapp.decorators import login_required
 from webapp.login import user_info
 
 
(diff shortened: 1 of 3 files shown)


webapp/cube/views.py
```diff
@@ -45,14 +46,22 @@
 )
 
 
+def is_authorized(user):
```
tbille, line 49: New decorator?

`@canonical_employees` or something like this

jpmartinspt: could be, this will be removed after as while, so I thought I would just keep it simple

tbille (approved):
code looks good

webapp/cube/views.py
tbille: ok why not.

PR #9175, opened 2021-02-03
UA: bug fixes

sowasred2012:
## Done

- Reduce the amount of calls we're making to the purchase preview endpoint, previously it would happen twice when launching the payment modal on /subscribe
- Explicitly set account and previous_purchase_id as `None` if we know the user isn't logged in
- Disable caching on /advantage/subscribe

## QA

- Check the [demo](https://ubuntu-com-9175.demos.haus/advantage/subscribe?test_backend=true)
- While not logged in, inspect the page source for "accountID" and "previousPurchaseID" variables, see that both values are empty strings.
- Log in
- Open network panel in dev tools
- Select a product, add it to cart, then click "Buy Now"
- Fill out the form, select "United Kingdom" as your country
  - Card details: 4242 4242 4242 4242 / CVC: 242 / Expiry: 04/24 / Zip: 42424
- Click "Pay"
- In the network panel, you should see only one request to "preview"

### DIFF
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -173,7 +173,7 @@ function attachCTAevents() {
 
       checkoutEvent(analyticsFriendlyProducts(), 1);
       setOrderInformation(cartItems, modal);
-      checkVAT();
+      checkVATdebounce();
     }
 
     if (isRenewalCTA || isShopCTA) {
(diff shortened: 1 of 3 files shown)


ClementChaumel (approved):
JS looks fine! :+1:

alnvdl (approved):
LGTM! I see the `Cache-Control` header in effect. Shouldn't the same cache policy apply to `"/advantage"` as well, just to have the behavior between both UA pages be predictable?

sowasred2012:
> LGTM! I see the `Cache-Control` header in effect. Shouldn't the same cache policy apply to `"/advantage"` as well, just to have the behavior between both UA pages be predictable?

I want to verify that this fixes the issue(s) on the shop before we update that too - if it does, I'll raise another PR that does that :+1:

PR #9222, opened 2021-02-10
Update customerInfo object before API call

### LINKED ISSUE #9194: Cannot change address in the shop
Once an address is set in a first purchase, I cannot change it in a subsequent purchase. All requests made to the backend contain the original address that was set during the first purchase. The request that gets sent to ua-contracts only contains the first, unmodified address.

Steps to reproduce:
- Make a purchase
- Make another purchase and try to replace the address somehow
- Watch the request logs

ClementChaumel:
## Done

- Fixed a bug where the backend would be called with an outdated address

## QA

- https://ubuntu-com-9222.demos.haus/advantage/subscribe?test_backend=true
- log in
- Make a purchase
- Make another purchase and try to replace the address somehow
- Watch the request logs

## Issue / Card

Fixes #9194

### DIFF
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -362,6 +362,8 @@ function applyTotals() {
 }
 
 function applyLoggedInPurchaseTotals() {
+  setCustomerInfo();
+
   if (currentTransaction.type == "purchase") {
     const formData = new FormData(form);
     const country = formData.get("Country");


alnvdl:
I tested this here, and the address change is working now, thanks!
It'd be nice to have tests for this modal/form to avoid regressions like these.

alnvdl (approved):
I tested this here, and the address change is working now, thanks!
It'd be nice to have tests for this modal/form to avoid regressions like these.

anthonydillon (approved):
LGTM :+1:

anthonydillon:
@ClementChaumel I agree could you add a test for this?

PR #9249, opened 2021-02-16
Build warning payment issue

ClementChaumel:
## Done

- Added a warning to /advantage if a user has monthly subs and the payment method is invalid
- Fixed the links to /subscribe so they keep the `?test_backend=true` parameter if it's present

[Design](https://app.zeplin.io/project/601d6513b99f24a8412ecfa5/screen/6024087b77a6eab466624106)

## QA

- Not sure how to QA that yet. @ me on mattermost so I can show you!

## Issue / Card

Fixes #3800

## Screenshots

image

## Note 

- The text wraps using the copy provided and the vanilla component.
If we remove the `.` then it fits, what should we do?
image

- The link is broken but will be implemented in https://github.com/canonical-web-and-design/ubuntu.com/pull/9248

### DIFF
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -46,7 +46,7 @@
         </div>
       </div>
       {% if subscriptions['has_monthly'] %}
-        {% if not default_payment_method["id"] %}
+        {% if payment_method_warning %}
           <div class="row">
             <div class="p-notification--caution">
               <p class="p-notification__response" role="status">
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -45,7 +45,7 @@
       </div>
       {% if subscriptions['has_monthly'] %}
         {% if not default_payment_method["id"] %}
-          <div class="row" style="margin-top: 1rem;">
+          <div class="row">
             <div class="p-notification--caution">
               <p class="p-notification__response" role="status">
                 <span class="p-notification__status">Payment method:</span>You need to <a href="/advantage/payment-methods{% if is_test_backend %}?test_backend=true{% endif %}">update your payment methods</a> to ensure there is no interruption to your Ubuntu Advantage subscriptions.
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -36,6 +36,8 @@
     </div>
   </section>
 
+  <hr></hr>
+
   {% if enterprise_contracts %}
     <section class="p-strip is-shallow" style="overflow: visible;">
       <div class="row u-vertically-center">
@@ -48,7 +50,7 @@
           <div class="row">
             <div class="p-notification--caution">
               <p class="p-notification__response" role="status">
-                <span class="p-notification__status">Payment method:</span>You need to <a href="/advantage/payment-methods{% if is_test_backend %}?test_backend=true{% endif %}">update your payment methods</a> to ensure there is no interruption to your Ubuntu Advantage subscriptions.
+                <span class="p-notification__status">Payment method:</span>You need to <a href="/advantage/payment-methods{% if is_test_backend %}?test_backend=true{% endif %}">update your payment methods</a> to ensure there is no interruption to your Ubuntu Advantage subscriptions
               </p>
             </div>
           </div> 
@@ -89,7 +91,7 @@
       {% endif %}
     </section>
 
-    <section class="p-strip u-no-padding--top is-shallow">
+    <section class="p-strip is-shallow">
       <div class="row u-hide--medium u-hide--large">
         <div class="col-12">
           {% for account in enterprise_contracts %}
@@ -375,9 +377,10 @@
             </table>
           {% endfor %}
         </div>
-      <div class="col-12 u-align--right">
-        <a href="/advantage/subscribe{% if is_test_backend %}?test_backend=true{% endif %}" class="p-button--positive u-align--right" >Add new subscription</a>
-      </div>
+        <div class="col-12 u-align--right">
+          <a href="/advantage/subscribe{% if is_test_backend %}?test_backend=true{% endif %}" class="p-button--positive u-align--right" >Add new subscription</a>
+        </div>
+        <hr></hr>
       </div>
     </section>
   {% endif %}
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -380,10 +380,15 @@
         <div class="col-12 u-align--right">
           <a href="/advantage/subscribe{% if is_test_backend %}?test_backend=true{% endif %}" class="p-button--positive u-align--right" >Add new subscription</a>
         </div>
-        <hr></hr>
       </div>
     </section>
   {% endif %}
+
+  <section class="p-strip u-no-padding--top is-shallow" style="overflow: visible;">
+    <div class="row">
+      <hr></hr>
+    </div>
+  </section>
 
   <section class="p-strip u-no-padding--top" style="overflow: visible;">
     <div class="row">
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -7,7 +7,7 @@
 {% block content %}
 
 {% if user_info %}
-  <section class="p-strip--suru-topped">
+  <section class="p-strip--suru-topped u-no-padding--bottom">
     <script>
       dataLayer.push({
         'event' : 'GAEvent',
@@ -36,7 +36,7 @@
     </div>
   </section>
 
-  <hr></hr>
+  <hr class="p-separator" />   
 
   {% if enterprise_contracts %}
     <section class="p-strip is-shallow" style="overflow: visible;">
@@ -68,7 +68,7 @@
               <div class="row">
                 <div class="row col-start-large-2 col-10">
                   <h5>{{subscriptions["current_subscription_number"]}} subscriptions are set to auto-renew monthly</h5>
-                  <hr>
+                  <hr class="p-separator" />
                   <div> 
                     <label class="p-radio">
                       <input id="auto-renewal-on" type="radio" class="p-radio__input" name="auto-renewal"  {% if subscriptions['is_auto_renewal_enabled'] %}checked=""{% endif %} aria-labelledby="enable" value="on">
@@ -380,15 +380,11 @@
         <div class="col-12 u-align--right">
           <a href="/advantage/subscribe{% if is_test_backend %}?test_backend=true{% endif %}" class="p-button--positive u-align--right" >Add new subscription</a>
         </div>
+        <hr class="p-separator" />
       </div>
     </section>
   {% endif %}
 
-  <section class="p-strip u-no-padding--top is-shallow" style="overflow: visible;">
-    <div class="row">
-      <hr></hr>
-    </div>
-  </section>
 
   <section class="p-strip u-no-padding--top" style="overflow: visible;">
     <div class="row">
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -44,6 +44,15 @@
         </div>
       </div>
       {% if subscriptions['has_monthly'] %}
+        {% if not default_payment_method["id"] %}
+          <div class="row" style="margin-top: 1rem;">
+            <div class="p-notification--caution">
+              <p class="p-notification__response" role="status">
+                <span class="p-notification__status">Payment method:</span>You need to <a href="/advantage/payment-methods{% if is_test_backend %}?test_backend=true{% endif %}">update your payment methods</a> to ensure there is no interruption to your Ubuntu Advantage subscriptions.
+              </p>
+            </div>
+          </div> 
+        {% endif %}
         <div class="row">
           <div class="col-8 col-medium-3">
             <p class="u-no-margin--bottom">Your next recurring payment of <strong><span class="js-format-price">{{subscriptions["next_payment"]["ammount"]}}</span></strong> will be taken on <strong>{{subscriptions["next_payment"]["date"]}}</strong>.</p>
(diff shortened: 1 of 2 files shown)


anthonydillon:
I am hitting a `HTTPError` when calling the function `get_customer_info`. The response is `404 Client Error`.

templates/advantage/index.html
```diff
@@ -44,6 +44,15 @@ <h2>Your {% if subscriptions['total_subscriptions'] > 1 %}{{subscriptions['total
         </div>
       </div>
       {% if subscriptions['has_monthly'] %}
+        {% if not default_payment_method["id"] %}
+          <div class="row" style="margin-top: 1rem;">
```
anthonydillon, line 48: It's not good practise to put margins or padding on rows as they should just act as a clearfix. Maybe this needs to be within its own strip?

anthonydillon (changes requested):

anthonydillon:
I would suggest handling the response errors with something like this:
```python
def get_customer_info(self, account_id):
        try:
            response = self._request(
                method="get",
                path=f"v1/accounts/{account_id}/customer-info/stripe",
            )
        except HTTPError as error:
            if (
                error.response.status_code == 401
                or error.response.status_code == 404
            ):
                return error.response.json()
            else:
                raise error

        return response.json()
```
Then you need to handle the error into the frontend.

templates/advantage/index.html
nottrobin: @ClementChaumel correct me if I'm wrong, but I think this was the result of a long discussion and much back-and-forth with @anasereijo.

I think maybe @anthonydillon you should talk to @anasereijo directly, face to face, about this issue. Unless I've got the wrong end of the stick @ClementChaumel?

templates/advantage/index.html
```diff
@@ -367,7 +376,7 @@ <h4>Ubuntu 16.04 and 18.04 LTS</h4>
           {% endfor %}
         </div>
       <div class="col-12 u-align--right">
-        <a href="/advantage/subscribe" class="p-button--positive u-align--right" >Add new subscription</a>
+        <a href="/advantage/subscribe{% if is_test_backend %}?test_backend=true{% endif %}" class="p-button--positive u-align--right" >Add new subscription</a>
```
nottrobin, line 379: I'm trying to remember the reasons why we implemented this `test_backend=true` feature. Can anyone remember? @sowasred2012 ?

If we're at the point where we're adding significant modifications to templates just to support our testing, I think maybe we should have another look at whether this is a good solution.

nottrobin:
> ``` python3
> if (
>     error.response.status_code == 401
>     or error.response.status_code == 404
> ):
> ```

Two points on this:

- Why do we need to catch both `401` and `404`? I heard about the need for `404`, but are we sure catching `401` is also necessary?
- The fact that you're returning `error.response.json()` suggests to me that you're expecting that `json()` to have useful content. If that's the case, could you maybe inspect that content too, to see if you can test based on that content to differentiate this "you don't have stripe data" `404` from a "this API endpoint doesn't exist" `404`?

Actually @ClementChaumel if you're able to, could you show us an example of what the shape of the response actually looks like in these `401` and `404` cases that you care about?

nottrobin:
is there any information about the changes in this PR that I can read somewhere else? There's not much context here...

templates/advantage/index.html
ClementChaumel: This is the PR that introduced it https://github.com/canonical-web-and-design/ubuntu.com/pull/8888
I don't remember how the whole conversation went but I think it was because using `.env` to set the API url to the test backend when developing is bad practice as we should be as close to the real environment. so we agreed a query param was a good way to do it

ClementChaumel:
> Actually @ClementChaumel if you're able to, could you show us an example of what the shape of the response actually looks like in these `401` and `404` cases that you care about?

Sure here's the response for a 404
```json
{
   "code":"not found",
   "message":"account \"blipblop\" not found",
   "traceId":"f17e0d70-ddda-4d4b-9c6e-e0b6f79a043a"
}
```

templates/advantage/index.html
ClementChaumel: I think you might be thinking of something else? :thinking:  I just added that here to match the design. I don't remember discussing this issue

ClementChaumel:
> is there any information about the changes in this PR that I can read somewhere else? There's not much context here...

@nottrobin This PR adds a the warning you can see in the screenshot IF a customer has at least one monthly-renewing subscription AND their payment method is faulty / missing

templates/advantage/index.html
sowasred2012: @nottrobin @ClementChaumel the reason was as soon as we were no longer working on a feature branch and were instead creating pull requests to be merged into master, the way that we had been setting the staging environment - hardcoding the appropriate values, then updating them to production when the feature branch was eventually merged into master - became very unwieldy. This solution seemed like a good compromise at the time, as Clem says.

anasereijo:
Thanks, @ClementChaumel, this is looking great. Just a few little things (I hope):

- [ ] yes, please remove `.` 

some comments related to the table (sorry if this is the wrong PR)
- [ ] remove the `u-no-padding--top` from the table so it gets a bit more padding at the top
- [ ] can we have an `hr` below the table, as per my design?

anthonydillon:
The reason to explicitly targeting 401 and 404 is due to the expected responses in the API docs: https://github.com/CanonicalLtd/ua-contracts/blob/master/docs/contracts.yaml#L2344

nottrobin:
@anthonydillon but we expect the 404 to happen in a normal user flow right? As in, we can see an expected user flow where they would encounter the 404 and we know what to do with it. This is why we're catching it.

The 401, on the other hand, is presumably a standard "you're not authorised to use this API" response. Why is this a special case on this particular endpoint?

anthonydillon:
@ClementChaumel lets not handle the 401 until we have a reason too.

anthonydillon:
> The text wraps using the copy provided and the vanilla component.
If we remove the . then it fits, what should we do?
image

I would remove the words "Ubuntu Advantage" and just leave subscriptions as its clear that is what we are talking about.

templates/advantage/index.html
```diff
@@ -366,13 +377,19 @@ <h4>Ubuntu 16.04 and 18.04 LTS</h4>
...
+        <div class="col-12 u-align--right">
+          <a href="/advantage/subscribe{% if is_test_backend %}?test_backend=true{% endif %}" class="p-button--positive u-align--right" >Add new subscription</a>
+        </div>
       </div>
     </section>
   {% endif %}
 
+  <section class="p-strip u-no-padding--top is-shallow" style="overflow: visible;">
+    <div class="row">
+      <hr></hr>
```
anthonydillon, line 389: A horizontal rule is a self-closing tag: `<hr />`

anthonydillon (approved):
LGTM with two tiny suggestions and a wording suggestion above.

anthonydillon:
By the way I QA but changing the conditional from `{% if not default_payment_method["id"] %}` to `{% if default_payment_method["id"] %}`

templates/advantage/index.html
nottrobin: Ah then yes I'm thinking of something else =) ignore me.

wgx:
ux +1

alnvdl:
I don't think the condition being checked on this PR are ever possible:

> Added a warning to /advantage if a user has monthly subs and the payment method is invalid

If you have a monthly subscription, the payment method ID will never be null/undefined because in order for you to have a subscription, you must have set a payment method before. And one cannot remove a payment method, so once set, it's forever set.

Instead, the user should be alerted that they have a purchase stuck due to a payment issue. Detecting that a purchase is stuck can be obtained by calling [`GetSubscriptions`](https://github.com/CanonicalLtd/ua-contracts/blob/develop/docs/contracts.yaml#L2648), which returns a list of subscriptions. Then for each of these subscriptions, you have to look at whether there is a pending purchase (see the `pendingPurchases` attribute).

If you find a pending purchase for any active subscription, then you need to get that purchase to see what must be done next. [This document explains what to do after yo get that purchase]([https://github.com/CanonicalLtd/ua-contracts/blob/develop/docs/shop.md#workflows]):

> Cases (b) above can be detected by looking at the purchase object: if the status of the purchase is processing and the invoice payment status is need_3ds_authorization, the frontend should get the payment intent secret in the invoice object embedded in the purchase and authorize 3DS with Stripe. This should automatically unblock the purchase.
>
> Cases (c) above can be detected by looking at the purchase object: if the status of the purchase is processing and the invoice payment status is need_another_payment_method, the frontend should ask for a new payment method, set it for the purchase account (via SetStripeCustomerInfo) and retry the invoice payment (via RetryPurchaseStripeInvoicePayment).
>
> Cases (d) and (e) above can be detected by looking at the purchase and subscription objects. If the purchase status is inactivated, the only thing to do is to display to the customer that their purchase was canceled and that no charge was made, redirecting them to try again. If the subscription status is deactivated, the customer may also be told that their service may be affected unless they make a new purchase.

You can find an example of code that does this check in ua-contracts-web as well:
https://github.com/CanonicalLtd/ua-contracts/blob/develop/web/src/Shop.js#L651-L660

wgx:
>Instead, the user should be alerted that they have a purchase stuck due to a payment issue. 

We're talking about the same thing. If there is a payment issue on the subscription, we should set a flag in the account `PAYMENT_ISSUE=True` and if that is set on the user object, display the warning in the UI.

The remedy is for the user to visit the Payment methods view and update their payment card.

wgx:
@alnvdl 

So I've read some docs and I see 2 possible ways around this. 

1. We can see the expiry date on a card when we tokenise it. If the card on the user's Subscription is expiring, we could notify the user before anything happens in Stripe, or...

2. Implement a Stripe webhook. Stripe tells us each time a subscription charge is either paid or fails. If it fails we set the users' `Payment_Issue` to `True`, send them an email and display the banner in the UI. 

https://stripe.com/docs/billing/subscriptions/webhooks 


cc @frankban

alnvdl:
> @alnvdl
> 
> So I've read some docs and I see 2 possible ways around this.
> 
> 1. We can see the expiry date on a card when we tokenise it. If the card on the user's Subscription is expiring, we could notify the user before anything happens in Stripe, or...

We could check the card is about to expire and offer a friendly warning before payment fails, but I'd say it's a nice-to-have. For renewals, payments will be triggered automatically by Stripe, and what matters here is to offer a way for the customer to resume these specific payments. Please note that it's not enough to just replace the card if a purchase/renewal failed: one must also retry the invoice payment.

[The Purchase React component](https://github.com/CanonicalLtd/ua-contracts/blob/develop/web/src/Shop.js#L932) encapsulates the logic for **asynchronously** resolving a purchase (meaning the user may leave the session and come back later) that got stuck due to a failed payment (or one that requires extra authorization). A similar thing willl need to be done here.

> 2. Implement a Stripe webhook. Stripe tells us each time a subscription charge is either paid or fails. If it fails we set the users' `Payment_Issue` to `True`, send them an email and display the banner in the UI.
> 
> https://stripe.com/docs/billing/subscriptions/webhooks

We already use these webhooks. You can access the result of our interpretation of webhook events by calling [`GetSubscriptions`](https://github.com/CanonicalLtd/ua-contracts/blob/develop/docs/contracts.yaml#L2648), which returns a list of subscriptions. Then for each of these subscriptions, you have to look at the `pendingPurchases` attribute. A pending purchase is the entity we create to represent that fact that a payment failed or is otherwise stuck due to some other reason (e.g., 3DS required).

If you find any pending purchase in the returned subscriptions, call `GetPurchase` to see what action is required next for that specific purchase. We already interpret all Stripe information for you in ua-contracts and return a clear instruction via our API (ask for another credit card, ask for 3DS authorization). See more details at: https://github.com/CanonicalLtd/ua-contracts/blob/develop/docs/shop.md#workflows

Please note that we also use purchases to represent automatic renewals that will be done every month, and these renewals may get stuck for several reasons (card expired, card canceled, insufficient funds, bank did not authorize, 3DS required, etc). But whichever the failure, there's always only two things we need do:
- ask for another card, OR
- ask that the current card be authorized via 3DS

So, in summary:
- Look at all subscriptions: if you find any pending purchase, display this warning with a link to the purchase ID.
- When the user clicks this link, they are taken to a page that lets them input a new card which will be set as their default. At the same time, if the reason of failure was not 3DS, the user must be able to retry the payment for that purchase to unblock it.

anthonydillon:
@ClementChaumel what is the plan for this one?

ClementChaumel:
I've updated the check to look for a pending purchase as @alnvdl suggested. It should now display the warning if there is any pending purchase at all. I don't know how set up my account to be in such a state though. @alnvdl can you confirm that's the right thing to look for? thanks!

templates/advantage/index.html
```diff
@@ -44,6 +46,15 @@ <h2>Your {% if subscriptions['total_subscriptions'] > 1 %}{{subscriptions['total
         </div>
       </div>
       {% if subscriptions['has_monthly'] %}
+        {% if payment_method_warning %}
+          <div class="row">
+            <div class="p-notification--caution">
+              <p class="p-notification__response" role="status">
+                <span class="p-notification__status">Payment method:</span>You need to <a href="/advantage/payment-methods{% if is_test_backend %}?test_backend=true{% endif %}">update your payment methods</a> to ensure there is no interruption to your Ubuntu Advantage subscriptions
```
alnvdl, line 53: Technically, it might happen that 3DS got stuck as well. In that case, the payment method would be fine, it's just that we need to authorize it.

This would happen if the user closes (or otherwise loses) the session right when they were supposed to authorize the purchase, and then they come back and want to resume the purchase. Or when a monthly renewal requires 3DS for some reason (should be rare, but it's theoretically possible).

Also, when a purchase is stuck due to a missing payment method, the corrective action is to ask for a new card, set it (SetStripeCustomerInfo) and then retry the payment of that purchase's invoice.

All of that being said, I think this could be better phrased as "There is a pending purchase on your account that requires further input from you. Click here to resolve it."

alnvdl (approved):
LGTM as a first step, but we need to start considering how to resolve pending purchases as well (probably in a follow-up).

templates/advantage/index.html
wgx: I don't agree - the wording is too complex for what is a very rare edge case. Please keep the copy as-is.

PR #9252, opened 2021-02-17
Update purchase to subscription matching

albertkol:
## Done

- The way a new contract gets attached to a subscription had to be updated. 
- Before you could always make the assumption that from all the subscriptions you get from that account, the first one (and the only one) is the right subscription to attach the new contract to it
- Now that users can have up to two subscriptions (yearly and monthly), we can no longer rely on that assumption. 
- The purchase form had to be adapted to cover the new situations.

## QA

- Go to https://ubuntu-com-9252.demos.haus/advantage?backend_testing=true make a purchase. It should work.

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/web-squad/issues/3833

### DIFF
--- a/static/js/src/advantage/contracts-api.js
+++ b/static/js/src/advantage/contracts-api.js
@@ -114,6 +114,7 @@ export async function postPurchaseData(
       account_id: accountID,
       products: products,
       previous_purchase_id: previousPurchaseId,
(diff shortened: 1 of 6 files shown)


ClementChaumel (approved):
LGTM!

PR #9261, opened 2021-02-19
CUBE: Change EDX API enrollments endpoint

jpmartinspt:
## Done

The endpoint we were using is not stable in production so we are using
the older version of it as it is currently more stable.

## QA

- Verify the page is still working and you are enrolled in the courses you are supposed to be enrolled in.


pmahnke (approved):
+1

PR #9262, opened 2021-02-19
Bump cookie policy to version v3.2.0

anthonydillon:
## Done
Bump cookie policy to version v3.2.0

## QA
- Load the demo
- Clear your cookies
- Refresh and see the policy
- Click around and see it work
- You can revoke it from the footer to change
- You should be able to scroll while the modal is visible


anthonydillon:
This is done

PR #9267, opened 2021-02-22
Add is_hidden column to usn

### LINKED ISSUE #9263: Add "is_hidden" column for USNs

albertkol:
## Done

- Add new column to usn table
- add BaseFilterQuery class is overriding the default Query class and applies a default filtering based on the set rules of the Model set inside the `_base_filters`.
   
If we wish to bypass this default filtering, we would call
    `without_default_filtering` on the query:

`usn = db_session.query(Notice).without_default_filters().get(notice_id)`

## QA

- Have a database
- Do not fetch changes
- `docker-compose up -d` and `dotrun`
- Browse to http://0.0.0.0:8001/security/notices/USN-4417-2 # can be any id.
- You should get a 200
- Let's make USN-4417-2 private:
```sql
docker exec -it db psql -U postgres
select id,is_hidden from notice where id='USN-4417-2'; -- you will see the is_hidden value is set to false
update notice set is_hidden=True where id='USN-4417-2'; -- set is_hidden value to true
select id,is_hidden from notice where id='USN-4417-2'; -- you will see the is_hidden value is set to true 
```
- Browse to http://0.0.0.0:8001/security/notices/USN-4417-2
- You will get a 404

## Issue / Card

Fixes #9263

### DIFF
--- a/webapp/security/alembic/versions/8008e46e6ea8_add_is_hidden_column_for_notices.py
+++ b/webapp/security/alembic/versions/8008e46e6ea8_add_is_hidden_column_for_notices.py
@@ -1,0 +1,32 @@
+"""add_is_hidden_column_for_notices
+
+Revision ID: 8008e46e6ea8
+Revises: 3e0a3f15abb9
+Create Date: 2021-02-19 12:39:11.459139
+
+"""
+from alembic import op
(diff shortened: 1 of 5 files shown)


carkod (approved):
LGTM

PR #9268, opened 2021-02-22
CUBE: Update mappings to use the prod issuer IDs

### LINKED ISSUE #36: Odd styling on Juju and MAAS row
We are overriding .two-col width which we shouldnt do and there is an equal-heights issue which could be removed.

cloud reference architecture cloud ubuntu

jpmartinspt:
## Done

Update mappings to use the prod issuer IDs

## QA

Use the demo to make sure /cube/microcerts is still working


bethcollins92 (approved):
LGTM 🥇

PR #9269, opened 2021-02-22
Add cancel subscription

albertkol:
## Done

- Add cancel subscription.
- `DELETE /advantage/subscribe`:
```json
{
        "account_id": "account_id",
        "product_listings": [
            "product_listing_id_1", "product_listing_id_2",
        ],
        "previous_purchase_id": "previous_purchase_id",
}
```

## QA
Not easy to test until we have a form that allows us to buy monthly products.

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/web-squad/issues/3793

### DIFF
--- a/webapp/advantage.py
+++ b/webapp/advantage.py
@@ -221,11 +221,23 @@ class AdvantageContracts:
     def purchase_from_marketplace(
         self, marketplace: str, purchase_request: dict
     ) -> dict:
-        response = self._request(
-            method="post",
-            path=f"v1/marketplace/{marketplace}/purchase",
-            json=purchase_request,
-        )
+        try:
+            response = self._request(
+                method="post",
+                path=f"v1/marketplace/{marketplace}/purchase",
+                json=purchase_request,
+            )
+        except HTTPError as http_error:
+            api_response = http_error.response.json()
+
+            if (
+                "cannot remove all subscription items"
+                in api_response["message"]
+            ):
+                raise CannotCancelLastContractError
+
+            raise
+
         return response.json()
 
     def preview_purchase_from_marketplace(
@@ -238,6 +250,14 @@ class AdvantageContracts:
         )
 
         return response.json()
+
+    def cancel_subscription(self, subscription_id: str) -> dict:
+        response = self._request(
+            method="delete",
+            path=f"v1/subscriptions/{subscription_id}",
+        )
+
+        return response.json() if response.status_code != 200 else None
 
 
 class UnauthorizedError(Exception):
@@ -252,3 +272,7 @@ class UnauthorizedError(Exception):
 
     def asdict(self):
         return self.__dict__
+
+
+class CannotCancelLastContractError(Exception):
+    pass
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -27,7 +27,11 @@ from canonicalwebteam.search.views import NoAPIKeyError
 
 # Local
 from webapp.login import empty_session, user_info
-from webapp.advantage import AdvantageContracts, UnauthorizedError
+from webapp.advantage import (
+    AdvantageContracts,
+    UnauthorizedError,
+    CannotCancelLastContractError,
+)
 from webapp.decorators import store_maintenance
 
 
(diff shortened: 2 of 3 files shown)


SirSamTumless:
@alnvdl and @frankban let me know if you want to take a look at this.

webapp/advantage.py
```diff
@@ -239,6 +239,14 @@ def preview_purchase_from_marketplace(
 
         return response.json()
 
+    def cancel_subscription(self, subscription_id: str) -> dict:
```
frankban, line 242: Docstring?

webapp/views.py
```diff
@@ -767,6 +767,116 @@ def post_advantage_subscriptions(preview):
     return flask.jsonify(purchase), 200
 
 
+def cancel_advantage_subscriptions():
```
frankban, line 770: Docstring here as well?
```
"""Downsell to 0 all product listings provided in the request.

If required, cancel the subscription.
"""
```

webapp/views.py
```diff
@@ -767,6 +767,116 @@ def post_advantage_subscriptions(preview):
...
+            user_token,
+            token_type=("Macaroon" if user_token else "Bearer"),
+            api_url=api_url,
+        )
+    else:
+        return flask.jsonify({"error": "authentication required"}), 401
+
+    payload = flask.request.json
+
+    if not (
```
frankban, line 790: The request could just be 
```
{
  account_id: $id,
  previous_purchase_id: $id.
  listings: [$listing1, $listing2, ...]
}
```
?
And to avoid looking up values more than once, you could do:
```python
account_id = payload.get("account_id")
previous_purchase_id = payload.get("previous_purchase_id")
listings = payload.get("listings", [])

if not (account_id and previous_purchase_id and listings):
            return flask.jsonify({"error": "bad request"}), 400
```

webapp/views.py
```diff
@@ -767,6 +767,116 @@ def post_advantage_subscriptions(preview):
...
+            extra={"payload": payload}
+        )
+        return (
+            flask.jsonify(
+                {"error": "could not retrieve account subscriptions"}
+            ),
+            500,
+        )
+
+    if not monthly_subscriptions:
```
frankban, line 819: Shouldn't you check monthly_subscriptions["subscriptions"] ?
What do you think about:
```python
subs = monthly_subscriptions.get("subscriptions") if monthly_subscriptions else None
id not subs:
         return flask.jsonify({"error": "no monthly subscriptions found"}), 400
monthly_subscription = subs[0]
```

webapp/views.py
```diff
@@ -767,6 +767,116 @@ def post_advantage_subscriptions(preview):
...
+    }
+
+    try:
+        purchase = advantage.purchase_from_marketplace(
+            marketplace="canonical-ua", purchase_request=purchase_request
+        )
+    except HTTPError as http_error:
+        api_response = http_error.response.json()
+
+        if "cannot remove all subscription items" in api_response["message"]:
```
frankban, line 845: I think this check could be done by `advantage.purchase_from_marketplace`, that could return a specific exception for this case, like `CannotCancelLastContractError`, that you check here and react accordingly. What do you think?

frankban (commented):
Looks good with some minor comments.
It would be nice to add tests for the advantage client changes and probably also for the view.
Thanks!

webapp/views.py
albertkol: Done.

webapp/views.py
albertkol: I changed `if not monthly_subscriptions` to `if not monthly_subscriptions.get("subscriptions")`, well spotted. 

monthly_subscriptions.get("subscriptions") already returns `None` by default if it cannot find the key in the dictionary, so the if-else is not needed.

alnvdl:
@albertkol You can test this by using ua-contracts-web to buy a monthly subscription: http://10.48.130.194:8080/staging
Go to Shop -> Monthly plans and buy something and then try to cancel it using the demo of this PR.

webapp/views.py
```diff
@@ -767,6 +767,116 @@ def post_advantage_subscriptions(preview):
...
+    purchase_request = {
+        "accountID": account_id,
+        "purchaseItems": [
+            {
+                "productListingID": product["product_listing_id"],
+                "metric": "active-machines",
+                "value": 0,
+                "delete": True,
+            }
+            for product in payload.get("products")
```
alnvdl, line 833: I think calling this field products, which then contains product listings is a bit misleading. Products are one thing, product listings are another. Perhaps just call it `product_listings` and have it be a flat list of product listing IDs?

alnvdl (approved):
Mostly LGTM, just one suggestion.

As I said in another comment, this can be tested using ua-contracts-web to trigger a monthly purchase.

webapp/views.py
albertkol: Good idea. Made the change.

webapp/views.py
albertkol: Done.

albertkol:
I managed to test it easier with CURL requests, thanks to @alnvdl 

From a QA point of view, it can be merged, we will do more testing once the front-end will allow us to make purchases and cancellations. 

@frankban I made all the changes you proposed except the docblocks which I am not a big fan of. We do not have any docblocks for any endpoints/views either. This one might benefit from having one, but we will see.. 

If there are any further feedback please let me know, otherwise I will merge this.

PR #9300, opened 2021-02-23
Updated button in sub heading

### LINKED ISSUE #3838: On mobile the header tabs should always be displayed

SirSamTumless:
## Done

Updated button in the sub heading to "watch the webinar" linking to “https://ubuntu.com/smartstart” with “https://ubuntu.com/engage/ubuntu-core-20-webinar”

## QA

- Check the changes against the [copy doc](https://docs.google.com/document/d/1QoM0Yqx4JfDJ2G3OPt5bds-ximG1jASmU_PWX4iff3E/edit?ts=602e6dca#) and accept the correct changes

Fixes [#3838](https://github.com/canonical-web-and-design/web-squad/issues/3838)

### DIFF
--- a/templates/core/index.html
+++ b/templates/core/index.html
@@ -12,7 +12,7 @@
       <h1>Secure embedded devices</h1>
       <p class="p-heading--four u-sv3">Give your device its own app store</p>
       <p>
-        <a class="p-button--positive" href="/core/smartstart">Get to market faster</a>
+        <a class="p-button--positive" href="/engage/ubuntu-core-20-webinar">Watch the webinar</a>
       </p>
       <p class="u-sv3">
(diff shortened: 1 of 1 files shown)


pmahnke (approved):
+1

PR #9307, opened 2021-02-24
Add is_hidden default to false

albertkol:
## Done

- Fix error, because is_hidden is not defaulting to false correctly.

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -249,9 +249,7 @@ def _update_notice_object(notice, data):
     notice.published = data["published"]
     notice.references = data["references"]
     notice.instructions = data["instructions"]
-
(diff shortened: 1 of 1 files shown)


webapp/security/views.py
```diff
@@ -249,6 +249,7 @@ def _update_notice_object(notice, data):
     notice.published = data["published"]
     notice.references = data["references"]
     notice.instructions = data["instructions"]
+    notice.is_hidden = False
```
jpmartinspt, line 252: Could we not just do `notice.is_hidden = data.get("is_hidden", False)` and remove the if?

albertkol: Changed!

jpmartinspt (approved):

PR #9310, opened 2021-02-24
Update payment summary and remove cart functionality

anthonydillon:
## Done
- Update payment summary section
- Add monthly subscription support
- Remove the cart from the advantage purchase flow
- Adjust the displayed cost in summary with 2 digits

## QA
- Go to https://ubuntu-com-9310.demos.haus/advantage/subscribe?test_backend=true

**Check the payment summary looks like [the design](https://app.zeplin.io/project/601d6513b99f24a8412ecfa5/screen/602ba60be0395723916f27ec)** 
- Select a desktop contract
- Add some quantity
- Select no support
- Check the "Your chosen plan" section matches [the design](https://app.zeplin.io/project/601d6513b99f24a8412ecfa5/screen/602ba60be0395723916f27ec)

**Removing the cart**
- See that switching between Monthly and Annual works as expected
- Check the payment modal appears in place of the cart
- Check the information is correct in the payment modal **I am yet to update the bottom of the payment modal**

**Disabling billing period**
- Cancel and change the support to standard and advanced
- See the summary switched to yearly and the option to change billing period is removed.
- Check the payment modal is correct

## Issue / Card
Fixes https://github.com/canonical-web-and-design/web-squad/issues/3850
Fixes https://github.com/canonical-web-and-design/web-squad/issues/3851


anthonydillon:
Don't merge is set as it only works with staging API at the moment.

anthonydillon:
@anasereijo this one is ready for another review.

wgx:
UX plus-one, based on how it appears now on demo 👍🏽

anasereijo:
@anthonydillon this is looking much better, just a couple more things:

- [ ] can we have padding-top to the "switch to annual..." text? Around 8px should be fine
on mobile:
- [ ] can you hide the illustration as we discussed yesterday, so the number of servers is next to the plan?
- [ ] add a bit more space between the price and the buy button?
Screenshot 2021-02-26 at 09 24 02

anthonydillon:
> can you hide the illustration as we discussed yesterday, so the number of servers is next to the plan?

This is is difficult due to the section being rendered via JS. Think this is ok?
image

anasereijo:
> > can you hide the illustration as we discussed yesterday, so the number of servers is next to the plan?
> 
> This is is difficult due to the section being rendered via JS. Think this is ok?
> image

That's a shame but I guess that's ok for now. We can have a look later again

anasereijo:
@anthonydillon I wanted to give you a +1 but this still needs to be done:


> * [x]  can we have padding-top to the "switch to annual..." text? Around 8px should be fine

anasereijo:
LGTM

static/js/src/ua-product-selector.js
```diff
@@ -375,34 +232,17 @@ function productSelector() {
       style: "currency",
       currency: currency,
       currencyDisplay: "symbol",
-      minimumFractionDigits: 0,
-      maximumFractionDigits: 0,
+      minimumFractionDigits: 2,
+      maximumFractionDigits: 2,
```
ClementChaumel, line 236: I believe 2 is the default for the en-US locale

ClementChaumel (approved):
Looks good to me! :+1:

ClementChaumel:
I cannot buy while being logged in

I get those errors
image

Here is the response from the failing API call
```json
{
   "api_response":{
      "code":"bad request",
      "message":"previous purchase different from the expected: got pAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA want pAMyal0kV6fwXBm1j1E097MDDgzwPBoUJ-sJ4mbwYvyc",
      "traceId":"40f5ce66-1084-4abf-8821-85141cbbec96"
   },
   "purchase_request":{
      "accountID":"aBe5SKIDeTMFTwNJY0X9BNrLIKd0tFeMrnciM2llSS_0",
      "previousPurchaseID":"",
      "purchaseItems":[
         {
            "metric":"active-machines",
            "productListingID":"lAItB58PXVMHumEC2kgZKN9HPKYLL1agdjuUXkbiytBk",
            "value":4
         }
      ]
   }
}
```
And here's the part of the code that is trying to access the customer's name

```
function fetchCustomerInfo(accountId) {
  getCustomerInfo(accountId)
    .then((res) => {
      const { name, address } = res.customerInfo;
      customerInfo = { ...customerInfo, name, address };
      setFormElements();
      isCustomerInfoSet = true;
    })
    .catch((e) => console.error(e));
}
```


This is what the UI looks like when it happens
image
No subtotal, VAT or total since the preview call fails

alnvdl:
I had the same problem as Clement. First purchase went well, but when I tried to purchase again, it failed. The post to /preview failed with a 500:
```
{ code: "bad request", message: "previous purchase different from the expected: got pAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA want pADwkr7sj2WEDcRDMNgqX80CMyY-hSKsZvRHtheRGTnA", traceId: "5f278f63-1276-40f7-867c-51f10153b6b0" }
```

The previous purchase ID is required so that both the frontend and the backend can agree on which subscription state they are basing the purchase modification, without the frontend having to specify the whole subscription state every time.

I think what is happening here is that the previous purchase ID was being globally set in the window, an that's not the case anymore. Still, we need to carry it over somehow from subscription state (in case the subscription state exists). Each subscription will have its own last purchase ID if it exists, so you should select accordingly.

Other than that, I think removing the cart actually makes for a nicer experience for 95% of our customers, which is great. For the other 5%, it will make it temporarily more annoying until the interface gives the user the option to reuse the same card as last time.

anthonydillon:
@ClementChaumel @alnvdl thanks, I believe I have resolved all the issues.

ClementChaumel:
LGTM!

albertkol:
It's selecting the correct `previous_purchase_id` and allows for monthly product purchases. :+1:

But I found some mini-issues with the html&css:

image

image

anthonydillon:
@albertkol thanks, not much I can do about this without rewriting a lot of the grid. Definitely, something to try and improve in the future.

PR #9313, opened 2021-02-25
Add contract details

albertkol:
## Done

The `enterprise_contracts` has now:
-  `enterprise_contract["is_detached"]` -> "true" or "false"
-  `enterprise_contract["period"]` -> "yearly" or "monthly" 
-  `enterprise_contract["is_cancelable"]` -> "true" or "false" 
-  `enterprise_contract["price_per_unit"]` -> {"currency": "USD", "value": 20000} 

There is now the concept of `is_detached`, which if it's `True`, then it means that the contract is either a legacy contract or an already cancelled contract. Either case you should not show the cancelation/upsize-downsize view on expand. `period`, `price_per_unit` and `is_cancelable` will not exist if `is_detached` is `True`.

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -407,19 +407,6 @@ def search_snaps():
 
 @store_maintenance
 def advantage_view():
-    accounts = None
-    personal_account = None
-    enterprise_contracts = {}
-    entitlements = {}
-    subscriptions = {
-        "total_subscriptions": 0,
-        "has_monthly": False,
-        "next_payment": {},
-    }
-    new_subscription_start_date = None
-    new_subscription_id = None
-    open_subscription = flask.request.args.get("subscription", None)
-    payment_method_warning = None
     is_test_backend = flask.request.args.get("test_backend", False)
 
     stripe_publishable_key = os.getenv(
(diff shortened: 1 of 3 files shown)


webapp/views.py
```diff
@@ -619,6 +690,15 @@ def advantage_view():
     )
 
 
+def get_contract_period(contract):
+    text_from = contract["contractInfo"]["effectiveFrom"]
+    text_to = contract["contractInfo"]["effectiveTo"]
+    date_from = dateutil.parser.parse(text_from)
+    date_to = dateutil.parser.parse(text_to)
+
+    return "yearly" if (date_to - date_from).days > 31 else "monthly"
```
ClementChaumel, line 699: This could be bad for contract nearing their ends

ClementChaumel (approved):

webapp/views.py
```diff
@@ -453,21 +454,65 @@ def advantage_view():
...
+                    "price": listing["price"],
+                    "quantity": 0,
+                }
+
+            monthly_products[listing["productID"]] = {
+                "id": listing["id"],
+                "metric": listing["allowanceMetric"],
+                "price": listing["price"],
+                "quantity": 0,
+            }
```
alnvdl, line 483: If the product is yearly, you will end up adding it to both `yearly_products` and to `monthly_products`.

webapp/views.py
```diff
@@ -453,21 +454,65 @@ def advantage_view():
 
             raise http_error
 
+        listings_response = advantage.get_marketplace_product_listings(
+            "canonical-ua"
+        )
+
+        product_listings = listings_response.get("productListings")
+
+        monthly_products = {}
```
alnvdl, line 463: Please, name these `*_listings` (products are one thing, listings is another).

webapp/views.py
```diff
@@ -453,21 +454,65 @@ def advantage_view():
...
+        yearly_products = {}
+        listings_to_products = {}
+
+        for listing in product_listings:
+            listings_to_products[listing["id"]] = listing["productID"]
```
alnvdl, line 468: This is the other way around: you want to map from the product -> listingID + metric + quantity.

webapp/views.py
```diff
@@ -453,21 +454,65 @@ def advantage_view():
...
-            for subscription in monthly_subscriptions.get("subscriptions", []):
+            monthly_subscriptions = []
+            yearly_subscriptions = []
+
+            for subscription in all_subscriptions.get("subscriptions", []):
+                period = subscription["subscription"]["period"]
+                previous_purchase_ids[period] = subscription["lastPurchaseID"]
+
+                if subscription["subscription"]["period"] == "yearly":
+                    yearly_subscriptions.append(subscription)
```
alnvdl, line 504: `continue` or `else` missing.

webapp/views.py
```diff
@@ -453,21 +454,65 @@ def advantage_view():
...
 
+        listings_response = advantage.get_marketplace_product_listings(
+            "canonical-ua"
+        )
+
+        product_listings = listings_response.get("productListings")
+
+        monthly_products = {}
+        yearly_products = {}
+        listings_to_products = {}
```
alnvdl, line 465: This should be name products_to_listings. Sorry for the misleading name, this issue came from my pastebin.

webapp/views.py
```diff
@@ -453,21 +454,65 @@ def advantage_view():
...
+                period = subscription["subscription"]["period"]
+                previous_purchase_ids[period] = subscription["lastPurchaseID"]
+
+                if subscription["subscription"]["period"] == "yearly":
+                    yearly_subscriptions.append(subscription)
+
+                monthly_subscriptions.append(subscription)
+
+            for subscription in monthly_subscriptions:
+                for purchase in subscription["purchasedProductListings"]:
```
alnvdl, line 509: please name this variable to something like `purchasedProductListing` or `ppl` to be more accurate.

webapp/views.py
```diff
@@ -453,21 +454,65 @@ def advantage_view():
...
+            if listing["period"] == "yearly":
+                yearly_products[listing["productID"]] = {
+                    "id": listing["id"],
+                    "metric": listing["allowanceMetric"],
+                    "price": listing["price"],
+                    "quantity": 0,
+                }
+
+            monthly_products[listing["productID"]] = {
+                "id": listing["id"],
```
alnvdl, line 479: You may want to keep the whole product listing object here, so that you can use it later for easier lookups.

webapp/views.py
```diff
@@ -495,7 +540,37 @@ def advantage_view():
...
+            for subscription in yearly_subscriptions:
+                for purchase in subscription["purchasedProductListings"]:
+                    listing_id = purchase["productListing"]["id"]
+                    purchased_product_id = listings_to_products[listing_id]
+                    yearly_products[purchased_product_id][
+                        "quantity"
+                    ] = purchase["value"]
+
             for contract in account["contracts"]:
+                contract["period"] = get_contract_period(contract)
```
alnvdl, line 552: It is impossible to determine the period for a contract. The best you can do is determine whether it is affected by a monthly subscription, by a yearly subscription, or both. For that, look into `monthly_subscriptions[contract.products[0]]`, and if you find anything there, it means that contract is currently affected by a monthly subscription.

The same applies to yearly, of course.

If you are interested in displaying prices for both, perhaps you could append them to an `affectedBy` list with the details (product listing object).

webapp/views.py
```diff
@@ -495,7 +540,37 @@ def advantage_view():
...
+
+                period_products = (
+                    monthly_products
+                    if contract["period"] == "yearly"
+                    else yearly_products
+                )
```
alnvdl, line 558: Again, this is not possible. It could be affected by both.

webapp/views.py
```diff
@@ -495,7 +540,37 @@ def advantage_view():
...
+
+                product_name = contract["contractInfo"]["products"][0]
+                contract_product = period_products[product_name]
+                contract["is_cancelable"] = False
```
alnvdl, line 562: `is_cancellable` could be determined by a single lookup:
`monthly_subscriptions[contrac.products[0]]`. If it exists, then we have a monthly listing to cancel on that contract.

webapp/views.py
albertkol: Fixed!

webapp/views.py
albertkol: Removed.

webapp/views.py
albertkol: Fixed.

albertkol:
@alnvdl I made drastic changes to the whole view. Mainly cleaning up. 

The main concept is still the same.
1) Fetch all product listings to and map them by `{productID}-{period}` so I can later add the price to the contract.
2) Fetch all subscriptions and split them by period
3) From my subscription arrays, create `monthly_purchased_products` and `yearly_purchased_products`. These will each have the key as `productID` and the value as `quantity`. Assuming quantity might be needed.
4) Fetch all contracts map them to the products purchased.
5) If contract["purchasedID"] exists in both add the contract twice to `enterprise_contracts`. 

Potential problems:
- displaying start date and end date can be wrong
- displaying number of machines can be wrong
- contract token is exactly the same for multiple rows

PR #9333, opened 2021-03-02
Add auto renewal endpoint

albertkol:
## Done

- Add auto-renewal endpoint:
`POST /advantage/set-auto-renewal`
```
{
  "should_auto_renew": true/false
}
```

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -1238,22 +1238,84 @@ def post_payment_method():
 
 
 def post_auto_renewal_settings():
+    user_token = flask.session.get("authentication_token")
+    is_test_backend = flask.request.args.get("test_backend", False)
 
-    # try:
-    #     # TODO Call the API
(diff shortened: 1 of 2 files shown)


webapp/views.py
```diff
@@ -1238,22 +1238,84 @@ def post_payment_method():
...
+                flask.current_app.extensions["sentry"].captureException(
+                    extra={
+                        "subscription_id": subscription["subscription"]["id"],
+                        "api_response": http_error.response.json(),
+                    }
+                )
+                return (
+                    flask.jsonify(
+                        {
+                            "error": "could change auto renewal settings",
```
ClementChaumel, line 1309:
```suggestion
                            "error": "could not change auto renewal settings",
```

This one is my bad but it should be "could not change"

ClementChaumel (approved):
LGTM!

PR #9353, opened 2021-03-04
Dynamic meta image

pmahnke:
## Done

- Default meta image per page based on the title

## QA

- Test different pages from ubuntu.com on the following validators and make sure that the image is rendered properly:
  - https://developers.facebook.com/tools/debug
  - https://cards-dev.twitter.com/validator (for some reason I'm getting an error on this one)
  - https://metatags.io/

## Screenshots

https://ubuntu-com-9353.demos.haus/kubernetes
image


requirements.txt
```diff
@@ -27,4 +27,5 @@ ubuntu-release-info==21.1
 launchpadlib==1.10.13
 macaroonbakery==1.3.1
 sortedcontainers==2.3.0
-vcrpy-unittest==0.1.7
\ No newline at end of file
+vcrpy-unittest==0.1.7
+pillow
```
anthonydillon, line 31: Do we need this?

webapp/metaimage.py
```diff
@@ -0,0 +1,37 @@
...
+from PIL import ImageFont
+from PIL import ImageDraw
+
+
+def share_meta_image():
+    static_folder = "static/files/"
+    title = "Ubuntu rules"
+    title = flask.request.args.get("title", "")
+    overlay_image = "meta_image_blank.jpg"
+    # overlay_image = flask.request.args.get("overlay_image", "")
```
anthonydillon, line 17: I know this is WIP but some comments here

anthonydillon (commented):
This is great is there still work to be done here?

anthonydillon:
This is what I see from the Twitter card validator:
image

requirements.txt
pmahnke: it probably needs to be pinned to a version, but this is the PIL package that draws (I think)

pmahnke:
@anthonydillon I saw this as well, I assume this is because it takes too long to render? or is it a header issue?

anthonydillon:
Maybe because the image is not strictly 1:2 ratio?

requirements.txt
nottrobin: yes this should be pinned - I imagine renovate might come round and pin it for you if you don't ;)

webapp/app.py
```diff
@@ -254,8 +255,11 @@ def utility_processor():
     view_func=appliance_install,
 )
 app.add_url_rule(
-    "/appliance/portfolio",
-    view_func=appliance_portfolio,
+    "/appliance/portfolio", view_func=appliance_portfolio,
+)
+
+app.add_url_rule(
+    "/static/files/meta_image.png", view_func=share_meta_image, methods=["GET"]
```
nottrobin, line 262: It's misleading to put this in `/static/files` when it's not a static endpoint. Why wouldn't this be `ubuntu.com/generate-meta-image` or something?

pmahnke: @nottrobin no reason, I wasn't 100% sure what I was doing at first... as long as the headers are right and it is cacheable, I don't think it matters?

nottrobin: No indeed. Just thought I'd mention it 🙂

nottrobin:
Oh if it's because takes too long to render then that could have big implications

anthonydillon:
@pmahnke @nottrobin what is the plan with this one?

nottrobin:
It needs scheduling =)

goulinkh:
@nottrobin @anthonydillon Do you mind reviewing this PR, I changed the image resolution from to 1200x630 (Which I think it's the standard resolution) and specified the absolution image URL as I think that was the issue.

nottrobin:
From memory this PR isn't more than a proof of concept. Generating this image will be pretty expensive, so the solution we go live with needs to be able to reliably generate each image only once and then serve it from some kind of cache. I'm also not sure ubuntu.com is the right place for it, it would probably make more sense as a feature of the assets server. So I expect this to be a fair bit more work to get right.

I also haven't really heard anyone asking for this since Peter first proposed it (although it certainly is a neat idea). I'd be happy to add it to the systems squad backlog, but it would be fairly far down the list.

webapp/metaimage.py
```diff
@@ -0,0 +1,34 @@
...
+    title = flask.request.args.get("title", "")
+    overlay_image = "meta_image_blank.png"
+
+    # wrap text
+    wrapper = textwrap.TextWrapper(width=23)
+    wrapped_title = "\n".join(wrapper.wrap(text=title))
+
+    img = Image.open(str(os.path.join(static_folder, overlay_image)))
+    draw = ImageDraw.Draw(img)
+    # font = ImageFont.truetype(<font-file>, <font-size>)
```
anthonydillon, line 22: Left in this commented code

anthonydillon (approved):
Code looks good with one small comment. QA is good. Some content improvements from the site itself.

nottrobin:
@anthonydillon are you disagreeing with me?

anthonydillon:
@nottrobin Sorry missed your message above. Yes, the performance hit of this should be carefully considered. Technically and from a QA point of view, the original issues have been resolved.

nottrobin:
Added to backlog: https://docs.google.com/spreadsheets/d/1HH7Elz5ayh4sjH6avh0Z8_4riqlJsZ9JNMxxlesH8Go/edit#gid=1343426497

image

PR #9362, opened 2021-03-05
Added login to the header of the website

anthonydillon:
## Done
Added login to the header of the website

## QA
- Open the demo
- If you see your name in the header awesome, go to /logout and head back to the demo to see the logged out view
- Otherwise, click Signin and see that it brings you back to the same page logged in
- Check different screen sizes
- Check it is in the correct place as [the design](https://zpl.io/VxNjdOk)

## Issue / Card
Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9288

## Screenshots
image

image

image

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -38,6 +38,7 @@ from webapp.cube.views import cube_home, cube_microcerts
 
 from webapp.views import (
     accept_renewal,
+    account_query,
     advantage_view,
     advantage_shop_view,
     advantage_thanks_view,
@@ -170,6 +171,7 @@ def utility_processor():
 
 # Simple routes
 app.add_url_rule("/sitemap.xml", view_func=sitemap_index)
+app.add_url_rule("/account.json", view_func=account_query)
 app.add_url_rule("/advantage", view_func=advantage_view)
 app.add_url_rule("/advantage/subscribe", view_func=advantage_shop_view)
 app.add_url_rule(
@@ -605,12 +607,17 @@ openstack_docs.init_app(app)
 
 @app.after_request
 def cache_headers(response):
-    """
-    Set cache expiry to 60 seconds for homepage and blog page
-    """
+    # Set cache to private to prevent caching
 
     if flask.request.path in ["/core/build"]:
         response.cache_control.private = True
+
+    if flask.request.path in ["/account.json"]:
+        response.cache_control.private = True
+        response.headers[
+            "Cache-Control"
+        ] = "no-cache, no-store, must-revalidate"
+        response.headers["Pragma"] = "no-cache"
 
     if flask.request.path.startswith("/advantage"):
         response.cache_control.private = True
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,4 @@
+{
+  "singleQuote": true,
+  "trailingComma": "es5"
+}
--- a/static/js/src/header-account.js
+++ b/static/js/src/header-account.js
@@ -1,0 +1,12 @@
+var accountContainer = document.querySelector('.js-account');
+if (accountContainer) {
+  fetch('/account.json')
+    .then(response => response.json())
+    .then(data => {
+      if (data.account === null) {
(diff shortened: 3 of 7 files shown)


webapp/app.py
```diff
@@ -605,13 +607,18 @@ def takeovers_index():
...
+    # Set cache to private to prevent caching
 
     if flask.request.path in ["/core/build"]:
         response.cache_control.private = True
 
+    if flask.request.path in ["/account.json"]:
+        response.cache_control.private = True
+        response.headers[
+            "Cache-Control"
+        ] = "no-cache, no-store, must-revalidate"
```
nottrobin, line 619: Rather than chucking everything at the problem, let's work out what we actually need. If you have no-store, you shouldn't need no-cache as well. In fact I think there's a chance that no-cache might override no-store in some cases.

I don't know if there's any relevant client that would need the `pragma` header as well as the `cache-control` one. I think probably not.

I don't know how both interact with must-revalidate.

Then there's the question of whether we actually want to prevent caching completely here. We probably do, but we could consider allowing the user's browser to cache the response for a bit. I'll think and read about it a bit.

nottrobin (changes requested):
I feel this is a little too MVP to go live. If I see my name there, I expect to be able to do *something* with it, even if that's just to log out.

anthonydillon:
I think its a good first step. within a day or two, there will be a dropdown from the name. After that, at some point, we will add service information to it. All small incremental steps.

anasereijo:
@anthonydillon did you put 'sign in' instead of 'login' on purpose? I remember a discussion about this but not sure what was the final decision. Maybe @cassiocassio remembers.

anasereijo:
@anthonydillon this is looking good, but I agree with @nottrobin feels odd just having my name there and not being able to do anything with it. Maybe it can go live after the dropdown is implemented?

anthonydillon:
@anasereijo @nottrobin I have added a basic logout link so the users can do that.

anthonydillon:
Unfortunately it's a bespoken pattern so I'll need to build it from scratch. I am trying to build small incremental branches.

nottrobin:
This is probably out of scope for this pr, but as has been noted in other places, logout shouldn't log you out of all of sso. It should just delete the local cookie.

nottrobin:
It is out of scope. Filed issue: https://github.com/canonical-web-and-design/ubuntu.com/issues/9366

nottrobin (approved):
I like this now as an MVP. The log out link provides an action against your name, making it actually feel like a functional element, so the plain text name no longer feels pointless.

For me this is fine for now, with the drop-down to be implemented soon.

pmahnke:
@anasereijo when I play with this, I find the name on the right of the Search a bit strange.  Specially as search doens't grow.  Is that just me?

anasereijo:
> @anasereijo when I play with this, I find the name on the right of the Search a bit strange. Specially as search doens't grow. Is that just me?

That's not just you, I also agree.

This is the issue with building elements separately - design wise doesn't work like that. When I designed the login I also changed the search to make them look nice together.

anasereijo:
> Unfortunately it's a bespoken pattern so I'll need to build it from scratch. I am trying to build small incremental branches.

I'd prefer to wait until we have the dropdown with the logout link.

anthonydillon:
Ok will have to find another way.

cassiocassio:
Establishing the visual alignment of the right hand margin is important for almost all our layouts.

So we can't move the white, obvious search box and break the grid.

We also did benchmarking and sense-checking, and agreed login / userID belonged on the outside edge - the whole page content belongs to <userID>. On the left: <Ubuntu>, on the right <username>.

So that's a motivation - along with space constraints - to redesign search as the compact icon - so it looks right on the inside.

cassiocassio:
-1 to putting "logout" text in the main navigation. Feels peculiarly important.

nottrobin:
@cassiocassio why did you reopen? What are you expecting with this PR?

PR #9374, opened 2021-03-10
add 1.20 release to k8s chart

evilnick:
## Done

- Add dates for 1.20 k8s to release-cycle chart

N.B. The length of support has changed due to upstream changes. There will be changes to the chart in future to indicate a 'security patch support' period, though this is currently waiting approval (at which point I will update the text also).

## QA

- See the chart at <http://0.0.0.0:8001/about/release-cycle#charmed-kubernetes-release-cycle> renders correctly

## Screenshots

image

### DIFF
--- a/static/js/src/chart-data.js
+++ b/static/js/src/chart-data.js
@@ -1287,7 +1287,7 @@ export var kubernetesReleases = [
     startDate: new Date("2020-01-07T00:00:00"),
     endDate: new Date("2020-12-07T00:00:00"),
     taskName: "Kubernetes 1.17",
-    status: "CHARMED_KUBERNETES_SUPPORT",
+    status: "CHARMED_KUBERNETES_EXPIRED_SUPPORT",
   },
   {
     startDate: new Date("2020-03-24T00:00:00"),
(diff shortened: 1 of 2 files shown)


anthonydillon:
@evilnick looks good, could you also add the 1.20 release to [the small screen table](https://github.com/canonical-web-and-design/ubuntu.com/blob/master/templates/about/release-cycle.html#L578):
image

static/js/src/chart-data.js
```diff
@@ -1287,7 +1287,7 @@ export var kubernetesReleases = [
     startDate: new Date("2020-01-07T00:00:00"),
     endDate: new Date("2020-12-07T00:00:00"),
     taskName: "Kubernetes 1.17",
-    status: "CHARMED_KUBERNETES_SUPPORT",
+    status: "CHARMED_KUBERNETES_EXPIRED_SUPPORT",
   },
   {
     startDate: new Date("2020-03-24T00:00:00"),
```
anthonydillon, line 1293: Hasn't 1.18 also expired?

anthonydillon (changes requested):

static/js/src/chart-data.js
evilnick: no. The support is n-2 effectively, so 1.18 is supported until 1.21 is released

evilnick: I will update the small screen table and the values with my best guess for release dates of 1.21. Unfortunately this table is going to get way messier because of changes to the upstream supported versions (they now run for a year instead of 9 months) vs our support (n-2). It is likely our support terms will be updated in the near future to address some of this

evilnick:
@anthonydillon thanks for pointing out the html table stuff. I believe that last commit fixes all the dates. For now, anyhow

anthonydillon (approved):
Thanks LGTM :+1: 
LGTM :+1:

PR #9389, opened 2021-03-15
Change cancel contract

albertkol:
## Done

- Change cancel contract to only accept one `product_listing_id`. Before it was accepting a list of ids, but we only ever cancel one product at a time, so it's pointless.


ClementChaumel (approved):
:+1:

PR #9391, opened 2021-03-15
Fix security endpoints bugs

### LINKED ISSUE #9390: Fix update and delete security endpoints

albertkol:
Fix pocket error name. It was copy-pasted wrongly. Resulting in a 500.

Add `without_default_filters` to the get query for `update_notice` and
`delete_notice`.

We filter out by default notices with `is_hidden = True` in all select
queries. For update and delete we need to add `without_default_filters`
to bypass that default filter, otherwise we will always get a 404,
making both endpoints unusable.

Updates and deletes can only be done by canonical people, so
no need for the filter.

## Issue / Card

Fixes #9390

### DIFF
--- a/webapp/security/schemas.py
+++ b/webapp/security/schemas.py
@@ -54,7 +54,7 @@ class Component(String):
 
 class Pocket(String):
     default_error_messages = {
-        "unrecognised_component": (
+        "unrecognised_pocket": (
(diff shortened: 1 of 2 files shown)


ClementChaumel (approved):
:+1:

PR #9394, opened 2021-03-15
Sync front and back ends

ClementChaumel:
## QA
Go to https://ubuntu-com-9394.demos.haus/advantage?test_backend=true and make some changes to your subscriptions using the table. (424242424242424242... is a test card if you need one)
Then try to toggle auto-renewal on and off and see that the table reacts accordingly

### General
- [x]  Deal with detached contracts
A detached contract is either a cancelled monthly contract or a legacy contract. In either case we should not give access to resizing and cancellation features (maybe disabling the accordion). 

### Cancellation 
- [x] 1. Does not send `previous_purchase_id`
**Solution**: Make it `previous_purchase_ids["monthly"]`
- [x] 2. No need to send `period`, as it's always going to be `monthly`.
**Solution**: Delete it
- [x] 3. Only needs to send the `product_listing_id` as a string
We are always cancelling one product at a time. No need to make it a list of one product listing.
You seem to be sending the contract id for some reason. That needs fixing too.
- [x] 4. `account_id` seems hard-coded -- Not sure what you mean here. Maybe it was an earlier version?

_Details_:
**Request:** `DELETE /advantage/subscribe`
Schema of payload:
```
{
    "account_id": String(required=True),
    "previous_purchase_id": String(required=True),
    "product_listing_id": String(required=True),
}
```

### Auto-renewal
- [x] 1.  Reading the auto-renewal status
**Solution**: Use `subscriptions["is_auto_renewal_enabled"]` to auto select the right status.
- [x] 2. Missing auto-renewal endpoint call

_Details_:
**Request:** `POST /advantage/set-auto-renewal`
Schema of the payload: 
```
{
    "should_auto_renew": Boolean()
}
```
If `should_auto_renew` is not specified, I default to `False`. Maybe throwing a validation error is better?

### Resizing
- [x] 1. Number of machines is not pre-filled
**Solution**: Use `contract["machineCount"]` to get that number.
- [x] 2. Missing prices per contract
**Solution**: You can use `contract["price_per_unit"]` to get that number. If the contract is detached (`contract["is_detached"] == True `) it won't exist.
- [x] 3. Update the resizing endpoint call
Add `resizing` property and mark it as `true`. The default behaviour of a shop purchase is to **increase** the number of existing machines by whatever number you purchased. Resizing needs to **override** the existing number of machines. `resizing` will be used to differentiate between the two actions. If missing it defaults to `False`.

_Details_:
**Request:** `POST /advantage/subscribe`
Schema of the payload:
```
{
    "account_id": String(required=True),
    "period": String(enum=["monthly", "yearly"], required=True),
    "previous_purchase_id": String(required=True),
    "products": List(Nested(ProductSchema), required=True),
    "resizing": Boolean(),
}
ProductSchema:
{
    name = String() ## not required
    period = String(enum=["monthly", "yearly"]) ## not required
    price = Int() ## not required
    product_listing_id = String(required=True)
    quantity = Int(required=True)
}
```

Fixes https://github.com/canonical-web-and-design/web-squad/issues/3937


albertkol:
This is missing the `subscriptions` to `monthly_information` rename.

In hindsight it was not worth the change. I get 500 because of it. (I assume)

ClementChaumel:
@albertkol it was because of the `total_enterprise_contracts` that I didn't update. It should be fine now.

albertkol:
I resized, cancelled (both last and not last contract), enabled and disabled auto-renewed and it worked every time. I think it's good to go.

:+1:

albertkol (approved):

PR #9395, opened 2021-03-15
Change auto renewal parameter location

albertkol:
## Done

- Change location of the parameter for the auto renewal endpoint.


ClementChaumel (approved):
:+1:

PR #9425, opened 2021-03-19
Update caching rules with flask-base v0.8.0

nottrobin:
Use the [new caching defaults](https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/pull/45) from flask-base 0.8.0, for improved client performance and freshness of content.

Also simplify caching rules as follows:

- For private pages, simply use `cache-control: no-store`, as all the other options (`no-cache`, `must-revalidate`, `Pragma: nocache`, `max-age=0`) are simply less complete ways of doing the same thing.
- Remove the `max-age=61` special cache override for `/` and `/blog`, as the new default is `max-age=60` which is even fresher
  - *Note: The reason it was `61` and not `60` is because we used to use Squid as our proxy cache, and for some bizarre reason Squid wouldn't cache pages with `max-age=60` or less*
 
QA
--

Click around, refresh pages a fair bit, check they also seem to respond quickly. I guess https://ubuntu-com-9425.demos.haus/blog would be a good place to test this.

You can also try hitting the demo with `curl -I`, or use the network tab in the browser tools, to check the `cache-control` headers returned are as expected.

It's particularly important to make sure the private pages aren't going to be cached:

``` bash
$ curl -I https://ubuntu-com-9425.demos.haus/cube
cache-control: no-store

$ curl -I https://ubuntu-com-9425.demos.haus/core/build
cache-control: no-store

$ curl -I https://ubuntu-com-9425.demos.haus/advantage
cache-control: no-store
```

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -396,10 +396,6 @@ def takeovers_json():
     takeovers = get_takeovers()
     response = flask.jsonify(takeovers["active"])
     response.cache_control.max_age = "300"
-    response.cache_control._set_cache_value(
-        "stale-while-revalidate", "360", int
-    )
-    response.cache_control._set_cache_value("stale-if-error", "600", int)
 
     return response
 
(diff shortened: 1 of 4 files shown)


webapp/app.py
```diff
@@ -656,27 +652,10 @@ def cache_headers(response):
...
-        response.cache_control.private = True
-
-    if flask.request.path.startswith("/advantage"):
-        response.cache_control.private = True
-        response.headers[
-            "Cache-Control"
-        ] = "no-cache, no-store, must-revalidate"
-        response.headers["Pragma"] = "no-cache"
-
-    if flask.request.path in ["/", "/blog"]:
```
jpmartinspt, line 669: This one was to keep right?

nottrobin: No, this sets max-age=61. The new default is mas-age=60. I don't think we need to override it any more.

anthonydillon (approved):
Code looks good and QA :+1:

PR #9434, opened 2021-03-22
Move Certification Api requests

carkod:
## Done

- Move certification Api requests
- Add tests
- Cleanup names

## QA

- https://ubuntu-com-9434.demos.haus/certification should still work
- Check that tests pass
- Check that code matches https://github.com/canonical-web-and-design/certification.ubuntu.com/blob/master/webapp/api.py

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/166

### DIFF
--- a/tests/test_certification.py
+++ b/tests/test_certification.py
@@ -1,0 +1,26 @@
+# Packages
+from vcr_unittest import VCRTestCase
+
+# Local
+from webapp.app import app
+
+
+class TestCertification(VCRTestCase):
+    def _get_vcr_kwargs(self):
+        """
+        This removes the authorization header
+        from VCR so we don't record auth parameters
+        """
+        return {
+            "decode_compressed_response": True,
+            "filter_headers": ["Authorization", "Cookie"],
+        }
+
+    def setUp(self):
+        app.testing = True
+        self.client = app.test_client()
+        return super().setUp()
+
+    def test_certification_home(self):
+        response = self.client.get("/certification")
+        self.assertEqual(response.status_code, 200)
--- a/webapp/certification/api.py
+++ b/webapp/certification/api.py
@@ -1,5 +1,6 @@
 from requests import Session
 
+
 class CertificationAPI:
     """
     Method names and properties to describe and map directly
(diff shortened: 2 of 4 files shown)


webapp/certification/api.py
```diff
@@ -0,0 +1,182 @@
...
+        response = self.session.get(
+            f"{self.base_url}/{path.strip('/')}/?format=json", params=params
+        )
+
+        # Raise any HTTP errors
+        response.raise_for_status()
+
+        return response
+
+    def certifiedmakes(
```
jpmartinspt, line 32: @nottrobin any reason you chose not to use snake case for the methods names in this class? If not can we? I feel like it's a bit more readable.

carkod: My guess is that the Api has properties called this `certifiedmakes`, `certifiedmodeldevices`, `componentsummaries` so we followed this naming. But I guess we can change our methods. Updated

tests/test_certification.py
```diff
@@ -0,0 +1,38 @@
...
+    def setUp(self):
+        app.testing = True
+        self.client = app.test_client()
+        return super().setUp()
+
+    def test_certification_home(self):
+        response = self.client.get("/certification")
+        self.assertEqual(response.status_code, 200)
+
+    def test_home_api_endpoint(self):
```
jpmartinspt, line 30: Is this test really needed? Once the you move the logic into the home view `test_certification_home` should fail if the response returned from the API is not enough to build the homepage. I also can't seem to spot where in `CertificationAPI` do we need that key for it to work properly , am I missing something?

carkod: I think it's a good first example test for future calls to the API but I could do it in the next PR.
Yes that key is in line 75 of `api.py`

jpmartinspt: On line `75` you are just passing `certifiedmodels` as a parameter to `_get` so it can build the URL. I don't think at any point the class requires `certifiedmodels` to be in the response payload to function.

carkod: Removed.

jpmartinspt (approved):

PR #9441, opened 2021-03-23
Update the agreement file downloads to use assets

anthonydillon:
## Done
Update the agreement file downloads to use assets

## QA
- Go to /legal/contributors/agreement
- Check that the two "download as a PDF" links work from assets

## Issue / Card
Fixes https://github.com/canonical-web-and-design/canonical.com/issues/367


bethcollins92 (approved):
Links both LGTM 👍🏼 

Not related but the border at the top of the form is missing but it also is on live:

image

is this the same for you or is it a chrome thing?

anthonydillon:
Yeah there are some custom styling removing the top border with `fieldset + fieldset` which might have a knock on effect if I change now.

PR #9442, opened 2021-03-24
Overview page

carkod:
## Done

Desktop, Server, Iot, SoC data for overview page

## QA

https://ubuntu-com-9442.demos.haus/certification should show all the data in json format

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/165

### DIFF
--- a/webapp/certification/views.py
+++ b/webapp/certification/views.py
@@ -12,4 +12,60 @@ api = CertificationAPI(
 
 
 def certification_home():
-    return flask.render_template("certification/index.html")
+
+    certified_releases = api.certified_releases(limit="0")["objects"]
+    certified_makes = api.certified_makes(limit="0")["objects"]
+
+    # Desktop section
+    desktop_releases = []
+    desktop_vendors = []
+
+    # SoC section
+    soc_releases = []
+    soc_vendors = []
+
+    # IoT section
+    iot_releases = []
+    iot_vendors = []
+
+    for release in certified_releases:
+        if int(release["desktops"]) > 0 or int(release["laptops"]) > 0:
+            desktop_releases.append(release)
+
+        if int(release["smart_core"] > 1):
+            iot_vendors.append(release)
+
+        if int(release["soc"] > 1):
+            soc_releases.append(release)
+
+    for vendor in certified_makes:
+        if int(vendor["desktops"]) > 0 or int(vendor["laptops"]) > 0:
+            desktop_vendors.append(vendor)
+
+        if int(vendor["smart_core"] > 1):
+            iot_releases.append(vendor)
+
+        if int(vendor["soc"] > 1):
+            soc_vendors.append(vendor)
+
+    # Server section
+    server_releases = {}
+    server_vendors = api.vendor_summaries_server()["vendors"]
+
+    for vendor in server_vendors:
+        for release in vendor["releases"]:
+            if release in server_releases:
+                server_releases[release] += vendor[release]
+            else:
+                server_releases[release] = vendor[release]
+
+    return flask.render_template(
+        "certification/index.html",
+        desktop_releases=desktop_releases,
+        desktop_vendors=desktop_vendors,
+        server_releases=server_releases,
+        iot_releases=iot_releases,
+        iot_vendors=iot_vendors,
+        soc_releases=soc_releases,
+        soc_vendors=soc_vendors,
+    )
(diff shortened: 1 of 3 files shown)


bethcollins92:
QA LGTM 👍🏼

webapp/certification/views.py
```diff
@@ -12,4 +12,44 @@
 
 
 def certification_home():
-    return flask.render_template("certification/index.html")
+
+    # Desktop section
+    desktop_releases = []
+    desktop_vendors = []
+
+    for release in api.certified_releases(limit="0")["objects"]:
```
jpmartinspt, line 20: I think the certifiedreleases endpoint also supports queries like `?desktops__gte=1&laptops__gte=1`

jpmartinspt:
This page will need 7 HTTP requests to load right now, I don't want to block any work, but it's possible that we could benefit from getting all `certified_releases` and `certified_makes` and group the info as we need on the view. Just make sure to consider any cases where you think we could reduce the number of calls/improve things before you merge the feature branch.

webapp/certification/views.py
```diff
@@ -12,4 +12,46 @@
...
+
+    # SoC section
+    soc_releases = api.certified_releases(soc__gte="1")["objects"]
+    soc_vendors = api.certified_makes(soc__gte="1")["objects"]
+
+    return flask.render_template(
+        "certification/index.html",
+        desktop_releases=desktop_releases,
+        desktop_vendors=desktop_vendors,
+        server_releases=server_releases,
```
jpmartinspt, line 52: not passing down `server_vendors`

carkod: I just actually removed that, because it's not required in the design/wireframe

image

carkod:
> This page will need 7 HTTP requests to load right now, I don't want to block any work, but it's possible that we could benefit from getting all `certified_releases` and `certified_makes` and group the info as we need on the view. You don't need to address this now but please consider any cases where you think we could reduce the number of calls/improve things before you merge the feature branch.

Refactored

webapp/certification/views.py
jpmartinspt: Ah ok. It's still on the template.

carkod: Renamed, because I think it's confusing, this is actually raw data from the API, not an empty array like the others.

carkod: Removed from the template. I think beth is already aware of this anyway.

webapp/certification/views.py
```diff
@@ -12,4 +12,69 @@
...
-    return flask.render_template("certification/index.html")
+
+    certified_releases = api.certified_releases(limit="0")["objects"]
+    certified_makes = api.certified_makes(limit="0")["objects"]
+
+    # Desktop section
+    desktop_releases = []
+    desktop_vendors = []
+
+    for release in certified_releases:
```
jpmartinspt, line 23: If you declare all the list at the tops you can do all the appends in these two for loops

webapp/certification/views.py
```diff
@@ -12,4 +12,69 @@
...
+        if int(release["desktops"]) > 0 or int(release["laptops"]) > 0:
+            desktop_releases.append(release)
+
+    for vendor in certified_makes:
+        if int(vendor["desktops"]) > 0 or int(vendor["laptops"]) > 0:
+            desktop_vendors.append(vendor)
+
+    # Server section
```
jpmartinspt, line 31: If all you need for the overview are the counts you can refactor the for loop to be
```
    # Server section
    server_releases = {}
    server_vendors = api.vendor_summaries_server()["vendors"]

    for vendor in server_vendors:
        for release in vendor["releases"]:
            if release in server_releases:
                server_releases[release] += vendor[release]
            else:
                server_releases[release] = vendor[release]
```

jpmartinspt: It would also be interesting to know if we could just use the output of `https://certification.canonical.com/api/v1/certifiedreleases/?servers__gte=1`. The counts seem slightly different between the two endpoints so it might be that we need to use `vendor_summaries`

carkod: Not sure what you mean but I did the math and it matches exactly

webapp/certification/views.py
carkod: Thanks that's a good one

jpmartinspt (approved):

carkod:
Thanks!

PR #9443, opened 2021-03-24
Fix tax_id parameter

albertkol:
I assumed tax_id would be a string. It turns out it's a dictionary. I
updated the validation to expect a dictionary instead.

## QA

- Go to https://ubuntu-com-9443.demos.haus/advantage/subscribe?test_backend=true
- Select and item and get to the modal
- Open the network tab
- Insert a VAT number: 123 you should get an error (correctly, as it's an invalid number) from the calls: `POST /advantage/customer-info` and `POST /advantage/customer-info-anon` will fail
- Insert GB 123 123 123 it should work this time then finish the payment

### DIFF
--- a/webapp/advantage/schemas.py
+++ b/webapp/advantage/schemas.py
@@ -25,6 +25,11 @@ class AddressSchema(Schema):
     state = String()
 
 
+class TaxIdSchema(Schema):
+    type = String()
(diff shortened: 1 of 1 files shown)


ClementChaumel (approved):
:book: LGTM

PR #9448, opened 2021-03-24
Start trial

albertkol:
## Done

`POST /advantage/post-guest-trial` with
```
email: string
acount_name: string
name: string
address: {
  city: string
  country: string
  line1: string
  postal_code: string
  state: string
}
products: [
  {
    product_listing_id: string
    quantity: int
  }
]
```
It will store the info of the purchase in the session. Once the user
logs in and arrives on `/advantage` it will trigger the
`GET /advantage/save-guest-trial` endpoint with that data.
This second endpoint will fetch an `account_id` for the user(using email and
account_name), update the customer info(using name and address) and then
make the trial (using account_id and products).

`POST /advantage/post-trial` with
```
acount_id: string
products: [
  {
    product_listing_id: string
    quantity: int
  }
]
```
For logged in users, we use their current `account_id` and we create a
trial immediately with the info in `products`.

## QA

We QA when the front-end is ready.

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/web-squad/issues/3948


albertkol:
This was replaced by canonical-web-and-design/ubuntu.com#9511

PR #9475, opened 2021-03-26
Certification search

carkod:
## Done

Search results for all certification

## QA

- Search using queries .e.g.:
https://ubuntu-com-9475.demos.haus/certification?text=lenovo&form=Desktop&release=20.04+LTS&vendors=Lenovo

- Pagination example:
https://ubuntu-com-9475.demos.haus/certification?text=lenovo&limit=20&offset=21
if no parameters are passed, limit=20 by default

- Filter numbers example:
https://ubuntu-com-9475.demos.haus/certification?text=lenovo&page=2&limit=20

List of query params you can use:
- text (search box)
- form (Desktop, Laptops, Components, Ubuntu Core, Soc, Server)
- release (with "LTS" included)
- vendors (I'll talk do David why he prefers vendorsname)
- offset
- limit (default 20)

Let me know if I am missing anything thanks

## Issue / Card

Fixes https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/147


bethcollins92:
Data LGTM 👍🏼

webapp/certification/helpers.py
```diff
@@ -0,0 +1,29 @@
+def get_pagination_page_array(page, total_pages):
```
jpmartinspt, line 1: I don't think you need this, look at how we do pagination for CVE's and Tutorials, there are shared templates to achieve this.

carkod: But the difference is that in this one we are doing it with `pages` instead of `limit` and `offset`?

jpmartinspt: There are two templates on the codebase one of them uses pages the other uses offsets. I like offset + limit. I actually don't know if we prefer one or the other maybe @nottrobin knows?

carkod: I am sure robin prefers the offset + limt. But I thought that one is harder for Front-end needs more calculation, but I guess @bethcollins92 take the code from [CVE](https://github.com/canonical-web-and-design/ubuntu.com/blob/master/templates/security/cve/index.html#L88).

jpmartinspt:
Looks like there has been a bad rebase? A lot of new changes.

webapp/certification/views.py
```diff
@@ -71,13 +86,96 @@ def certification_home():
...
+            "certification/search-results.html",
+            results=results,
+            query=query,
+            form=forms,
+            releases=releases,
+            form_filters=form_filters,
+            release_filters=release_filters,
+            vendor_filters=vendor_filters,
+            vendors=vendors,
+            query_string="&".join(query_items),
```
jpmartinspt, line 163: Why do we need this variable?

webapp/certification/views.py
```diff
@@ -71,13 +86,96 @@ def certification_home():
...
+            release_filters[release] = 0
+
+        vendor_filters = {}
+        for vendor in all_vendors:
+            vendor_filters[vendor] = 0
+
+        results = models_response["objects"]
+
+        # Populate filter numbers
+        for index, model in enumerate(results):
```
jpmartinspt, line 137: I'm not entirely sure I get this, but if what we want to do here is to get the number of results per filter type should we not be incrementing by one instead?
```
form_filters[model["category"]] += 1
release_filters[model["release"]] += 1
vendor_filters[model["make"]] += 1
```

carkod: Done.

jpmartinspt (approved):

PR #9489, opened 2021-03-30
advantage: implement view as functionality

frankban:
## Done

advantage: implement view as functionality.

This will allow admins to view /advantage as a specific user by providing the `email` query.

## QA

- Go to http://0.0.0.0:8001/advantage?test_backend=true , log in and check that you see your contracts as usual.
- Now, assuming you have the contract ACL in ua-contracts staging, go to http://0.0.0.0:8001/advantage?test_backend=true&email=allan.vidal@canonical.com and you should see Allan's contracts.

### DIFF
--- a/webapp/advantage/views.py
+++ b/webapp/advantage/views.py
@@ -35,7 +35,7 @@ session = talisker.requests.get_session()
 
 
 @advantage_checks(check_list=["is_maintenance", "view_need_user"])
-@use_kwargs({"subscription": String()}, location="query")
+@use_kwargs({"subscription": String(), "email": String()}, location="query")
 def advantage_view(**kwargs):
     is_test_backend = kwargs.get("test_backend")
     api_url = kwargs.get("api_url")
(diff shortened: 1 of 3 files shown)


webapp/advantage/views.py
```diff
@@ -121,9 +123,14 @@ def advantage_view(**kwargs):
                 }
 
         for contract in account["contracts"]:
-            contract_id = contract["contractInfo"]["id"]
-            contract["token"] = advantage.get_contract_token(contract_id)
             contract_info = contract["contractInfo"]
+            if not contract_info.get("items"):
```
albertkol, line 127: I am not sure I understand what this `if` or `TODO` is about.
..and is this work to be done on our end or on your end?

If this is something we have to do, I'd rather have an issue created for it and have the whole `if` removed.

frankban: Our end. We might have some leftover contracts with no items, and therefore without EffectiveFrom, EffectiveTo etc, leading to key errors in the code below. This check makes the frontend ignore such contracts.

albertkol: Could we have the API not return contracts with no items instead?

frankban: there should not be contracts with no items, hence the TODO for me. Once fixed on ua-contracts we can remove this check.

PR #9498, opened 2021-04-01
Certification hardware details

carkod:
## Done

Hardware details page, a.k.a Models-release hardware details page.
Wireframe
hardware-details-wireframe

## QA

Check /certification/202102-28707/20.04%20LTS

## Issue / Card
Fixes https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/173

### DIFF
--- a/webapp/certification/views.py
+++ b/webapp/certification/views.py
@@ -15,6 +15,69 @@ api = CertificationAPI(
 )
 
 
+def certification_hardware_details(canonical_id, release):
+
+    try:
+        models = api.certified_models(canonical_id=canonical_id)["objects"][0]
+    except KeyError:
+        abort(404)
+
+    model_releases = api.certified_model_details(
+        canonical_id=canonical_id, limit="0"
+    )["objects"]
+
+    # Release section
+    release_details = next(
+        (
+            detail
+            for detail in model_releases
+            if detail["certified_release"] == release
+        ),
+        None,
+    )
+    if not release_details:
+        abort(404)
+
+    model_devices = api.certified_model_devices(
+        canonical_id=canonical_id, limit="0"
+    )["objects"]
+
+    hardware_details = {}
+    for device in model_devices:
+        device_info = {
+            "name": (
+                f"{device['make']} {device['name']}"
+                f" {device['subproduct_name']}"
+            ),
+            "bus": device["bus"],
+            "identifier": device["identifier"],
+        }
+
+        category = device["category"]
+        if category not in ["BIOS", "USB"]:
+            category = category.capitalize()
+
+        if category not in hardware_details:
+            hardware_details[category] = []
+
+        hardware_details[category].append(device_info)
+
+    # default to category, which contains the least specific form_factor
+    form_factor = release_details.get("form_factor", category)
+
+    return render_template(
+        "certification/hardware-details.html",
+        canonical_id=canonical_id,
+        model_name=models["model"],
+        form=models["category"],
+        form_factor=form_factor,
+        vendor=models["make"],
+        major_release=models["major_release"],
+        hardware_details=hardware_details,
+        release_details=release_details,
+    )
+
+
 def certification_model_details(canonical_id):
     models = api.certified_models(canonical_id=canonical_id)["objects"]
 
@@ -76,18 +139,16 @@ def certification_model_details(canonical_id):
                             }
                         )
 
-    # Build model name
-    model_names = [model["model"] for model in models]
-
-    category = models[0]["category"]
     # default to category, which contains the least specific form_factor
-    form_factor = model_release and model_release.get("form_factor", category)
+    form_factor = model_release and model_release.get(
+        "form_factor", models[0]["category"]
+    )
 
     return render_template(
         "certification/model-details.html",
         canonical_id=canonical_id,
-        name=", ".join(model_names),
-        category=category,
+        name=models[0]["model"],
+        category=models[0]["category"],
         form_factor=form_factor,
         vendor=models[0]["make"],
         major_release=models[0]["major_release"],
--- a/templates/certification/hardware-details.html
+++ b/templates/certification/hardware-details.html
@@ -1,0 +1,35 @@
+<h1>This is the hardware details page</h1>
+
+<h2>ID (Canonical ID)</h2>
+{{canonical_id}}
+
+<h2>Name:</h2>
+{{ name }}
+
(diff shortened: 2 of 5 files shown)


bethcollins92:
If I look at the data on https://ubuntu-com-9498.demos.haus/certification/201906-27091/18.04%20LTS compared to the live site https://certification.ubuntu.com/hardware/201906-27091, it seems there is some keys missing. For example card reader, or Cdrom, Audio etc. Is there another variable with this data in that you're not printing on the demo?

carkod:
> If I look at the data on https://ubuntu-com-9498.demos.haus/certification/201906-27091/18.04%20LTS compared to the live site https://certification.ubuntu.com/hardware/201906-27091, it seems there is some keys missing. For example card reader, or Cdrom, Audio etc. Is there another variable with this data in that you're not printing on the demo?

Yep, I wasn't using the device API so didn't get that data. You have now a new section `hardware_details` which contains all this info.

bethcollins92:
QA LGTM

webapp/certification/views.py
```diff
@@ -15,6 +15,99 @@
...
+                                f"{device['make']} {device['name']}"
+                                f" {device['subproduct_name']}"
+                            ),
+                            "bus": device["bus"],
+                            "identifier": device["identifier"],
+                        }
+                    )
+
+    # Build model name
+    model_names = next(model["model"] for model in models)
```
jpmartinspt, line 89: Why not models[0]["model"]?

jpmartinspt: or `model["model"]` if you take the suggestion at the top.

webapp/certification/views.py
```diff
@@ -15,6 +15,99 @@
 )
 
 
+def certification_hardware_details(canonical_id, release):
+    models = api.certified_models(canonical_id=canonical_id)["objects"]
```
jpmartinspt, line 19: You could do:
```
try:
    model = api.certified_models(canonical_id=canonical_id)["objects"][0]
except KeyError:
    abort(404)
```
and then you would can just replace occurences of `models[0]` with `model`

webapp/certification/views.py
```diff
@@ -15,6 +15,99 @@
...
+        device_info = {
+            "name": (
+                f"{device['make']} {device['name']}"
+                f" {device['subproduct_name']}"
+            ),
+            "bus": device["bus"],
+            "identifier": device["identifier"],
+        }
+
+        category = device["category"]
```
jpmartinspt, line 55: I think we could future proof a bit by checking only capitalising strings the are not all uppercase. I'm assuming that that's the reason we do the current check, if not disregard.
```python3
    category = (
        device["category"]
        if device["category"].isupper()
        else device["category"].capitalize()
    )
```

webapp/certification/views.py
```diff
@@ -15,6 +15,99 @@
 )
 
 
+def certification_hardware_details(canonical_id, release):
+    models = api.certified_models(canonical_id=canonical_id)["objects"]
+
+    if not models:
+        abort(404)
+
+    model_devices = api.certified_model_devices(
```
jpmartinspt, line 24: You could probably delay this call until you know there are `release_details`

webapp/certification/views.py
```diff
@@ -15,6 +15,99 @@
...
+                            "identifier": device["identifier"],
+                        }
+                    )
+
+    # Build model name
+    model_names = next(model["model"] for model in models)
+
+    category = models[0]["category"]
+    # default to category, which contains the least specific form_factor
+    form_factor = release_details and release_details.get(
```
jpmartinspt, line 93: No need to check for `release_details` was we abort every requests where it is falsy.

webapp/certification/views.py
```diff
@@ -15,6 +15,99 @@
...
+    for device_category, devices in release_details.items():
+        if (
+            device_category in ["video", "processor", "network", "wireless"]
+            and devices
+        ):
+            device_category = device_category.capitalize()
+
+            components[device_category] = []
+
+            if device_category in components:
```
jpmartinspt, line 75: This condition is always going to be true since you execute `components[device_category] = []` just before.
It can probably be removed.

webapp/certification/views.py
```diff
@@ -15,6 +15,99 @@
...
+            category = category.capitalize()
+
+        if category not in hardware_details:
+            hardware_details[category] = []
+
+        hardware_details[category].append(device_info)
+
+    # Components section
+    components = {}
+    for device_category, devices in release_details.items():
```
jpmartinspt, line 66: A slight variation with the comments I mentioned bellow addressed. What do you think?

```python3
components = {}
for category in ["Video", "Processor", "Network", "Wireless"]:
    devices = release_items.get(category)
    if not devices:
        continue
    
    components[category] = []
    for device in devices:
        components[category].append(
            {
                "name": (
                    f"{device['make']} {device['name']}"
                    f" {device['subproduct_name']}"
                ),
                "bus": device["bus"],
                "identifier": device["identifier"],
            }
        )
```

carkod: Looks good, loops is shorter. but I think front-end will have to capitilize. done.

webapp/certification/views.py
carkod: addressed in the devices loop refactor

webapp/certification/views.py
carkod: Moved down. Done

webapp/certification/views.py
carkod: Done

webapp/certification/views.py
carkod: Done

webapp/certification/views.py
jpmartinspt: I capitalised the categories on the for loop :)

webapp/certification/views.py
```diff
@@ -15,6 +15,89 @@
...
+            category = category.capitalize()
+
+        if category not in hardware_details:
+            hardware_details[category] = []
+
+        hardware_details[category].append(device_info)
+
+    # Components section
+    components = {}
+    for category in ["video", "processor", "network", "wireless"]:
```
jpmartinspt, line 67: If you look in the suggestion I made the categories were capitalised, no need to do it on the frontend.

carkod: Sorry, just talked to Beth about this, we don't need a components section anymore. The `hardware_details` is enough

PR #9514, opened 2021-04-07
Add login and username to the navigation header

### LINKED ISSUE #9289: [Global login in navigation] Build the static dropdown from username

anthonydillon:
## Done
Add login and username to the navigation header

## QA
- Go to the demo
- See there is an action "Sign in" in the header
- Click it and see you return to the same page signed in
- Check your name in the top right
- Click your name and see your drop-down of local services
- Check different viewports (Small, medium and large) __There is no drop-down on small screens__

## Issue / Card
Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9288
Fixes #9289 

## Screenshots
Small
image

Medium
image

Large
image

Menu active
image


anasereijo:
@anthonydillon thanks for doing this so quickly 🥇 

I have some suggestions but no blockers, so it can wait till the next stage of this project if you don't have time now.

- [ ] Is it possible to remove the top shadow from the login dropdown to match what we do with the nav dropdowns? It's just the top one. See below

Screenshot 2021-04-07 at 15 37 23

Screenshot 2021-04-07 at 15 48 41


- [ ] why the height of each element isn't 48px as per [vanilla](https://vanillaframework.io/docs/examples/patterns/navigation/subnav-dark#)?

Screenshot 2021-04-07 at 15 51 28
Screenshot 2021-04-07 at 15 51 39

- [ ] I thought we'd have Account settings link above Logout

- [ ] I think this is more of a UX question (@cassiocassio): currently, we can open the login dropdown whilst there's another nav dropdown opened - does this make sense? Should the nav dropdown close if we click login?

Screenshot 2021-04-07 at 15 54 35


As we discussed earlier, we should get some time (UX and Visual) to work on smaller screens designs but what we have now is ok for a first version.

anthonydillon:
@anasereijo Great points thanks, addressed them all.

anasereijo:
👍 thanks!

ziheliu214:
after clicking 'sign in' this screen is presented, but before confirming my action my profile is already showing on top right. The cancel button is not working too.
Screenshot 2021-04-07 at 17 41 36

Different view point works. ✅

anthonydillon:
@ziheliu214 good point. The way it works is you are logged in to SSO but you are signing into the ubuntu.com service so this step is required.

clagom:
@anthonydillon I see this at 840px width, between breaking points
Screenshot 2021-04-08 at 13 32 23

anthonydillon:
Yeah that is an existing issue. I'll file an issue to fix it.

nottrobin:
This is shaping up really nicely. I'll be super happy to get this merged, it'll be a huge boon to the site. I love the architecture, it's nice and simple and neat. Great work.

Just a few things:

### Suggested changes

- Shouldn't the "CUBE microcertifications" link go to `/cube/microcerts` rather than `/cube`? Going to `/cube` is rather confusing because there's nothing about that page that is specific to a logged-in user.

- Why is it that we don't display the drop-down in mobile view? If this is the only way to get this merged quickly, I think it's fine. But if I switched from my desktop to my mobile I would certainly expect that I could still use the header to get to my UA Subscriptions, and might be quite frustrated to discover that I can't.

### Non-blocking observations

- It's a shame that after clicking "account settings" and landing on login.ubuntu.com there's no way to get back to ubuntu.com. The styling is similar enough that it feels like you're on the same site, but you're not. So it's a bit disorientating to discover that the nav, search, logo link etc. are missing. Maybe the old "external site" icon could have helped here, although I don't know whether we're allowed to use it any more. I feel strongly that this **shouldn't** block this PR, but we should follow up afterwards on the best way to improve the user experience.

- Similarly, it's a shame that the account name is missing initially, and when it appears it pushes the search to the left. It means the page visibly jumps on every page load. Even though it doesn't actually push content down so it probably won't upset Google, I find it quite jarring and I don't think it looks super professional. I do feel that the top-right is the correct place for the account info so I wouldn't change that.
   - I just wonder if maybe we could hide the search until the account info appears, with a timeout that lets it appear anyway? 'cos if that element is gonna jump it's not super useable anyway until the account arrived.

anthonydillon:
> Why is it that we don't display the drop-down in mobile view? If this is the only way to get this merged quickly, I think it's fine. But if I switched from my desktop to my mobile I would certainly expect that I could still use the header to get to my UA Subscriptions, and might be quite frustrated to discover that I can't.

This is due to a lack of UX and design for how the small screen interaction should work. Therefore as you say this is a short term solution to keep us moving.

clagom:
> > Why is it that we don't display the drop-down in mobile view? If this is the only way to get this merged quickly, I think it's fine. But if I switched from my desktop to my mobile I would certainly expect that I could still use the header to get to my UA Subscriptions, and might be quite frustrated to discover that I can't.
> 
> This is due to a lack of UX and design for how the small screen interaction should work. Therefore as you say this is a short term solution to keep us moving.

@anasereijo can you iterate on this point please? I think it's mainly a visual implementation of the same drop-down also on mobile. CC @SirSamTumless

anthonydillon:
> I just wonder if maybe we could hide the search until the account info appears, with a timeout that lets it appear anyway? 'cos if that element is gonna jump it's not super useable anyway until the account arrived.

I agree and am tempted to try this but think it might take some more thinking. For example,
- Originally I thought switching search and user positions would be best but the top right is a better place for it.
- Maybe switch to a visualisation of the user such as Google does (First letter with a colour generated by the full name) which would mean nothing moves. The add the full name to the top of the dropdown.
- etc :shrug:

nottrobin (approved):

PR #9534, opened 2021-04-12
Add country to ensurePurchaseAccount

alnvdl:
We will soon introduce a requirement in ua-contracts so that all endpoints sending us shop customer information must always include both the name and the country. This is to ensure we don't create invalid entries in Stripe and Salesforce.

It seems this requirement is currently respected everywhere here, except in `ensurePurchaseAccount` (which is fine, it never required such field). Including this field in the request now before we make it mandatory should have no harmful effects: it will just be ignored for now.

## QA

- Go to https://ubuntu-com-9534.demos.haus/advantage/subscribe?test_backend=true
- Try to make a guest purchase in the UA shop; it should still succeed, and both the Flask app and ua-contracts should be receiving the country.

### DIFF
--- a/static/js/src/ua-payment-modal.js
+++ b/static/js/src/ua-payment-modal.js
@@ -749,7 +749,8 @@ function handleGuestPaymentMethodResponse(data) {
   ensurePurchaseAccount(
     customerInfo.email,
     customerInfo.accountName,
-    paymentMethod.id
+    paymentMethod.id,
+    customerInfo.address.country
   ).then((data) => {
     if (data.code) {
       // an error was returned, most likely cause
--- a/static/js/src/advantage/contracts-api.js
+++ b/static/js/src/advantage/contracts-api.js
@@ -15,7 +15,8 @@ export async function getPurchase(purchaseID) {
 export async function ensurePurchaseAccount(
   email,
   accountName,
-  paymentMethodID
+  paymentMethodID,
+  country
 ) {
   const queryString = window.location.search; // Pass arguments to the flask backend eg. "test=backend=true"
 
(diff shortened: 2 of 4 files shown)


frankban (approved):
The code looks good, but is this really required? Don't we have the address mandatory already when setting customer info?

alnvdl:
> The code looks good, but is this really required? Don't we have the address mandatory already when setting customer info?

@frankban No fields are currently mandatory when setting customer info. We will be making some of them mandatory now. Do you believe the entire address should be mandatory?

albertkol:
@alnvdl is it possible to rebase this branch on the `ua-monthly-subs-and-resizing`? 
I hope this does not need to make it to prod soon.

static/js/src/ua-payment-modal.js
```diff
@@ -746,7 +746,8 @@ function handleGuestPaymentMethodResponse(data) {
   ensurePurchaseAccount(
     customerInfo.email,
     customerInfo.accountName,
-    paymentMethod.id
+    paymentMethod.id,
+    customerInfo.address.country
```
ClementChaumel, line 750: I'm refactoring this file at the moment but I'll keep this requirement in mind. Thanks!

alnvdl:
> @alnvdl is it possible to rebase this branch on the `ua-monthly-subs-and-resizing`?
> I hope this does not need to make it to prod soon.

Ok, I'll rebase it, and it does not need to go to production soon.

alnvdl:
@albertkol @ClementChaumel I rebased this on top of `ua-monthly-subs-and-resizing`.

ClementChaumel (approved):

PR #9540, opened 2021-04-14
Text updates on /openstack section

pmahnke:
## Done

- Text changes and additions based on Tytus review

## QA

     - http://0.0.0.0:8001/openstack/compare
         - One minor problem I've already raised offline is that the "a" in the "OpenStack distributions: a comparison" title should be moved to the next line. Is that something you could make happen, please?
             - I have added a break, but again, normally not a good practice with flowing text. 
     - http://0.0.0.0:8001/openstack/features
          - In the sub-title ("Total automation. Architectural freedom. Enterprise dashboards. Stability and security above all."), I'd suggest that we move the "Enterprise" word to the next line.
              - I have added the line breaks; however, normally that is a fool's errand on a responsive site.
          - In the "Finance and government architecture" section, I'd suggest we move "Why financial institutions use Ubuntu" CTA one line down so that we keep all CTAs in the same line.
              - This will not work well on responsive websites and there are 100's of things like this all over the site like this.  Perhaps make the intro text longer?
          - I'm not sure why you guys decided to make the "Read our MAAS ebook" a green button. This is one of the least relevant CTAs on this page. If we want to make more green buttons on the page, that would be rather the OpenStack upgrades webinar than the MAAS ebook.
              - I changed the button to white. 
          - The "Economics are the ultimate measure of effectiveness in the modern compute substrate; your private cloud has to be as efficient as your public clouds or it won’t attract internal workloads." paragraph under the "Compute" section is not in the copy doc and should be removed.
              - I fixed a few headings to add the 'With' into the link.
              - I also moved the 'With's into the href... looks odd 
     - http://0.0.0.0:8001/openstack
         - It looks like the new page is missing 4 paragraphs from the copy doc: everything from "Every OpenStack version supported" to "Carrier-grade private cloud"
         - The "Further reading" section was supposed to feature articles tagged with "openstack" and "private cloud" keywords while it doesn't seem to - using the 'OpenStack' tag, we can only do a single tag... looks like you are using lots of similar tags on blog posts, that doesn't do anything really.  For example - "bare metal , Charmed OpenStack , Charms , cloud , OpenStack , Private cloud" would look better to me as "bare metal, OpenStack, Juju, Private cloud"

- See that the changes are as described


SirSamTumless:
One thing that was missed is the "In the "_Finance and government architecture" section, I'd suggest we move "Why financial institutions use Ubuntu" CTA one line down so that we keep all CTAs in the same line_" is still miss-aligned.

pmahnke:
@SirSamTumless re your comment 'This will not work well on responsive websites and there are 100's of things like this all over the site like this. Perhaps make the intro text longer?'

pmahnke:
@SirSamTumless can you re-review and approve (if it is ok)?

SirSamTumless (approved):

PR #9547, opened 2021-04-15
Update instruction copy

### LINKED ISSUE #9517: Beta shop says I don't need to attach machines to set up UA services
When I go to the beta shop at https://ubuntu-com-9240.demos.haus/advantage, it says "You don’t need to attach machines before setting up UA services." for Ubuntu 16.04 and 18.04.

Screenshot from 2021-04-07 12-06-34

Is this because the UA client hasn't been officially released for these series yet?

Perhaps this will be addressed soon, but I would expect this to either link to documentation on how to actually set up access to services with the "old" UA client, or to assume that the new UA client is already released in this copy, if it is expected to be released along with the beta shop.

I think it is misleading as it is though; as a customer I would read this and think that 16.04 and 18.04 automatically enable all the UA services with no special instructions required -- which is not the case.

ClementChaumel:
## Done

- Update instruction copy for 16.04 and 18.04 according to https://github.com/canonical-web-and-design/ubuntu.com/issues/9517#issuecomment-820183733

## QA

-  https://ubuntu-com-9547.demos.haus/advantage?test_backend=true

## Issue / Card

Fixes #9517

## Screenshots

image


ClementChaumel:
@wgx PTA at the remixed copy

wgx:
LGTM 👍🏽

anasereijo:
LGTM 👍

anasereijo:
@albertkol please review

PR #9549, opened 2021-04-15
Add a null check for the login menu when logged out

anthonydillon:
## Done
Add a null check for the login menu when logged out

## QA
- Open the demo and login
- Check there is no error in the console
- Check the login menu works 
- Open the demo in a private window with the console open
- Check there is no error in the console

### DIFF
--- a/static/js/src/navigation.js
+++ b/static/js/src/navigation.js
@@ -273,22 +273,24 @@ if (accountContainer && accountContainerSmall) {
       function setupAllContextualMenus(contextualMenuToggleSelector) {
         // Setup all menu toggles on the page.
         var userToggle = document.querySelector(contextualMenuToggleSelector);
-        setupContextualMenu(userToggle);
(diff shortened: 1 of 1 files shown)


static/js/src/navigation.js
```diff
@@ -273,22 +273,24 @@ if (accountContainer && accountContainerSmall) {
...
-            userToggle.contains(event.target) ||
-            contextualMenu.contains(event.target)
-          );
-
-          if (clickOutside) {
-            toggleMenu(userToggle, false);
-          }
-        });
+        if (userToggle) {
+          setupContextualMenu(userToggle);
```
anthonydillon, line 277: Most of the changes are whitespace due to the if statement wrapping this code. Add `?w=1` to the diff page to see the true changes.

ClementChaumel (approved):
Looks good! 
I didn't know about this `?w=1` trick, thanks!

Issue #176, opened 2016-04-08
small screen /download/server/arm page should not have pictos on small screen in bottom sections


pmahnke:
## Context

don't show pictos in 'For developers' and 'Ready to deploy now' sections on the small screen
## Process
1. go to /download/server/arm on a small screen
## Expected result

no pictos

PR #9551, opened 2021-04-15
Refactor search results page

bethcollins92:
## Done

- Add form factor headers only if user selects form factor from secondary nav, or selects the "All ubuntu certified <form> hardware" link from the overview page. If user selects filters and clicks `update` the header should now be the standard search results header.
- Fix SoC path in `/certification/views.py`
- Fix error by null checking `forms` in `/certification/views.py`
- Add `filters=True` flag to URL to be able to differentiate between form factor view and search results view. 

## QA

- Start here: https://ubuntu-com-9551.demos.haus/certification
- Click through to the search results from the "All ubuntu certified `<form>` hardware" links from the cards on the overview page. This should show the form-factor specific headings. 
- Use the secondary nav to select "Desktop" or "IoT" etc - these should show the specific form-factor headings - then select a filter and click update and you should be back to the generic search results page.
- Go back to the overview page and select from the drop down and type something to do a normal search. The header should be the generic search results header. 
- See the [wireframe](https://docs.google.com/document/d/1vIf1vC1mg72JbzmahrcKCaRZN-qMLCgin_i9iqjUxpM/edit#heading=h.b6q302go5m5l) for a description on the differences between the generic search results page and the form-factor pages.

#### Known issues:
- nav not highlighting correct page - @carkod will work on fixing [this](https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/178) next week. 
- Numbers on overview page are incorrect: [issue](https://github.com/canonical-web-and-design/ubuntu.com/issues/9554) filed here

## Issue / Card

Fixes[ #176](https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/176)

### DIFF
--- a/webapp/certification/views.py
+++ b/webapp/certification/views.py
@@ -136,7 +136,9 @@ def certification_home():
             iot_releases.append(release)
 
         if int(release["soc"] > 1):
-            release["path"] = f"/certification?form=SoC&release={version}"
+            release[
+                "path"
+            ] = f"/certification?form=Server%20SoC&release={version}"
             soc_releases.append(release)
 
     for vendor in certified_makes:
@@ -180,7 +182,7 @@ def certification_home():
             if request.args.getlist("form")
             else None
         )
-        if "Models" in forms:
+        if forms and "Models" in forms:
             forms = ",".join(
                 ["Desktops", "Laptops", "Ubuntu Core", "Server", "Server SoC"]
             )
--- a/webapp/certification/views.py
+++ b/webapp/certification/views.py
@@ -177,15 +177,18 @@ def certification_home():
         offset = request.args.get("offset", default=0, type=int)
         filters = request.args.get("filters", default=False, type=bool)
 
+        selected_forms = request.args.getlist("form")
+        if "SoC" in selected_forms:
+            selected_forms.remove("SoC")
+            selected_forms.append("Server SoC")
+
         forms = (
-            ",".join(request.args.getlist("form"))
-            if request.args.getlist("form")
+            ",".join(selected_forms)
+            if selected_forms
             else None
         )
         if forms and "Models" in forms:
-            forms = ",".join(
-                ["Desktops", "Laptops", "Ubuntu Core", "Server", "Server SoC"]
-            )
+            forms = None
         releases = (
             ",".join(request.args.getlist("release"))
             if request.args.getlist("release")
@@ -211,7 +214,7 @@ def certification_home():
             "Desktop": 0,
             "Ubuntu Core": 0,
             "Server": 0,
-            "Server SoC": 0,
+            "SoC": 0,
         }
         release_filters = defaultdict(lambda: 0)
         for release in all_releases:
@@ -225,7 +228,10 @@ def certification_home():
 
         # Populate filter numbers
         for model in results:
-            form_filters[model["category"]] += 1
+            if model["category"] == "Server SoC":
+                form_filters["SoC"] += 1
+            else:
+                form_filters[model["category"]] += 1
             release_filters[model["release"]] += 1
             vendor_filters[model["make"]] += 1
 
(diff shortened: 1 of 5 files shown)


webapp/certification/views.py
```diff
@@ -173,13 +175,14 @@ def certification_home():
...
+        filters = request.args.get("filters", default=False, type=bool)
 
         forms = (
             ",".join(request.args.getlist("form"))
             if request.args.getlist("form")
             else None
         )
-        if "Models" in forms:
+        if forms and "Models" in forms:
             forms = ",".join(
```
codersquid, line 186: The API does a search for all forms when nothing is passed, so in this case you can just leave it as None.

You can see an example by leaving out `category` as a query parameter:
https://certification.staging.canonical.com/api/v1/certifiedmodels/?format=json

codersquid (changes requested):
I have an inline comment below. I also found a typo in L196 where it is getting a list from `vendor` instead of `vendors` which causes a search with vendors to get ignored.

There's an additional problem there, and I'll open an issue for it.

webapp/certification/views.py
bethcollins92: This has been resolved - we need to pass this so the check can be done in the front end to default tick each category on the search results page.

anthonydillon:
@bethcollins92 did it not working making these separate pages?

bethcollins92:
> @bethcollins92 did it not working making these separate pages?

@anthonydillon no I spoke to @carkod and he said it would mean repeating a lot of code in the backend, he spoke to Robin and it was decided they would make some changes so `navigation.yaml` accepts queries

anasereijo:
- [ ] when I select the sec nav items they reload the page but the links should become orange
Screenshot 2021-04-19 at 09 54 35

- [ ] when I am on [Ubuntu Certified SoC hardware](https://ubuntu-com-9551.demos.haus/certification?form=Server%20SoC),  if I select HP (for example) and then click 'update' search takes me back to 'Overview' page with a list of results that are not HP. I am wondering if the search results should still be in SoC hardware page and have a 'No results' message?

carkod:
"Lenovo" seems to be always selected by default in the filters no matter what I search

bethcollins92:
> * [ ]  when I am on [Ubuntu Certified SoC hardware](https://ubuntu-com-9551.demos.haus/certification?form=Server%20SoC),  if I select HP (for example) and then click 'update' search takes me back to 'Overview' page with a list of results that are not HP. I am wondering if the search results should still be in SoC hardware page and have a 'No results' message?

Good spot @anasereijo, there was a typo for the vendor, now the correct vendor will show.
With regards to keeping the same "Desktop header" when you select a vendor, I've filed another [issue](https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/181) to look at this.

bethcollins92:
Thanks @carkod, I've pushed the changes to fix this

bethcollins92:
> When I visit https://ubuntu-com-9551.demos.haus/certification?form=Server%20SoC It shows both Server and Server SoC as ticked.

Thanks @codersquid  - Just fixed now

carkod:
@bethcollins92 I would suggest maybe to do less commits to avoid having a very difficult rebase, so we can remember and group things correctly, maybe needs to be discussed with @anthonydillon , but there are commits like fixing lint that could be squeezed into the previous commit with `git commit --amend` and grouping some commits like "Fixing SoC", which is kind of duplicated. But you don't have to do this in this one, just giving some suggestions.

bethcollins92:
> @bethcollins92 I would suggest maybe to do less commits to avoid having a very difficult rebase against master, so we can remember and group things correctly, maybe needs to be discussed with @anthonydillon , but there are commits like fixing lint that could be squeezed into the previous commit with `git commit --amend` and grouping some commits like "Fixing SoC", which is kind of duplicated. But you don't have to do this in this one, just giving some suggestions.

Sounds like a good suggestion thanks @carkod. Would be able to pair with me while I do it as I'm not 100% sure and want to make sure I do it correctly?

carkod (approved):
LGTM

PR #9558, opened 2021-04-16
Feature flag esm apps

ClementChaumel:
## Done

- Added a feature flag around ESM apps to hide the option and allow us to release UA monthly

drive by: 
- Aligned the tech support link with the rest of the form

## QA

- go to https://ubuntu-com-9558.demos.haus/advantage/subscribe?test_backend=true
- check that everything is working correctly and the "Feature coverage" section is missing.
- go to https://ubuntu-com-9558.demos.haus/advantage/subscribe?test_backend=true&esm_apps=true
- check that the feature coverage is showing and working as expected.

### DIFF
--- a/static/js/src/advantage/subscribe/reducers/form-reducer.js
+++ b/static/js/src/advantage/subscribe/reducers/form-reducer.js
@@ -117,7 +117,7 @@ const formSlice = createSlice({
     },
     changeFeature(state, action) {
       const urlParams = new URLSearchParams(window.location.search);
-      const isAppsEnabled = urlParams.get("ESMApps") === "true";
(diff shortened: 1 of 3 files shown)


albertkol (approved):
I would call the flag "enable_esm_apps" or "esm_apps" instead of the capitalised "ESMApps", to follow the example set by "test_backend". But I leave it up to you.

ClementChaumel:
> I would call the flag "enable_esm_apps" or "esm_apps" instead of the capitalised "ESMApps", to follow the example set by "test_backend". But I leave it up to you.

Good catch! I'll update it

PR #9569, opened 2021-04-19
Fix number of paid subscriptions

### LINKED ISSUE #9516: Beta shop heading doesn't match subscription list
When I sign in to https://ubuntu-com-9240.demos.haus/advantage (beta UA shop), it says I have 5 paid subscriptions but only lists 3.

Screenshot from 2021-04-07 12-03-03

albertkol:
## Done

- Fix number of paid subscriptions

## QA

- It's a bit tricky. Check if the number of rows in the table matches the number in the header.

## Issue / Card

Fixes #9516

### DIFF
--- a/webapp/advantage/views.py
+++ b/webapp/advantage/views.py
@@ -212,7 +212,8 @@ def advantage_view(**kwargs):
                 else:
                     enterprise_contract.append(contract)
 
-                total_enterprise_contracts += 1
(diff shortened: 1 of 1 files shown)


ClementChaumel (approved):
Every day I discover a new attribute contracts can have. Today it's "expired". LGTM :+1:

PR #9585, opened 2021-04-21
Default to 0 if the machine number field is empty

ClementChaumel:
## Done

- Default to 0 if the machine number field is empty

## QA

- go to https://ubuntu-com-9585.demos.haus/advantage?test_backend=true
- click a subscription name 
- click "Change subscription"
- Delete the number of machine in the field
- check that it instead of going blank defaults to the min value for the field.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9465

### DIFF
--- a/static/js/src/advantage/manage-subscription.js
+++ b/static/js/src/advantage/manage-subscription.js
@@ -195,7 +195,7 @@ function handleCancelChangesClick(id, VPSize) {
 
 function handleChange(e, id, VPSize) {
   const defaultValue = Number.parseInt(e.target.defaultValue);
-  let newValue = Number.parseInt(e.target.value);
+  let newValue = Number.parseInt(e.target.value) || 0;
 
   const unitPrice = Number.parseFloat(e.target.dataset.unitPrice / 100);
   const nextPayment = Number.parseFloat(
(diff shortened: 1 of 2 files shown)


albertkol:
@ClementChaumel @wgx  It works.. but.. do you really want to make it work this way?

If have 1 machine of product X and I want to change it to 5.
I would backspace the 1 and insert a 5. But you don't let me do that. I have to use the arrows or write 15 then remove the 1.

Can't you make it go back to the initial value when I un-focus the input field, only if it's empty?

What do you think?

wgx:
I'd not seen this before. Agree with you @albertkol  - it should, if an invalid entry is present, default back to the "initial value" - to prevent users accidentally changing their entitlements through errors.

ClementChaumel:
> @ClementChaumel @wgx It works.. but.. do you really want to make it work this way?
> 
> If have 1 machine of product X and I want to change it to 5.
> I would backspace the 1 and insert a 5. But you don't let me do that. I have to use the arrows or write 15 then remove the 1.
> 
> Can't you make it go back to the initial value when I un-focus the input field, only if it's empty?
> 
> What do you think?

You're right! I'm used to only clicking on the arrow because it's quicker but of course, that's not how a real user would probably use it.

albertkol (approved):
Please remove the `console.logs`. Then it's good to go.

ClementChaumel:
:see_no_evil:

PR #9588, opened 2021-04-21
Fix broken state after failed purchase

### LINKED ISSUE #9463: UA shop: Changing payment method and then changing subscription leaves user in a confused/broken state
**Tested with**: https://ubuntu-com-9240.demos.haus/advantage/subscribe?test_backend=true
**Steps to reproduce**:
1. Purchase 2x uai-essential-desktop on an empty account
2. Change the subscription: increase it to 4
3. Change credit card to 4000 0000 0000 0341
4. Try to change quantity to 6 using the existing contract accordion

**Result**: no error or alert is shown, subscription disappears, see screenshot. Customer cannot make new purchases.
purchase_stuck

albertkol:
## Done

- select every single subscription of the account, but filter out the deactivated ones. (locked subscriptions were excluded by mistake)
- both yearly and monthly subscriptions can have a locked status, but before we were only dealing with monthly ones
- if a user encounters payment issues, through resizing or purchasing, the transaction will finish they get redirected to /advantage and the notification saying that a payment issue has happen will show up.

## QA

1.
- Make a new purchase yearly and/or monthly product with this card number: 4000000000000341
- You will reach /advantage and you will see the notification
2.
- Change your default credit card to 4000000000000341
- Then resize a product
- You will reach /advantage and you will see the notification

## Issue / Card

Fixes #9463

### DIFF
--- a/templates/advantage/index.html
+++ b/templates/advantage/index.html
@@ -55,16 +55,16 @@
           <h2>Your {% if total_enterprise_contracts > 1 %}{{total_enterprise_contracts}} {% endif %}paid subscription{% if total_enterprise_contracts == 1 %}{% else %}s{% endif %}</h2>
         </div>
       </div>
+      {% if payment_method_warning %}
+        <div class="row">
+          <div class="p-notification--caution">
+            <p class="p-notification__response" role="status">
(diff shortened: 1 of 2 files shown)


ClementChaumel:
It does what you say it should but is that enough to get off this "broken state"?

albertkol:
@ClementChaumel this one only makes the broken state more verbose.

If a payment fails they will have the notification sending them to `/payment-methods`.

Unlocking the subscription will be done as part of https://github.com/canonical-web-and-design/web-squad/issues/3950

PR #4034, opened 2018-09-10
Rotate the DE takeover with Financial Services takeover

anthonydillon:
## Done
Swap the AI takeover with the Financial Services takeover on the German homepage.

## QA
- Check out this feature branch
- Run the site using the command `./run serve`
- View the site locally in your web browser at: [http://0.0.0.0:8001/index_de](http://0.0.0.0:8001/index_de)
- Run through the following [QA steps](https://github.com/canonical-webteam/practices/blob/master/workflow/qa-steps.md)
- Refresh and see you get the Financial Services takeover and the German one.

Fixes #4033

PR #9593, opened 2021-04-22
Add country parameter

albertkol:
## Done

- Fix bug where country was not an accepted parameter.

## QA

- Have the console open
- Make a purchase as a guest
- If perview returns 200, it works

## Issue / Card

Fixes #4034


ClementChaumel (approved):
nice!

PR #9600, opened 2021-04-22
Fix vendor param issue

carkod:
## Done

Fix vendor params for certified models endpoint

## QA

- Go to https://ubuntu-com-9600.demos.haus/certification?text=&form=SoC&vendor=Applied+Micro+Circuits+Corp.&vendor=Qualcomm+Inc&filters=True Some results show up

## Issue / Card

Fixes https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/179

### DIFF
--- a/webapp/certification/api.py
+++ b/webapp/certification/api.py
@@ -78,6 +78,8 @@ class CertificationAPI:
                 "offset": offset,
                 "level": level,
                 "major_release__in": major_release__in,
+                # If vendor is a list, requests will tranform into
+                # &vendor=item&vendor=item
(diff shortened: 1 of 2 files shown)


SirSamTumless:
@codersquid could you give this a review when you get a chance

codersquid:
I wasn't familiar with how the requests library handles lists when passed as a param and had to check the requests src and then had to check urlencode's src. (that's a really nice feature)

Do you think it calls for a comment about how a list will be handled as a multivalued query parameter?

PR #9606, opened 2021-04-23
Remove default quantity for small screens

ClementChaumel:
## Done

- Set the default quantity to 0 on small screens so the product summary doesn't show until the user has chosen a quantity

## QA

- go here https://ubuntu-com-9606.demos.haus/advantage/subscribe?test_backend=true
- resize your browser until you get the small VP view (875px)
- refresh and fill the form
- The summary and buy button should remain hidden until you increase the quantity
- Verify that it isn't the case on large VP sizes

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9573

### DIFF
--- a/static/js/src/advantage/subscribe/reducers/form-reducer.js
+++ b/static/js/src/advantage/subscribe/reducers/form-reducer.js
@@ -1,11 +1,15 @@
 import { createSlice } from "@reduxjs/toolkit";
 
+const isSmallVP =
+  Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0) <
+  875;
+
 const initialFormState = {
   type: "physical",
   version: "18.04",
   feature: "infra",
   support: "unset",
-  quantity: 1,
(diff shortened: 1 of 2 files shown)


elioqoshi:
Works as expected, thanks @ClementChaumel !

albertkol:
@ClementChaumel do you need to rebase this one? ..or why do I get the same behaviour being unable to delete the number in the input field. So if I want to change the number from 1 to 5, I have to insert 15 then delete the 1.

ClementChaumel:
> @ClementChaumel do you need to rebase this one? ..or why do I get the same behaviour being unable to delete the number in the input field. So if I want to change the number from 1 to 5, I have to insert 15 then delete the 1.

Ah that's a different one, I'll fix it too

ClementChaumel:
Done! PTAL @albertkol

albertkol (approved):

PR #9609, opened 2021-04-23
Persist redux state to local storage

ClementChaumel:
## Done

- Persist redux state to local storage so the form remembers the user's selection.

## QA

- go here https://ubuntu-com-9609.demos.haus/advantage/subscribe?test_backend=true
- make some changes to the form
- refresh/navigate to another page and go back
- check that the changes are still here
- make the purchase
- go back to /subscribe
- check that the form "forgot" about your last purchase and is reset to the default state

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9605

## Question

@wgx @elioqoshi The way it's done now, the form will remember "forever" the selection. Do we want to clear it at some point? (on a successful purchase or another action?)

### DIFF
--- a/static/js/src/utils/persitState.js
+++ b/static/js/src/utils/persitState.js
@@ -1,0 +1,21 @@
+export function saveState(state, stateName) {
+  console.log("write");
+  try {
+    const serializedState = JSON.stringify(state);
+    localStorage.setItem(stateName, serializedState);
+  } catch (err) {
+    // ignore write errors
+  }
+}
+
+export function loadState(stateName, slice, defaultState) {
+  try {
+    const serializedState = localStorage.getItem(stateName);
+    if (serializedState === null) {
+      return defaultState;
+    }
+    return JSON.parse(serializedState)[slice];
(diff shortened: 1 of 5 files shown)


wgx:
Works great for me. A successful purchase should clear the users' selection from local storage.

static/js/src/utils/persitState.js
```diff
@@ -0,0 +1,21 @@
+export function saveState(state, stateName) {
+  console.log("write");
```
anthonydillon, line 2: Left a console log

anthonydillon (approved):
Code and QA :+1: with a small comment in the code

static/js/src/utils/persitState.js
ClementChaumel: Whoops, nice catch! Thanks

ClementChaumel:
@wgx I added the clearing of the state once a purchase is completed. Can you QA again please? Thanks!

PR #9636, opened 2021-04-27
Accept binary packages for notices

albertkol:
## Done

- Accept binary packages for notices. These should not be displayed on the website, but they should be accessible through the API.

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/ubuntu-com-security-api/issues/46


carkod (approved):
LGTM

PR #9651, opened 2021-04-28
Hide monthly options if no products are available

ClementChaumel:
## Done

- Added a check to add the period dropdown if no monthly products are present

## QA

- check here, https://ubuntu-com-9651.demos.haus/advantage/subscribe?test_backend=true select server, essential and see that you can choose monthly or yearly at the bottom
- Check here, https://ubuntu-com-9651.demos.haus/advantage/subscribe select server essential and see that there is no dropdown for the billing period

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9647

### DIFF
--- a/static/js/src/advantage/subscribe/reducers/form-reducer.js
+++ b/static/js/src/advantage/subscribe/reducers/form-reducer.js
@@ -22,7 +22,7 @@ const initialFormState = {
       advanced: 127500,
     },
   },
-  hasMultiplePeriods: false,
+  hasMonthly: false,
 };
 
 const prefixMap = {
@@ -31,14 +31,14 @@ const prefixMap = {
   "infra+apps": "uaia",
 };
 
-function hasMultiplePeriods(productID) {
-  let numberOfPeriods = 0;
+function hasMonthly(productID) {
+  let containsMonthly = false;
   productsArray.forEach((product) => {
-    if (product[1].productID === productID) {
-      numberOfPeriods++;
+    if (product[1].productID === productID && product[1].period == "monthly") {
+      containsMonthly = true;
     }
   });
-  return numberOfPeriods > 1;
+  return containsMonthly;
 }
 
 function matchingProduct(testItem, selectedProductID, selectedBillingPeriod) {
@@ -130,7 +130,7 @@ const formSlice = createSlice({
         state.support = "essential";
       }
       state.product = getProduct(state);
-      state.hasMultiplePeriods = hasMultiplePeriods(state.product.productID);
+      state.hasMonthly = hasMonthly(state.product.productID);
     },
     changeVersion(state, action) {
       state.version = action.payload;
@@ -152,7 +152,7 @@ const formSlice = createSlice({
         state.billing = "yearly";
       }
       state.product = getProduct(state);
-      state.hasMultiplePeriods = hasMultiplePeriods(state.product.productID);
+      state.hasMonthly = hasMonthly(state.product.productID);
     },
     changeSupport(state, action) {
       state.support = action.payload;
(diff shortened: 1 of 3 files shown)


static/js/src/advantage/subscribe/reducers/form-reducer.js
```diff
@@ -5,6 +5,15 @@ const isSmallVP =
   Math.max(document.documentElement.clientWidth || 0, window.innerWidth || 0) <
   875;
 
+const productsArray = Object.entries(window.productList);
+function hasMonthly() {
```
anthonydillon, line 9: We should check if the selected product has a monthly option not just if any do.

ClementChaumel: You're right it makes more sense this way. I updated it

static/js/src/advantage/subscribe/reducers/form-reducer.js
```diff
@@ -28,7 +31,15 @@ const prefixMap = {
...
-const productsArray = Object.entries(window.productList);
+function hasMonthly(productID) {
+  let numberOfPeriods = 0;
+  productsArray.forEach((product) => {
+    if (product[1].productID === productID) {
+      numberOfPeriods++;
+    }
+  });
+  return numberOfPeriods > 1;
+}
```
anthonydillon, line 42: Can we be more precise with the logic? If they add daily this would return true. Here is an example of what I mean;

```js 
function hasMonthly(productID) {
  let containsMonthly = false;
  productsArray.forEach((product) => {
    if (product[1].productID === productID && product[1].period == "monthly") {
      containsMonthly = true;
    }
  });
  return containsMonthly;
}
```

anthonydillon (changes requested):

static/js/src/advantage/subscribe/reducers/form-reducer.js
ClementChaumel: The naming of the function is maybe not right but if they add daily we would still want to display the billing period dropdown, right?

anthonydillon: I think the name is perfectly correct and clear for the functional logic it returns. If they add daily this will return true even if the options are ["yearly", "daily"].

anthonydillon: If we are in a rush, this is ok. But, it would be better to have logic that knows what it's displaying. 

I would like this function to return an array of periods for the selected product so we can check the length to hide the select and use the array to add items to the select box in a clear known way. If this is blocking the launch it would work but please revisit as this could come back to bite us in the future.

ClementChaumel: You're right it's better this way I updated it

static/js/src/advantage/subscribe/renderers/form-renderer.js
```diff
@@ -184,11 +190,15 @@ function renderSummary(state) {
...
   }
 
-  // Monthlty is only available for infra and essential
-  if (state.feature !== "infra" || state.support !== "essential") {
-    billingSection.classList.add("u-hide");
-  } else {
+  if (periods.length > 1) {
     billingSection.classList.remove("u-hide");
+    let options = "";
+    periods.forEach((period) => {
```
anthonydillon, line 196: Sorry, if I am not being clear but truly dynamic will also cause issues as we need middleware logic. What I meant by "add items to the select box in a clear known way" was to conditionally add known options if the period exists in the array.

anthonydillon (approved):
Awesome thanks :+1:

PR #9659, opened 2021-04-29
Fix period being hardcoded to monthly for resizing

ClementChaumel:
## Done

- Pass the correct period instead of always passing monthly to the resizing endpoint

## QA

- go here https://ubuntu-com-9659.demos.haus/advantage?test_backend=true
- log in
- resize a contract
- check in the network tab that the right period is specified


albertkol (approved):
:+1:

PR #9660, opened 2021-04-29
Fix payment warning notification

albertkol:
## Done

- Fix payment warning notification. The problem is that the `payment_method_warning` would take the value of the last subscription. Now the if conditions if you want to update it or not.


ClementChaumel (approved):
:+1:

PR #9662, opened 2021-04-29
Redirect certification old site to new microsite

carkod:
## Done

Redirect certification.ubuntu.com to www.ubuntu.com/certification

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/ubuntu.com/issues/9429


jkfran:
Right now this could create a redirection loop, because:
ubuntu.com/certification -> certification.ubuntu.com
certification.ubuntu.com -> ubuntu.com/certification

Please check why `ubuntu.com/certification` is redirecting to `certification.ubuntu.com` and whatever it is you will need to disable it.

Also, you will need to add `certification.ubuntu.com` in the `extraHost` section in this file, and please add the same rule for `certification.staging.ubuntu.com`

konf/site.yaml
```diff
@@ -204,6 +204,9 @@ production:
     if ($host = 'cloud.ubuntu.com' ) {
       rewrite ^ https://ubuntu.com/public-cloud$request_uri? permanent;
     }
+    if ($host = 'certification.ubuntu.com' ) {
+      rewrite ^ https://ubuntu.com/certification$request_uri? permanent;
```
nottrobin, line 208: We'll also need to redirect all the old URL structures to their new pages.

konf/site.yaml
```diff
@@ -366,6 +369,9 @@ staging:
     if ($host = 'insights.staging.ubuntu.com' ) {
       rewrite ^ https://staging.ubuntu.com/blog$request_uri? permanent;
     }
+    if ($host = 'certification.staging.ubuntu.com/' ) {
+      rewrite ^ https://ubuntu.com/certified$request_uri? permanent;
```
nottrobin, line 373: This should surely redirect to staging?

```suggestion
      rewrite ^ https://staging.ubuntu.com/certified$request_uri? permanent;
```

nottrobin:
@carkod could you add a redirect from `/certification` to `/certified` in `redirects.yaml`?

redirects.yaml
```diff
@@ -123,17 +123,22 @@ category/product/landscape/?: https://landscape.canonical.com/
...
 certified/laptops/?: "/certified?category=Laptop"
 certified/servers/?: "/certified?category=Server"
 certified/devices/?: "/certified?category=Device"
 certified/socs/?: "/certified?category=SoC"
-cc/?: "/security/certifications"
-cc-eal/?: "/security/certifications"
-cis-audit/?: "/security/certifications"
-cloud/awsome/?: "https://launchpad.net/awsome"
-cloud/azure/?: "/public-cloud"
-cloud/bootstack/?: "/openstack/managed"
```
nottrobin, line 136: Why have you removed these?

konf/site.yaml
carkod: The queries have been done in the view and the URL paths have been changed in the `redirects.yaml`.

redirects.yaml
carkod: Thanks sorry, mistake.

redirects.yaml
```diff
@@ -123,11 +123,22 @@ category/product/landscape/?: https://landscape.canonical.com/
...
 certified/laptops/?: "/certified?category=Laptop"
 certified/servers/?: "/certified?category=Server"
 certified/devices/?: "/certified?category=Device"
 certified/socs/?: "/certified?category=SoC"
+certified/desktop/?: "/certified?category=Desktop"
+certified/server/?: "/certified?category=Server"
+certified/iot/?: "/certified?category=Device"
+certified/soc/?: "/certified?category=SoC"
+certified/hardware/(?P<canonical_id>.*): "/certified/{canonical_id}"
+certified/components/(?P<component_id>.*): "/certified/component/{component_id}"
```
nottrobin, line 137: We use "component" singular in the new URL scheme? This should be plural.

redirects.yaml
```diff
@@ -123,11 +123,22 @@ category/product/landscape/?: https://landscape.canonical.com/
...
+certified/catalog/search/?: "/certified"
+certified/desktop/models/?: "/certified"
```
nottrobin, line 139: Shouldn't this go to `"/certified?category=Desktop"`?

redirects.yaml
carkod: No, these are redirects targeted to the old site. So whenever someone hits /certified/compoents/<id> it should redirect to /certified/component/<id> (in singular) on our site

redirects.yaml
carkod: Ok, I think the reason why I did this is because on certificaiton.ubuntu.com/desktop/models all links have `?category=Desktop`

webapp/certified/views.py
```diff
@@ -0,0 +1,355 @@
...
+            else:
+                server_releases[release] = vendor[release]
+
+    if request.args:
+        query = request.args.get("q", default=None, type=str)
+
+        # Old site replacements
+        if request.args.get("query"):
+            query = request.args.get("query", default=None, type=str)
+
```
nottrobin, line 273: Instead of handling these query parameters directly, could we send a `301` redirect to tell the client the new URL scheme?

```suggestion
        # Old site replacements
        if request.args.contains(("query", "vendors")):
           args = request.args
           args.replace("query", "q")
           flask.redirect(f"/certified?{args.join(';')}")
```

redirects.yaml
```diff
@@ -123,6 +123,22 @@ category/product/landscape/?: https://landscape.canonical.com/
 category/product/(server|ubuntu-server-edition)/?: /server
 category/product/ubuntu-advantage/?: /support
 category/product/ubuntu-light/?: http://lubuntu.net/
+# old certification.ubuntu.com redirects
+certified/desktops/?: "/certified?category=Desktop"
+certified/laptops/?: "/certified?category=Laptop"
```
nottrobin, line 128: This doesn't exist.

redirects.yaml
```diff
@@ -123,6 +123,22 @@ category/product/landscape/?: https://landscape.canonical.com/
...
+certified/servers/?: "/certified?category=Server"
+certified/devices/?: "/certified?category=Device"
+certified/socs/?: "/certified?category=SoC"
+certified/desktop/?: "/certified?category=Desktop"
```
nottrobin, line 132: This is actually equivalent to `/certified?category=Desktop&category=Laptop`

redirects.yaml
carkod: Sorry, just realized now, this actually does exist in the new site navigation, so I'll keep it there for now.

webapp/certified/views.py
carkod: So it was a bit tricky, as the `request.args` is an `ImmutableMultiDict`, so can't change it this way. I instead created a new dict, replaced the keys, then encoded the URL. It seems to work well. And the `flask.redirect` needs a `return`

PR #9667, opened 2021-04-29
Add docs navigation for OpenStack

anthonydillon:
## Done
Add docs navigation for OpenStack

## QA
- Open the demo and go to /openstack
- Click the docs navigation item and check docs load
- Click around docs and see the navigation is persistent


pmahnke (approved):
+1

PR #9669, opened 2021-04-30
Update instruction copy

### LINKED ISSUE #9517: Beta shop says I don't need to attach machines to set up UA services

ClementChaumel:
## Done

- Update instruction copy for 16.04 and 18.04 according to https://github.com/canonical-web-and-design/ubuntu.com/issues/9517#issuecomment-820183733

## QA

-  https://ubuntu-com-9669.demos.haus/advantage?test_backend=true

## Issue / Card

Fixes #9517

## Screenshots

image


ClementChaumel:
@lechsandecki here are the changes to the instructions. Please review it and provide instruction for the 16.04 and 18.04 changes thanks!

templates/advantage/table/_attached-machines.html
```diff
@@ -3,32 +3,49 @@
...
+              <span>Add the PPA</span>
+              <code>sudo add-apt-repository ppa:ua-client/stable<br/>
+              sudo apt-get update</code>
+              <hr />
+              <span>Install the UA client</span>
+            <code>sudo apt install ubuntu-advantage-tools</code>
+            </div>
+          </div>
+        <div class="advantage-table-section">
+          <p class="u-no-margin--bottom">Once installed or if you’re using Ubuntu 14.04 ESM or 20.04 LTS, first attach each machine.</p>
```
lechsandecki, line 19: I'd drop that. I suggest that, on this page, we keep the following:

to attach a machine: sudo ua abc
to detach a machine: sudo ua def

Please ensure that you have the latest version of the ua client installed: apt-get upgrade

templates/advantage/table/_attached-machines.html
```diff
@@ -3,32 +3,49 @@
...
```
lechsandecki, line 48: we also need to update the ESM screen (it's now 14.04 or 16.04, not just 14.04)
Similarly, Livepach can get activated via the ua client for each Ubuntu LTS now. We don't need separate "Ubuntu 16.04 LTS and 18.04 LTS" instructions anymore

lechsandecki (changes requested):

albertkol:
@ClementChaumel you broke the table

image

ClementChaumel:
I've updated the copy according to your review @lechsandecki, please take another look:
image
image
image

and I fixed the table @albertkol :angel:

lechsandecki:
+1, looks great. Can we make it live asap, please?

SirSamTumless:
Hi @lsandecki I don't think it to say "we are slow implementation" in the public channnel. We ask for your review yesterday and you've only just come abck.

lechsandecki (approved):

PR #9743, opened 2021-05-11
Pass customer name when setting customer info as guest

alnvdl:
This is needed for a new requirement we will be introducing in ua-contracts: that calls to set customer info must always contain the customer name and country. While that doesn't roll out to production in ua-contracts, it should still be safe to call this method with this added field.


ClementChaumel (approved):
LGTM

alnvdl:
Tested this by making guest and regular purchases on top of the old code, as well as in the new code (requires name + address), both are working.

PR #9769, opened 2021-05-18
Rename is_binary to is_visible

albertkol:
## Done

- Rename `is_binary` to `is_visible`.

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -47,7 +47,7 @@ def notice(notice_id):
             release_version = releases[codename]
             release_packages[release_version] = {}
             for package in pkgs:
-                if package.get("is_binary", False):
+                if not package.get("is_visible", True):
                     continue
 
                 name = package["name"]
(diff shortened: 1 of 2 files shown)


webapp/security/views.py
```diff
@@ -47,7 +47,7 @@ def notice(notice_id):
             release_version = releases[codename]
             release_packages[release_version] = {}
             for package in pkgs:
-                if package.get("is_binary", False):
+                if not package.get("is_visible", True):
```
jpmartinspt, line 50: Just to confirm that the intended change is if there is no `is_visible` we default to show the package?

jpmartinspt (approved):

webapp/security/views.py
albertkol: Yes. By default we show everything, and only if the `is_visible` this present and set to `false`. Then we hide it.

anthonydillon:
Is there an issue for this? If not can you add an Epic, Sprint and Milestone please.

PR #9774, opened 2021-05-18
Support query strings in navigation

carkod:
## Done

Currently `navigation.yaml` supports only paths, i.e. `/parent/child/grandchild`.
This will support query strings i.e. `/parent?page=child` so that we can match the entire URL rather than just the path
To make this work, the `navigation.yaml` needs to contain a parameter `match_query: True`

## QA

- Go to https://ubuntu-com-9774.demos.haus/certified?form=Desktop and click on the navigation links and make sure they are highlighted when you change them.
- Other links to check to make sure this feature doesn't break existent navigation
  - https://ubuntu-com-9774.demos.haus/server/docs
  - https://ubuntu-com-9774.demos.haus/kubernetes/compare (click on the different navigation links)
  - https://ubuntu-com-9774.demos.haus/security/cve 
  - https://ubuntu-com-9774.demos.haus/tutorials

## Issue / Card

Fixes https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/178

### DIFF
--- a/navigation.yaml
+++ b/navigation.yaml
@@ -1616,17 +1616,19 @@ ubuntu1604:
 certification:
   title: Hardware
   path: /certified
+  persist: True
+  match_query: True
 
   children:
     - title: Overview
-      path: /certified
+      path: /certified?
     - title: Laptops
-      path: /certified/laptops
+      path: /certified?category=Laptop
     - title: Desktops
-      path: /certified/desktops
+      path: /certified?category=Desktop
     - title: Servers
-      path: /certified/servers
+      path: /certified?category=Server
     - title: Devices
-      path: /certified/devices
+      path: /certified?category=Device
     - title: SoCs
-      path: /certified/socs
+      path: /certified?category=SoC
--- a/templates/templates/_navigation.html
+++ b/templates/templates/_navigation.html
@@ -1,4 +1,4 @@
-{% set breadcrumbs = get_navigation(request.path).breadcrumbs %}
+{% set breadcrumbs = get_navigation(request).breadcrumbs %}
 
 <header id="navigation" class="p-navigation">
   {% block header_banner %}
--- a/templates/templates/footer.html
+++ b/templates/templates/footer.html
@@ -1,4 +1,4 @@
-{% set nav_sections = get_navigation(request.path).nav_sections %}
+{% set nav_sections = get_navigation(request).nav_sections %}
 
 {% block footer %}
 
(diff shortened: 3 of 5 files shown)


navigation.yaml
```diff
@@ -1616,10 +1616,12 @@ ubuntu1604:
...
   path: /certified
+  persist: True
+  match_query: True
 
   children:
     - title: Overview
-      path: /certified
+      path: /certified?
     - title: Laptops
       path: /certified/laptops
```
nottrobin, line 1626: I suppose this needs changing to `/certified?category=Laptops`, and all the others? The Laptop, Desktop etc. currently aren't being highlighted.

navigation.yaml
```diff
@@ -1616,10 +1616,12 @@ ubuntu1604:
 certification:
   title: Hardware
   path: /certified
+  persist: True
+  match_query: True
 
   children:
     - title: Overview
-      path: /certified
+      path: /certified?
```
nottrobin, line 1624: Why does this have a question mark on the end?

nottrobin:
So I don't really have a good suggestion here, but I'm a bit worried about this. We know that we will want to change these pages for e.g. `/certified/category=Laptops` to `/certified/laptops` before too long. And at that point we'll have no need for this query string functionality. In fact we actively don't want it.

I would love it if we could come up with a way of solving this that was more ephemeral, that was easier to rip out when we need to and didn't risk other people relying on this functinoality in the meantime.

One option is that we simply don't do this work and put up with those navigation sections just not being highlighted for now, until we have dedicated pages.

@anthonydillon what do you think?

anthonydillon:
I agree and have the same concerns. I would prefer the navigation to be imperfect until we can switch to true destination pages. This will give us the impotence to get this change complete quickly. There is a real danger this behaviour will become relied on and we should actively work to discourage it.

carkod:
@nottrobin @anthonydillon So before doing this, I considered a few other alternatives e.g. matching the query string and changing the class to active, but eventually I settled with this solution because of two main reasons
- it is the simplest in terms of amount of code changed so it's easy to revert (only the `request.path` change affects a few more files), we can use this PR to revert or I can do it.
- I did a few tests and it is quite safe, mainly because it only works with the `match_query` in `navigation.yaml`, otherwise it will not pass the condition and will work as usual. So I believe this is cleaner than a possible JS solution.

**But** if you guys consider not having the "highlight" feature more important than messing with the navigation, I can close this. I just thought maybe it was more important to get this working.

navigation.yaml
carkod: Sorry, I thought it would work after rebasing with the changes made in `redirects.yaml` but it doesn't seem to work that way. Fixed now.

navigation.yaml
carkod: `request.full_path` doesn't match without the question mark (query string).

PR #9775, opened 2021-05-18
Certified overview laptops

carkod:
## Done

- Certified overview laptops. 

## QA

- Check https://ubuntu-com-9775.demos.haus for the new laptops card

## Issue / Card

Fixes https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/195

### DIFF
--- a/webapp/certification/views.py
+++ b/webapp/certification/views.py
@@ -196,6 +196,10 @@ def certified_home():
     certified_makes = api.certified_makes(limit="0")["objects"]
 
     # Desktop section
+    laptop_releases = []
+    laptop_vendors = []
+
+    # Desktop section
     desktop_releases = []
     desktop_vendors = []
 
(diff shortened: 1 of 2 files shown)


bethcollins92:
There are no numbers after the releases. Also the numbers seem to be exactly the same between laptops and desktops - double checking is it just a coincidence?
image

codersquid:
> Also the numbers seem to be exactly the same between laptops and desktops - double checking is it just a coincidence?

The numbers should not be the same. See https://certification.canonical.com/api/v1/certifiedmakes/?format=json

carkod:
> There are no numbers after the releases. Also the numbers seem to be exactly the same between laptops and desktops - double checking is it just a coincidence?
> image

Fixed. Also removed the path, since it's composed in the template, and removed the `filters=True` from there

bethcollins92:
QA LGTM - not sure if Sheila wants to check the python code?

webapp/certification/views.py
```diff
@@ -195,6 +195,10 @@ def certified_home():
     certified_releases = api.certified_releases(limit="0")["objects"]
     certified_makes = api.certified_makes(limit="0")["objects"]
 
+    # Desktop section
```
codersquid, line 198: Update the comment to say Laptops.

codersquid (commented):
I have a very minor comment about a comment. see inline comment.

codersquid (approved):
lgtm

Issue #192, opened 2016-04-08
/desktop/education hero unreadable [ubuntu/chrome, ipad/safari]


deadlight:
## Context

The new text sizing for tablet makes the hero on /desktop/education unreadable
## Process
- navigate to /desktop/education on medium viewport
## Expected result

Text is readable and in the light area of the background image
## Current result

The text flows over the dark part of the image and isn't readable.
## Note

We will need to craft the padding to make it work

PR #9784, opened 2021-05-19
[Certs] Refactor overview page

bethcollins92:
## Done

- First draft of refactoring overview page  - see list of tasks [here](https://docs.google.com/document/d/1FRi83535YcpAv3jJ2lb4dpcBNlviA2JomD0d9gD5yAQ/edit#)
- Add `persist: True` to nav to keep secondary nav on model details pages
- Replace download link section and fix accordion on model details page which was accidentally removed in a previous rebase

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/certified
- Check the page against the [doc](https://docs.google.com/document/d/1FRi83535YcpAv3jJ2lb4dpcBNlviA2JomD0d9gD5yAQ/edit#)
- This page is also being looked at again by UX so there is likely to be more changes in the future. 

## Issue / Card

Fixes [#192](https://github.com/canonical-web-and-design/certification.ubuntu.com/issues/192)

### DIFF
--- a/templates/certification/index.html
+++ b/templates/certification/index.html
@@ -1,17 +1,18 @@
 {% extends "certification/base_certification.html" %}
 
-{% block title %}Certified hardware | Ubuntu{% endblock %}
+{% block title %}Certified hardware{% endblock %}
 
 {% block meta_description %}Hardware that have been certified for use with Ubuntu.{% endblock meta_description %}
-{% block meta_copydoc %}https://docs.google.com/document/d/1cPdcdSLLawGMn1bSKjbbeJh_CMgpj7d8YPm7F40XP38/edit#{% endblock meta_copydoc %}
+{% block meta_copydoc %}https://docs.google.com/document/d/1pSkPo5fvgrLb-Tjiq_he9MAvPLFX79xiz-H9s2mSXgI/edit#{% endblock meta_copydoc %}
 
 {% block content %}
 
 <section class="p-strip--suru-bottomed">
   <div class="row u-vertically-center">
     <div class="col-7">
-      <h1>Ubuntu Certified hardware</h1>
-      <p>Looking for machines you can trust with Ubuntu? Ubuntu Certified hardware has passed our extensive testing and review process, ensuring that Ubuntu runs well out of the box, ready for your organisation. We work closely with OEMs to jointly make Ubuntu available on a wide range of desktops, laptops, servers, IoT, and SoC hardware.</p>
+      <h1>Ubuntu certified hardware</h1>
+      <p class="p-heading--4">Find machines you can trust with Ubuntu</p>
+      <p>Ubuntu certified hardware has passed our extensive testing and review process, ensuring that Ubuntu runs well out of the box, ready for your organisation. We work closely with OEMs to jointly make Ubuntu available on a wide range of desktops, laptops, servers, IoT, and SoC hardware.</p>
     </div>
     <div class="col-5 u-hide--small u-align-center">
       {{ image (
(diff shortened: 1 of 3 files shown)


templates/certification/index.html
```diff
@@ -130,9 +120,10 @@ <h2 class="p-card__title p-heading--3">Ubuntu Desktop certified hardware</h2>
         loading="lazy"
```
carkod, line 120: Maybe I don't understand this functionality well enough, but why are you using `lazy` in some and `auto` in others?

bethcollins92: I just thought as the top two could be visible on page load for some screens better not lazy load them

davegoddard42:
UX wise this is fine, and I've marked as +1. Although I think it's worth mentioning the links in the cards (see image below) have no sense of priority (they're all the same 'link' style), and don't stand out much against the body text. I don't feel as though I understand enough of the background (user motivation, use case etc) as to whether this is an issue or not. It's certainly readable and clear, hence the +1.

image

bethcollins92:
@davegoddard42 Thanks for your feedback. We will have chance to update this in the future so any changes can be made once you've had chance to look at the other pages etc.

carkod (approved):

nottrobin (approved):
This looks great. Thanks @bethcollins92.

thoughts, that are absolutely not blockers, but just things for your consideration. Feel free to file them as new issues if you agree with them so you can merge this now:

- Shouldn't the collapsed sections in the left hand filter bar for search results at least be expanded if any item in that section is ticked?
- Are we sure we want those sections collapsed? I can see the argument for collapsing the very long vendors list (although it should be expanded if you're actually filtering on a vendor). I would have thought it's pretty useful to have at least the "Category" always visible, and maybe the "Release" too.
- On the homepage, I would have thought more people are looking for laptops than desktops (from a quick search, it seems about 3 times as many laptops are sold). Laptops are first in the navigation for that reason. Is the reason we put desktops first on the homepage just 'cos that's what we have the text for? In which case, maybe we should file an issue about expanding the laptop text so we can put that section first?

davegoddard42:
@nottrobin 

cc @bethcollins92 

I've noted these points down. Will be aware of them when I work on the other pages

davegoddard42:
@nottrobin 

cc @bethcollins92 

I've noted these points down. Will be aware of them when I work on the other pages

davegoddard42:
@nottrobin 

cc @bethcollins92 

I've noted these points down. Will be aware of them when I work on the other pages

bethcollins92:
Thanks @nottrobin:

- The accordion definitely needs work I agree. It always has category expanded but not the others - I'll file an issue to look at fixing this and maybe @davegoddard42 will have an idea on improving this too. 
- The laptop coming second is exactly for the reason you said about the text. I'll file an issue to improve the text in that section and if UX agrees then we can put that one first.

carkod:
> Thanks @nottrobin:
> 
> * The accordion definitely needs work I agree. It always has category expanded but not the others - I'll file an issue to look at fixing this and maybe @davegoddard42 will have an idea on improving this too.
> * The laptop coming second is exactly for the reason you said about the text. I'll file an issue to improve the text in that section and if UX agrees then we can put that one first.

Also as part of potential UX review, I mentioned before whether having that download link in the models details (e.g. https://ubuntu-com-9784.demos.haus/certified/201808-26455) makes sense or not? Because this is a hardware page, a link to download Ubuntu is not related to the hardware? Unless there is a strong marketing reason behind it (do people really use it?)

carkod (approved):
LGTM

PR #9798, opened 2021-05-24
[CUBE] Fix progression hero and broken mobile view

### LINKED ISSUE #9789: Microcerts view on small screens does not work
When viewing the microcerts on a small screen the modified mobile table pattern does not work.

bethcollins92:
## Done

- Extract `.is-faded` to it's own class so it can be conditionally added to progression hero depending on module status
- Add media query to width of tables to fix mobile view
- Fix text overflow on cards by adding `white-space: normal` to the list of topics
- Filed an [issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/3754) on Vanilla

## QA

- Force all statuses of modules to passed by adding `course["status"] = "passed"` to line 127 of `webapp/cube/views.py`
- See the progression hero change to show passed microcerts
- Check the microcerts table on mobile view

## Issue / Card

Fixes #9789 

## Screenshots

none passed:

image

force passed all:

image

mobile view: 

image

### DIFF
--- a/static/sass/_pattern_cube-progression.scss
+++ b/static/sass/_pattern_cube-progression.scss
@@ -171,8 +171,7 @@
   }
 
   .is-faded {
-    // Without !important the background only applies to certain microcerts
-    background-color: rgba(255, 255, 255, 0.1) !important;
+    background-image: rgba(255, 255, 255, 0.1);
     clip-path: polygon(50% 0%, 100% 50%, 50% 100%, 0% 50%);
     opacity: 0.5;
 
--- a/static/sass/_pattern_cube-progression.scss
+++ b/static/sass/_pattern_cube-progression.scss
@@ -28,34 +28,14 @@
       width: map-get($rhombusSize, "width") * 1px;
     }
 
-    %is-faded {
-      background-color: rgba(255, 255, 255, 0.1);
-      clip-path: polygon(50% 0%, 100% 50%, 50% 100%, 0% 50%);
-      opacity: 0.5;
-
-      &--right {
-        @extend %is-faded;
-
-        clip-path: polygon(0% 34%, 100% 0%, 100% 66.6%, 0% 100%);
-      }
-
-      &--left {
-        @extend %is-faded;
-
-        clip-path: polygon(0% 0%, 100% 34%, 100% 100%, 0% 66.6%);
-      }
-    }
-
     &.is-architecture {
       @extend %diamond-badge;
-      @extend %is-faded;
 
       background-image: url(https://assets.ubuntu.com/v1/8b175afd-Architecture.svg);
     }
 
     &.is-bash {
       @extend %rhombus-badge;
-      @extend %is-faded--right;
 
       background: url(https://assets.ubuntu.com/v1/4b2c2de2-bash.svg);
       transform: translate(
@@ -66,7 +46,6 @@
 
     &.is-devices {
       @extend %rhombus-badge;
-      @extend %is-faded--left;
 
       background-image: url(https://assets.ubuntu.com/v1/9ba960cc-devices+and+files.svg);
       transform: translate(0, map-get($diamondSize, "height") * 0.5px);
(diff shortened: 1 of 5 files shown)


static/sass/_pattern_cube-progression.scss
```diff
@@ -202,4 +169,23 @@
       );
     }
   }
+
+  .is-faded {
+    // Without !important the background only applies to certain microcerts
```
anthonydillon, line 174: This is due to the use of `background: ...` as a style on elements such as ".p-cube-progression.is-bash".

The shorthand (`background:`) has a high speciality than (`background-color`, `background-image`, etc). It's generally good practice to use the most specific styling attribute if you are setting a single value. Because of strange issues like these.

Could you replace the use to `background:` with `background-image:` and you shouldn't need this important.

static/sass/_pattern_cube-table.scss
```diff
@@ -1,13 +1,20 @@
+// Media queries added to width of tds on mobile view cards
+// Bug: https://github.com/canonical-web-and-design/vanilla-framework/issues/3754
 .p-table--cube {
   th,
   td {
     &:nth-child(1) {
       overflow: visible;
-      width: 4%;
+
+      @media only screen and (min-width: $breakpoint-medium) {
```
anthonydillon, line 9: Can you not just wrap the entire block in a single media query?

static/sass/_pattern_cube-table.scss
```diff
@@ -121,6 +133,12 @@
 
     > .p-table--cube__badge {
       align-self: center !important;
+
+      @media only screen and (min-width: $breakpoint-medium) {
+        .is-passed {
```
anthonydillon, line 138: What is this for? It looks like this state class is used in conjunction with `.p-table--cube__badge` and so can not be matched as this would need `&.is-passed`.

anthonydillon: It looks like you are replacing the code below which is: 
```css
margin-left: -0.2rem;
```

anthonydillon (changes requested):
Much much better @bethcollins92. Few requests in the code.

static/sass/_pattern_cube-progression.scss
bethcollins92: Thanks for the explanation - very helpful

static/sass/_pattern_cube-table.scss
bethcollins92: Good spot thanks. This is hard to QA as when you pass a microcert a new "passed" orange badge is sent from badgr which has slightly different dimensions. I QA'd it at the time but since the alpha has gone live I've no longer passed any. I will replace this as come to think of it I think I needed this style on all screen sizes.

bethcollins92:
Thanks @anthonydillon - addressed those comments

templates/cube/microcerts.html
```diff
@@ -210,17 +210,18 @@ <h2>Microcertifications</h2>
...
-                style="margin-left: -0.2rem;"
-                {% endif %}
-              />
+            <img
+              class="p-table--cube__badge"
+              src="{{ module.badge.url }}"
+              alt="Badge for {{ module.name }}"
+              {% if module.status == 'passed' %}
+              style="margin-left: -0.2rem;"
+              {% endif %}
```
anthonydillon, line 219: I don't understand, should this switch out the image? This is what I see:
image

static/sass/_pattern_cube-table.scss
```diff
@@ -1,41 +1,44 @@
+// Media queries added to width of tds on mobile view cards
+// Bug: https://github.com/canonical-web-and-design/vanilla-framework/issues/3754
```
anthonydillon, line 2: Might be worth checking this with the latest Vanilla to see if needed.

anthonydillon (changes requested):
Few comments inline.

templates/cube/microcerts.html
bethcollins92: No, I'll double check with Joao but I think as we are fake passing the microcerts the "passed" microcert isn't being fetched from Badgr so (that's why it's hard to QA.) But I remember having to add this at the time of building as the dimensions of the passed badge were slightly different.

anthonydillon (approved):
Very cool, thanks @bethcollins92 :+1:

PR #9799, opened 2021-05-24
Gray out button

### LINKED ISSUE #9641: UA shop: can edit a subscription while an edit is already pending
**Steps to reproduce**:
The same as #9639 or #9640.

**What is expected**: if a previous purchase subscription is stuck for a subscription, the "Change subscription" button should be greyed out.

**What happens**: user can edit the subscription, which fails but causes the spinner to be displayed forever.

albertkol:
## Done

- Gray out change subscription button if there is a pending payment.

## QA

- Have a pending payment
- Check the change subscription button for both a yearly and a monthly subscription

## Issue / Card

Fixes #9641

### DIFF
--- a/templates/advantage/table/_manage-monthly.html
+++ b/templates/advantage/table/_manage-monthly.html
@@ -18,7 +18,7 @@
 
     <div class="change-subscription {% if mobile %}u-align--center{% endif %}">
       <p>Auto-renewal {% if monthly_information['is_auto_renewal_enabled'] %}enabled{% else %}disabled{% endif %}</p>
-      <button class="p-button--neutral js-change-subscription-button" data-id="{{contract["contractInfo"]["id"]}}--monthly" data-viewport="{% if mobile %}small{% else %}large{% endif %}">Change subscription</button>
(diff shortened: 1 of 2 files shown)


ClementChaumel:
So, this prevents any change on any contract if there is a pending payment. Shouldn't we be able to still downsize and cancel monthly ones?

albertkol:
@ClementChaumel not sure what to say.. Ideally we force the user to fix its payments asap, regardless if it comes form a monthly or a yearly subscription. That would be my choice. 

@wgx what do you think?
To give you more context. If a user has a pending payment against a subscription (stripe subscription, not a table row) it locks it, preventing any further purchases. A user can have max two (stripe) subscriptions: a yearly and a monthly. 
If your monthly subscription is locked you can still make purchases/resizes against your yearly one. Also this means that in theory a user can end up with a maximum of two pending payments.

This PR change makes it so that the "Change subscription" button from all table rows (monthly and yearly) get disabled, regardless which stripe subscription the pending payment is issues against.

wgx:
@albertkol @ClementChaumel 

I agree this isn't great long-term, but for this version it's fine. 

We're redesigning this UI - starting this week - so these controls will move anyway. 

A short term fix could be to add a tooltip to the cancelled button

> You cannot change this subscription because you have payment issues to resolve.

Bear in mind they will also have a big yellow notificaiton banner on their views with a link to resolve.

ClementChaumel (approved):

PR #9815, opened 2021-05-26
Certification master branch merge

carkod:
## Done

Certification/Certified project. 

## QA

All PRs have been reviewed. Look around for bugs, typos, code missing etc...


codersquid:
The model pages are missing the kernel and bios info.

webapp/certified/views.py
```diff
@@ -0,0 +1,352 @@
...
+
+def certified_component_details(component_id):
+
+    try:
+        component = api.component_summary(component_id)
+    except requests.exceptions.HTTPError as error:
+        if error.response.status_code == 404:
+            abort(404)
+        else:
+            current_app.extensions["sentry"].captureException()
```
codersquid, line 25: I notice here the exception is captured for sentry but not in other places. is that intentional?

webapp/certified/views.py
```diff
@@ -0,0 +1,352 @@
...
+
+        if int(vendor["smart_core"] > 1):
+            iot_vendors.append(vendor)
+
+        if int(vendor["soc"] > 1):
+            soc_vendors.append(vendor)
+
+    # Server section
+    server_releases = {}
+    server_vendors = api.vendor_summaries_server()["vendors"]
```
codersquid, line 258: why not treat servers the same way the other categories are treated? if you do that you can drop the expensive api call.

codersquid (commented):
I have a couple of questions inline.

carkod:
> The model pages are missing the kernel and bios info.

BIOS info is in the hardware details page, I believe this is what we agreed on in terms of design e.g. https://ubuntu-com-9815.demos.haus/certified/201708-25655/18.04%20LTS

Kernel info, not sure which part you mean exactly. Is it `This system was tested with 18.04 LTS, running the 4.15.0-20-generic kernel.` on the old site?

webapp/certified/views.py
codersquid: We discussed this and decided to leave it the same for now since the results are counted differently on the old site and we want to reproduce what the old site does.

webapp/certified/views.py
carkod: I think in this case we can only do this if there is an HTTP error. But Sentry will catch any 500 errors anyway.

webapp/certified/views.py
carkod: As discussed, data is no the same.

webapp/certified/views.py
```diff
@@ -52,7 +52,11 @@ def certified_hardware_details(canonical_id, release):
     try:
         models = api.certified_models(canonical_id=canonical_id)["objects"][0]
     except KeyError:
-        abort(404)
+        if error.response.status_code == 404:
```
codersquid, line 55: Here there is no error with a status code to check since it is not catching the http exception.

webapp/certified/views.py
```diff
@@ -0,0 +1,353 @@
...
+            arch = "64 Bit"
+
+        release_info = {
+            "name": f"Ubuntu {ubuntu_version} {arch}",
+            "kernel": model_release["kernel_version"],
+            "bios": model_release["bios"],
+            "level": model_release["level"],
+            "notes": model_release["notes"],
+            "version": ubuntu_version,
+            "download_url": get_download_url(model_release),
```
codersquid, line 144: The logic for choosing a download url is in the template now, so I think it is safe to drop this.

codersquid (commented):
I have an inline comment about dropping hte download_url bit now that it is done in the template.

but in addition to that, another person probably has to make the call but I think the download link on the model details page could be dropped now that this is a part of ubuntu.com and no longer certified.ubuntu.com.

webapp/certified/views.py
carkod: Done. Thanks.

codersquid (approved):
You've resolved all of my comments. lgtm.

templates/certified/model-details.html
```diff
@@ -0,0 +1,263 @@
...
+                {% endif %}
+                {% if category == "Server" %}
+                <a href="/download/server" class="p-button--neutral">Download</a>
+                {% endif %}
+                {% if category == "Server SoC" %}
+                <a href="/download/server/arm" class="p-button--neutral">Download</a>
+                {% endif %}
+              </p>
+            {% endif %}
+            {% if release.kernel_version %}
```
bethcollins92, line 101: I think this should be `release_details.kernel` looking at what's being sent to the template from the view? Also should we include the bios too going off what Sheila said?

webapp/certified/views.py
```diff
@@ -0,0 +1,350 @@
...
+        ubuntu_version = model_release["certified_release"]
+        arch = model_release["architecture"]
+
+        if arch == "amd64":
+            arch = "64 Bit"
+
+        release_info = {
+            "name": f"Ubuntu {ubuntu_version} {arch}",
+            "kernel": model_release["kernel_version"],
+            "bios": model_release["bios"],
```
bethcollins92, line 138: bios here - in `release_info`

templates/certified/model-details.html
carkod: All Done

PR #9830, opened 2021-05-27
Add redirects for certified microsite

carkod:
## Done

Redirect https://certification.ubuntu.com to https://ubuntu.com/certification

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9429

### DIFF
--- a/webapp/certified/views.py
+++ b/webapp/certified/views.py
@@ -2,9 +2,10 @@ import talisker.requests
 import talisker.sentry
 import requests
 import math
-from flask import request, render_template, abort, current_app
+from flask import request, render_template, abort, current_app, redirect
 from requests import Session
 from webapp.certified.api import CertificationAPI
+from urllib.parse import urlencode
 
 session = Session()
 talisker.requests.configure(session)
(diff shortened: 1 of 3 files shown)


webapp/certified/views.py
```diff
@@ -264,10 +265,21 @@ def certified_home():
 
     if request.args:
         query = request.args.get("q", default=None, type=str)
+
+        # Old site replacements
+        if set(request.args) & set(["query", "vendors"]):
+            # Convert ImmutableMultiDict into normal dict
+            old_params = request.args.to_dict()
```
nottrobin, line 272: If you're gonna manipulate this list in-place to replace old params with new ones, then `old_params` isn't a terribly accurate name. `params` would be better. Or even `parameters`.

nottrobin (approved):
This looks good. One optional comment in code, but once you've considered that feel free to merge.

Presumably the act of merging this and releasing it won't actually do the switch, because the final step is a DNS switch, is that right? I don't know if we need any more approval before we actually do the final DNS switch.

PR #9847, opened 2021-06-01
Refactor purchase modal

ClementChaumel:
## Done
Rewrote the modal as a react app using [react query](https://react-query.tanstack.com/), [Formik](https://formik.org/) and [React Vanilla components](https://canonical-web-and-design.github.io/react-components/?path=/docs/col--default-story)

The modal works the same way as before except for a couple of changes

## Changes
If a user is logged in and already registered a payment method with us, they won't have to go through the form again and instead directly land on the "recap / step 2" view 
image

If a user is buying from a VAT applicable country and we already know their VAT number, we prepopulate alongside the rest of the information
image

If a user's account is in a locked state (because a purchase failed) we tell them to go to /payment-method to fix it 
image
(thanks @albertkol )

Removed the 45s timer after a purchase which was causing issues and was made and replaced it with 5 spaced retries.

## QA

https://ubuntu-com-9847.demos.haus/advantage/subscribe?test_backend=true

Logged in / logged out 

Few things to check:
- successful purchase with normal card (4242424242424242);
- successful purchase triggering 3ds modal with the card (4000002760003184)
- failed purchase 3ds modal with the card (4000002760003184) 
- other test cards can be found here: https://stripe.com/docs/testing
- as a guest use an existing email

## Issue / Card

Fixes
https://github.com/canonical-web-and-design/commercial-squad/issues/11
https://github.com/canonical-web-and-design/ubuntu.com/issues/9638
https://github.com/canonical-web-and-design/ubuntu.com/issues/9370
https://github.com/canonical-web-and-design/commercial-squad/issues/2


SirSamTumless:
@ClementChaumel Please attach this to an issue

albertkol:
@wgx @elioqoshi @ClementChaumel 

Problem:
- Given I make a first ever purchase with my email address
- I insert all my details and click continue
- `ensurePurchaseAccount` will create an account in the background
- I click change my details
- I change details and press continue again
- I am forced to login because if the email address use already exists in the database we always force the user to login

Same problem but slightly different:
- Given I make a first ever purchase with my email address
- I insert all my details and click continue (with a faulty card 4000000000000341 or 4000002760003184)
- `ensurePurchaseAccount` will create an account in the background
- I the payment fails and I see the error message. I am allowed to edit my details.
- I change details and press continue again
- I am forced to login because if the email address use already exists in the database we always force the user to login

Maybe we can allow the user to continue the purchase without forcing the user to login if they have an account. Although we might run in different issues then, not sure... but I think it would be fine.

albertkol:
@wgx @elioqoshi @ClementChaumel 

After I press the button I cannot see the error message until I scroll up.

We could:
- move the error message above the submit button
- or scroll automatically up

Or you think it's fine?

wgx:
The error message should be visible to the user without scrolling.

albertkol:
@wgx @ClementChaumel 

Is the VAT field for all EU countries or just UK?

VAT is inconsistent:
- it never gets auto-filled
- changing from a country that pays VAT to a country that does not would trigger a total price update and vice-versa (I think it's fine if the user has no account until they press continue for the first time)
- you are pushing the VAT even if I switched to a country that does not pay VAT and the field is hidden

albertkol:
Browsing to /advantage/subscribe sometimes launches the modal directly. I think it has something to do with having made already a purchase and revisiting that link.

albertkol:
@wgx @ClementChaumel 

Short terms sending the user to `/payment-methods` is fine. 
image

But long term, we should immediately delete failed payments behind the scenes, so the user never walks away from the modal with pending payments and also allow him to retry as many times as they want in the modal. 

Just for context. This happens when your payment is made in the background, but it fails. The purchase gets attached to the subscription, and the user is forced to fix it, in order to proceed to do anything else.

static/js/src/advantage/contracts-api.js
```diff
@@ -306,3 +306,82 @@ export async function setAutoRenewal(value) {
...
+}
+
+export async function postLoggedInFreeTrial({
+  accountID,
+  name,
+  address,
+  productListingId,
+  quantity,
+}) {
+  const queryString = window.location.search; // Pass arguments to the flask backend eg. "test=backend=true"
```
anthonydillon, line 358: Typo again

static/js/src/advantage/subscribe/listeners/ui-event-listeners.js
```diff
@@ -1,16 +1,28 @@
...
+
+  const purchaseModalToggles = document.querySelectorAll(
+    "[aria-controls='purchase-modal']"
+  );
+  purchaseModalToggles.forEach((toggle) => {
+    toggle.addEventListener("click", (e) => {
+      e.preventDefault();
+      store.dispatch(togglePurchaseModal());
+    });
+  });
```
anthonydillon, line 27: I think this could do with a comment to explain its purpose?

static/js/src/advantage/subscribe/react/APICalls/PendingPurchase.jsx
```diff
@@ -0,0 +1,77 @@
...
+        pi_status,
+        pi_secret,
+        id: stripeInvoiceId,
+      } = res.stripeInvoices[0];
+
+      //Requires 3DS check
+      if (
+        pi_decline_code === "authentication_required" ||
+        (pi_status === "requires_action" && pi_secret)
+      ) {
```
anthonydillon, line 31: You could move this conditional to a function called `if (requires3DSCheck(..)) {` which would explain itself without the need for the comment.

static/js/src/advantage/subscribe/react/APICalls/Product.jsx
```diff
@@ -0,0 +1,37 @@
+import { useState } from "react";
+import { useQuery, useQueryClient } from "react-query";
+
+const useProduct = () => {
+  const [observerSet, setObserverSet] = useState(false);
+
+  const buyButton = document.querySelector("#buy-now-button");
```
anthonydillon, line 7: I assume in the future we will move to using `ref` but it would be better to target js-prefixed classes

static/js/src/advantage/subscribe/react/APICalls/RegisterPaymentMethod.jsx
```diff
@@ -0,0 +1,93 @@
...
+    const {
+      name,
+      organisation,
+      email,
+      address,
+      city,
+      country,
+      postalCode,
+      usState,
+      CAProvince,
```
anthonydillon, line 22: Should probably be `caProvince` to be consistent

static/js/src/advantage/subscribe/react/APICalls/RegisterPaymentMethod.jsx
```diff
@@ -0,0 +1,93 @@
...
+      taxID: {
+        type: "eu_vat",
+        value: VATNumber,
+      },
+    });
+
+    if (customerInfoRes.errors) {
+      //We ignore VAT errors but throw the others
+      if (JSON.parse(customerInfoRes.errors).code !== "tax_id_invalid")
+        throw new Error(JSON.parse(customerInfoRes.errors).code);
```
anthonydillon, line 81: Where has the curly brackets gone?

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
```diff
@@ -0,0 +1,319 @@
...
+      setStep(1);
+    }
+  }, [purchaseError]);
+
+  useEffect(() => {
+    if (pendingPurchase?.status === "done") {
+      // Remove the product selection from the local storage
+      localStorage.removeItem("ua-subscribe-state");
+
+      //redirect
```
anthonydillon, line 127: Could you expand on this comment please

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
```diff
@@ -0,0 +1,319 @@
...
+        queryClient.setQueryData(
+          "userInfo",
+          getUserInfoFromVariables(data, variables)
+        );
+
+        actions.setSubmitting(false);
+      },
+      onError: (error) => {
+        if (error.message === "email_already_exists") {
+          //Email already exists
```
anthonydillon, line 153: I think this is clear from the condition above

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
```diff
@@ -0,0 +1,319 @@
...
+  };
+
+  const onPayClick = () => {
+    purchaseMutation.mutate("", {
+      onSuccess: (data) => {
+        //start polling
+        setPendingPurchaseID(data);
+      },
+      onError: (error) => {
+        // An error happened
```
anthonydillon, line 186: I think this is clear

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
```diff
@@ -0,0 +1,319 @@
...
+        if (error.message.includes("can only make one purchase at a time")) {
+          setError(
+            <>
+              You already have a pending purchase. Please go to{" "}
+              <a href="/advantage/payment-methods">payment methods</a> to retry.
```
anthonydillon, line 191: Do we display pending purchases on the payment-methods page?

static/js/src/advantage/subscribe/react/components/FormRow.jsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Row, Col } from "@canonical/react-components";
+
+const formRow = ({ label, children, error }) => {
+  return (
+    <Row className="p-form__group  u-no-padding u-vertically-center">
+      <Col size="4" style={{ alignSelf: "start" }}>
+        <label className="">{label}</label>
```
anthonydillon, line 8: Unnecessary empty className here

static/sass/_pattern_renewal-modal.scss
```diff
@@ -181,7 +186,7 @@
     input,
     select,
     textarea {
-      margin-bottom: 0;
+      // margin-bottom: 0;
```
anthonydillon, line 189: Remove commented out code please

templates/advantage/subscribe/_purchase-modal.html
```diff
@@ -0,0 +1,6 @@
+<div class="p-modal p-modal--ua-payment is-details-mode"id="purchase-modal" style="display: none;">
```
anthonydillon, line 1: Add space between class and id

anthonydillon (approved):
Code looks much better, well done @ClementChaumel. Approved but a few comments in the code.

package.json
```diff
@@ -31,18 +31,21 @@
...
     "autoprefixer": "10.2.4",
     "babel-jest": "26.6.3",
     "babel-loader": "8.2.2",
     "concurrently": "6.0.0",
     "cypress": "5.6.0",
+    "esbuild": "0.9.2",
     "eslint": "7.19.0",
     "eslint-config-prettier": "7.2.0",
     "eslint-plugin-cypress": "2.11.2",
+    "eslint-plugin-react": "^7.23.2",
```
hatched, line 48: You probably want to pin these (aka remove the hat ^)

package.json
```diff
@@ -62,8 +65,15 @@
     "@canonical/cookie-policy": "3.2.0",
     "@canonical/global-nav": "2.4.5",
     "@canonical/latest-news": "1.2.0",
+    "@canonical/react-components": "^0.17.1",
     "@reduxjs/toolkit": "^1.5.0",
+    "@stripe/react-stripe-js": "^1.4.1",
+    "@stripe/stripe-js": "^1.15.0",
     "date-fns": "2.17.0",
+    "formik": "^2.2.8",
```
hatched, line 73: While we're using formik on the two app UI's right now we're evaluating alternatives that don't have the same performance and api restrictions.... Just something to keep in mind for later :)

static/js/src/advantage/contracts-api.js
```diff
@@ -306,3 +306,82 @@ export async function setAutoRenewal(value) {
...
+      ],
+    }),
+  });
+
+  if (response.ok) {
+    const data = await response.json();
+    return { ok: true, ...data };
+  } else {
+    console.error(`${response.status} - ${response.statusText}`);
+    return { ok: false };
```
hatched, line 347: When you can you want to keep the type and structure of a function return value the same so that consumers of the API don't have to be quite as defensive. 

In this case that would mean that both return objects would contain a data object, even if it's empty.

static/js/src/advantage/contracts-api.js
```diff
@@ -306,3 +306,82 @@ export async function setAutoRenewal(value) {
...
+        },
+      ],
+    }),
+  });
+
+  if (response.ok) {
+    const data = await response.json();
+    return { ok: true, ...data };
+  } else {
+    console.error(`${response.status} - ${response.statusText}`);
```
hatched, line 346: As this is an API layer this information might be interesting to the consumer of the API. Consider passing the full `response` object back to the user.

static/js/src/advantage/contracts-api.js
```diff
@@ -306,3 +306,82 @@ export async function setAutoRenewal(value) {
...
+      products: [
+        {
+          product_listing_id: productListingId,
+          quantity: quantity,
+        },
+      ],
+    }),
+  });
+
+  if (response.ok) {
```
hatched, line 342: I'm not sure that this response handling is valuable in the API as it's written. It's actually reducing the amount of data that the consumer is receiving but it's not providing any protection against the failure cases or if the json fails to parse. 

If you wanted to just return the data from the request then I might suggest a structure like...

`return { data, err };`

So you'd only ever return the successfully parsed data and an error message that would default to null so the consumer would not have to wrap this function in a try/catch

So it would take the usage from...

```js
let response = null;
try {
 response = await postGuestFreeTrial(...);
catch (e) {
 // handle error
}
// use response
```

to...

```js

const {data, err} = await postGuestFreeTrial(...);
if (err) {
 // handle error
}
//use data
```

I'd apply similar changes to the other function in this file as well.

static/js/src/advantage/countries-and-states.js
```diff
@@ -0,0 +1,368 @@
+export const vatCountries = [
```
hatched, line 1: Can these values be returned from the server instead of loaded into the bundle?

static/js/src/advantage/subscribe/listeners/ui-event-listeners.js
```diff
@@ -1,16 +1,28 @@
...
       e.preventDefault();
       store.dispatch(toggleOtherVersionsModal());
     });
   });
+
+  const purchaseModalToggles = document.querySelectorAll(
+    "[aria-controls='purchase-modal']"
+  );
+  purchaseModalToggles.forEach((toggle) => {
+    toggle.addEventListener("click", (e) => {
```
hatched, line 23: As this is an exported function, how do we know that we won't get multiple events attached?

static/js/src/advantage/subscribe/react/APICalls/PendingPurchase.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const stripe = useStripe();
+
+  const { isLoading, isError, isSuccess, data, error } = useQuery(
+    "pendingPurchase",
+    async () => {
+      const res = await getPurchase(pendingPurchaseID);
+
+      if (!res.stripeInvoices) {
+        throw new Error("Missing invoice");
```
hatched, line 17: There is no need to throw here, please handle the error gracefully instead of panicking

static/js/src/advantage/subscribe/react/APICalls/PendingPurchase.jsx
```diff
@@ -0,0 +1,77 @@
+import { useStripe } from "@stripe/react-stripe-js";
```
hatched, line 1: This is a `jsx` file with a capitalized name that's not a component. It appears to be a hook instead. Can you please collect the hooks into a centralized location and use the hook file naming convention of `usePendingPurchase.js`

static/js/src/advantage/subscribe/react/APICalls/PendingPurchase.jsx
```diff
@@ -0,0 +1,77 @@
...
+
+      //Card declined
+      if (pi_status === "requires_payment_method") {
+        const invoiceRes = await postInvoiceID(
+          "purchase",
+          pendingPurchaseID,
+          stripeInvoiceId
+        );
+
+        const error = Error(JSON.parse(invoiceRes.errors).decline_code);
```
hatched, line 48: You're mixing constructors with your thrown errors. But I'd prefer to never throw unless you want to application to halt, failing gracefully is preferred.

static/js/src/advantage/subscribe/react/APICalls/Preview.jsx
```diff
@@ -0,0 +1,44 @@
+import { useQuery } from "react-query";
```
hatched, line 1: See above comment on naming conventions.

static/js/src/advantage/subscribe/react/APICalls/Product.jsx
```diff
@@ -0,0 +1,37 @@
+import { useState } from "react";
```
hatched, line 1: naming convention :)

static/js/src/advantage/subscribe/react/APICalls/Product.jsx
hatched: This is not idiomatic React. You'll want to modify the call signature of this hook to accept a ref to the component that you want to use here.

static/js/src/advantage/subscribe/react/APICalls/Purchase.jsx
```diff
@@ -0,0 +1,37 @@
+import { useMutation } from "react-query";
```
hatched, line 1: This appears to be a utility function but it's a `jsx` file with a capitalized filename.

static/js/src/advantage/subscribe/react/APICalls/RegisterPaymentMethod.jsx
hatched: I'm surprised the linter allowed this - might be best to add that rule to the linter.

hatched (changes requested):
While I haven't finished the review I think there are a number of changes to be made to make this more react-like. If you wouldn't mind reviewing the comments I've made and applying those changes throughout the PR and then I can take another look. 

Thanks for all the work you put into this so far it's a great start 👍 

If you have any questions please feel free to ping.

wgx:
I've given this a once-over and it seems to be fine to me. As long as we replicate the experience of the last form then I have no concerns. I did log [a separate issue to address the Card fields](https://github.com/canonical-web-and-design/ubuntu.com/issues/9873) inconsistency.

package.json
ClementChaumel: I did try to use react-hook-form in the first place but the vanilla components weren't playing nice with it and so I turned to Formik instead

static/js/src/advantage/countries-and-states.js
ClementChaumel: @albertkol ? :smiley:

static/js/src/advantage/subscribe/react/APICalls/PendingPurchase.jsx
ClementChaumel: It's the way `react-query` handles errors https://react-query.tanstack.com/guides/query-functions#handling-and-throwing-errors

ClementChaumel:
> @albertkol 
> * it never gets auto-filled
I just tested it and it does auto-fill it for me when I'm logged in? We'll need to look into it

> * you are pushing the VAT even if I switched to a country that does not pay VAT and the field is hidden
I think Allan said it's fine but I can remove it

ClementChaumel:
> Browsing to /advantage/subscribe sometimes launches the modal directly. I think it has something to do with having made already a purchase and revisiting that link.


This is the part that "remembers" the state. If you leave the while on the modal and come back it will display the modal populated with the product you selected just like when you left it. @wgx is that confusing?

static/js/src/advantage/subscribe/react/APICalls/Product.jsx
ClementChaumel: In the future, hopefully, the product selector will share its state with the modal directly. I'll update it to use a ref and js- classes

ClementChaumel: Actually, I remember now why I queried the button here. 6 components are subscribed to this query and I wanted to only search for this button once 
image
Can you think of a better pattern or should I just 
```
  const buyButton = document.querySelector(".js-buy-now-button");
```
in all of them and pass the button node to the function?


Also, looking into it I'm not sure it's possible to use refs here as the button I'm targeting lives outside of the react app. Or if it is I don't know how to do it

static/js/src/advantage/subscribe/react/APICalls/RegisterPaymentMethod.jsx
ClementChaumel: I'm writing Python now! :smile: 
I think the linter is fine with it for one liners. I'll add them back in for consistency.

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
ClementChaumel: :see_no_evil: Whoops that was a leftover from when the error message was cryptic!

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
ClementChaumel: So, my reasoning here was: If the error comes from invalid form data and can be retried, we display it above the form to let you retry. But if the error has to do with your account state or anything else that is out of your control then there is no point in showing you the form again since re-filling it would do nothing.
@wgx  what do you think?

wgx: @ClementChaumel @anthonydillon Yep, this is fine. Payment methods is the only way to fix a "pending purchases" state - so this is good.

static/js/src/advantage/contracts-api.js
ClementChaumel: The free trial endpoints (which are not used yet) were results of a prior iteration. I'll bring them in line with the rest of the calls :+1:

static/js/src/advantage/subscribe/listeners/ui-event-listeners.js
ClementChaumel: Currently, `initUIControls` is only called once.
I could switch from `.addEventListener(...)` to `.onClick=(...)` to make sure we don't attach duplicate event listeners. Would that be fine or do you know of a better way?

package.json
hatched: I'm interested to hear about that issue as that one is currently on the top of the list :)

ClementChaumel: IIRC the issue was that vanilla react component don't forward the ref to the actual input field and so it wasn't possible to control it (reset / set to a specific value)
https://react-hook-form.com/get-started#Integratinganexistingform

static/js/src/advantage/subscribe/react/APICalls/Product.jsx
hatched: Ahh ok, can you add some detailed comments to the code to outline this reasoning.

elioqoshi:
Looks good from my side as well, thanks for this!

static/js/src/advantage/subscribe/react/components/Summary.jsx
```diff
@@ -0,0 +1,139 @@
...
+          <div>{product?.name}</div>
+        </Col>
+      </Row>
+      <Row className="u-no-padding u-sv1">
+        <Col size="4">
+          <div className="u-text-light">Machines:</div>
+        </Col>
+        <Col size="8">
+          <div>
+            {quantity} x {formatter.format(product?.price?.value / 100)}
```
hatched, line 33: If there is an issue when deriving these values then you might end up with some funny output here. Maybe this should be calculated outside of the return so that sane defaults could be returned in the event of an error.

static/js/src/advantage/subscribe/react/components/Summary.jsx
```diff
@@ -0,0 +1,139 @@
...
+                add(new Date(), {
+                  months: product?.period === "monthly" ? 1 : 12,
+                }),
+                DATE_FORMAT
+              )}
+            </div>
+          </Col>
+        )}
+      </Row>
+      {preview ? (
```
hatched, line 69: This is a bit of an excessive use of a ternary :)

Consider abstracting this logic out into separate render functions.

static/js/src/advantage/subscribe/reducers/form-reducer.js
```diff
@@ -115,7 +115,9 @@ function getProduct(state) {
 
 const formSlice = createSlice({
   name: "form",
-  initialState: loadState("ua-subscribe-state", "form", initialFormState),
+  initialState: {
+    ...loadState("ua-subscribe-state", "form", initialFormState),
```
hatched, line 119: I'm not sure what `loadState` returns but in this case these two values are pretty much identical if there are nested objects in there.

hatched (approved):
Thanks for keeping at this @ClementChaumel

static/js/src/advantage/subscribe/reducers/form-reducer.js
ClementChaumel: Good catch I was at some point overriding some attributes after the deconstruction but I'm not anymore so I'll revert it

static/js/src/advantage/subscribe/react/components/Summary.jsx
ClementChaumel: This component is conditionally rendered depending on the presence of those values so I think it's pretty safe. Do you think I should still add another layer of safety?
I'm not sure what sane default could be returned for a missing price on a purchase summary. In this case, the absolute worst thing that can happen (And don't think it can) is to display NaN

anthonydillon:
I think @albertkol is in the best place to perform the QA here

PR #9857, opened 2021-06-03
Fix slow page problem

albertkol:
## Done

- Remove unnecessary echo

## QA

- Go to https://ubuntu-com-9857.demos.haus/advantage does it take ages to load the page to see your canonical subscriptions? If no, it works.

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/13


SirSamTumless:
🔔  ⏰  🛎️  @ClementChaumel

ClementChaumel (approved):
Still a bit slow but nothing like live! GJ can't believe that was the reason!!

albertkol:
@ClementChaumel we still have about 5-8 seconds load time. But once the page is loaded you are free to go and do your thing. Another one of the pending PRs will bring a small optimisation that Allan suggested. And the rest will be dealt during the table redesign. 

On production, the table loads, but then you get stuck and have to wait for 10+ seconds to be able to do anything. Which this PR should solve.

nottrobin:
The title of this PR doesn't seem accurate...

PR #9889, opened 2021-06-10
Align the navigation height with the new Vanilla styling

anthonydillon:
## Done
- Increase the height of the navigation to 56px to match the latest in Vanilla
- Removed the duration of the animation of the links to make it feel faster

## QA
- Open the demo
- Check the header is 56px height
- Check live is shallower
- Check the dropdowns work
- Check the user menu aligns when open

## Issue / Card
Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9865

## Screenshots
image

### DIFF
--- a/static/sass/_pattern_navigation.scss
+++ b/static/sass/_pattern_navigation.scss
@@ -92,18 +92,18 @@ $ubuntu-logo-width--mobile: $ubuntu-logo-svg-width + 2 * $sph-inner;
     }
 
     & &__link-anchor {
-      @include vf-animation(all, brisk, in);
+      @include vf-animation(all, snap, in);
 
       color: $color-light;
       display: block;
 
       @media (max-width: $breakpoint-navigation-threshold) {
-        padding: $spv-inner--medium $grid-margin-width $spv-inner--medium
+        padding: $spv-inner--large $grid-margin-width $spv-inner--large
           $grid-margin-width;
       }
 
       @media (min-width: $breakpoint-navigation-threshold + 1) {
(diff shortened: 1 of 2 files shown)


anasereijo:
@anthonydillon thanks for doing this. just a couple of amendments.

- [ ] if you zoom in you'll that the height it's off by a few pixels:

Screenshot 2021-06-10 at 17 12 33

- [ ] when I click to open the dropdown seems like there's a bit of delay on turning the background white:

https://user-images.githubusercontent.com/36884067/121559951-f7c27000-ca0e-11eb-8a7a-cd9ea5462f03.mov

fitojb:
Now that the main navigation bar is taller, I would humbly suggest to make the secondary navigation thinner as well, so that the complete navigation is exactly as tall as it was before (which besides avoiding bumping up the above-the-fold factor, IMHO it shows hierarchy more clearly as an additional benefit).
I've mocked up what I had in mind, if you wish to consider it:
image

anasereijo:
> Now that the main navigation bar is taller, I would humbly suggest to make the secondary navigation thinner as well, so that the complete navigation is exactly as tall as it was before (which besides avoiding bumping up the above-the-fold factor, IMHO it shows hierarchy more clearly as an additional benefit).
> I've mocked up what I had in mind, if you wish to consider it:
> image

Hi @fitojb, thanks for your feedback and suggestions. We are currently working on the redesign of our navigation and we will take into consideration your suggestions! Cheers

carkod (approved):
LGTM

PR #9896, opened 2021-06-14
Fix lost renewal bug

albertkol:
## Done

- Fix lost renewal bug

## QA

- Go to https://ubuntu-com-9896.demos.haus?email=email@address.com
- The email address should have an account with both `lost` and `active` renewals

### DIFF
--- a/webapp/advantage/views.py
+++ b/webapp/advantage/views.py
@@ -767,7 +767,7 @@ def _make_renewal(advantage, contract_info):
         return None
 
     sorted_renewals = sorted(
-        (r for r in renewals if r["status"] != "closed"),
+        (r for r in renewals if r["status"] not in ["lost", "closed"]),
         key=lambda renewal: parse(renewal["start"]),
     )
 


anthonydillon:
@albertkol sorry your QA steps are too vague. I have gone to the link provided but the standard homepage. Could you provide more detail, please?

anthonydillon:
Why is there no issue for this?

albertkol:
@anthonydillon I created an issue. I thought it's small enough to pass by without one.

I was hoping @alnvdl would take a look instead. He would have more context.. i.e. if my solution is correct; or the email address to QA with;

Here is the context:
https://pastebin.canonical.com/p/HqhRhbVzQd/

alnvdl (changes requested):
This would work for the case in hand, but a more lasting solution would be to only display renewals that have `status == "pending"`.

As for reproducing this, one can test this by creating multiple renewals with different states (lost, pending, closed) in ua-contracts. The renewal guide in was updated a while ago to include such an example: https://github.com/CanonicalLtd/ua-contracts/blob/develop/docs/renewals.md

A list of the possible statuses is available here: https://github.com/CanonicalLtd/ua-contracts/blob/develop/internal/store/store.go#L888-L937

albertkol:
@alnvdl made the change. Back with you. I tested it to reproduce the problem with the email address from the pastebin. The renewal shows up now.

anthonydillon (approved):
LGTM to fix a live issue

PR #9900, opened 2021-06-14
Replaced the g-recatcha with two honeypot fields

anthonydillon:
## Done
- Replaced the g-recatcha with two honeypot fields

Do no merge until this one merges: https://github.com/canonical-web-and-design/ubuntu.com/pull/9590/

## QA
- Go to /engage/multi-cloud-business-guide and fill in the form and see if work
- Go back to /engage/multi-cloud-business-guide and inspect the form and add a value to the website or name hidden fields then submit and check it errors
- Go to /core/contact-us and check that works as before
- Go to /internet-of-things#get-in-touch and check that works as before.

## Issue / Card
Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/9753

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -18,6 +18,7 @@ from ubuntu_release_info.data import Data
 from canonicalwebteam.store_api.stores.snapstore import SnapStore
 from canonicalwebteam.launchpad import Launchpad
 from geolite2 import geolite2
+from requests import Session
 from requests.exceptions import HTTPError
 from canonicalwebteam.search.models import get_search_results
 from canonicalwebteam.search.views import NoAPIKeyError
@@ -25,11 +26,21 @@ from canonicalwebteam.search.views import NoAPIKeyError
 
 # Local
 from webapp.login import user_info
+from webapp.marketo import MarketoAPI
 
 
 ip_reader = geolite2.reader()
 session = talisker.requests.get_session()
 store_api = SnapStore(session=talisker.requests.get_session())
+
+marketo_session = Session()
+talisker.requests.configure(marketo_session)
+marketo_api = MarketoAPI(
+    "https://066-EOV-335.mktorest.com",
+    os.getenv("MARKETO_API_CLIENT"),
+    os.getenv("MARKETO_API_SECRET"),
+    marketo_session,
+)
 
 
 def _build_mirror_list():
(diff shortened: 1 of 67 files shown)


webapp/views.py
```diff
@@ -716,3 +734,62 @@ def sitemap_index():
...
+    form_fields = {}
+    for key, value in flask.request.form.items():
+        if value:
+            form_fields[key] = value
+
+    # Check honeypot values are not set
+    honeypots = {}
+    honeypots["name"] = flask.request.form.get("name")
+    honeypots["website"] = flask.request.form.get("website")
+    if honeypots["name"] != None and honeypots["website"] != None:
```
jpmartinspt, line 753: i think this can be simplified to `if form_fields.get("name") or form_fields.get("website") `, because both `None` and `""` evaluate to False.

webapp/views.py
```diff
@@ -716,3 +734,62 @@ def sitemap_index():
...
+        if honeypots["name"] != "" and honeypots["website"] != "":
+            raise BadRequest("Unexpected hotpot fields (name, website)")
```
jpmartinspt, line 755: honeypot instead of hotpot?

webapp/views.py
jpmartinspt: You might actually be able to `pop` right here too as if you add `None` as a default.

anthonydillon: I know this looks strange but it's because there is logically a difference between `None` and empty string here. 

1. The first check, checks if we are working with a form that contains honeypots or the legacy ones using recaptcha. 
2. The second that checks for empty string is actually testing if the honeypots have been triggered 

I am tempted to break these checks out into easy to read named simple functions. But that splits the code around more.

webapp/views.py
anthonydillon: Good spot, fixed

webapp/views.py
nottrobin: Since this wasn't clear to João, maybe you could add a comment explaining?

webapp/views.py
nottrobin: lol. I like hotpot though...

webapp/views.py
anthonydillon: Good idea, done.

anthonydillon:
@nottrobin @pmahnke would you mind reviewing this change please?

pmahnke:
It worked when I filled it in without honey pot fields - GOOD
It also worked when I filled in the 'website' field - BAD

Also, what do I have to do on the Marketo side to turn off the g-recaptcha work there?

anthonydillon:
Being split up and proposed as separate PRs

PR #9940, opened 2021-06-23
9494 add free trial to purchase modal

### LINKED ISSUE #9494: Add free trial option to the payment modal on /subscribe
When the product is available for a free trial, display a radio button to toggle between real purchase and trial on the payment modal.

ClementChaumel:
## Done

- Added ability to start a free trial when the product is eligible

## QA

- go here https://ubuntu-com-9940.demos.haus/advantage/subscribe?test_backend=true&esm_apps=true
- select infra + apps essential
- start a trial!

## Issue / Card

Fixes #9494 

## Screenshots
image

### DIFF
--- a/build.js
+++ b/build.js
@@ -30,7 +30,7 @@ for (const [key, value] of Object.entries(entries)) {
     sourcemap: true,
     outfile: "static/js/dist/" + key + ".js",
     target: ["chrome58", "firefox57", "safari11", "edge16"],
-    define: { "process.env.NODE_ENV": '"production"' },
+    define: { "process.env.NODE_ENV": '"dev"' },
   };
 
   esbuild.build(options).then((result) => {
--- a/build.js
+++ b/build.js
@@ -30,7 +30,7 @@ for (const [key, value] of Object.entries(entries)) {
     sourcemap: true,
     outfile: "static/js/dist/" + key + ".js",
     target: ["chrome58", "firefox57", "safari11", "edge16"],
-    define: { "process.env.NODE_ENV": '"dev"' },
+    define: { "process.env.NODE_ENV": '"production"' },
   };
 
   esbuild.build(options).then((result) => {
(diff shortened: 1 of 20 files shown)


wgx:
1. Form cannot be submitted - returns 405 method not allowed
2. Tab order - it's possible to (accidentally) tab into the disabled credit card fields
3. Product selector - it's not possible to click the hyperlink on the panel because the whole panel is already a button (see below, cc @elioqoshi ) - feel free to make separate issue to resolve this. 

image

elioqoshi:
Thanks for flagging. I opened #9961 for this

albertkol:
Free trials seems to be working both for logged in user and logged out user.

Few mentions:
1) Guest purchase should not make any `customer-info-anon` request, because we have no account, because there is no payment_method_id, because there is no card
2) VAT field never shows up
3) User already trialling a product is allowed to purchase the same trial as a guest again. The modal sends them to /thank-you; it will fail as soon at the trial purchase is triggered on login and sending the user to a 500. After a refresh the user will be back to /advantage. (we can discuss this later)

ClementChaumel:
> * VAT field never shows up

We don't need VAT for a free trial, do we?

albertkol:
Yes. I forgot. An extra reason not to have any `anon` requests in the background on changing the country.

anasereijo:
The form looks good 👍 

not sure if it's related to this PR but here it goes:

Can you wrap the getting started text? Looks weird having just the word address on its own.
The sign-in button, do we need to use ellipsis? it doesn't open an edit mode it just takes you to SSO -maybe this question is for Will?

Screenshot 2021-06-29 at 14 20 48

elioqoshi:
Form works well, thanks @ClementChaumel 

Little nitpick: I feel it's unfortunate there is scrolling on the modal as it's so long, but that's not fixable without changing the layout and @SirSamTumless would scold me for even thinking about that 🤐

albertkol (approved):
:+1:

PR #9946, opened 2021-06-23
Override the defaultState values instead of trusting blindly the local state

ClementChaumel:
## Done

Fix a bug where user who visited the site on a previous version would have a missing attribute in their localstorage state


albertkol (commented):
:+1: Seems to work.

albertkol (approved):

PR #9950, opened 2021-06-25
Add grace period

albertkol:
## Done

- `contract.contractInfo.grace_period_end` shows the formatted grace period of the contract. It will only show up if the contract is expired.


ClementChaumel (approved):
Thanks:

PR #9985, opened 2021-07-05
Check for defaultPaymentMethod instead of customerId to decide which step to show

ClementChaumel:
## Done

- Display the second step of the form only if there is a default payment method

## QA

- Go to https://ubuntu-com-9985.demos.haus/advantage/subscribe?test_backend=true 
- log in with an account that has no payment method set
- open the modal and check that you see the form
- log out
- Log in with an account with a payment method
- open the modal and  check that you see the summary + pay button

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/43


albertkol (approved):
:+1:

PR #9989, opened 2021-07-05
Fix modal loading forever for users with only salesforce contracts

ClementChaumel:
## Done

- Fix the modal for users who don't have a stripe account yet. (Users with only salesforce contracts)

## QA

- Ask Albert :joy: 

or 

- Create a new Ubuntu one account, ask Albert or Allan to add a renewal to your account
- go to /subscribe and open the modal
- check that it displays the form correctly

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/60

### DIFF
--- a/static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoAnon.jsx
+++ b/static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoAnon.jsx
@@ -22,21 +22,25 @@ const usePostCustomerInfoAnon = () => {
       state: country === "US" ? usState : caProvince,
     };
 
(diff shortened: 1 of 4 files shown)


albertkol (approved):
The modal loads now, but the preview still get's called immediately in the background 3 times. 
There is also a random customer-info-anon happening. 

image

But the issue is fixed, so I am approving this.

ClementChaumel:
> The modal loads now, but the preview still get's called immediately in the background 3 times.
> There is also a random customer-info-anon happening.
> 
> image
> 
> But the issue is fixed, so I am approving this.

I've removed the customer-info-anon one but I'll look at the preview calls in another PR :sweat_smile:

PR #9993, opened 2021-07-06
Fix missing grid-margin-width in Vanilla 2.32

bartaz:
## Done

- Replaces deprecated variable with correct use for Vanilla 2.32

This should fix Vanilla build errors when updating to v2.32 in #9937

## QA

- Check out this feature branch or use [demo](https://ubuntu-com-9993.demos.haus)
- Make sure navigation renders as expected


anthonydillon:
Ah I have just been updating sites to use the map instead.

PR #10037, opened 2021-07-15
Hide expiration date

### LINKED ISSUE #74: Home: rows are not going full width [IE9, Win 7]
Not an issue on live.

albertkol:
## Done

- Number of days to expiry can be flawed if a product was purchased both from a monthly and a yearly subscription, so we will hide it for now

## QA

- Have a monthly subscription on your testing account
- Login to https://ubuntu.com/advantage/subscribe?test_backend=true
- Disable auto-renewal for your account
- Take a look at your monthly product -> Click change subscription
- You will see a notification telling you how many days until expiry
- Go to: https://ubuntu-com-10037.demos.haus/advantage?test_backend=true
- Take a look at your monthly product -> Click change subscription
- You will **_NOT_** see a notification telling you how many days until expiry

## Issue / Card

Fixes #74

### DIFF
--- a/templates/advantage/table/_manage-monthly.html
+++ b/templates/advantage/table/_manage-monthly.html
@@ -45,13 +45,6 @@
     {% if not mobile %}
       <hr />
     {% endif %}
-    {% if not monthly_information['is_auto_renewal_enabled'] %}
-      <div class="p-notification--caution">
-        <p class="p-notification__response" role="status">
-          <span class="p-notification__status">End in {{contract["contractInfo"]["daysTillExpiry"] }} days.</span>Auto-renewal disabled.
-        </p>
-      </div>
-    {% endif %}
     <p class="p-muted-text" style="margin-bottom: .5rem">Machines</p>
     <div class="panel-input-wrapper">
       <input 


requirements.txt
```diff
@@ -1,7 +1,7 @@
 canonicalwebteam.flask-base==0.9.0
 alembic==1.5.4
 canonicalwebteam.http==1.0.3
-canonicalwebteam.blog==6.3.1
+canonicalwebteam.blog==6.4.0
```
anthonydillon, line 4: I don't understand, this is already the case on master.

templates/advantage/table/_manage-monthly.html
```diff
@@ -45,13 +45,15 @@ <h5 class="u-no-padding--top" {% if not mobile %}style="margin-bottom: .5rem"{%
     {% if not mobile %}
       <hr />
     {% endif %}
+    <!-- It conflicts with the total contract expiry date, if yearly purchase exists
```
anthonydillon, line 48: I thought we were only hiding it when its a month contract not entirely?

requirements.txt
albertkol: Weird. No idea how did made it here. I will rebase again.

templates/advantage/table/_manage-monthly.html
albertkol: The file I am changing is: `templates/advantage/table/_manage-monthly.html`

Everything I change in that file affects only monthly subscriptions. However yearly subscriptions do not have this notification, so technically I am hiding it for all subscriptions. 

Once the /advantage redesign kicks in, we can reconsider adding this back.

ClementChaumel: Do we want to just remove the code instead of comment it out? We can always look in git history if we need to reintroduce it

albertkol: @anthonydillon show I remove it? I had a feeling that if we remove it we would forget about it forever. So I chose to just comment it out.

anthonydillon: > The file I am changing is: templates/advantage/table/_manage-monthly.html

Ah thanks, makes sense.

> @anthonydillon show I remove it? I had a feeling that if we remove it we would forget about it forever. So I chose to just comment it out.

I would prefer to remove it as random commented out code quickly becomes confusing and is rarely useful when reimplementing.

anthonydillon (approved):
Ok once commented out code is removed.

Issue #4248, opened 2018-10-18
German takeover - OpenStack Made Easy


pmahnke:
This should go live with the German Webinar

anasereijo:
duplicate epic

PR #10066, opened 2021-07-22
[Certified improvements] - add functionality for expandable section

bethcollins92:
## Done

- Add option for accordion to have multiple tabs open if the `data-multiple-expanded` attribute is set to true
- NOTE: There will be functionality added in a future PR to only see 5 filters per group, and click "See all" to expand each group. As this is not in the scope of this PR, I have left the "Vendors" group closed on page load as the list of vendors is very long and users may miss the "update" button. In the PR where the "See all" functionality is added I will ensure all 3 groups are expanded on page load. 

## QA

- View page at: https://ubuntu-com-10066.demos.haus/certified?q=dell
- "Category" and "Certified Ubuntu release" groups should be expanded on page load, and "Vendors" should be collapsed. 
-  Click on the titles of each group and check they open and close as expected

## Issue / Card

Fixes [#4248](https://github.com/canonical-web-and-design/web-squad/issues/4248)

## Screenshots

Screenshot 2021-07-22 at 16 40 01


static/sass/_pattern_expandable-section.scss
```diff
@@ -0,0 +1,21 @@
+@mixin ubuntu-p-expandable-section {
```
sowasred2012, line 1: Not sure I understand this - it looks like it's aliasing existing Vanilla `p-accordion` classes to `p-expandable-section` classes, without adding anything custom. I think it might be that this class is being used so it can be referenced in the [js above](#diff-4877968da7344af822e0e9e508c5634c0572082f0319df7e7cf014d131855b13), without interfering with the existing accordion script, is that right?

bethcollins92: Yeah that's right. The styling needs to be the same as the `p-accordion` but the functionality different. As accordions aren't built to have more than one section open at a time, (which is what we need) I needed to write the javascript for this

bethcollins92: @sowasred2012 I've made the suggestions changes - not sure on the data attribute name but can't think of anything to explain it better if you have any suggestions!

templates/certified/search-results.html
```diff
@@ -55,9 +55,9 @@ <h1 class="u-sv3">Ubuntu certified
                </li>
                <li class="p-accordion__group">
                  <div role="heading" aria-level="2" class="p-accordion__heading">
-                   <button type="button" class="p-accordion__tab" id="tab2" aria-controls="tab2-section" aria-expanded="false">Certified Ubuntu release</button>
+                   <button type="button" class="p-accordion__tab" id="tab2" aria-controls="tab2-section" aria-expanded="true">Certified Ubuntu release</button>
                  </div>
-                 <section class="p-accordion__panel" id="tab2-section" aria-hidden="true" style="padding-left: 1rem;" aria-labelledby="tab2">
+                 <section class="p-accordion__panel" id="tab2-section" aria-hidden="false" style="padding-left: 1rem;" aria-labelledby="tab2">
                    {% for release_filter in release_filters %}
```
carkod, line 61: Can you we not have this `padding-left` in the `.scss`?

templates/certified/search-results.html
```diff
@@ -55,9 +55,9 @@ <h1 class="u-sv3">Ubuntu certified
                </li>
                <li class="p-accordion__group">
                  <div role="heading" aria-level="2" class="p-accordion__heading">
-                   <button type="button" class="p-accordion__tab" id="tab2" aria-controls="tab2-section" aria-expanded="false">Certified Ubuntu release</button>
+                   <button type="button" class="p-accordion__tab" id="tab2" aria-controls="tab2-section" aria-expanded="true">Certified Ubuntu release</button>
                  </div>
-                 <section class="p-accordion__panel" id="tab2-section" aria-hidden="true" style="padding-left: 1rem;" aria-labelledby="tab2">
+                 <section class="p-accordion__panel" id="tab2-section" aria-hidden="false" style="padding-left: 1rem;" aria-labelledby="tab2">
```
carkod, line 60: Can you we not have this padding-left in the .scss?

sowasred2012: I see that this is done in a few places throughout u.c - I think it's OK to have this small amount of inline style for nudges like this, however, because it's a repeated pattern, I also think it'd be worth suggesting an amendment to the accordion pattern in Vanilla, to introduce something like an `is-dense` modifier that reduces the padding the way this is doing. It might not get accepted, but we should have the conversation.

bethcollins92: Ok sure, I'll make an issue to suggest the amendment

sowasred2012 (approved):
LGTM!

PR #10075, opened 2021-07-27
Build typescript

huwshimi:
## Done

- Add TypeScript config according to the [esbuild docs](https://esbuild.github.io/content-types/#typescript).
- Emit errors when building TypeScript files.

## QA

- Check that the site builds and runs as it did.
- If you wish you could add a .ts or .tsx file to build.js and check that it builds (or emits errors) when running the site.

## Issue / Card

Fixes canonical-web-and-design/commercial-squad#83.


huwshimi:
@ClementChaumel esbuild already compiles TypeScript files, so this just PR adds errors checking and sets up the tsconfig.

ClementChaumel (approved):
Awesome!

PR #10077, opened 2021-07-27
Fix `buyingFor` radio being ignored if an organisation name is entered

ClementChaumel:
## Done

- Check buyingFor value instead of the presence of an organisation name to decide what to pass to the backend.

## QA

- Go to https://ubuntu-com-10077.demos.haus/advantage/subscribe?test_backend=true as a guest
- select a product
- open the modal
- enter an organisation name then click Myself radio
- Check that the organisation name is ignored and the customer name is used instead as the account name

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/81


albertkol (approved):
:+1:

PR #10089, opened 2021-07-28
Set up the Advantage page to mount a React app

huwshimi:
## Done

- Set up the Advantage page to mount a React app.

## QA

- Log in.
- You should see "Your subscriptions".
- When the React app has loaded it should say "Content" under the title.

## Issue / Card

Fixes canonical-web-and-design/commercial-squad#85.


tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "isolatedModules": true,
-    "jsx": "react-jsx",
+    "jsx": "react",
```
huwshimi, line 10: It seems `esbuild` does not support the new JSX transform yet so I've reverted this.

ClementChaumel: What was the issue?

ClementChaumel (approved):
:+1:

tsconfig.json
huwshimi: It seems that esbuild has to add support for the React 17 JSX transform so I guess it just wasn't going to build with `"react-jsx"` (there were conflicting errors about needing React to be imported/not imported).

More details in the issue: https://github.com/evanw/esbuild/issues/334.

PR #4276, opened 2018-10-24
Responsive takeover and k8 design complete as per design spec

inuwa:
## Done

Created the k8 german takeover for K8 and added responsiveness.
## QA

- Check out this feature branch
- Run the site using the command `./run serve`
- View the site locally in your web browser at: [http://0.0.0.0:8001/](http://0.0.0.0:8001/)
- Run through the following [QA steps](https://github.com/canonical-webteam/practices/blob/master/workflow/qa-steps.md)
- Change browser to german refresh until K8 German takeover shows, check that it match the (design)[https://app.zenhub.com/workspace/o/canonical-websites/www.ubuntu.com/issues/4231]

## Issue / Card

Fixes https://github.com/canonical-websites/www.ubuntu.com/issues/4232

## Screenshots
screenshot 2018-10-24 at 11 33 04
screenshot 2018-10-24 at 11 32 43

PR #10092, opened 2021-07-28
[Certified improvements] - add pagination select 

bethcollins92:
## Done

- Update page based on [design](https://app.zeplin.io/project/60c1ca887b99f18a380e57e9/screen/60f169f4009bd51281d0a3ff)
- Add functionality to be able to change the number of results per page

### Design notes @anasereijo : 
- The width of the pagination dropdown is wider than design as when 100 was selected it was cutting off the 0 (I made it as small as possible).
- On the table, I couldn't have the "CATEGORY" heading touching the edge of the table without right aligning it (which we can't do with the content of that column as they aren't numbers), as when the screen size changes slightly it causes the ellipses. I've made it work for the main screen sizes, let me know what you think. 

## QA

- View page at: https://ubuntu-com-10092.demos.haus/certified?q=dell
- Check page against [design](https://app.zeplin.io/project/60c1ca887b99f18a380e57e9/screen/60f169f4009bd51281d0a3ff)
- Change both results per page dropdowns (above and below table) and check page updates as expected.

## Issue / Card

Fixes [#4276](https://github.com/canonical-web-and-design/web-squad/issues/4276)

## Screenshots

Screenshot 2021-07-28 at 12 14 32

### DIFF
--- a/templates/certified/search-results.html
+++ b/templates/certified/search-results.html
@@ -6,7 +6,7 @@
 
 {% block content %}
 
-<form class="form">
+<form name="searchResults">
  <section class="p-strip--suru-topped">
    <div class="u-fixed-width">
      <h1 class="u-sv3">Ubuntu certified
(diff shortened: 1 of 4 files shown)


bethcollins92:
@carkod this test is failing: 
```    def test_search_results(self):
        response = self.client.get(
            "/certified?q=xps&category=Laptop&vendor=Dell"
        )
        self.assertEqual(response.status_code, 200)```
I've checked that url path locally and on the demo and they both give a 200 so not sure why it's failing

**Comment by anasereijo** [comment]
this is looking great, thanks!

**Comment by davegoddard42** [comment]
Looks great, thanks

File: templates/certified/search-results.html [hunk 1]
```diff
@@ -159,6 +201,21 @@ <h3>Why not try widening your search?</h3>
 </form>
 
 <script>
+  //function to ensure only the option which has been changed is appended to the URL
+  function updateResultsPerPage() {
+    let resultsDropdowns = document.querySelectorAll(".p-results-per-page")
+    resultsDropdowns.forEach(resultsDropdown => {
```
carkod, line 207: I don't think you need 2 loops for this. If you use the `event.target` you may be able to grab the specific `p-results-per-page` HTML element and do the second loop only. Pseudo-code:
In the handler:
```
... onchange=updateResultsPerPage(event) ...
```
```
function (event) {
  let resultsDropdown = event.target;
  const options = resultsDropdown.querySelectorAll("option");
  ....
}
```

bethcollins92: This doesn't work for this case as within that function I need to be able to access the `resultsDropdown` that has been clicked and also the `resultsDropdown` that hasn't been clicked. The one that hasn't been clicked has to be prevented from being appended to the URL (otherwise 2 x `&limit=...` are added), so therefore I set the `name` attr of the `<select>` which isn't changed to `""` to prevent it being appended when the form is submitted. If you can think of another way to avoid both being appended I open to ideas :)

carkod:
> @carkod this test is failing in `text_certification.py`:
> 
> ```
>  def test_search_results(self):
>         response = self.client.get(
>             "/certified?q=xps&category=Laptop&vendor=Dell"
>         )
>         self.assertEqual(response.status_code, 200)
> ```
> 
> I've checked that url path locally and on the demo and they both give a 200 so not sure why it's failing

Because you have added `limit` argument to the API now, you have to regenerate the cassette. So can you delete the search results cassette (/tests/cassettes/TestCertification.test_search_results.yaml) and do `dotrun test-python` please?

templates/certified/search-results.html
carkod: Ok, this sounds like you have to add the `event.preventDefault();` after the `let resultsDropdown = event.target;` Since you are using the same function in both places, it should prevent it from triggering by default on both and then it will only submit when you do it explicitely.

webapp/certified/views.py
```diff
@@ -331,6 +331,7 @@ def certified_home():
             vendor=vendors,
             query=query,
             offset=offset,
+            limit=limit,
```
carkod, line 334: FYI. This is causing the python test to fail, because now the data source the code is asking for is different, so need to regenerate the cassette.

templates/certified/search-results.html
bethcollins92: This also doesn't work - I think it's because when the form is submitted it checks for which option is selected and appends it to the URL, and because both drop downs have an option selected it adds them both. So even if you prevent default, when the other `resultsDropdown ` is submitted it also submits the selected option from the `resultsDropdown` which isn't interacted with

webapp/certified/views.py
bethcollins92: Thank you! fixed it

carkod (approved):

PR #10096, opened 2021-07-29
Get subscriptions data

huwshimi:
## Done

- Add subscription data to the window.
- Fetch data from the window and insert into the react query store.

## QA

- [Turn on dev mode](https://github.com/canonical-web-and-design/ubuntu.com/blob/master/HACKING.md#javascript).
- Click on the devtools.
- You should see some data that has been loaded.

## Issue / Card

Fixes canonical-web-and-design/commercial-squad#100.


package.json
```diff
@@ -56,6 +57,7 @@
     "postcss-cli": "8.3.1",
     "prettier": "2.2.1",
     "prop-types": "15.7.2",
+    "react-test-renderer": "17.0.2",
```
huwshimi, line 60: This dep is required by the hooks testing library. See: https://react-query.tanstack.com/guides/testing.

ClementChaumel (approved):
LGTM, just a suggestion: if you have time, consider merging the dependency changes on master and then rebase ua-subscriptions on top of it to avoid future conflicts. (and so I can use it for my tests!)

If you're too busy don't worry I'll do it!

huwshimi:
> LGTM, just a suggestion: if you have time, consider merging the dependency changes on master and then rebase ua-subscriptions on top of it to avoid future conflicts. (and so I can use it for my tests!)
> 
> If you're too busy don't worry I'll do it!

No problems, I'll put up a PR.

PR #10104, opened 2021-07-30
Add API types

huwshimi:
## Done

- Move the contracts api files to be in the same dir.
- Add generated types from the contracts service.
- Add types for the advantage page.

## QA

N/A

## Issue / Card

Fixes canonical-web-and-design/commercial-squad#128.


static/js/src/advantage/api/contracts-types.ts
```diff
@@ -0,0 +1,1694 @@
+/**
+ * This file was largely auto-generated by openapi-typescript.
+ * It may be easiest to regenerated this file from the contracts API definitions
+ * rather than updating this file directly.
+ * This file should remain consistent with the contracts API, any types that
+ * need to be augmented should be done in types.ts.
+ */
```
huwshimi, line 7: @ClementChaumel these types will probably help you when/if you start migrating your components.

My approach is to define these types exactly as they come from the contracts API and then in the types.ts file I'm defining types that include any additional data that gets added in the view.py.

albertkol (approved):
Looking at these types makes me realise how much unnecessary stuff the front-end receives. 

The back-end will be worked on soon. After that we should maybe sit together at some point and filter out the unnecessary stuff.

huwshimi:
> Looking at these types makes me realise how much unnecessary stuff the front-end receives.
> 
> The back-end will be worked on soon. After that we should maybe sit together at some point and filter out the unnecessary stuff.

Yeah good idea, I don't think all those types will be used, I just imported them all in case Clem needed other parts of the API.

PR #10106, opened 2021-07-30
Update payment methods page

albertkol:
## Done

With the release of `/account/invoices`, I took the time to look at `/account/payment-methods` and bring them in line with each other. `/payment-methods` was originally intended to be part of the `/advantage` domain. Now it's part of the `/account` pages, making it a higher level page, independent of product names.

**At the moment this page allows only editing existing card information. It does not allow for adding a card, if you do not already have one set.**

### Front-end/Visual changes:
1) Updating the meta title and description.
2) Removing the Ubuntu Advantage title
image
3) Removing the "You are signed in as". That was forgotten there. It's been remove already from all other pages.
4) Making the Payment methods title an h1. Similar to the title on /invoices.
5) Mini fix CSS to bring the Update button in line with the card input field
6) Remove front-end `user_info` checks. Back-end forbids access to `/payment-methods` for guest users.

### Back-end changes:
1) Fixing errors with people with Stripe accounts or Stripe account customer info reach the page.
- If a user with no Stripe account reaches `/payment-methods`. They get the 500 page.
- If a user with a Stripe account, but no Stripe customer info reaches `/payment-methods` page they will get see page, but will get a 500 if they try to provide a card. That is because the API expects name, address info to be submitted together with the card info.
Instead of a 500 we show this:
image

2) Update the POST `/advantage/payment-method` link to `/account/payment-methods`

## QA

1. Accessing /account/payment-methods without stripe account
- Have a Ubuntu One account without a stripe account; You can achieve this the easiest by creating a new UO account and log in with it then go to `/account/payment-methods`
- Log in with your UO account
- Go to https://ubuntu-com-10106.demos.haus/account/payment-methods?test_backend=true
- You will see the page showing you no cards set

2. Accessing /account/payment-methods without default payment method:
- Have an account attached to your UO user but default payment method set yet; You should be able to reproduce if you use your main canonical account on production. You will have the company UA account attached, but no payment methods attached to it. 
- Log in with your UO account
- Go to https://ubuntu-com-10106.demos.haus/account/payment-methods?test_backend=true
- You will see the page showing you no cards set

3. Accessing the /account/payment-methods with everything you need:
- Have a stripe account
- Log in with your UO account
- Go to https://ubuntu-com-10106.demos.haus/account/payment-methods?test_backend=true
- You will see your card
- Update your card
- It should work

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/123

### DIFF
--- a/templates/account/payment-methods/index.html
+++ b/templates/account/payment-methods/index.html
@@ -77,7 +77,4 @@
 </script>
 <script src="{{ versioned_static('js/dist/ua-payment-methods.js') }}" type="module" defer></script>
 
-
-{% with first_item="_ua_got_questions", second_item="_ua_legal", third_item="_ua_further_reading" %}{% include "shared/contextual_footers/_contextual_footer.html" %}{% endwith %}
-
 {% endblock content %}
--- a/templates/account/payment-methods/index.html
+++ b/templates/account/payment-methods/index.html
@@ -9,7 +9,7 @@
 
 {% block content %}
 
-<section class="p-strip--suru-topped card-management-section">
+<section class="p-strip--suru-topped p-strip-account-page card-management-section">
   <div class="row">
     <div class="col-12">
 
--- a/templates/account/payment-methods/index.html
+++ b/templates/account/payment-methods/index.html
@@ -1,71 +1,53 @@
 {% extends "advantage/base_advantage.html" %}
 
-{% block title %}Ubuntu Advantage for Infrastructure{% endblock %}
-{% block meta_description %}Ubuntu Advantage for Infrastructure offers a single, per-node packaging of the most comprehensive software, security and IaaS support in the industry, with OpenStack support, Kubernetes support included, and Livepatch, Landscape and Extended Security Maintenance to address security and compliance concerns.{% endblock %}
-{% block meta_copydoc %}https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.nleu1cdig11l{% endblock meta_copydoc %}
+{% block title %}Canonical payment method{% endblock %}
+{% block meta_description %}Mange your payment method for your Canonical services.{% endblock %}
+
+{% block head_extra %}
+<meta name="robots" content="noindex, nofollow">
+{% endblock %}
 
 {% block content %}
 
-{% if user_info %}
-  <section class="p-strip--suru-topped">
-    <div class="row">
-      <div class="col-12">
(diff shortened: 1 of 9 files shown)


pmahnke:
@albertkol could you explain more about what you did in this PR in the Done section, it is not a very good explaination of the work.

albertkol:
@anthonydillon could you please review the meta info for the `/payment-methods` page. I copy-pasted the text you suggested for the `/invoices` page, and tweaked it a bit.

albertkol:
@elioqoshi could you please review the UX and Design. 

This page at this point is a sister page of the `/invoices` that you reviewed last week. So the visuals should be consistent. You can read the visual changes I made to this page above. Also I give explanations of the UX for users without stripe account. These can be temporary until we properly address them later. 

One thing I want you to notice is the white space. For `/invoices` you mentioned that the page needs a `min-height`. This page has that set to 600 I think. For `/invoices` we went with just increasing the padding top and bottom a bit. Whichever solution we go with should be applied to both. Let me know which one you prefer.

codedfile (approved):

codedfile (approved):

codedfile (approved):
#albertkol:fix-payment-methods-page

templates/account/payment-methods/index.html
```diff
@@ -1,71 +1,53 @@
 {% extends "advantage/base_advantage.html" %}
 
-{% block title %}Ubuntu Advantage for Infrastructure{% endblock %}
-{% block meta_description %}Ubuntu Advantage for Infrastructure offers a single, per-node packaging of the most comprehensive software, security and IaaS support in the industry, with OpenStack support, Kubernetes support included, and Livepatch, Landscape and Extended Security Maintenance to address security and compliance concerns.{% endblock %}
-{% block meta_copydoc %}https://docs.google.com/document/d/1K65n3NpZCh7EWDXKLhG31seGJ2jKbW6lqw3BwV2mlnU/edit#heading=h.nleu1cdig11l{% endblock meta_copydoc %}
+{% block title %}Canonical payment method{% endblock %}
+{% block meta_description %}Mange your payment method for your Canonical services.{% endblock %}
```
anthonydillon, line 4: Reads well to me, thanks :+1:

ClementChaumel (approved):

elioqoshi:
Works as expected. Thanks @albertkol 

Small thing only, when updating a card, the notification tells you that the page is reloading. I think that's unnecessary:
image

Could the notification be displayed after the page has been reloaded, only once, saying "Card has been updated?" If the user would navigate away from it, it would not be displayed again, unless they would update the card again. Does that make sense?

+1 as this is not a big thing, though would be good to update it.

PR #10107, opened 2021-07-30
[Certified improvements] - Add clear filters button

bethcollins92:
## Done

- Add placeholders to search bar
- Add clear filters button which reloads page, clears all filters and fetches all results

## QA

- View page at:  https://ubuntu-com-10107.demos.haus/certified?q=
- Select some filters and click "Apply filters" 
- Select the "Clear filters button"
- See there are no filters applied and the page is displaying all results

## Issue / Card

Fixes [#4253](https://github.com/canonical-web-and-design/web-squad/issues/4253)

### DIFF
--- a/templates/certified/search-results.html
+++ b/templates/certified/search-results.html
@@ -21,7 +21,7 @@
    </div>
    <div class="u-fixed-width">
      <div class="p-search-box">
-       <input class="p-search-box__input" type="text" name="q" value="{{ query or '' }}">
+       <input class="p-search-box__input" type="text" name="q" value="{{ query or '' }}" placeholder="Search hardware">
        <button type="submit" class="p-search-box__button"><i class="p-icon--search"></i></button>
      </div>
    </div>
@@ -135,6 +135,7 @@
              </ul>
            </aside>
            <button href="#" class="p-button--positive p-update-results" id="apply-filters" type=submit disabled>Apply filters</button>
+           <button class="p-button--neutral p-update-results" onclick="return clearFilters()">Clear filters</button>
          </div>
        </nav>
      </div>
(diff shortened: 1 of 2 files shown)


davegoddard42:
@bethcollins92 - The title of the page (H1) changes when things are selected - this shouldn't change but I'm aware that might not be the scope of this ticket (clearing filters). Same with the selection of items and collapsing the accordion - it should show a little chip highlighting how many things are selected. 

Please ignore if this is not relevant to the ticket you're doing here! The filter stuff works as expected.

bethcollins92:
Thanks @davegoddard42 ! I'm tackling each small task at a time but if you think clearing the filters make the title change too obvious/an issue I can fix it in this one (was going to remove that logic once we built the specific /certified/<category> pages). With regards to the little chip - that's an issue within itself as it will be quite a bit of work

davegoddard42:
Sorry @bethcollins92 , still learning how we break things down. Let me give it a UX+1, tackle the other bits where you think it's best.

bethcollins92:
@davegoddard42 No probs at all! Thanks for the feedback

templates/certified/search-results.html
```diff
@@ -202,6 +203,10 @@ <h3 class="p-heading--4">No results - why not try widening your search?</h3>
 
 <script>
 
+ function clearFilters() {
+  location.assign(`${location.protocol}//${location.host}/certified?q=`)
```
carkod, line 207: Do you really need `location.protocol and location.host?` does it not work with `/`?

bethcollins92: it works with `/certified?q=` - updated, thanks!

carkod (approved):

PR #10122, opened 2021-08-04
Add header buttons

huwshimi:
## Done

- Add header buttons.
- Make header responsive.

## QA

- Clicking the buttons should take you to the appropriate place.
- Resize your screen down to the small size and the buttons should drop below the title and become stacked.

## Issue / Card

Fixes: canonical-web-and-design/commercial-squad#88.
Fixes: canonical-web-and-design/commercial-squad#89.

## Screenshots

Screen Shot 2021-08-04 at 1 36 12 pm
Screen Shot 2021-08-04 at 1 36 24 pm


static/sass/_pattern_subscribe.scss
```diff
@@ -2,6 +2,41 @@
...
+
+    @media only screen and (max-width: $breakpoint-small) {
+      // At small screens make the buttons closer to the title.
+      padding-top: $spv-outer--medium;
+
+      [class^="p-button"] {
+        // At small screens make the buttons fill the screen.
+        width: 100%;
+      }
+    }
```
ClementChaumel, line 30: This already happens for `$breakpoint-x-small` is that a requirement that they should do that for `small` as well?

templates/advantage/index.html
```diff
@@ -19,8 +19,13 @@
...
     <div class="row">
-      <div class="col-12">
-        <h1>Your subscriptions</h1>
+      <div class="col-12 p-subscriptions__header">
+        <h1 class="u-no-margin--bottom">Your subscriptions</h1>
+        <div class="p-subscriptions__header-buttons">
+          <a class="p-button--positive" href="/advantage/subscribe{% if get_test_backend %}?test_backend=true{% endif %}">Buy new subscription</a>
+          <a class="p-button--neutral" href="/account/invoices{% if get_test_backend %}?test_backend=true{% endif %}">Invoices</a>
+          <a class="p-button--neutral" href="/account/payment-methods{% if get_test_backend %}?test_backend=true{% endif %}">Payment methods</a>
+        </div>
```
ClementChaumel, line 28: Any reason why the header isn't part of the React app?
Not sure if there would be any drawbacks (except having the header load faster) but you'd be able to use your hook for urls

ClementChaumel (commented):
Also, small nitpick about `_pattern_subscribe.scss` we tend to use `subscribe` to describe anything related to the product selector at /advantage/subscribe not the /advantage page itself

static/sass/_pattern_subscribe.scss
huwshimi: Yeah OK, I'll remove that and try and get a design review at some point for the in-between screen sizes.

huwshimi:
> Few things to consider instead of creating new custom classes:
> 
> * Could you use the grid to align elements left and right and use alignment classes to display the buttons next to eachother?
> * if you use, `u-equal-height` and `u-vertically-center` you can align elements horizontally.
> 
> Using these Vanilla patterns could reduce the custom local styling and help them out comes to updates as everyone is familiar with this usage.

Thanks, I'll give that a go. I was trying to make these better at the in-between screen size, but maybe I'll request a design review in the future to go over the middle screen sizes.

templates/advantage/index.html
huwshimi: I thought as the header is static I'd leave it in the template so that (like you say) it'll be visible while the content component loads.

If you think it'd be better to have it in React then I'll follow up with another PR to move it over (just so this PR doesn't get too large).

PR #10125, opened 2021-08-04
Add count to closed accordion

bethcollins92:
## Done

- Add counter to accordion tab when the accordion is closed
- Move javascript functionality into it's own file

## QA

- View page at: https://ubuntu-com-10125.demos.haus/certified?q=
- Select filters on the LHS and close the accordion - check the chip is visible on each section with the correct number of filters applied. 
- Check when there are no filters applied the chip isn't visible. 

## Issue / Card

Fixes [#4250](https://github.com/canonical-web-and-design/web-squad/issues/4250)

## Screenshots

Screenshot 2021-08-04 at 11 20 40

### DIFF
--- a/build.js
+++ b/build.js
@@ -21,6 +21,7 @@ let entries = {
   openstackChart: "./static/js/src/openstack-chart.js",
   uaSubscribe: "./static/js/src/advantage/subscribe/react/app.jsx",
   "cloud-price-slider": "./static/js/src/cloud-price-slider.js",
+  "certification-results": "./static/js/src/certification-results.js",
 };
 
 for (const [key, value] of Object.entries(entries)) {
--- a/build.js
+++ b/build.js
@@ -21,7 +21,7 @@ let entries = {
   openstackChart: "./static/js/src/openstack-chart.js",
   uaSubscribe: "./static/js/src/advantage/subscribe/react/app.jsx",
   "cloud-price-slider": "./static/js/src/cloud-price-slider.js",
-  "certification-results": "./static/js/src/certification-results.js",
+  "certified-search-results": "./static/js/src/certified-search-results.js",
 };
 
 for (const [key, value] of Object.entries(entries)) {
--- a/static/js/src/certification-results.js
+++ b/static/js/src/certification-results.js
@@ -1,0 +1,133 @@
+function enableApplyFilters() {
+  const filtersSelected = [];
+  const filters = document.querySelectorAll(".js-enable-apply-filters");
+  filters.forEach((filter) => {
+    filter.addEventListener("change", () => {
+      if (!filtersSelected.includes(filter)) {
+        filtersSelected.push(filter);
+      } else {
+        const index = filtersSelected.indexOf(filter);
+        filtersSelected.splice(index, 1);
+      }
+      const button = document.querySelector(".js-apply-filters");
+      if (filtersSelected.length) {
+        button.removeAttribute("disabled");
+      } else {
+        button.disabled = true;
+      }
+    });
+  });
+
+  function displayFilterCount() {
+    const accordion = document.querySelector(".js-filter-count");
+    const accordionTabs = accordion.querySelectorAll(".p-accordion__tab");
+    accordionTabs.forEach((tab) => {
+      tab.addEventListener("click", () => {
+        let categoryFilterCount = 0;
+        let vendorFilterCount = 0;
+        let releaseFilterCount = 0;
+        filtersSelected.forEach((filter) => {
+          if (filter.id === "category-filter") {
+            categoryFilterCount++;
+          } else if (filter.id === "vendor-filter") {
+            vendorFilterCount++;
+          } else if (filter.id === "release-filter") {
+            releaseFilterCount++;
+          }
+        });
+        const chip = tab.querySelector(".p-round-chip");
+        chip.classList.toggle("u-hide");
+
+        const categoryChipValue = tab.querySelector("#category-count");
+        if (categoryChipValue) {
+          categoryChipValue.innerHTML = categoryFilterCount;
+        }
+        const vendorChipValue = tab.querySelector("#vendor-count");
+        if (vendorChipValue) {
+          vendorChipValue.innerHTML = vendorFilterCount;
+        }
+        const releaseChipValue = tab.querySelector("#release-count");
+        if (releaseChipValue) {
+          releaseChipValue.innerHTML = releaseFilterCount;
+        }
+
+        // Add line break in title to keep chip on same line
+        const header = tab.querySelector(".p-certification-header");
+        if (header.innerHTML.includes("Certified")) {
+          if (!chip.classList.contains("u-hide")) {
+            header.innerHTML = "Certified Ubuntu <br> release";
+          } else {
+            header.innerHTML = "Certified Ubuntu release";
+          }
+        }
+      });
+    });
+  }
+  displayFilterCount();
+}
+
+enableApplyFilters();
+
+//function to ensure only the option which has been changed is appended to the URL
+function updateResultsPerPage() {
+  let resultsDropdowns = document.querySelectorAll(".p-results-per-page");
+  resultsDropdowns.forEach((resultsDropdown) => {
+    const options = resultsDropdown.querySelectorAll("option");
+    options.forEach((option) => {
+      if (option.selected) {
+        if (option.value === resultsDropdown.dataset.limit) {
+          resultsDropdown.name = "";
+        }
+        searchResults.submit();
+      }
+    });
+  });
+}
+
+function toggleVendorsList() {
+  const vendorSection = document.querySelector("#vendor-section");
+  const vendorPanel = vendorSection.querySelector(".p-accordion__panel");
+  const toggleVendorButtons = document.querySelectorAll(".js-reveal-vendors");
+  toggleVendorButtons.forEach((button) => {
+    button.addEventListener("click", () => {
+      window.scrollTo(0, 360);
+      vendorPanel.classList.toggle("is-collapsed__vendors");
+      toggleVendorButtons.forEach((button) => {
+        button.classList.toggle("u-hide");
+      });
+    });
+  });
+}
+toggleVendorsList();
+
+function toggleVersionsList() {
+  const versionSection = document.querySelector("#version-section");
+  const versionPanel = versionSection.querySelector(".p-accordion__panel");
+  const toggleVersionButtons = document.querySelectorAll(".js-reveal-versions");
+  toggleVersionButtons.forEach((button) => {
+    button.addEventListener("click", () => {
+      versionPanel.classList.toggle("is-collapsed__versions");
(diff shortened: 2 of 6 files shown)


build.js
```diff
@@ -21,6 +21,7 @@ let entries = {
   openstackChart: "./static/js/src/openstack-chart.js",
   uaSubscribe: "./static/js/src/advantage/subscribe/react/app.jsx",
   "cloud-price-slider": "./static/js/src/cloud-price-slider.js",
+  "certification-results": "./static/js/src/certification-results.js",
```
carkod, line 24: shouldn't this be named maybe `certified-search-results` or similar?

static/js/src/certification-results.js
```diff
@@ -0,0 +1,142 @@
...
+    const tab = filterSection.querySelector(".p-accordion__tab");
+    tab.addEventListener("click", () => {
+      const links = filterSection.querySelector(".js-toggle-links");
+      if (links) {
+        links.classList.toggle("u-hide");
+      }
+    });
+  });
+}
+toggleShowAllLinks();
```
carkod, line 142: I think maybe you should stack all this functions calls together?, so we can clearly see at one glance which functions are being executed in this script.
```
...
toggleVendorsList();
toggleShowAllLinks();
toggleVersionsList();
```

templates/certified/search-results.html
```diff
@@ -127,13 +134,13 @@ <h2 class="p-heading--4">
...
-                  <p class="p-reveal-versions is-dense"><a> Show all versions ({{release_filters|length - 5}})</a></p>
-                  <p class="p-reveal-versions is-dense u-hide"><a>Show fewer</a></p>
+                  <p class="js-reveal-versions is-dense"><a> Show all versions ({{release_filters|length - 5}})</a></p>
+                  <p class="js-reveal-versions is-dense u-hide"><a>Show fewer</a></p>
                 </span>
                </li>
              </ul>
            </aside>
-           <button href="#" class="p-button--positive p-update-results" id="apply-filters" type=submit disabled>Apply filters</button>
+           <button href="#" class="p-button--positive p-update-results js-apply-filters" type=submit disabled>Apply filters</button>
```
carkod, line 143: Buttons [don't have `href` property](https://www.w3.org/TR/WCAG20-TECHS/H91.html#H91_mappings)

templates/certified/search-results.html
```diff
@@ -200,133 +207,16 @@ <h3 class="p-heading--4">No results - why not try widening your search?</h3>
  </section>
 </form>
 
+
```
carkod, line 210: Is this space necessary?

templates/certified/search-results.html
```diff
@@ -200,133 +207,16 @@ <h3 class="p-heading--4">No results - why not try widening your search?</h3>
...
 <script>
 
  function clearFilters() {
   location.assign(`/certified?q=`)
    return false
  }
```
carkod, line 216: Any reason you are not moving this to the script?

templates/certified/search-results.html
```diff
@@ -200,133 +207,16 @@ <h3 class="p-heading--4">No results - why not try widening your search?</h3>
...
-
-  sideNavigations.forEach(setupSideNavigation);
-}
-
-setupSideNavigations('.p-side-navigation, [class*="p-side-navigation--"]');
-
 </script>
 
+<script src="{{ versioned_static('js/src/certification-results.js') }}" type="module" defer></script>
+
```
carkod, line 221: I don't think you need `type="module"` unless you are using the Common JS `import module` syntax

carkod:
@spencerbygraves any design comments? so that Beth can address them all together

carkod:
# QA
Not too important but the outline looks cut off. If you can easily fix it, go ahead, if not I think it's fine to leave it
Screenshot 2021-08-04 at 12 33 04

bethcollins92:
Thanks @carkod, made those changes. I've made an [issue](https://github.com/canonical-web-and-design/ubuntu.com/issues/10129) for the checkbox highlight issue. It seems to only happen on Macs

static/js/src/certified-search-results.js
```diff
@@ -0,0 +1,147 @@
...
+      if (filtersSelected.length) {
+        button.removeAttribute("disabled");
+      } else {
+        button.disabled = true;
+      }
+    });
+  });
+
+  function displayFilterCount() {
+    const accordion = document.querySelector(".js-filter-count");
```
carkod, line 27: Sorry, didn't see this before, but indentation seems a bit off here?

carkod (approved):

static/js/src/certified-search-results.js
bethcollins92: This function is within the `enableApplyFilters()` function as it needs to access the `filtersSelected` array

spencerbygraves:
Thanks @bethcollins92, a few things from me...

- On the counter, I think it would be better if it can be aligned with the first line of text if it wraps. Is that possible?
- Can we also change the way the hover works please. Currently when you hover on a closed section title, the grey colour is the same as the counter so it disappears (see the bottom image). On hover can we make the counter darker like below. @lyubomir-popov said he can help you achieve this if needed

image

I've noticed some other problems with the side nav for another issue.

@ziheliu214 has recently done a search and filter pattern which works differently to this, so this may need to change more significantly. cc @clagom 

- The text for section titles should be the same as on the [Vanilla side navigation](https://vanillaframework.io/docs/examples/patterns/side-navigation/default)
- The section titles have a hover state when closed but not when open
- The checkboxes and text are not aligned
- Can we remove the margin from the hrs, it is leaving a white gap when you hover, and makes the text look misaligned

image

- Remove the left margin from filter buttons

Thanks

lyubomir-popov:
@ziheliu214 @clagom any reason why this shouldn't use the search and filter / chips pattern in vanilla? 
Having a counter of applied filters in a collapsed accordion is much less clear than havign chips showing exactly which filters have been applied, being able to selectively remove filters by clicking on the chip [x] button, positioning a [clear] button next to the applied filters that you can see, etc etc

ziheliu214:
> @ziheliu214 @clagom any reason why this shouldn't use the search and filter / chips pattern in vanilla?
> Having a counter of applied filters in a collapsed accordion is much less clear than havign chips showing exactly which filters have been applied, being able to selectively remove filters by clicking on the chip [x] button, positioning a [clear] button next to the applied filters that you can see, etc etc

Hi, yes I think we can use Search and Filter here.
Here we have:
- 3 filtering categories: Category, Vendor, Release
- each category have a maximum 30 (? saw from the number on "show all vendors") filtering items 

So its all valid.

lyubomir-popov:
@clagom how should we proceed? keep fixing glitches in the existing implementation, or switch to the consistent vanilla search and filter pattern?

clagom:
@lyubomir-popov @ziheliu214 on Charmhub the filter on the left doesn't populate the search field either: https://charmhub.io/?base=all&filter=big-data
The numeric count is only visible when the accordion is closed. Adding a chip in the search field when users click on the check-filter on the left would suffice but would require much more complexity in terms of implementation I assume.
Also, if we implement chips, would we then have two ways of adding and removing filter? One with the checkbox and one within the search input? Or would just be there to remove the filter that is added from with check inputs only?
I would suggest to fix the issue first and work on the other points, what do you think?

ziheliu214:
> @lyubomir-popov @ziheliu214 on Charmhub the filter on the left doesn't populate the search field either: https://charmhub.io/?base=all&filter=big-data
> The numeric count is only visible when the accordion is closed. Adding a chip in the search field when users click on the check-filter on the left would suffice but would require much more complexity in terms of implementation I assume.
> Also, if we implement chips, would we then have two ways of adding and removing filter? One with the checkbox and one within the search input? Or would just be there to remove the filter that is added from with check inputs only?
> I would suggest to fix the issue first and work on the other points, what do you think?

In my point of view, Charmhub is a unique case that it can work without S&F - It only has 1 filtering category and 10 items within it. Indexing is not difficult and users can view the filters on the left intuitively. 

But for this issue, search and filter is the best option as it has 3 filtering categories and each list is long.

clagom:
I think we can explore as a follow up @ziheliu214, including @davegoddard42 who worked on the UX here. Would be a major change and I don't think we want to block this issue for a long time.

davegoddard42:
Sorry, just reading through now. I'm happy to explain further the design decisions I took to get to this solution, but @clagom and @ziheliu214  pretty much covered it. 

Is the charmhub page shared the only example of the S&F or should I look at others?

lyubomir-popov:
@clagom @ziheliu214 @davegoddard42 Should the default here be collapsed? Looking at the zeplin design the default seems to be to show all filters expanded, which would mitigate the problem of seeing a number but not the actual filters.

On the vanilla side, there's opportunity to improve / clarify things as well - when should you:

1) put the filters behind a click in the search field
2) put them directly on the page (this one is easy, when there's not sa lot of them and they aren't caegorized?) 
3) when should they be displayed in an accordion (which potentially also hides them depending on choice of default)

I'll work with @bethcollins92 to polish this version. There are quite a few things to fix (@spencerbygraves mentioned most above, I have a few of my own that I'll list separately), so definitely not a small task. A lot of it will be snowflake code as vanilla doesn't cover it). 

But it would be good if @ziheliu214 @davegoddard42 you could help clarify the above 3 points and update the documentation when you get a chance? Then we can add the polished version of this as a vanilla example too.

davegoddard42:
@lyubomir-popov the default state of the accordion is every section open, but with a 'show all [x]' link to expand out the entire section.

lyubomir-popov:
@ziheliu214 @davegoddard42 I've created a vanilla [issue](https://github.com/canonical-web-and-design/vanilla-framework/issues/3911) with the outstanding questions raised here.

bethcollins92:
Hi @spencerbygraves - I've addressed:
- changing the shade of the counter chip when hovering the accordion
- removing the margin from the bottom of the `<hr>`
- Moving the chip to the top line of the accordion in the certified releases filter title

I'll create a new issue now to address the other problems

spencerbygraves:
@bethcollins92 great, thanks very much for doing the updates 👍

lyubomir-popov:
@bethcollins92 a few small glithces, feel free to move to a separate issue if they are outside of the scope here:

1) The checkboxes sit a bit too high compared to their respective labels:
image

for comparison, in vainlla: 
image

2) What does the "Show all versions (2)" link do? when I click it it jkust jumps a little bit.

3) there's a lign between accordion headers and the contents of that section. I understand the desire to distinguishbetween header and contents, but a line going all the way makes it look like an unrelated section. I would propose a lign that starts where the text starts, which implies internal grouping within an accordion section, rather than a separate accordion section:
image

To be honest I'm not sure if it's even needed as the checkboxxes are not easily confused with the header anyway...

bethcollins92:
As this may be lost in the comments - here is the [issue](https://github.com/canonical-web-and-design/web-squad/issues/4315) to fix remaining problems

PR #10135, opened 2021-08-05
Append the test backend flag to URLs

huwshimi:
## Done

- Fetch and store the test backend flag.
- Append the test backend flag to URLs.

## QA

- In the payment method notification click on "update your payment methods".
- You should get taken to /account/payment-methods?test_backend=true
- Visit /advantage
- In the payment method notification click on "update your payment methods".
- You should get taken to /account/payment-methods

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/136

### DIFF
--- a/static/js/src/advantage/react/hooks/useUsingTestBackend.test.tsx
+++ b/static/js/src/advantage/react/hooks/useUsingTestBackend.test.tsx
@@ -1,0 +1,27 @@
+import React from "react";
+import { renderHook, WrapperComponent } from "@testing-library/react-hooks";
+import type { ReactNode } from "react";
+import { QueryClient, QueryClientProvider } from "react-query";
+import { useUsingTestBackend } from "./useUsingTestBackend";
+
+describe("useUsingTestBackend", () => {
+  let queryClient: QueryClient;
+  let wrapper: WrapperComponent<ReactNode>;
+
+  beforeEach(() => {
+    queryClient = new QueryClient();
+    wrapper = ({ children }) => (
+      <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+    );
+  });
+
+  it("can return the pending purchase id from the store", async () => {
+    queryClient.setQueryData("usingTestBackend", true);
+    const { result, waitForNextUpdate } = renderHook(
+      () => useUsingTestBackend(),
+      { wrapper }
+    );
+    await waitForNextUpdate();
+    expect(result.current.usingTestBackend).toBe(true);
+  });
+});
--- a/static/js/src/advantage/react/hooks/useUsingTestBackend.test.tsx
+++ b/static/js/src/advantage/react/hooks/useUsingTestBackend.test.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { PropsWithChildren } from "react";
 import { renderHook, WrapperComponent } from "@testing-library/react-hooks";
 import type { ReactNode } from "react";
 import { QueryClient, QueryClientProvider } from "react-query";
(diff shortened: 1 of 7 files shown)


static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.test.tsx
```diff
@@ -8,7 +8,7 @@ describe("SubscriptionCancel", () => {
   it("handles closing the modal", () => {
     const onClose = jest.fn();
     const wrapper = shallow(<SubscriptionCancel onClose={onClose} />);
-    wrapper.find(Modal).invoke("close")();
+    wrapper.find(Modal).invoke("close")!();
```
ClementChaumel, line 11: What is that syntax?

static/js/src/advantage/react/hooks/useUsingTestBackend.ts
```diff
@@ -0,0 +1,9 @@
+import type { UsingTestBackend } from "advantage/api/types";
+import { useQuery } from "react-query";
+
+export const useUsingTestBackend = () => {
+  const { data: usingTestBackend } = useQuery<UsingTestBackend>(
+    "usingTestBackend"
+  );
+  return { usingTestBackend };
```
ClementChaumel, line 8: How are you accessing the window object here?

static/js/src/advantage/react/hooks/useUsingTestBackend.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+  let wrapper: WrapperComponent<ReactNode>;
+
+  beforeEach(() => {
+    queryClient = new QueryClient();
+    wrapper = ({ children }) => (
+      <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+    );
+  });
+
+  it("can return the pending purchase id from the store", async () => {
```
ClementChaumel, line 18: The name of the test seem wrong

static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.test.tsx
huwshimi: This is a [non-null assertion](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). Because `close` is an optional prop we need to tell Typescript to call the function only if it exists (which it will in this case if we've passed the `onClose` function correctly).

static/js/src/advantage/react/hooks/useUsingTestBackend.ts
huwshimi: I'm only accessing the window object in [`useLoadWindowData`](https://github.com/canonical-web-and-design/ubuntu.com/pull/10135/files#diff-934076135df0b5c96a13328ac7f3355537dcfd4dc21aad9d7bede7b8e6377d21R17) so that at every other point in the app (like here and in tests) I can get/add data to/from the store and not worry about setting variables on the window.

Once the API is available I'll be able to remove the window hook and add API calls to these hooks.

Do you think that make sense?

static/js/src/advantage/react/hooks/useUsingTestBackend.test.tsx
huwshimi: Whoop! Thanks.

static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.test.tsx
ClementChaumel: Oh interesting, thanks!

static/js/src/advantage/react/hooks/useUsingTestBackend.ts
ClementChaumel: Ah ok I see! It might be a bit overkill to have an ad-hoc hook just to access one query?
What I would probably do is just use this in the consuming components
```jsx
  const { data: usingTestBackend } = useQuery<UsingTestBackend>(
    "usingTestBackend"
  );
```

Is there some advantage I'm missing to wrap this in a custom hook? Or maybe it'll make more sense when the data actually comes from an API call

huwshimi: Yeah with this really simple flag there's not a huge amount of benefit at the moment except:
a) the type is set in one place and,
b) the data is consumed consistently (window -> window fetch hook -> react query -> hook to get data from react query) so that it's obvious to always use the hook to get the data vs a mix of custom hooks and `useQuery`,
c) the hook is set up ready to add the API calls when they're ready.

Does that sound reasonable?

ClementChaumel (approved):

PR #10141, opened 2021-08-06
Add testing factories for the contract API

huwshimi:
## Done

- Added test factory dep.
- Added testing factories for the API and contract API types.
- Updated window data hook to use the types and factories.

## QA

N/A

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/145.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "eslint-plugin-cypress": "2.11.2",
     "eslint-plugin-react": "7.23.2",
     "exports-loader": "2.0.0",
+    "fishery": "1.4.0",
     "full-icu": "1.3.1",
     "jest": "^26.6.3",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -56,6 +56,7 @@
     "eslint-plugin-cypress": "2.11.2",
     "eslint-plugin-react": "7.23.2",
     "exports-loader": "2.0.0",
+    "fishery": "1.4.0",
```
huwshimi, line 59: @ClementChaumel @albertkol I picked this factory lib as it supports js as well as ts so we can start using it for the js React files and then if they get migrated to ts the factories will already be there.

ClementChaumel (approved):
:+1:

PR #10158, opened 2021-08-09
Add sentry reporting in the frontend

ClementChaumel:
## Done

- Added the react sentry dependency
- Added monitoring for all runtime errors
- Added explicit monitoring for unknown card and payment errors
- Removed useless `else` statements in util function

## QA

- TURN OFF YOUR ADBLOCKER
- Pull the branch
- Make some changes that would cause an error (like calling an undefined function) in the react payment modal app
- run `dotrun`
- go to http://0.0.0.0:8001/advantage/subscribe?test_backend=true
- open the modal and watch it crash
- go to https://sentry.is.canonical.com/canonical/ubuntu-com/ and check that the error was correctly reported.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/10

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,8 @@
     "@canonical/latest-news": "1.2.0",
     "@canonical/react-components": "0.20.0",
     "@reduxjs/toolkit": "1.5.0",
+    "@sentry/react": "^6.11.0",
+    "@sentry/tracing": "^6.11.0",
     "@stripe/react-stripe-js": "1.4.1",
     "@stripe/stripe-js": "1.15.0",
     "date-fns": "2.17.0",
--- a/static/js/src/advantage/error-handler.js
+++ b/static/js/src/advantage/error-handler.js
@@ -186,25 +186,35 @@ export function getErrorMessage(error) {
     message?.includes("Your card has insufficient funds")
   ) {
     return "That card doesn’t have enough funds to make this payment. Please contact your card issuer, or try a different card.";
-  } else if (INCORRECT_NUMBER_CODES.includes(code)) {
+  }
+  if (INCORRECT_NUMBER_CODES.includes(code)) {
     return "That card number is incorrect. Check the number and try again.";
-  } else if (INCORRECT_CVC_CODES.includes(code)) {
+  }
+  if (INCORRECT_CVC_CODES.includes(code)) {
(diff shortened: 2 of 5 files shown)


anthonydillon:
Few questions:
- Why remove the `else`? Are you expecting to hit multiple conditions in a single pass? It will be slower to do all the checks.
- Could you only push errors in production and so dev work does not trigger a sentry error?
-

ClementChaumel:
> Few questions:
> 
> * Why remove the `else`? Are you expecting to hit multiple conditions in a single pass? It will be slower to do all the checks.
Following each `if` there is a `return` statement so the next ones are ignored anyway 

> * Could you only push errors in production and so dev work does not trigger a sentry error?
Sure but maybe we could look at creating [environments in Sentry](https://docs.sentry.io/product/sentry-basics/environments/)?
Or another way to filter out everything coming from something else than prod would be to add a filter in the allowed domains section of the project setting.
What do you think?

anthonydillon:
Yea sure, I was just suggesting the production reporting because that is what we do in other places.

albertkol:
@anthonydillon @ClementChaumel 

If front-end sentry gets blocked by add blocker. It might be worth implementing this solution:
https://docs.sentry.io/platforms/javascript/guides/react/troubleshooting/#using-the-tunnel-option 

I do not mind pushing a commit with the endpoint in python for this purpose.

We could make it generic for the whole site `https://ubuntu.com/sentry` (not just advantage)

Or should we discuss it during our dev :ketch-up: ?

albertkol:
There seems to be a sentry call always happening on /advantage/subscribe that throws 403 with Add block off:
image

How does the environment separation work? Do you define it on sentry by domain name or in the code when you make the request?

When I look at this error:
image

The env is prod when the url is 0.0.0.0:8001.

albertkol:
@ClementChaumel any idea why this thing is showing up on page load? 
image

ClementChaumel:
> @ClementChaumel any idea why this thing is showing up on page load?
> image

I'm not sure to be honest I think it's trying to set up the envelope which allows it to send messages to the sentry server https://develop.sentry.dev/sdk/envelopes/

albertkol (approved):

PR #10161, opened 2021-08-09
Update some more forms to work with the new endpoint

jpmartinspt:
## Done

Update some more forms to work with the new endpoint by removing unneeded fields.


anthonydillon (approved):
LGTM

PR #10163, opened 2021-08-10
Add renewal settings component.

huwshimi:
## Done

- Add renewal settings component.

## QA

- Click one of the 'Renewal settings' contextual menus.
- You should see the form.
- Clicking cancel should close the form.

## Issue / Card

Fixes: canonical-web-and-design/commercial-squad#97.

## Screenshots

Screen Shot 2021-08-10 at 3 55 44 pm

### DIFF
--- a/static/js/src/advantage/react/components/Subscriptions/SubscriptionList/ListGroup/ListGroup.tsx
+++ b/static/js/src/advantage/react/components/Subscriptions/SubscriptionList/ListGroup/ListGroup.tsx
@@ -1,27 +1,35 @@
-import { ContextualMenu } from "@canonical/react-components";
-import React from "react";
+import React, { useRef, useState } from "react";
 import type { ReactNode } from "react";
+
+import RenewalSettings from "../RenewalSettings";
 
 type Props = {
   children: ReactNode;
   title: string;
 };
 
-const ListGroup = ({ children, title }: Props): JSX.Element => (
-  <div className="p-subscriptions__list-group">
-    <div className="p-subscriptions__list-group-title">
-      <span className="p-text--x-small-capitalised u-align-text--small-to-default u-no-margin--bottom">
-        {title}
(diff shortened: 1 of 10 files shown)


static/js/src/advantage/react/components/Subscriptions/SubscriptionList/ListGroup/ListGroup.tsx
```diff
@@ -1,27 +1,35 @@
...
+  return (
+    <div className="p-subscriptions__list-group">
+      <div
+        className="p-subscriptions__list-group-title"
+        ref={(element) => {
+          positionNode.current = element;
+          // Fire state change so that the menu rerenders now that the ref
+          // exists in the DOM.
+          setRefReady(true);
+        }}
```
ClementChaumel, line 23: Can you explain what's going on here?

huwshimi: Yeah, so because React uses a shadow DOM, when we set up a ref e.g. `ref={positionNode}` there is no way to know when that ref exists in the real DOM (refs by their nature don't cause rerenders). But because you usually use refs to interact with DOM elements React has provided a pattern called ["Callback refs"](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) so that you can do something when the element gets rendered.

In our case we set the current ref to the element and because we want to the contextual menu to update it's position/size from that element we also do a state update (`setRefReady`) to cause the components to rerender now that the position node exists.

Does that make sense?

ClementChaumel: Oh ok, I didn't know about the callback ref, thanks for the explanation!

ClementChaumel (approved):

PR #10167, opened 2021-08-10
Add footer_item_small to make adding footer lists more explicit

pmahnke:
## Done

- Broke the old `_footer_item.html` into two partials, one for navigation sections with sub-pages and one with no sub-pages.  It used to be based on the number of subpages; however, that doesn't work in cases where there are hidden pages like contact-us and thank-you perfectly.

## QA

- Look at the footer and see that all the sections that should be expanded are and those that are not are not

## Demo

https://ubuntu-com-10167.demos.haus/

FYI @mmanciop

### DIFF
--- a/static/sass/_pattern_footer.scss
+++ b/static/sass/_pattern_footer.scss
@@ -181,7 +181,7 @@
     padding: 0;
 
     &.second-level-nav-small {
-      @media only screen and (max-width: $breakpoint-small) {
+      @media only screen and (max-width: $breakpoint-medium) {
         margin-left: -0.7rem;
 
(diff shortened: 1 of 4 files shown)


anthonydillon (approved):
LGTM :+1: I see that is has made observility a soft link. 

A few small things I noticed:
- The `p-footer-list--margin` doesn't have any styling I can find. Do we still need that class?
- Also there is an issue at medium screens but is present on live.
ubuntu-com-10167 demos haus_

pmahnke:
@anthonydillon can you look again? I think I fixed both

PR #4317, opened 2018-10-31
Flattened /legal/ubuntu-advantage section

pmahnke:
## Done

- Flattened legal ubuntu-advantage section
- Updated navigation
- Updated redirects
- Added some special redirects requested by legal team

## QA

- Check out this feature branch
- Run the site using the command `./run serve`
- look at the files in the /legal directory and compare to this [spreadsheet](https://docs.google.com/spreadsheets/d/12kuuGgwRqN3k7wlRc4QKIuZifjvxbynizO_LEjKjBq4/edit#gid=0)

* test the link [/legal/ubuntu-advantage-service-description](http://0.0.0.0:8001/legal/ubuntu-advantage-service-description) and [the old link](http://0.0.0.0:8001/legal/ubuntu-advantage/service-description) redirects also test that [/uasd](http://0.0.0.0:8001/uasd) redirects as well
* test the link [/legal/ubuntu-advantage-service-description/ja](http://0.0.0.0:8001/legal/ubuntu-advantage-service-description/ja) and [the old link](http://0.0.0.0:8001/legal/ubuntu-advantage/service-description-ja) redirects
* test the link [/legal/ubuntu-advantage-assurance](http://0.0.0.0:8001/legal/ubuntu-advantage-assurance) and [the old link](http://0.0.0.0:8001/legal/ubuntu-advantage/assurance) redirects
* test the link [/legal/ubuntu-advantage-service-terms](http://0.0.0.0:8001/legal/ubuntu-advantage-service-terms) and [the old link](http://0.0.0.0:8001/legal/ubuntu-advantage/service--terms) redirects

## Issue / Card

Fixes #4200

PR #10171, opened 2021-08-10
Limit number of filters by category

carkod:
## Done

Show vendors and filters only for that specific category #4317

## QA

- Check https://ubuntu-com-10171.demos.haus/certified/desktops and check only shows filters that apply to that category (that have results)

## Issue / Card

Fixes https://github.com/canonical-web-and-design/web-squad/issues/4317

### DIFF
--- a/webapp/certified/views.py
+++ b/webapp/certified/views.py
@@ -395,12 +395,46 @@ def create_category_views(category, template_path):
     Helper function to create multiple /certified/<page> category views
 
     Keyword arguments:
-    category -- must be categories accepted by API (Desktop, Laptop, SoC, \
+    category -- must be categories accepted by API (Desktop, Laptop, \
     Server, Ubuntu Core, Server SoC)
     template_path -- full template path (e.g. certified/search-results.html)
     """
-    certified_releases = api.certified_releases(limit="0")["objects"]
-    certified_makes = api.certified_makes(limit="0")["objects"]
+    if category == "Desktop":
+        certified_releases = api.certified_releases(
+            limit="0", desktops__gte=1
+        )["objects"]
+        certified_makes = api.certified_makes(limit="0", desktops__gte=1)[
+            "objects"
+        ]
+    elif category == "Laptop":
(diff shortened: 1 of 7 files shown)


webapp/certified/views.py
```diff
@@ -395,12 +395,28 @@ def create_category_views(category, template_path):
...
+        certified_releases = api.certified_releases(limit="0", laptops__gte=1)["objects"]
+        certified_makes = api.certified_makes(limit="0", laptops__gte=1)["objects"]
+    elif category == "Server SoC":
+        certified_releases = api.certified_releases(limit="0", soc__gte=1)["objects"]
+        certified_makes = api.certified_makes(limit="0", soc__gte=1)["objects"]
+    elif category == "Ubuntu Core":
+        certified_releases = api.certified_releases(limit="0", smart_core__gte=1)["objects"]
+        certified_makes = api.certified_makes(limit="0", smart_core__gte=1)["objects"]
+    elif category == "Server":
+        certified_releases = api.certified_releases(limit="0", servers_gte=1)["objects"]
```
codersquid, line 415: You need an additional _ for this. `servers__gte`. Same is true for the certified_makes call below.

codersquid (changes requested):

codersquid:
While looking at the demo I noticed that some of the filter counts to show less are negative. For example, https://ubuntu-com-10171.demos.haus/certified/laptops

That shows -2 for vendor and -3 for release.

I guess it wasn't obvious before because there were so many vendors in the results.

carkod:
> While looking at the demo I noticed that some of the filter counts to show less are negative. For example, https://ubuntu-com-10171.demos.haus/certified/laptops
> 
> That shows -2 for vendor and -3 for release.
> 
> I guess it wasn't obvious before because there were so many vendors in the results.

Fixed

webapp/certified/views.py
carkod: Good one, thanks!

codersquid (approved):
lgtm

bethcollins92:
Looking good thanks @carkod 🎉

PR #10175, opened 2021-08-11
WIP: Add a JSON endpoint for mirrors list

bartaz:
## Done

- [List of work items including drive-bys.]

## QA

## Issue / Card

Fixes #

## Screenshots

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -487,6 +487,14 @@ def account_query():
     )
 
 
+def mirrors_query():
+    """
+    A JSON endpoint to request list of Ubuntu mirrors
+    """
+
+    return flask.jsonify(_build_mirror_list())
+
+
 def build_tutorials_index(session, tutorials_docs):
     def tutorials_index():
         page = flask.request.args.get("page", default=1, type=int)
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -71,6 +71,7 @@ from webapp.views import (
     sixteen_zero_four,
     openstack_install,
     marketo_submit,
+    mirrors_query,
 )
 
 from webapp.advantage.views import (
@@ -291,6 +292,7 @@ def utility_processor():
 # Simple routes
 app.add_url_rule("/sitemap.xml", view_func=sitemap_index)
 app.add_url_rule("/account.json", view_func=account_query)
+app.add_url_rule("/mirrors.json", view_func=mirrors_query)
 app.add_url_rule("/marketo/submit", view_func=marketo_submit, methods=["POST"])
 app.add_url_rule("/advantage", view_func=advantage_view)
 app.add_url_rule("/advantage/subscribe", view_func=advantage_shop_view)


webapp/views.py
```diff
@@ -48,7 +48,7 @@
 )
 
 
-def _build_mirror_list():
+def _build_mirror_list(all=False):
```
nottrobin, line 51: How about doing it the other way around? Something like `nearby=False`, or `use_geoip=False`?

bartaz: Good call, updated to `local` as discussed.

albertkol:
In what situation do we need all mirror links? @nottrobin @bartaz 

Looking at the way the `_build_mirror_list` works, it's returning the mirror links of the `ip_location` or `[]`, if there is no `ip_location` in the header of the request. When `[]` is returned we default the `releases.ubuntu.com` mirror.

Or is it more of a aesthetic decision to have `/mirrors.json` return all the mirrors?

webapp/views.py
```diff
@@ -64,7 +64,15 @@ def _build_mirror_list():
         flask.request.headers.get("X-Real-IP", flask.request.remote_addr)
```
albertkol, line 64: Two lines above. `country_code` gets initialised with `NO_COUNTRY_CODE`. It's redundant and can be removed.

albertkol:
I would imagine if we do `/mirrors.json?local=false` we would like to show all mirrors.

If `/mirrors.json?local=false` is set we still show only local mirrors.

webapp/views.py
```diff
@@ -64,7 +64,15 @@ def _build_mirror_list():
...
     )
 
-    if ip_location and "country" in ip_location:
+    if not local:
```
albertkol, line 67: Instead of the `if` and `elif`. I'd just do a quick return of the `mirror_list` inside each if. But it's up to you.

webapp/views.py
```diff
@@ -64,7 +64,15 @@ def _build_mirror_list():
...
+            mirror_list.append(
+                {
+                    "link": mirror["link"],
+                    "bandwidth": mirror["mirror_bandwidth"],
+                }
+            )
+    elif ip_location and "country" in ip_location:
         country_code = ip_location["country"]["iso_code"]
 
         for mirror in mirrors:
```
albertkol, line 78: The if condition below is getting split on 3 lines. 

I'd extract `mirror["mirror_countrycode"]` as a variable `mirror_countrycode`. 
..and extract `mirror["link"].startswith("https")` as a variable `is_https`.
Rewrite the condition as `if mirror_countrycode == contry_code and is_https:` 

But I leave it up to you.

albertkol (commented):
Some small things.

albertkol:
> Or is it more of a aesthetic decision to have `/mirrors.json` return all the mirrors?

Now I remember. We wanted all the mirrors so we can `linkcheck` them, right?

bartaz:
> > Or is it more of a aesthetic decision to have `/mirrors.json` return all the mirrors?
> 
> Now I remember. We wanted all the mirrors so we can `linkcheck` them, right?

@albertkol Yes, it's to make it possible to check if mirrors are working.

webapp/views.py
bartaz: Good comments, I was not changing these from what it was before, but I can clean it up a bit while I'm on it.

webapp/views.py
```diff
@@ -487,6 +498,23 @@ def account_query():
     )
 
 
+def mirrors_query():
+    """
+    A JSON endpoint to request list of Ubuntu mirrors
+    """
+    local = flask.request.args.get("local", default=False)
+
+    if not local or local == "false" or local == "False" or local == "no":
```
albertkol, line 507: No idea it would be so painful working with bool query parameters in flask.

Anyway.. I would have wrote this condition instead:
```
if local and local.lower() != "true":
    local = False
```

But it's fine. Up to you again.

albertkol (approved):
Looks good! :+1:

webapp/views.py
carkod: This is not a boolean for Python, it's a string

Issue #8583, opened 2020-10-23
Write a test to ensure mirrors are loading properly


nottrobin:
Following https://github.com/canonical-web-and-design/ubuntu.com/pull/8549, the second time the mirrors have broken, we should ensure there's a test to make sure mirrors don't get accidentally broken again.

nottrobin:
@SirSamTumless as far as I know, @bartaz deliberately didn't close this issue because it's not wholly resolved by #10190. This should be placed back in ice box.

PR #10190, opened 2021-08-12
Add cron job to check Ubuntu download mirrors

bartaz:
## Done

- Added the cron job to check in `/mirrors.json` returns a list of links

## QA

This is a cron job in GH action, so not sure how to QA it.

The job was tested running on staging here:
https://github.com/canonical-web-and-design/ubuntu.com/runs/3321716261?check_suite_focus=true

## Issue / Card

Part of #8583

### DIFF
--- a/.github/workflows/mirrors.yaml
+++ b/.github/workflows/mirrors.yaml
@@ -1,0 +1,18 @@
+name: Ubuntu mirrors on live
+
+on:
+  schedule:
+    - cron: "20 7 * * *"
+
+jobs:
+  check-mirrors:
+    if: github.repository == 'canonical-web-and-design/ubuntu.com'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check if /mirrors.json returns list of links
+        run: curl https://ubuntu.com/mirrors.json | jq .[].link -e
+
+      - name: Send message on failure
+        if: failure()
+        run: curl -X POST -F "workflow=${GITHUB_WORKFLOW}" -F "repo_name=${GITHUB_REPOSITORY}" -F "action_id=${GITHUB_RUN_ID}" ${{ secrets.BOT_URL }}?room=web--design


bartaz:
@nottrobin @albertkol The job created here simply checks if mirrors.json returns a non-empty list of mirrors. If JSON will not be returned, or malformed, or array would be empty (not contain any object with a `.link` prop we expect) the test will fail.

I guess it's good for start. If we want to extend it we could try to fetch every mirror. I tried that (getting just head to check if they return 200), but some of them redirect, some timeout, etc. So it's hard to say on what we want to fail here.

The other question is how often we want to run it. Currently I copied the cron schedule from some other job (every morning). If we don't check individual mirrors it's probably fine to do it often.

albertkol:
@bartaz the easiest way to test it is to have your new git action set as a `on: pull_request` rather than a cronjob. Once you are happy with it have it set to a cron.

Are you sure it returns an error if the array is empty? Running it locally with an empty array seems to be finish without errors.
```bash
fylon @ fylon ~/Code/ubuntu.com (mirrors-check)
└─ $ ▶ curl http://0.0.0.0:8001/mirrors.json | jq .[].link -e
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     3  100     3    0     0   1000      0 --:--:-- --:--:-- --:--:--  1000
```
Anyway.. I guess there is some value in checking if we have a valid json coming from `/mirrors.json`. But I thought the main thing we want to test is that the `https` urls return 200.

bartaz:
> @bartaz the easiest way to test it is to have your new git action set as a on: pull_request rather than a cronjob. Once you are happy with it have it set to a cron.

@albertkol Yes, I did test it first as PR action, for example here: https://github.com/canonical-web-and-design/ubuntu.com/runs/3321716261?check_suite_focus=true
What I meant was that I'm not sure what QA steps to give on the action with cron job.

As for failing on empty JSON, I'm quite sure I had it fail on that locally, but can verify.

bartaz:
From my tests in the command line it seems that current test:

- fails when:
  - curl fails to load mirrors.json
  - mirrors.json contents are not valid JSON
  - mirrors.json returns valid, but empty array `[]`
  - mirrors.json returns valid JSON, which is not an array with objects that have 'link' property
- passes when:
  - mirrors.json is completely empty, is an empty string
  - mirrors.json is  a valid array of objects with links

My suggestion (because I don't have much time to continue on this now), would be to merge this PR to have a simple first test for mirrors, and continue the work later to:

- only return "https" mirrors in mirrors.json
- update the GH action test to verify error codes of all returned mirrors (expecting 200, making sure to follow any redirects)

nottrobin (approved):
Sure, LGTM

PR #10191, opened 2021-08-12
Fix cypress tests to ensure they pass

bethcollins92:
## Done

- Bumped Cypress to latest version
- Set cookies on the test files to prevent the cookie policy from blocking the page
- Fixed up tests in `form_spec.js` and `navigation_spec.js`

## QA

- For the post to marketo status 200 to pass you need to add the credentials to a local file named `cypress.env.json`. Please ping me so I can tell you where to find them
- In your terminal run `yarn run cypress open --config-file tests/cypress/cypress.json --config baseUrl=http://0.0.0.0:8001/`
- This will open a cypress window - click on the `form_spec.js` and then the `navigation_spec.js` to see the tests pass.
- Note: if the form field is out of view when it's attempted to be selected it will fail - you need to scroll as the test fills out the fields 

## Issue / Card

Fixes [#4365](https://github.com/canonical-web-and-design/web-squad/issues/4365)


bethcollins92:
@pmahnke I mentioned you in the comments, not sure whether it pinged you or not. Is the smartquotes.js something we definitely need to keep? There was an error locally and I couldn't get the cypress tests to even start running because of it. I can open another issue to uncomment this out and try and fix the error if so.

pmahnke:
It would be slightly hard to kill it as we have now straightened all our
quotes.

Could we get it fixed up?  The author already took a patch from @Anthony
Dillon ***@***.***>.

tests/cypress/forms_spec.js
```diff
@@ -98,11 +104,27 @@ context("Marketo dynamic form", () => {
...
 
-    cy.wait(10000); // eslint-disable-line
+    cy.wait(3000); // eslint-disable-line
     cy.get("#mktoForm_1251")
       .submit()
       .should((page) => {
-        expect(page[0].action).to.equal("/marketo/submit");
+        expect(page[0].action).to.include("/marketo/submit");
       });
+    cy.get(".p-modal__close").click();
```
petermakowski, line 113: Might be better to use something that's visible to the user instead of a class name. I think `Close active modal` aria-label would work (something like `cy.findByLabelText("Close active modal")`.

package.json
```diff
@@ -49,7 +49,7 @@
     "babel-jest": "26.6.3",
     "babel-loader": "8.2.2",
     "concurrently": "6.0.0",
-    "cypress": "5.6.0",
+    "cypress": "8.3.1",
```
petermakowski, line 52: 💥

petermakowski (commented):
Well done on fixing this 👍

tests/cypress/forms_spec.js
petermakowski: You can find some good examples of how you can write cypress tests in conjunction with @testing-library/cypress here: 

- https://testing-library.com/docs/cypress-testing-library/intro/#examples 
- https://github.com/testing-library/cypress-testing-library/blob/97939da7d4707a71049884c0324c0eda56e26fc2/cypress/integration/find.spec.js

bethcollins92: Great - Thanks @petermakowski

tests/cypress/forms_spec.js
```diff
@@ -98,11 +104,27 @@ context("Marketo dynamic form", () => {
...
+  });
+
+  it("should successfully complete download server", () => {
+    cy.visit("/download/server/s390x");
+
+    cy.get('input[name="firstName"]').type("Test");
```
petermakowski, line 119: I'd suggest using `findByLabelText` (so looking for an actual label that the user will be able to read) instead of a input's name. You want to avoid testing implementation details as much as possible.

```suggestion
    cy.findByLabelText(/First name:/).type("Test");
```

tests/cypress/forms_spec.js
```diff
@@ -12,97 +12,110 @@ const getIframeBody = () => {
...
-    cy.get('select[name="Country"]').select("Colombia");
-    cy.get('input[name="company"]').type("Test");
+    cy.findByLabelText(/First name:/).type("Test");
+    cy.findByLabelText(/Last name:/).type("Test");
+    cy.findByLabelText(/Email address:/).type("test@test.com");
+    cy.findByLabelText(/Mobile\/cell phone number:/).type("07777777777");
+    cy.findByLabelText(/Country:/).select("Colombia");
+    cy.findByLabelText(/Company:/).type("Test");
+    // cy.findByLabelText(/Job title:/).type("Test");
     cy.get('input[name="title"]').type("Test");
```
bethcollins92, line 48: Thanks for your review @petermakowski - I've made the tests more user centric. Just one problem I couldn't solve. here for some reason doing `cy.findByLabelText(/Job title:/).type("Test");` wasn't finding the element even though that is the label text. Am I missing something really obvious?

petermakowski: Oh interesting. This test actually caught a bug on this page!

Element IDs should be unique. Looks like there are multiple elements with an id "title" that the label refers to. 

image
image

I'd suggest updating the id of the label to be unique, for example to "job-title" instead of "title".

bethcollins92:
@petermakowski ready for re-review

petermakowski (approved):
LGTM. Appreciate you took the time to not only fix, but also improve, making these more user-centric. Well done 👍

petermakowski:
@bethcollins92  By the way, cypress has a `scrollIntoView` method that can solve problems you mentioned in the description: https://docs.cypress.io/api/commands/scrollintoview

bethcollins92:
> @bethcollins92 By the way, cypress has a `scrollIntoView` method that can solve problems you mentioned in the description: https://docs.cypress.io/api/commands/scrollintoview

ahh - that's good to know. I wondered what was happening there. Thank you!

bethcollins92:
@anthonydillon Please could you QA +1 this if the tests passed for you?

anthonydillon:
Tests pass locally for me. Thanks @bethcollins92

PR #10196, opened 2021-08-16
Fail the build when esbuild has an error

huwshimi:
## Done

- Fail the build when esbuild has an error.

## QA

- Check that the demo service has successfully built the branch.
- If you wish you could also test that this fails locally by adding `import nothing from "nothing";` to a js file and running `dotrun build-js` and you should see a failure.


anthonydillon (approved):
LGTM :+1:

PR #10200, opened 2021-08-16
Fix results per page

carkod:
## Done

Fix limit appending twice.
Refactored loops and unnecessary conditions.

## QA

- Check /certified/desktops or any search results page, and try the different page sizes (20, 40, all...) and make sure the URL does not append the `limit` query twice.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/web-squad/issues/4316

### DIFF
--- a/static/js/src/certified-search-results.js
+++ b/static/js/src/certified-search-results.js
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+
 function clearFilters() {
   location.assign(`/certified?q=`);
   return false;
@@ -73,14 +75,19 @@ function enableApplyFilters() {
 // function to ensure only the option which has been changed is appended to the URL
 function updateResultsPerPage() {
   const searchResults = document.querySelector(".js-search-results");
-  let resultsDropdowns = document.querySelectorAll(".p-results-per-page");
-  resultsDropdowns.forEach((resultsDropdown) => {
-    const options = resultsDropdown.querySelectorAll("option");
-    options.forEach((option) => {
-      if (option.selected) {
-        searchResults.submit();
-      }
-    });
+  const pageSizeTop = document.getElementById("page-size-top");
+  const pageSizeBottom = document.getElementById("page-size-bottom");
+
+  pageSizeTop.addEventListener("change", (e) => {
+    // Needs to be set because the other dropdown is a dummy
+    searchResults.submit();
+  });
+
+  pageSizeBottom.addEventListener("change", (e) => {
+    // Avoids submitting 2 redundant fields
+    let pageSizeTopChange = new Event("change");
+    pageSizeTop.value = e.target.value;
+    pageSizeTop.dispatchEvent(pageSizeTopChange);
   });
 }
 
@@ -133,3 +140,4 @@ enableApplyFilters();
 toggleVersionsList();
 toggleVendorsList();
 toggleShowAllLinks();
+updateResultsPerPage();
--- a/templates/certified/desktops.html
+++ b/templates/certified/desktops.html
@@ -41,7 +41,7 @@
       <div class="col-5 u-align--right">
         {% if total_results > 0 %}
         <p style="display: inline-block; padding-right: 0.5rem;">Results per page</p>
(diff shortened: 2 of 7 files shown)


bethcollins92:
@carkod - This is still having the same problem I was having. When you click "Apply filters", both dropdowns are in the same form so they both get appended to the URL - I just did it and here is the url `https://ubuntu-com-10200.demos.haus/certified/laptops?q=&limit=20&vendor=Dell&vendor=HP&limit=20`

carkod:
> @carkod - This is still having the same problem I was having. When you click "Apply filters", both dropdowns are in the same form so they both get appended to the URL - I just did it and here is the url `https://ubuntu-com-10200.demos.haus/certified/laptops?q=&limit=20&vendor=Dell&vendor=HP&limit=20`. The bottom drop down needs to be outside the form. I can take a look but I'm on Vanilla iteration for this week and next

If we pull out the bottom dropdown out of the form it's not going to work, because it doesn't have a form to submit to. Instead what I did is to make it a complete dummy, but yes this is not accessible.

bethcollins92:
> > @carkod - This is still having the same problem I was having. When you click "Apply filters", both dropdowns are in the same form so they both get appended to the URL - I just did it and here is the url `https://ubuntu-com-10200.demos.haus/certified/laptops?q=&limit=20&vendor=Dell&vendor=HP&limit=20`. The bottom drop down needs to be outside the form. I can take a look but I'm on Vanilla iteration for this week and next
> 
> If we pull out the bottom dropdown out of the form it's not going to work, because it doesn't have a form to submit to. Instead what I did is to make it a complete dummy, but yes this is not accessible.

If you submit the form from within the `updateResultsPerPage()` and call this when the option is selected then this should still  submit the form when you select a pagination?

carkod:
> > > @carkod - This is still having the same problem I was having. When you click "Apply filters", both dropdowns are in the same form so they both get appended to the URL - I just did it and here is the url `https://ubuntu-com-10200.demos.haus/certified/laptops?q=&limit=20&vendor=Dell&vendor=HP&limit=20`. The bottom drop down needs to be outside the form. I can take a look but I'm on Vanilla iteration for this week and next
> > 
> > 
> > If we pull out the bottom dropdown out of the form it's not going to work, because it doesn't have a form to submit to. Instead what I did is to make it a complete dummy, but yes this is not accessible.
> 
> If you submit the form from within the `updateResultsPerPage()` and call this when the option is selected then this should still submit the form when you select a pagination?

not sure what you mean by "within the `updateResultsPerPage()` ", but pagination is working.

bethcollins92:
> > > > @carkod - This is still having the same problem I was having. When you click "Apply filters", both dropdowns are in the same form so they both get appended to the URL - I just did it and here is the url `https://ubuntu-com-10200.demos.haus/certified/laptops?q=&limit=20&vendor=Dell&vendor=HP&limit=20`. The bottom drop down needs to be outside the form. I can take a look but I'm on Vanilla iteration for this week and next
> > > 
> > > 
> > > If we pull out the bottom dropdown out of the form it's not going to work, because it doesn't have a form to submit to. Instead what I did is to make it a complete dummy, but yes this is not accessible.
> > 
> > 
> > If you submit the form from within the `updateResultsPerPage()` and call this when the option is selected then this should still submit the form when you select a pagination?
> 
> Not sure what you mean. The pagination is working and the filters are also working?

If you go to https://ubuntu-com-10200.demos.haus/certified/laptops and select `Dell` then apply filters, then look at the URL the pagination is in the URL twice. Also, when I select a pagination in the drop down the page nothing changes - does it work for you in the demo?

carkod:
@bethcollins92 all done! Can you re-review?

bethcollins92:
@carkod Nearly! I think you missed adding the `page-size-bottom` class to the devices page, as it's the only one still broken. https://ubuntu-com-10200.demos.haus/certified/devices?q=&limit=40&limit=20

carkod:
> @carkod Nearly! I think you missed adding the `page-size-bottom` class to the devices page, as it's the only one still broken. https://ubuntu-com-10200.demos.haus/certified/devices?q=&limit=40&limit=20

Well spotted! Thanks. Should be ready now

bethcollins92 (approved):
LGTM! woo thanks @carkod

PR #10206, opened 2021-08-18
Fix invoices page

albertkol:
## Done
- Change 1:
Some invoices do not come with a total cost field. This change makes it so that we handle those situations more graceful. If total is missing we just display `-`.
- Change 2:
The table is overriding the footer. Replacing `height` with `min-height`.
- Change 3:
Canonical CUBE invoices should not show show `(annual)` in the service column.

## QA

- Be someone with access to the Canonical Staff account
- Go to https://ubuntu-com-10206.demos.haus
- Login
- Go to https://ubuntu-com-10206.demos.haus/account/invoices
- Change 1: The page should work
- Change 2: The table should not overlap the footer
- Change 3: Canonical CUBE should not show `(Annual)`

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -949,7 +949,7 @@ body {
 }
 
 .p-strip-account-page {
-  height: 600px;
+  min-height: 600px;
 
   @media (max-width: $breakpoint-small) {
     height: auto;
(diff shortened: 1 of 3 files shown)


wgx:
Works, no longer 500'd

wgx:
QA+1 👍🏽

bethcollins92 (approved):
LGTM 🎉

PR #10209, opened 2021-08-19
Backend new advantage init

albertkol:
## Done

Initialising the work for the new advantage page. Adding the classes and functions files.

### Classes
We are splitting our classes in two types:
- **Primitives** - classes that are built from extracting the necessary properties of the API json responses. Primitives are built by parser functions
Examples of primitives: `Contract`, `ContractItem`, `Subscription`, `Product`, `Account`
- **Models** - classes that are built by processing primitive classes. These are refined structures designed specifically for Ubuntu Advantage
Examples of models: `UserSubscription`, `Listings`

This is so that we have a clear separation between objects that belong to the domain of the API we consume and the object that belong to the domain of Ubuntu Advantage.

### Functions
We are splitting our functions in three files:
- **Parsers** - functions that accept raw json data and convert it to a primitive or a list of primitives. Name ideally starting with `parse_`
- **Builders** - functions that build specific structures. Name ideally starting with `build_`
- **Helpers** - functions that are not parsers or builders; they have a more general purpose
For example: `to_dict` transforms Objects and List of Object to Dictionary

We are splitting the functions depending on the action they fulfil.

### Folder structure
The `primitives`, `builders`, `parsers` and helpers will all reside inside a `ua-contracts` folder, since they all belong to the API’s domain with the goal of creating our `models`.

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/178


webapp/advantage/ua_contracts/helpers.py
```diff
@@ -0,0 +1,40 @@
+def to_dict(structure, class_key=None):
```
nottrobin, line 1: I don't understand the point of this function. Could you give some examples of where you need it? What are the input structures you have and what are the output structures you're expecting?

tests/test_advantage.py
```diff
@@ -2,7 +2,7 @@
 
 from requests.exceptions import HTTPError
 
-from webapp.advantage.api import (
+from webapp.advantage.ua_contracts.api import (
```
nottrobin, line 5: Maybe you don't actually need the word "api" here - maybe it could just be `ua_contracts`? Or whatever the best way to describe their system is.

webapp/advantage/models.py
```diff
@@ -0,0 +1,65 @@
+from typing import List
+
+
+class Entitlement:
```
nottrobin, line 4: Could you add docstrings to these classes so I can understand what they're for?

nottrobin (changes requested):
This broadly looks great, I'd just like a little more detail around the purpose of `to_dict`, and descriptions for `Entitlement`s and `Listing`s, so I can tell why they're needed.

albertkol:
`to_dict` It's a code snippet I found on the internet transforming Objects, and List of Object to dictionaries.

I use it later on in my parser functions tests. Parsers receive jsons and return Objects or List of Objects. I use `to_dict` to check if the instance the parser returned matches values with the instance of the object I expected. It would look something like: `self.assertEquals(to_dict(expected_obj), to_dict(actual_obj))`. 

If it ends up being useful just for testing, I am happy to move it inside the `tests` folder, somewhere.

albertkol:
`Listings` - Are what Ubuntu Advantage presents the user as a products. The API has two entities for this: `Product` and `ProductListing`. I combine them together and get `Listings`. Basically a `Listing` is `ProductListing` + `ProductListing.Product.name`

`Entitlement` are properties of the `UserSubscription`. 
They are those little labels at the bottom: `ESM infra`, `24/7 Support`. 
image

`Listings`, `Entitlement` and `UserSubscription` are all concepts that exist inside the Ubuntu Advantage domain. We will likely have more in the future f.x:`Invoice`, `Payment`.

tests/test_advantage.py
albertkol: It is. The name of the folder is just `ua_contracts`. Inside it we have a file called `api`.

PR #10214, opened 2021-08-23
Filter out CUBE contracts

albertkol:
## Done

- Filter out CUBE contracts form /advantage

## QA

- Login to https://ubuntu.com production
- Go to /advantage you will see the Canonical Staff account with lots of CUBE contracts
- Switch to the demo site: https://ubuntu-com-10214.demos.haus
- Go to /advantage you will not see the Canonical Staff account with any CUBE contracts

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/182

### DIFF
--- a/webapp/advantage/api.py
+++ b/webapp/advantage/api.py
@@ -60,7 +60,11 @@ class UAContractsAPI:
     def get_account_contracts(self, account_id: str):
         try:
             response = self._request(
-                method="get", path=f"v1/accounts/{account_id}/contracts"
+                method="get",
+                path=(
+                    f"v1/accounts/{account_id}/contracts"
+                    f"?productTags=ua&productTags=classic&productTags=pro"
+                ),
             )
         except HTTPError as error:
             if error.response.status_code == 401:


webapp/advantage/api.py
```diff
@@ -60,7 +60,11 @@ def get_accounts(self, email=""):
     def get_account_contracts(self, account_id: str):
         try:
             response = self._request(
-                method="get", path=f"v1/accounts/{account_id}/contracts"
+                method="get",
+                path=(
+                    f"v1/accounts/{account_id}/contracts"
+                    f"?productTags=ua&productTags=classic&productTags=pro"
```
ClementChaumel, line 66: Does this means `("ua" || "classic" || "pro")`? It's a weird syntax I would've thought that query params were unique

albertkol: @ClementChaumel yes, it's works like an or. This is an array query param of the name `productTags`. Back-end will allow us to give the values comma separated in the future, but for now, this is what worked when I tested.

albertkol: It works. We also do it on https://ubuntu.com/security/cve?version=impish&status=released&version=hirsute&status=released

I was the proud dev to implement it.

ClementChaumel (approved):

PR #10215, opened 2021-08-23
Fix price preview recalculation 

petermakowski:
## Background
- the VAT error occurs because the [checkVAT function](https://github.com/canonical-web-and-design/ubuntu.com/blob/master/static/js/src/advantage/subscribe/react/components/PaymentMethodForm.jsx#L93-L119) that's calling the API is being called even when VAT is undefined (e.g. hasn't been entered by the user yet). 
image

## Done
  - invoke the checkVAT function only if a VAT country is selected and VAT is defined

## QA

- go to https://ubuntu-com-10215.demos.haus/advantage/subscribe?test_backend=true
- make sure you're logged in
- select "buy" and fill out the form
  - select United Kingdom as a country
  - enter incorrect VAT (any random string will do)
- you should see an error `Error: That VAT number is invalid. Check the number and try again.`
- enter a valid VAT (e.g. GB 123 1234 12 123)
- error should disappear
- there should be no `VAT is missing` errors in the console

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/176

### DIFF
--- a/static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoForPurchasePreview.jsx
+++ b/static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoForPurchasePreview.jsx
@@ -1,7 +1,7 @@
 import { useMutation } from "react-query";
 import { postCustomerInfoForPurchasePreview } from "../../../contracts-api";
 
-const usePostCustomerInfoAnon = () => {
+const usePostCustomerInfoForPurchasePreview = () => {
   const mutation = useMutation(async (formData) => {
     const {
       name,
@@ -22,28 +22,24 @@ const usePostCustomerInfoAnon = () => {
       state: country === "US" ? usState : caProvince,
     };
 
-    if (VATNumber) {
-      const res = await postCustomerInfoForPurchasePreview(
-        window.accountId,
-        name,
-        addressObject,
-        {
-          type: "eu_vat",
-          value: VATNumber,
-        }
-      );
-
-      if (res.errors) {
-        throw new Error(JSON.parse(res.errors).code);
+    const res = await postCustomerInfoForPurchasePreview(
+      window.accountId,
+      name,
+      addressObject,
+      {
+        type: "eu_vat",
+        value: VATNumber,
       }
+    );
 
-      return res;
+    if (res.errors) {
+      throw new Error(JSON.parse(res.errors).code);
     }
 
-    throw new Error("VAT is missing");
+    return res;
   });
 
   return mutation;
 };
 
-export default usePostCustomerInfoAnon;
+export default usePostCustomerInfoForPurchasePreview;
--- a/static/js/src/advantage/subscribe/react/components/PaymentMethodForm.jsx
+++ b/static/js/src/advantage/subscribe/react/components/PaymentMethodForm.jsx
@@ -115,7 +115,9 @@ function PaymentMethodForm({ setCardValid }) {
   }, [values.country]);
 
   useEffect(() => {
-    checkVATDebounced(values);
+    if (vatCountries.includes(values.country) && values.VATNumber) {
+      checkVATDebounced(values);
+    }
   }, [values.country, values.VATNumber]);
 
   return (
--- a/static/js/src/advantage/subscribe/react/components/PaymentMethodForm.jsx
+++ b/static/js/src/advantage/subscribe/react/components/PaymentMethodForm.jsx
@@ -25,14 +25,14 @@ import { getErrorMessage } from "../../../error-handler";
 import { Field, Form, useFormikContext } from "formik";
 import { useQueryClient } from "react-query";
 
-import usePostCustomerInfoAnon from "../APICalls/usePostCustomerInfoAnon";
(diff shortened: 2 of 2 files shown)


ClementChaumel:
So, in this case, the API is not being called if there is no VAT number. I fixed that in a [previous commit](https://github.com/canonical-web-and-design/ubuntu.com/commit/638bfb1f10745d4ac15d5032516fa2cd4b9b9455) and that's the error that you see in the console.
I chose to add the check in the hook so it can be reused easily and without having to add the check to each consuming component. 


I think the bug we need to fix is this one:
If you try this
- go to the demo
- log in
- select UK with a valid tax number
- Wait for the total to update and show you tax amount
- Switch country to one in which VAT is not applicable
- It will not refresh and still display a tax amount.

And that's because the `usePostCustomerInfoAnon` instantly fails if there is no VAT number provided
But in the `checkVAT` function, where we declare what should happen depending on the state of `usePostCustomerInfoAnon`'s mutation we only invalidate (refresh) the preview if it's a success

```jsx
  const checkVAT = (formValues) => {
    mutation.mutate(formValues, {
      onSuccess: () => {
        queryClient.invalidateQueries("preview");
      },
      onError: (error) => {
        if (error.message === "tax_id_invalid") {
          setErrors({
            VATNumber:
              "That VAT number is invalid. Check the number and try again.",
          });
        }
      },
    });
  };
```
And we want the preview to refresh if we change to a non VAT country. So we either want to return an empty success from within the mutation if we detect that the country doesn't use VAT or maybe invalidate the preview on error and success alike in which case we could use the [`onSettled`](https://react-query.tanstack.com/guides/mutations#mutation-side-effects) side effect.

static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoForPurchasePreview.jsx
```diff
@@ -22,28 +22,24 @@ const usePostCustomerInfoAnon = () => {
       state: country === "US" ? usState : caProvince,
     };
 
-    if (VATNumber) {
```
petermakowski, line 25: I had to remove this check in order for the data to be updated on the back-end (even if no VAT number supplied) so that the subsequent preview call returns the correct price (for example when switching countries). This fixes another issue as well: https://github.com/canonical-web-and-design/commercial-squad/issues/174

static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoForPurchasePreview.jsx
```diff
@@ -22,28 +22,24 @@ const usePostCustomerInfoAnon = () => {
...
+        value: VATNumber,
       }
+    );
 
-      return res;
+    if (res.errors) {
+      throw new Error(JSON.parse(res.errors).code);
     }
 
-    throw new Error("VAT is missing");
```
petermakowski, line 43: I found this to be a little confusing, the API does not return an error when VAT number is missing. This seems like a perfectly valid call to make (as explained above).

static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoForPurchasePreview.jsx
```diff
@@ -1,7 +1,7 @@
 import { useMutation } from "react-query";
 import { postCustomerInfoForPurchasePreview } from "../../../contracts-api";
 
-const usePostCustomerInfoAnon = () => {
+const usePostCustomerInfoForPurchasePreview = () => {
```
petermakowski, line 4: I renamed the function to reflect the underlying API function name. What do you think @ClementChaumel ? Personally I find it much more descriptive.

static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoForPurchasePreview.jsx
petermakowski: I believe this is behaving correctly now:

Kapture 2021-08-24 at 08 17 26

static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoForPurchasePreview.jsx
ClementChaumel: Yeah the reason I added that was to make an effort to not call the API when we know it's not going to do anything. The condition should probably have been VATApplicableCountry && vatNumber
But I guess it's fine to let it go through as well

ClementChaumel (approved):

static/js/src/advantage/subscribe/react/APICalls/usePostCustomerInfoForPurchasePreview.jsx
petermakowski: Correct me if I'm wrong, but If it was `VATApplicableCountry && vatNumber` the issue would still occur. Looks like we always need to make 2 calls (postCustomerInfoForPurchasePreview and then preview) if any data that could affect the preview price has changed (including if we switched to a non-vat-applicable country). Otherwise preview will return response based on stale form data.

PR #10220, opened 2021-08-24
Fix guest user vat calculation

petermakowski:
## Done
- ensure user has an account before updating the account

## Background
- currently the app is making a request to update the account for guest users (when `accountId` is undefined)
  - this causes `customer-info-anon`endpoint to throw an `unauthorised` error

## QA

- log out
- Go to https://ubuntu-com-10220.demos.haus/advantage/subscribe?test_backend=true
- make a selection and press "Buy now"
- verify `customer-info-anon` isn't being called before you press "Continue"
- Verify `customer-info` and `preview` are called after you press "Continue" and correct price is displayed

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/172


anthonydillon (approved):
Code and QA LGTM :+1:

PR #10225, opened 2021-08-24
Add /advantage/user-subscriptions endpoint

albertkol:
## Done

- Adding `tests.advantage.helpers` make function: `make_subscription`, `make_contract`. 
These will create mock instances of primitive classes. These will be used for testing in reproducing scenarios.
- Add `advantage.ua_contract.builders` functions: `build_user_subscriptions`, `build_initial_user_subscriptions`, `build_final_user_subscriptions`. These functions have the goal of transforming step by step primitives into models. In this case `UserSubscription` models.
- Add `advantage.ua_contract.helpers`  functions. These are smaller functions used by `builders` in the process of creating model objects. 
- Add `GET /advantage/user-subscriptions` endpoint

## QA

- Have an account with susbcriptions
- Login on (`?test_backend=true` is optional): https://ubuntu-com-10225.demos.haus
- Go to: https://ubuntu-com-10225.demos.haus/advantage/user-subscriptions 
- You should see json response of your user_subscriptions
- Use the view as feature and look at other user's "user subscriptions": https://ubuntu-com-10225.demos.haus/advantage/user-subscriptions?email=<email_address>

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/184


webapp/advantage/ua_contracts/parsers.py
```diff
@@ -86,6 +87,7 @@ def parse_entitlements(raw_entitlements: dict) -> List[Entitlement]:
 
 def parse_contract_items(raw_items: dict) -> List[ContractItem]:
     items = []
+    raw_items = raw_items or []
```
carkod, line 90: I might be wrong, but if this is a dict, which I assume is `{}` but the alternative is an empty `[]` these are 2 different types? So shouldn't this be `raw_items or None`, since this is what you've done this in the other functions and args?

tests/advantage/helpers.py
```diff
@@ -49,3 +58,210 @@ def get_fixture(file: str) -> dict:
...
+    account_id: str = None,
+    marketplace: str = None,
+    period: str = None,
+    status: str = None,
+    last_purchase_id: str = None,
+    pending_purchases: List[str] = None,
+    items: List[SubscriptionItem] = None,
+) -> Subscription:
+    return Subscription(
+        id=id or "sAaBbCcDdEeFfGg",
```
carkod, line 74: I'm not going to ask you to change everything, but just as a suggestion isn't it more brief to say?:
```
def make_subscription(
    id: str = "aAaBbCcDdEeFfGg",
```
And then simply
```
id=id
```
If I'm right if `id` is `None` it will always go to the `or` option.

carkod:
I used Clement's email for user's subscriptions and got 500? Is this expected?

albertkol:
> I used Clement's email for user's subscriptions and got 500? Is this expected?

@ClementChaumel is always breaking my stuff. Let me check.

albertkol:
> I used Clement's email for user's subscriptions and got 500? Is this expected?

You get a 500 probably because you are missing the permission to view @ClementChaumel subscriptions. Sorry about it.

@ClementChaumel could you please QA and confirm that /advantage/user-subscriptions is working for you?

carkod:
> > I used Clement's email for user's subscriptions and got 500? Is this expected?
> 
> You get a 500 probably because you are missing the permission to view @ClementChaumel subscriptions. Sorry about it.
> 
> @ClementChaumel could you please QA and confirm that /advantage/user-subscriptions is working for you?

I can try someone else's, but Is this going to happen with all emails I try?

albertkol:
> > > I used Clement's email for user's subscriptions and got 500? Is this expected?
> > 
> > 
> > You get a 500 probably because you are missing the permission to view @ClementChaumel subscriptions. Sorry about it.
> > @ClementChaumel could you please QA and confirm that /advantage/user-subscriptions is working for you?
> 
> I can try someone else's, but Is this going to happen with all emails I try? I thought of using Clement's because I guess it's going to give me a lot more json data than anybody else.

Being able to view someone else's subscription is a special permission provided by the API team. I think they only gave a few of web team people permission to it.

webapp/advantage/ua_contracts/parsers.py
albertkol: I changed the `raw_items: dict` to `raw_items: List[Dict] = None` as it's more accurate.
Made this change for other parsers: `parse_entitlements`, `parse_subscriptions`, `parse_product_listings`.

`raw_items = raw_items or []` this line sets the `raw_items` to empty list, so that the loop on the next line won't throw a 500, because we try to iterate a None value.

Added tests for all these parsers in case the parameter received is a None value or an empty list.

tests/advantage/helpers.py
albertkol: When I wrote it, it went through my mind to do something like this. 

I did not want to have the hard-coded values set as default inside the function definition. So I decided to go for this option instead. Either version works, yours would probably be more optimal, as you don't need the `or` statements.

carkod: It does work, but aren't you doing the same thing with more code?

carkod (approved):

PR #10232, opened 2021-08-25
Certified fix bus identifier

carkod:
## Done

Only show identifier for usb and pci

## QA

Check that /certified/202010-28295 no longer shows identifier for bus not PCI or USB

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/10210

## Screenshots

fix-bus-identifier


templates/certified/model-details.html
```diff
@@ -118,7 +118,7 @@ <h4>Hardware</h4>
                   <tr>
                     <th colspan="2" class="u-text--muted">{{ hardware_subtitle }}</th>
                     <td colspan="8">{% for value in values %} 
-                      <p style="margin-bottom: 0.5rem;">{{ value.name }} {{ value.bus }}({{ value.identifier }})</p>
+                      <p style="margin-bottom: 0.5rem;">{{ value.name }} {{ value.bus }}{% if value.bus in ["usb", "pci"] %} ({{ value.identifier }}){% endif %}</p>
```
codersquid, line 121: I think I'd put the bus inside the conditional as well. I don't think it is valuable to show the bus type when there is no identifier.

codersquid (changes requested):

templates/certified/model-details.html
carkod: Sure. Done.

codersquid (approved):
Thanks for the fix.

PR #10241, opened 2021-08-27
use certified category pages instead of search

codersquid:
'See all' for each category should go to the category's landing page rather than a generic category search.

## QA

- Visit https://ubuntu-com-10241.demos.haus/certified
- Click on each 'See all' for each category and verify that it goes to each category's landing page rather than performing a search.

### DIFF
--- a/templates/certified/index.html
+++ b/templates/certified/index.html
@@ -104,7 +104,7 @@
         </li>
         {% endfor %}
       </ul>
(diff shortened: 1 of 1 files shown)


carkod:
@SirSamTumless I think @bethcollins92 can review this

bethcollins92 (approved):
LGTM - thanks for spotting and fixing that @codersquid

PR #10246, opened 2021-08-27
Load static data for vendor pages

carkod:
## Done

I took a guess looking at the [design](https://app.zeplin.io/project/60c1ca887b99f18a380e57e9/screen/60f169f4a5400b0f1506c19f), but obviously if the copy content they give us does not fit this data structure, we'll have to make changes.

But I thought this would be the best way to unblock this issue.

If @bethcollins92 thinks we can work with this, let's go ahead and review and merge, if not feel free to close this PR or suggest changes.

## QA

Best way to QA this at this point is to do it locally, as we don't have any content.

- Add `{{vendors_strip}}` to the `search-results.html` template, preferrably in the header, so you can see the json data clearly.
- Do the same for any category views e.g. `desktops.html` (the variable should be available in any of these templates)

## Issue / Card

Fixes https://github.com/canonical-web-and-design/web-squad/issues/4260


carkod:
@SirSamTumless @bethcollins92 

I took a guess looking at the [design](https://app.zeplin.io/project/60c1ca887b99f18a380e57e9/screen/60f169f4a5400b0f1506c19f), but obviously if the copy content they give us does not fit this data structure, we'll have to make changes.

But I thought this would be the best way to unblock this issue, since I will be on Vanilla squad.

If @bethcollins92 thinks we can work with this, let's go ahead and review and merge, if not feel free to close this PR or suggest changes.

webapp/certified/views.py
```diff
@@ -358,6 +360,12 @@ def certified_home():
         # Pagination
         total_results = models_response["meta"]["total_count"]
 
+        # Add static vendor data
+        with open("webapp/certified/vendors_data.yaml") as vendors_data:
+            vendors_strip = yaml.load(
```
carkod, line 365: Feel free to suggest a better name for this variable, I couldn't come up with anything else quickly, we are using the word "vendor" in so many variations...

bethcollins92: Maybe `vendors_info`? can we not use `vendors_data` as you've already used it?

bethcollins92 (approved):
LGTM for now, obviously will need to take another look when we get the vendor info

carkod:
So I am not sure what's the problem with the Cypress test, it's complaining about advantage not being able to buy a subscription but there is no advantage code changed here.

PR #10255, opened 2021-08-31
Display free token data from the new API

huwshimi:
## Done

- Update the free token card to display data from the new user subscriptions API.

## QA

- Visit [/advantage?test_backend=true](https://ubuntu-com-10255.demos.haus/advantage?test_backend=true).
- Check that the 'Free Personal Token' card is displayed in the list (both cards will show the free data for now). The data shown in the card should be exactly the same, but it is coming from the new API.

## Issue / Card

Fixes canonical-web-and-design/commercial-squad#209

### DIFF
--- a/static/js/src/advantage/api/enum.ts
+++ b/static/js/src/advantage/api/enum.ts
@@ -1,19 +1,27 @@
 export enum SupportLevel {
-  NONE = "n/a",
-  ESSENTIAL = "essential",
-  STANDARD = "standard",
-  ADVANCED = "advanced",
+  None = "n/a",
+  Essential = "essential",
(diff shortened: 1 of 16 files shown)


ClementChaumel (approved):
LGTM small nitpick 

whichever token is not selected gets a scrollbar which makes the content jump a bit.

huwshimi:
> LGTM small nitpick
> 
> whichever token is not selected gets a scrollbar which makes the content jump a bit.

Thanks, I just happened to notice that and pushed a fix to https://github.com/canonical-web-and-design/ubuntu.com/pull/10256.

PR #10256, opened 2021-08-31
Scroll to the subscription content if it is off the screen

huwshimi:
## Done

- Scroll to the subscription content if it is off the screen.

## QA

- Visit [/advantage?test_backend=true](https://ubuntu-com-10256.demos.haus/advantage?test_backend=true).
- Scroll down until all the subscription details is visible on the screen.
- Click on a subscription and the window shouldn't scroll.
- Click on the top subscription.
- Scroll down until just the free personal token card on the left is visible.
- Click on the free personal token card.
- You should get scrolled up until the top of the details is visible.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/93.

## Screenshots

scroll

### DIFF
--- a/static/js/src/advantage/react/components/Subscriptions/Content/Content.tsx
+++ b/static/js/src/advantage/react/components/Subscriptions/Content/Content.tsx
@@ -32,7 +32,7 @@ const Content = () => {
       // https://github.com/canonical-web-and-design/commercial-squad/issues/101
       // This only sets the selected token and does not set the modal to active
       // to prevent the modal appearing on first load on mobile.
-      setSelectedToken("ua-sub-123");
+      setSelectedToken("ua-sub-0");
     }
   }, [selectedToken, setSelectedToken]);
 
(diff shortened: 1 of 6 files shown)


huwshimi:
@wgx what do you think? Is this somewhat how you imagined it?

ClementChaumel (approved):

wgx:
@huwshimi Would it be possible to test this with a longer list of subscriptions. This behaviour is designed to fix an edge case where the user has 20 (or 100!) contract items in that list.

huwshimi:
> @huwshimi Would it be possible to test this with a longer list of subscriptions. This behaviour is designed to fix an edge case where the user has 20 (or 100!) contract items in that list.

I've updated this to have 20 ua subs (+ the free sub). At the moment the list is set to the height of the content on the right, but do we need that to allow more to appear on the page before it starts scrolling?

wgx:
@huwshimi thanks - I think this will work great, and is nice and usable. UX+1

PR #10272, opened 2021-09-01
Add trials label to /advantage table

ClementChaumel:
## Done

- Added back in the free trial row element
- Added a "Free trial label" before the free trial row

## QA

- Go to http://0.0.0.0:8001/advantage?test_backend=true with a free trial
- it should look like this
- it won't at the moment because the backend is not yet implemented

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/164

### DIFF
--- a/templates/advantage/table/_free-trial.html
+++ b/templates/advantage/table/_free-trial.html
@@ -1,0 +1,41 @@
+<div class="col-start-large-2 col-10">
+  {% if mobile %}
+    <hr class="p-separator" />
+  {% endif %}
+  <div class="row">
+    {% if contract['contractInfo']['status'] == "expired" %}
+
+      <div class="p-notification--caution">
+        <p class="p-notification__response">
+         <span class="p-notification__status">Your free trial expired.</span>You will lose access to this Ubuntu Advantage subscription on {{ contract['contractInfo']['grace_period_end'] }}.
+        </p>
+      </div>
+
+    {% endif %}
+
+    <div class="advantage-table-section">
+      <div class="advantage-table-instruction">
+        <span>Started on</span>
+        <p>{{ contract['contractInfo']['createdAtFormatted'] }}</p>
+        <hr />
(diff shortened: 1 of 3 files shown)


templates/advantage/table/_free-trial.html
```diff
@@ -0,0 +1,41 @@
+<div class="col-start-large-2 col-10">
+  {% if mobile %}
+    <hr class="p-separator" />
+  {% endif %}
+  <div class="row">
+    {% if contract['contractInfo']['status'] == "expired" %}
+
+      <div class="p-notification--caution">
+        <p class="p-notification__response">
+         <span class="p-notification__status">Your free trial expired.</span>You will lose access to this Ubuntu Advantage subscription on {{ contract['contractInfo']['grace_period_end'] }}.
```
petermakowski, line 10: Can't seem to find this in the copy doc. Could we add it there to make changes easier in the future?

petermakowski (commented):
Looks good, but it's not clear how can I test it if back-end isn't available yet. Do I need to alter a few lines of code in order to see it? Could you add instructions to the description 🙏

templates/advantage/table/_free-trial.html
ClementChaumel: I don't think any of the "update your subscription" copy made it into the copy doc.
Good call though! We should add them

ClementChaumel:
> Looks good, but it's not clear how can I test it if back-end isn't available yet. Do I need to alter a few lines of code in order to see it? Could you add instructions to the description

I think the easiest way to see what it looks like would be to reverse the condition 
```
{% if contract['is_trialled'] %}
```
to 
```
{% if not contract['is_trialled'] %}
```

Not great, I know :/

petermakowski (approved):
Maybe it's meant to be this way, but just wanted to note that the label isn't vertically aligned which is most apparent on mobile.

 image

PR #10276, opened 2021-09-01
Refactor advantage decorator

albertkol:
## Done

- Extract the `stripe_publishable_key` from the decorator and make it globally accessible to the front end
- Refactor the advantage_check decorator to instantiate the `UAContractsAPI` object
- Refactor views to use the new advantage_decorator

Documentation of the decorator: https://docs.google.com/document/d/10H9Fu-BvHEVzFvsfpP5czGMFUHf_JhnG_dXZJVML_JA/edit#

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/215

### DIFF
--- a/webapp/advantage/decorators.py
+++ b/webapp/advantage/decorators.py
@@ -1,0 +1,91 @@
+import os
+from distutils.util import strtobool
+from functools import wraps
+
+import flask
+from flask import g
+import talisker.requests
+
+from webapp.advantage.ua_contracts.api import UAContractsAPI
+from webapp.login import user_info
+
+PERMISSION_LIST = {
+    "user": "Endpoint needs logged in user.",
+    "user_or_guest": "Endpoint needs user or guest token.",
+}
+
+RESPONSE_LIST = {
+    "html": "Returns user friendly HTML response.",
+    "json": "Returns json response.",
+}
+
+
+def get_api_url(is_test_backend) -> str:
+    if is_test_backend:
+        return flask.current_app.config["CONTRACTS_TEST_API_URL"]
+
+    return flask.current_app.config["CONTRACTS_LIVE_API_URL"]
+
+
+def advantage_decorator(permission=None, response="json"):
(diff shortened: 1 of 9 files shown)


webapp/advantage/views.py
```diff
@@ -6,14 +6,14 @@
 
 from dateutil.parser import parse
 import flask
-import talisker.requests
 import pytz
+from flask import g
```
carkod, line 10: You can also do `from flask.g import api` if that's the only thing you are going to use

webapp/advantage/decorators.py
```diff
@@ -0,0 +1,91 @@
...
+            # init API instance
+            g.api = UAContractsAPI(
+                session=session,
+                authentication_token=(user_token or guest_token),
+                token_type=("Macaroon" if user_token else "Bearer"),
+                api_url=get_api_url(is_test_backend),
+            )
+
+            if response == "html":
+                g.api.set_is_for_view(True)
```
carkod, line 85: Shouldn't this be on the top, in line 50, so you save yourself from adding this duplicated condition?

carkod (approved):

webapp/advantage/views.py
albertkol: It would have been nice if I could do it. But `g.api` is a global value I defined in the decorator. You can set globals with any name using `flask.g`. But I cannot import `api`, as it is not a package/variable.

webapp/advantage/decorators.py
albertkol: Line 50 is checking for both permissions and response. 

I want `g.api.set_is_for_view(True)` to be executed even if permission is not set at all.

carkod: Yes but what I mean is that you could do something like:

```
if response == "html":
  # do the logic
  if permission == "user";
  # do the "html" and "user" logic
```
But it's not a major thing, I don't think we gain any performance advantages

PR #10277, opened 2021-09-02
Remove old contracts API

huwshimi:
## Done

- Remove old contracts API types, hooks, factories etc. that are no longer being used.

## QA

N/A none of this is being used so if the tests pass that should be enough.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/211.


ClementChaumel (approved):
LGTM

PR #10280, opened 2021-09-02
Remove inline muted heading content strip styles

### LINKED ISSUE #10237: Update muted headings to use <h4> as per design advice
The pattern below, please change the headings (CASE STUDY) from muted headings to `<h4>` as suggested [here](https://github.com/canonical-web-and-design/vanilla-framework/pull/3935#issuecomment-903651888)

Screenshot 2021-08-26 at 09 58 19

Searching the codebase for `<p class="p-muted-heading u-sv3" style="color: #fff; line-height: 1.5rem;">` should find the places we are using this content strip

bethcollins92:
## Done

- This was something we were going to add to Vanilla - but it was [decided](https://github.com/canonical-web-and-design/vanilla-framework/pull/3935#issuecomment-905288532) that the muted heading doesn't work with this size of icon. Therefore it needs to be replaced with a `<h4>`.

## QA

- View demo at: 
  - /0_docs/content-strips
  - /core
  - /managed/cassandra
  - /managed/kafka
  - /openstack
  - /security/esm
  - /kubernetes
  - /security/livepatch
- Check it replicates this [suggestion](https://github.com/canonical-web-and-design/vanilla-framework/pull/3935#issuecomment-903651888)

## Issue / Card

Fixes #10237 

## Screenshots

Example on /security/esm

Screenshot 2021-09-02 at 14 07 15

### DIFF
--- a/templates/0_docs/content_strips.html
+++ b/templates/0_docs/content_strips.html
@@ -44,7 +44,7 @@
               attrs={"class": "p-heading-icon__img"},
             ) | safe
           }}
-          <p class="p-muted-heading u-sv3" style="color: #fff; line-height: 1.5rem;">CASE STUDY</p>
+          <p class="p-heading--4">Case study</p>
         </div>
         <h4>ITstrategen keeps their applications secure with ESM</h4>
         <p>The security of customer data is of utmost importance to ITstrategen. That's why Ubuntu is their server
@@ -73,7 +73,7 @@
               attrs={"class": "p-heading-icon__img p-heading-icon__img--small"},
               ) | safe
             }}
-            <h4 class="p-heading-icon__title">Datasheet</h4>
+            <h4>Datasheet</h4>
           </div>
           <h3 class="p-heading--four"><a class="p-link--external p-link--inverted" href="https://assets.ubuntu.com/v1/8db65a1e-Kubernetes-for-the-Enterprise.pdf">Kubernetes for the enterprise</a></h3>
         </div>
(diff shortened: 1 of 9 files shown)


minkyngkm:
@bethcollins92 Just wondering if the`/kubernetes` page also needs to be updated in the same way 🤔

bethcollins92:
> @bethcollins92 Just wondering if the`/kubernetes` page also needs to be updated in the same way 🤔

Yes - it does, thank you! I'll update and let you know when you can check again. Let me know if you can think of anywhere else

bethcollins92:
Thanks @minkyngkm - updated

templates/core/index.html
```diff
@@ -287,7 +287,7 @@ <h1>Secure embedded devices</h1>
             ) | safe
             }}
           </div>
-          <p class="p-muted-heading u-sv3" style="color: #fff; line-height: 1.5rem;">whitepaper</p>
+          <p class="p-heding--4">Whitepaper</p>
```
minkyngkm, line 290: A minor typo "heding" to "heading"

```suggestion
          <p class="p-heading--4">Whitepaper</p>
```

templates/0_docs/content_strips.html
```diff
@@ -73,7 +73,7 @@ <h3>Webinars and case studies</h3>
               attrs={"class": "p-heading-icon__img p-heading-icon__img--small"},
               ) | safe
             }}
-            <h4 class="p-heading-icon__title">Datasheet</h4>
+            <h4>Datasheet</h4>
```
minkyngkm, line 76: Just wondering why in this page only `<h4>` is used here instead of `<p class="p-heading--4">`

bethcollins92: This page is a strange one as it's kind of secret to us, and it's not really a web page in a particular order that makes sense it's just a list of different examples. (I also noticed there was a `<h3>` before it so semantically it makes sense to have a `<h4>`)

templates/core/index.html
bethcollins92: Good spot thanks!

minkyngkm (approved):
LGTM! 🙌

PR #10285, opened 2021-09-02
Update recaptcha on the /legal/contributors/agreement form

pmahnke:
## Done

- Form assembly updated the recaptcha and I had to update the /legal/contributors/agreement form
- There was also an issue with the email the form was sending from, so I updated that on form assembly ([see this RT](https://rt.admin.canonical.com/Ticket/Display.html?id=132756))

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/legal/contributors/agreement
- See that the form isn't horrible looking (it needs a lot of css and classes from FA)
- See that the form submits in the inspector networking tab

### DIFF
--- a/templates/legal/contributors/agreement.html
+++ b/templates/legal/contributors/agreement.html
@@ -1,9 +1,16 @@
 {% extends "legal/_base_legal.html" %}
 
+{% block title %}Contributor agreement form | Contributors{% endblock %}
+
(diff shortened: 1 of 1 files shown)


bethcollins92 (approved):
LGTM with one small comment - the top border of the form is missing (not sure if this is intentional?)

Screenshot 2021-09-03 at 08 57 37

I checked in dev tools it's the `border-top-width: 0 !important;` on the fieldset causing it if you did want to keep the border.

Other than that the form looks good and successfully submits

PR #10304, opened 2021-09-06
Expiry notifications

huwshimi:
## Done

- Created a component to handle displaying expiry notifications.
- Added notifications to the header, subscription card and details panel.

## QA

I'm not sure if this can be QAed as I don't think the backoffice has a way to modify the purchase dates.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/90.
Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/104.
Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/117.

## Screenshots

Screen Shot 2021-09-06 at 10 56 51 am
Screen Shot 2021-09-06 at 5 14 31 pm
Screen Shot 2021-09-06 at 5 14 37 pm


static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+    expect(wrapper.find("[data-test='is_expired-large']").exists()).toBe(true);
+    expect(wrapper.find("[data-test='is_expiring-large']").exists()).toBe(
+      false
+    );
+  });
+
+  it("shows the highest priority notification when only showing one", () => {
+    const statuses = userSubscriptionStatusesFactory.build({
+      is_in_grace_period: true,
+      is_expired: true,
```
petermakowski, line 77: I noticed that you use snake case for lots of properties which seems inconsistent with how most JS code is structured in this repo (at least in advantage camelCase seems to dominate).

Is snake_case the direction we'd like to move towards now? It might be a good idea to ask other members of the team how they feel about the change (unless such conversation already happened!).

static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.tsx
```diff
@@ -0,0 +1,132 @@
...
+  NotificationProps
+>;
+
+type Messages = Record<
+  StatusKey,
+  { [K in ExpiryNotificationSize]?: { message: ReactNode; title?: ReactNode } }
+>;
+
+// The expiry status keys in priority order.
+const STATUS_KEYS = [
```
petermakowski, line 37: What do you think about renaming this variable instead of relying on a comment? Some examples below.
```suggestion
const STATUS_KEYS_BY_PRIORITY = [
```

```suggestion
const ORDERED_STATUS_KEYS
```

static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.tsx
```diff
@@ -0,0 +1,132 @@
...
+      title: "Your subscription has expired.",
+    },
+    [ExpiryNotificationSize.Small]: {
+      message: "Subscription expired",
+    },
+  },
+};
+
+const ExpiryNotification = ({
+  multiple = false,
```
petermakowski, line 84: I noticed that the `is`/`has` naming convention for booleans is followed in multiple places at canonical, not only this repo, but also react-components. Might be a good idea to rename to `shouldAllowMultiple` or similar for consistency.

static/js/src/advantage/react/components/Subscriptions/Notifications/Notifications.test.tsx
```diff
@@ -21,7 +21,7 @@ describe("Notifications", () => {
     expect(wrapper.find("[data-test='pendingPurchase']").exists()).toBe(true);
   });
 
-  it("does not diplay a pending purchase notification when there is no id", () => {
+  it("does not display a pending purchase notification when there is no id", () => {
```
petermakowski, line 24: NIce job on fixing small issues as you go. 👍

static/js/src/advantage/react/hooks/useUserSubscriptions.ts
```diff
@@ -30,6 +33,31 @@ export const selectFreeSubscription = (subscriptions: UserSubscription[]) =>
...
+ */
+const hasStatus = (
+  subscriptions: UserSubscription[],
+  status: keyof UserSubscriptionStatuses
+) => subscriptions.some(({ statuses }) => statuses[status]);
+
+/**
+ * Get a summary of statuses for all subscriptions.
+ */
+export const selectStatusesSummary = (
```
petermakowski, line 47: As a side note, I find it slightly confusing that these functions are placed in `hooks` directory as technically they're not hook but selector functions.

static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+      is_in_grace_period: true,
+    });
+    const wrapper = mount(
+      <ExpiryNotification
+        size={ExpiryNotificationSize.Large}
+        statuses={statuses}
+      />
+    );
+    expect(
+      wrapper.find("[data-test='is_in_grace_period-large']").exists()
```
petermakowski, line 48: I think we'd get more value from this test if we were testing results that the user actually cares about instead (actual text that's displayed).

static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+      is_expired: true,
+    });
+    const wrapper = mount(
+      <ExpiryNotification
+        size={ExpiryNotificationSize.Large}
+        statuses={statuses}
+      />
+    );
+    expect(wrapper.find("[data-test='is_expired-large']").exists()).toBe(true);
+  });
```
petermakowski, line 35: I think we'd get more value from this test if we were testing results that the user actually cares about instead (actual text that's displayed).

What you're testing here is that your logic for generatic test ids is functioning properly. It's possible for this test to pass even if the large/small notification isn't displayed correctly.

Parts that are probably more important than a test ID aren't covered (like displaying a title for large notifications).

petermakowski (approved):
Good to see some more TypeScript. Added quite a few below and would appreciate if you could address/reply. Don't want to block you though, so approving this ✅

static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.test.tsx
huwshimi: These values are from the API, they're in snake case and we don't transform the values to/from camel case as we interact with the API.

In this case the factory is building an API response (so even if we did transform to/from camel case the factory would still use snake case to mimic the API).

In the past we had some projects that transformed the case for the API values but we've found it much easier and less error prone to work with the data as it's sent/received by the API.

There's no intention to use snake case for anything other than API data.

static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.tsx
huwshimi: Good idea!

static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.tsx
huwshimi: Good idea. Done.

static/js/src/advantage/react/hooks/useUserSubscriptions.ts
huwshimi: Yeah I wasn't quite sure where to put them. They're designed to work with the React Query select property (I can't link directly to the select docs, but it's on this page: https://react-query.tanstack.com/reference/useQuery) so I left them in the same file as the hook they're used with.

They're used like this: https://github.com/canonical-web-and-design/ubuntu.com/blob/ua-subscriptions/static/js/src/advantage/react/components/Subscriptions/SubscriptionList/SubscriptionList.tsx#L29

Maybe it would make sense to create a dir for API endpoint with a hook and selector file.

static/js/src/advantage/react/components/Subscriptions/ExpiryNotification/ExpiryNotification.test.tsx
huwshimi: Thanks I really did write some terrible tests here. I reckon this is a tricky one to find a balance between writing maintainable tests and testing what we're actually producing. I'll see what I can do to fix these up.

PR #10317, opened 2021-09-07
Changes to modal step 1

ClementChaumel:
## Done

- Updated step 1 according to [the issue](https://github.com/canonical-web-and-design/commercial-squad/issues/167)
- Refactored the steps of the modal
- Removed any calls to update customer info anon as we don't display any preview related content in the first step anymore.
 

## QA

- go on [/advantage/subscribe?test_backend=true](https://ubuntu-com-10317.demos.haus/advantage/subscribe?test_backend=true)
- check that the first step matches the specs
- check that I didn't break anything

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/167

## Screenshots

image


static/js/src/advantage/subscribe/react/PurchaseModal.jsx
```diff
@@ -235,102 +100,26 @@ const PurchaseModal = () => {
...
+          />
+        )}
+        <button
+          ref={buttonRef}
+          aria-controls="purchase-modal"
+          aria-hidden="true"
+          style={{ position: "absolute", left: "-9999px", opacity: 0 }}
+        >
+          Cancel
+        </button>
```
ClementChaumel, line 120: This is a workaround to keep and use the event listener set by the js form. (The displaying of the modal is done by the form behind it)

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
```diff
@@ -199,34 +88,9 @@ const PurchaseModal = () => {
...
-              details and try again. Contact{" "}
-              <a href="https://ubuntu.com/contact-us">Canonical sales</a> if the
-              problem persists.
-            </>
-          );
-        }
-      },
-    });
+  const closeModal = () => {
+    console.log(buttonRef.current);
```
petermakowski, line 92: Probably shouldn't go to production 🙈

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
petermakowski: Could we display the elements that can trigger the modal via react instead/connect them with the redux store to avoid such workarounds?

static/js/src/advantage/subscribe/react/app.tsx
```diff
@@ -18,13 +27,13 @@ Sentry.init({
...
 
-const stripePromise = loadStripe(window.stripePublishableKey);
+const stripePromise = loadStripe(window.stripePublishableKey ?? "");
 const oneHour = 1000 * 60 * 60;
 const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       refetchOnWindowFocus: false,
-      refetchOnmount: false,
+      refetchOnMount: false,
```
petermakowski, line 36: nice to see more TS 👍 
seems like you found a bug too? Was`refetchOnmount` with incorrect casing even working before?

static/js/src/advantage/subscribe/react/components/ModalParts/ModalBody.tsx
```diff
@@ -0,0 +1,35 @@
+import React from "react";
+import { Col, Notification, Spinner } from "@canonical/react-components";
+
+type ModalBody = {
+  error?: React.ReactNode;
+  children: React.ReactChild;
```
petermakowski, line 6: Any reason why you opted for `ReactChild` instead of `ReactNode`? It's generally best to use `React.ReactNode` for typing children. 
https://github.com/typescript-cheatsheets/react#useful-react-prop-type-examples
```suggestion
  children: React.ReactNode;
```

static/js/src/advantage/subscribe/react/components/ModalParts/ModalFooter.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { Row, ActionButton } from "@canonical/react-components";
+
+type ModalFooter = {
+  children: React.ReactChild;
```
petermakowski, line 5: As above

static/js/src/advantage/subscribe/react/components/ModalSteps/StepOne.tsx
```diff
@@ -0,0 +1,55 @@
...
+import useStripeCustomerInfo from "../../APICalls/useStripeCustomerInfo";
+import PaymentMethodForm from "../PaymentMethodForm";
+import useProduct from "../../APICalls/useProduct";
+import usePreview from "../../APICalls/usePreview";
+import ModalHeader from "../ModalParts/ModalHeader";
+import ModalBody from "../ModalParts/ModalBody";
+import ModalFooter from "../ModalParts/ModalFooter";
+
+type StepOneProps = {
+  error: React.ReactNode | null;
```
petermakowski, line 14: `ReactNode` can be `null` be default
```suggestion
  error: React.ReactNode;
```
> type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;

petermakowski (commented):
Seems like this change means that we no longer display VAT number validation to the user  - ever. Just making sure that this what we want?

static/js/src/advantage/subscribe/react/app.tsx
ClementChaumel: I doubt it was! Thanks TS!

static/js/src/advantage/subscribe/react/components/ModalParts/ModalBody.tsx
ClementChaumel: I was fooled by the name! Thanks!

static/js/src/advantage/subscribe/react/components/ModalParts/ModalFooter.tsx
ClementChaumel:
```suggestion
  children: React.ReactNode;
```

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
ClementChaumel: :see_no_evil:

ClementChaumel:
> Seems like this change means that we no longer display VAT number validation to the user - ever. Just making sure that this what we want?

I think that's ok since we ignore incorrect VAT numbers anyway. @wgx ?

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
ClementChaumel: Yes, but to do that the easiest thing would be to bring all the product selector to the react app. And for the record, I'm planning to do that in a near future*

albertkol:
I am not sure what changes are made here, but since I am using this modal a lot for testing, I'd like to ask a few questions.

There seem to be a lot of big changes happening here. Are these affecting only the free trial flow or also the normal purchase? @ClementChaumel 

> Removed any calls to update customer info anon as we don't display any preview related content in the first step anymore

The  `customer info anon` calls are made to save the new VAT information immediately after they change. Then we call the `preview` to get an updated total cost with TAX included on step 1.

Do we no longer wish to display the total cost with TAX on step 1 any more? @wgx @ClementChaumel 

> Seems like this change means that we no longer display VAT number validation to the user - ever. Just making sure that this what we want?

For as much as I am aware, while we do validate VAT when we call `customer info anon`, we never really tell the user, that there was an error specifically with the VAT. We just say something generic. I thought this was something we could actually improve on.

> I think that's ok since we ignore incorrect VAT numbers anyway?

Not sure what you mean by ignore. @wgx do we not care to store VAT info for the new opt-out trials. If I remember right for opt-in trials we decided against it.

ClementChaumel:
> There seem to be a lot of big changes happening here. Are these affecting only the free trial flow or also the normal purchase? @ClementChaumel

This affects both free trial and regular purchase

> Do we no longer wish to display the total cost with TAX on step 1 any more? @wgx @ClementChaumel

That's correct

> For as much as I am aware, while we do validate VAT when we call `customer info anon`, we never really tell the user, that there was an error specifically with the VAT. We just say something generic. I thought this was something we could actually improve on.

We do check on the frontend but only for logged in users. (as we need an email address to be able to call that endpoint)
image

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
```diff
@@ -199,34 +88,10 @@ const PurchaseModal = () => {
...
-              problem persists.
-            </>
-          );
-        }
-      },
-    });
+  const closeModal = () => {
+    // I have no idea what's going on here, but without it the modal won't close for guest users. ¯\_(ツ)_/¯
+    buttonRef.current.click();
+    buttonRef.current.click();
```
petermakowski, line 94: We should fix this before merge. I'll try and look into this as well to get a better idea what's going on here.

petermakowski:
You have some prettier errors to fix in `usePostCustomerInfoForPurchasePreview.tsx` @ClementChaumel

ClementChaumel:
> You have some prettier errors to fix in `usePostCustomerInfoForPurchasePreview.tsx` @ClementChaumel

For some reason my VScode plugin just dies sometimes and stop running it on save

static/js/src/advantage/subscribe/react/PurchaseModal.jsx
petermakowski: image
@ClementChaumel  Looks like you have a duplicate `Cancel` button (each step has a ModalFooter + you define an extra one in `PurchaseModal`.

petermakowski: Please check if this solves the issue 🙏 https://github.com/ClementChaumel/ubuntu.com/pull/1

ClementChaumel: Yeah, so the clickHandler is set by the vanilla JS form. and I found that it only managed to attach to a button declared on PurchaseModal.jsx. And not any children so what I'm doing here is to create an invisible cancel button on PurchaseModal and then create a function that clicks this button whenever any of the other two (from ModalFooter) is clicked.

But it's proved flaky at best. I think I'm going to switch to just adding the closeModal function to the window object instead.
Again, not ideal. The best thing to do would be to just incorporate the form in the react app. But then again we already use the window object for a bunch of stuff so I figure it's not a horrible solution? What do you think?

static/js/src/advantage/subscribe/listeners/ui-event-listeners.js
```diff
@@ -16,16 +16,15 @@ export default function initUIControls(store) {
     });
   });
 
-  // Find every "purchase-modal" control (buy button and cancel button)
-  // and make them toggle the display of the purchase modal when clicked
+  // Add close modal function to window object so the React modal can use it
+  window.handleClosePurchaseModal = () => {
```
petermakowski, line 20: Let's leave this as is for now, but please refactor as soon as you have some time.

I don't see a reason why the two below coudn't be merged into one entry point with a single redux store:

```js
{ productSelector: "./static/js/src/advantage/subscribe/product-selector.js",
uaSubscribe: "./static/js/src/advantage/subscribe/react/app.jsx", }
```

static/js/src/advantage/subscribe/react/components/PaymentMethodForm.jsx
```diff
@@ -100,8 +99,8 @@ function PaymentMethodForm({ setCardValid }) {
           });
         }
       },
-      onSuccess: () => {
-        queryClient.invalidateQueries("preview");
+      onSettled: () => {
+        setSubmitting(false);
```
petermakowski, line 103: Doesn't `setSubmitting(false)` need to be called in `onError`, too?

petermakowski (approved):

static/js/src/advantage/subscribe/listeners/ui-event-listeners.js
petermakowski, line 20:
```suggestion
  window.handleTogglePurchaseModal = () => {
```

static/js/src/advantage/subscribe/react/app.tsx
```diff
@@ -8,6 +8,16 @@ import { Elements } from "@stripe/react-stripe-js";
 import { loadStripe } from "@stripe/stripe-js";
 import PurchaseModal from "./PurchaseModal";
 
+declare global {
+  interface Window {
+    stripePublishableKey?: string;
+    isGuest?: boolean;
+    accountId?: string;
+    previousPurchaseIds?: string[];
+    handleClosePurchaseModal?: () => void;
```
petermakowski, line 17:
```suggestion
    handleTogglePurchaseModal?: () => void;
```

static/js/src/advantage/subscribe/react/components/ModalParts/ModalFooter.tsx
```diff
@@ -0,0 +1,27 @@
...
+const ModalFooter = ({ children }: ModalFooter) => {
+  return (
+    <footer className="p-modal__footer">
+      <Row className="u-no-padding">
+        <ActionButton
+          className="col-small-2 col-medium-2 col-start-medium-3 col-start-large-7 col-3 u-no-margin"
+          appearance="neutral"
+          aria-controls="purchase-modal"
+          style={{ textAlign: "center" }}
+          onClick={window.handleClosePurchaseModal}
```
petermakowski, line 17:
```suggestion
          onClick={window.handleTogglePurchaseModal}
```

petermakowski (commented):
This should probably be `handleTogglePurchaseModal`, you can open the modal as well if you call the function again.

static/js/src/advantage/subscribe/react/components/PaymentMethodForm.jsx
ClementChaumel: OnSettled is called both on Error and on Success!

petermakowski: I see. All good then 👍

static/js/src/advantage/subscribe/listeners/ui-event-listeners.js
ClementChaumel: I tried something similar but couldn't get it to work

PR #10326, opened 2021-09-08
Ensure drawer is closed on page reload (small screens)

### LINKED ISSUE #10305: Filter drawer opens on page reload certified search results
See comment [here](https://github.com/canonical-web-and-design/ubuntu.com/pull/10278#issuecomment-912928621).

bethcollins92:
## Done

- Added a small amount of right padding to the buttons to ensure there is space between them and the right side of the drawer on small screens
- Added a function to ensure the drawer closes when the page reloads after applying/clearing filters
- Noticed the clear filters button on the generic search page takes you back to `/certified` which isn't the expected behaviour but this is out of scope for this PR. Issue created [here](https://github.com/canonical-web-and-design/ubuntu.com/issues/10325)

## QA

- View page at: https://ubuntu-com-10326.demos.haus/certified/desktops or /certified/laptops or /certified/devices (any in the secondary nav)
- Reduce your screen size until you see the toggle drawer button. 
- Click the button to open the filter drawer
- Apply some filters and click "Apply"
- When the page reloads with the new filters applied the filter drawer should be closed. 
- Click toggle drawer button again and select "Clear" to clear the filters
- On page reload the drawer should close.

## Issue / Card

Fixes #10305

### DIFF
--- a/static/js/src/certified-search-results.js
+++ b/static/js/src/certified-search-results.js
@@ -143,10 +143,10 @@ function toggleShowAllLinks() {
 }
 
 function hideDrawerPageReload() {
-  const closeDrawerButton = document.querySelector(
-    ".p-side-navigation__toggle--in-drawer"
-  );
-  closeDrawerButton.click();
+  if (window.location.href.includes("drawer")) {
+    const closeDrawerButton = document.querySelector("#toggle-filters");
+    closeDrawerButton.click();
+  }
 }
 
 enableApplyFilters();
@@ -154,3 +154,4 @@ toggleVersionsList();
 toggleVendorsList();
 toggleShowAllLinks();
 updateResultsPerPage();
+hideDrawerPageReload();
--- a/static/js/src/certified-search-results.js
+++ b/static/js/src/certified-search-results.js
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 
 function clearFilters() {
+  hideDrawerPageReload();
   let objUrl = new URL(window.location);
   objUrl.search = "";
   window.location.assign(objUrl);
@@ -139,6 +140,13 @@ function toggleShowAllLinks() {
       }
     });
   });
+}
+
+function hideDrawerPageReload() {
+  const closeDrawerButton = document.querySelector(
+    ".p-side-navigation__toggle--in-drawer"
+  );
+  closeDrawerButton.click();
 }
 
 enableApplyFilters();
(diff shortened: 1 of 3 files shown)


static/js/src/certified-search-results.js
```diff
@@ -141,6 +142,13 @@ function toggleShowAllLinks() {
   });
 }
 
+function hideDrawerPageReload() {
+  const closeDrawerButton = document.querySelector(
+    ".p-side-navigation__toggle--in-drawer"
```
carkod, line 147: I would use an `id` here? Would it be possible that another HTML item maybe around with the same class before this one?

templates/certified/search-results.html
```diff
@@ -136,9 +136,9 @@ <h2 class="p-heading--4">
               </ul>
               <hr style="margin-bottom: 1rem;">
             </aside>
-            <div class="u-align--right">
+            <div class="u-align--right" style="padding-right: 0.5rem">
               <button class="p-button--neutral p-update-results u-no-margin--right" onclick="return clearFilters()">Clear</button>
-              <button class="p-button--positive p-update-results js-apply-filters" type="submit" disabled>Apply</button>
+              <button class="p-button--positive p-update-results js-apply-filters" type="submit" onclick="hideDrawerPageReload()" disabled>Apply</button>
```
carkod, line 141: Shouldn't we execute this only when necessary? When the toggle actually shows?

carkod:
I couldn't find any examples in Vanilla, but is this toggle a Vanilla component? I was thinking maybe this needs to be fixed in Vanilla also, I have seen this issue somewhere else, I believe in Anbox-app, so maybe a generic fix would be better

carkod (approved):

carkod:
In this case, if you apply filters in a small screen size, after the panel closes, the results do get out of the viewport after reloading the page. But like I said last time, I think this is default browser behavior.

static/js/src/certified-search-results.js
bethcollins92: fixed

templates/certified/search-results.html
bethcollins92: I've updated the function - let me know if this is what you meant

static/js/src/certified-search-results.js
```diff
@@ -141,8 +142,19 @@ function toggleShowAllLinks() {
   });
 }
 
+function hideDrawerPageReload() {
+  const applyFilters = document.querySelectorAll(".js-apply-filters");
+  applyFilters.forEach((button) => {
```
carkod, line 147: Are we using this argument anywhere? If not maybe we should remove it. If we had linting, this probably would be highlighted

bethcollins92: Good spot - I thought I was going to add an event listener but then realised I just want it to hide on page reload so didn't need it. removed.

PR #10327, opened 2021-09-08
Account users/user row

petermakowski:
## Done

- Add User Row with 'regular', 'editing' and 'disabled' state
  - Actions do not interact with any backend, edit, deletion functionality etc. will be implemented separately
- add `@testing-library/jest-dom` (already recommended by @hatched here: https://docs.google.com/document/d/1ysVlPByzVY-NBUjGic1WupHojUl30P6cf3Xe-OVRRPg/edit#heading=h.qqiuv5t530vo)
- refactor `setupTests` to TS and add to `tsconfig.json` to enable TS suggestions.

### regular:
image
### edit:
image
### disabled:
image

## QA

- Log in
- Go to https://ubuntu-com-10327.demos.haus/advantage/users
- confirm the state of each row can be toggled via edit/cancel/save buttons

## Issue / Card

https://github.com/canonical-web-and-design/commercial-squad/issues/199

## Screenshots
image
image


static/js/src/advantage/users/components/TableView/TableView.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    { ...mockUser, id: "3", email: "angela@ecorp.com", role: "Technical" },
+  ];
+
+  render(<TableView users={users} />);
+  screen.getByLabelText("Edit user karen@ecorp.com").click();
+  expect(screen.getByLabelText("Edit user angela@ecorp.com")).toBeDisabled();
+
+  screen.getByRole("button", { name: "Cancel" }).click();
+
+  await waitFor(() =>
```
hatched, line 37: 👍

static/js/src/advantage/users/components/TableView/TableView.tsx
```diff
@@ -0,0 +1,58 @@
...
+  if (userInEditMode === null) {
+    return "regular";
+  } else if (userId !== userInEditMode) {
+    return "disabled";
+  } else {
+    return "editing";
+  }
+};
+
+const TableView: React.FC<{ users: Users }> = ({ users }) => {
```
hatched, line 22: Typically we rely on the TS inference for React component definitions. So you should be able to remove `React.FC` the compiler will infer the correct types with this syntax:

```suggestion
type Props = {
  users: Users;
}

const TableView = ({ users }: Props) => {
```

Could you also make this change elsewhere in this PR plz. You can read more about how `React.FC` [is not necessary here](https://github.com/facebook/create-react-app/pull/8177)

hatched (changes requested):
Thanks @petermakowski just a few easy changes requested below :)

static/js/src/advantage/users/components/TableView/TableView.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  screen.getByText("user@ecorp.com");
+  screen.getByText("Admin");
+  screen.getByText("10/01/2020");
+  screen.getByText("15/02/2021");
+});
+
+it("allows to edit only a single user at a time", async () => {
+  const users: User[] = [
+    { ...mockUser, id: "1", email: "karen@ecorp.com", role: "Billing" },
```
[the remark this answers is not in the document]
petermakowski: Do you mean by clone to "copy and paste the entire object"? Because I'm already using the spread operator which clones each value of the object (as they're all primitives - strings) instead of pointing to reference.

I think I might use a JSON file for mock data instead. 🤔

static/js/src/advantage/users/components/TableView/TableView.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    createdAt: "2020-01-10T10:00:00Z",
+    lastLoginAt: "2021-02-15T13:45:00Z",
+  };
+
+  render(<TableView users={[testUser]} />);
+
+  screen.getByText("user@ecorp.com");
+  screen.getByText("Admin");
+  screen.getByText("10/01/2020");
+  screen.getByText("15/02/2021");
```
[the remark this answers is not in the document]
petermakowski: If you know that `getBy` throws an error if nothing's returned (as that's how the library works), expect seems redundant.

But I agree, better to be explicit.

ClementChaumel:
Would it be possible to add some spacing inside the rows so it doesn't jump when it switches to edit mode?

ClementChaumel:
The card view doesn't have labels for the rows. It's a bit confusing what each is 
image

petermakowski:
> Would it be possible to add some spacing inside the rows so it doesn't jump when it switches to edit mode?

I assumed it's done on purpose - to fit in more rows at a time. But it would make for a smoother UX when editing (unless I add some transitions for animating the height, but that would require quite a lot of effort - tables aren't the easiest to animate). Thoughts  @wgx ?

wgx:
> > Would it be possible to add some spacing inside the rows so it doesn't jump when it switches to edit mode?
> 
> I think it's a good idea but I assumed it's done on purpose - to fit in more rows at a time. Thoughts @wgx ?

The 'jump' doesn't affect usability, but I agree it would be nice to vertically space the rows *as if* they are the height of the Select - to remove the jumping. Not essential, but would look smoother. @petermakowski

petermakowski:
> > > Would it be possible to add some spacing inside the rows so it doesn't jump when it switches to edit mode?
> > 
> > 
> > I think it's a good idea but I assumed it's done on purpose - to fit in more rows at a time. Thoughts @wgx ?
> 
> The 'jump' doesn't affect usability, but I agree it would be nice to vertically space the rows _as if_ they are the height of the Select - to remove the jumping. Not essential, but would look smoother. @petermakowski

Okay, will update. Thanks!

static/js/src/advantage/users/components/TableView/components.tsx
```diff
@@ -0,0 +1,172 @@
...
+        Cancel
+      </Button>
+      <Button
+        small
+        dense
+        appearance="positive"
+        onClick={() => handleEditSubmit(user.id)}
+      >
+        Save
+      </Button>
```
ClementChaumel, line 53: You might want to use the [ActionButton](https://canonical-web-and-design.github.io/react-components/?path=/docs/actionbutton--default-story) here so you can use its loading state while you wait for a response from the backend

ClementChaumel (approved):
:+1:

ClementChaumel:
Oh, I also realised that the target for this PR is master. Is that correct? Should this go on a feature branch instead?

petermakowski:
> Oh, I also realised that the target for this PR is master. Is that correct? Should this go on a feature branch instead?

I think it's best to merge everything incrementally to master (agile and all that). We don't link to this URL from anywhere anyway. Discussed this with @anthonydillon a while ago and he agreed.

static/js/src/advantage/users/components/TableView/components.tsx
```diff
@@ -0,0 +1,192 @@
+import React from "react";
+import { format } from "date-fns";
+import CSS from "csstype";
```
hatched, line 3: I don't see this dep in the package.json. It's best to be explicit about our dependencies, can you add it there plz.

static/js/src/advantage/users/components/TableView/components.tsx
```diff
@@ -0,0 +1,192 @@
...
+          <i className="p-icon--delete" aria-label="delete"></i>
+        </button>
+      ) : null}
+    </div>
+  );
+};
+
+type UserRoleProps = UserVariantProps;
+
+const getVisibilityStyles = (isVisible: boolean): CSS.Properties => ({
```
hatched, line 81: Typically we don't do any styles in JS but I'll leave this decision ultimately up to the project owners :)

hatched (approved):

static/js/src/advantage/users/components/TableView/components.tsx
petermakowski: Will do!

static/js/src/advantage/users/components/TableView/components.tsx
petermakowski: CSS-in-JS fan here 🙈 
But I do try to reuse the existing styles/use vanilla classes whenever possible.

PR #10328, opened 2021-09-08
User subscription in trial

albertkol:
## Done

- Update /advantage/user-subscriptions with the new trial rules
- The trial user subscription should be listed as its own item on /advantage and not be aggregated with the rest of the purchases for the same product listing

## QA

- Will be easier to test when trials are out.

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/224


ClementChaumel (approved):
LGTM

PR #10331, opened 2021-09-09
Get the test backend flag from the URL

huwshimi:
## Done

- Get the test backend flag from the URL.
- Remove old code for fetching the flag from the window.

## QA

- Visit [/advantage?test_backend=true](https://ubuntu-com-10331.demos.haus/advantage?test_backend=true) and log in
- Open your network dev tools.
- Check that there is a call to `/advantage/user-subscriptions?test_backend=true`.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/225.


static/js/src/advantage/react/hooks/useURLs.test.tsx
```diff
@@ -5,6 +5,14 @@ import { QueryClient, QueryClientProvider } from "react-query";
 
 import * as useURLsModule from "./useURLs";
 
+const setSearch = (search: string) => {
+  const location = { ...window.location, search };
+  Object.defineProperty(window, "location", {
+    writable: true,
+    value: location,
+  });
```
petermakowski, line 13: We should clean up after this test. Modifying globals directly can affect other tests and lead to some hard-to-debug issues.

I'd suggest using a combination of `jest.spy` and `mockImplementation` (or similar) instead.
https://jestjs.io/docs/jest-object#jestspyonobject-methodname

petermakowski (changes requested):

static/js/src/advantage/react/hooks/useURLs.test.tsx
huwshimi: Thanks @petermakowski, that was my initial idea too (to use a `spyOn`) but unfortunately TypeScript won't allow a spy to be assigned to the location.

I've updated the tests to restore the location after each test. The other approach would be to use a dep such as this: https://www.npmjs.com/package/jest-location-mock to allow us to use `location.assign()` but I was trying to avoid introducing more deps.

petermakowski (approved):

PR #10336, opened 2021-09-09
added fix for search function on /tutorials

mtruj013:
## Done

Changed key in `views.py` line 570 from "topic" to "link" to stop search from returning 505 on /tutorials

## QA

- View the site locally in your web browser at: https://ubuntu-com-10336.demos.haus/tutorials
- Run a search

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/10323

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -567,7 +567,7 @@ def build_tutorials_index(session, tutorials_docs):
                     identifier = result["link"][start:end]
                     if start != -1:
                         for doc in tutorials:
-                            if identifier in doc["topic"]:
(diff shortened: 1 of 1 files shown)


mtruj013:
@jkfran Hi! Just wondering if you think this is an acceptable fix for this issue? We couldn't figure out why `topic` stopped working in the first place, is that something you think should be investigated?

carkod (approved):
Strange thing is how is it working when the topic doesn't even exist as a property... It fails with any search on my local
LGTM Thanks!

PR #10340, opened 2021-09-10
Gray out the logo and secondary nav

### LINKED ISSUE #9914: expanding search: hide nav elements
when the search is expanded it should be the only item "visible" - everything else should have the `-overlay` class.

How it looks like: Global nav, Ubuntu logo and secondary nav are "visible" and it should be grayed out

Screenshot 2021-06-17 at 11 52 53

Screenshot 2021-06-17 at 11 52 00

---
*Reported from: https://ubuntu.com/*

minkyngkm:
## Done

- Made Ubuntu logo and secondary nav grayed out when search bar is active 

## QA

## Issue / Card

Fixes #9914 

## Screenshots

image

### DIFF
--- a/templates/templates/_navigation.html
+++ b/templates/templates/_navigation.html
@@ -2,7 +2,6 @@
 
 <header id="navigation" class="p-navigation">
   {% block header_banner %}
-  <div class="dropdown-window-overlay fade-animation"></div>
   <div class="p-navigation__row">
     <div class="p-navigation__banner">
       <div class="p-navigation__logo">
@@ -78,7 +77,7 @@
   </div>
   {% endblock header_banner %}
 </header>
-<div class="main-content dropdown-window-overlay fade-animation"></div>
+<div class="dropdown-window-overlay fade-animation"></div>
 <div class="dropdown-window slide-animation">
   <div class="u-hide" id="enterprise-content"></div>
   <div class="u-hide" id="developer-content">  </div>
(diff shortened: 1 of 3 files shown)


minkyngkm:
@anthonydillon @anasereijo Should the global nav on the very top be grayed out? Because the navbar is from dependency @canonical/global-nav

anthonydillon:
This has affected the mega menu:
image

anthonydillon (changes requested):
QA issue

anasereijo:
> @anthonydillon @anasereijo Should the global nav at the very top be grayed out also? Because the navbar is from dependency @canonical/global-nav

Ideally yes, the global nav should also be grayed out, but I understand if it's not possible to do it or very complicated so willing to leave it like that.

anthonydillon:
Thanks @minkyngkm looking better. Found another issue. 
- Open and close the Product menu in the global nav
- Click search and you can see the content below the global nav
image

minkyngkm:
Thanks for QAing @anthonydillon  - Fixed that issue but I had to give some margins to the dropdown menu. When the search bar is active, the dropdown menu is open from the product button and then the screen gets smaller, the dropdown menu is not completely hidden because of the global nav bar's opacity. 
Screenshot 2021-09-13 at 10 52 40 am

It was a little bit tricky to update the styles because the global nav is from dependency. It looks okay for me but since the margin is added slightly, it might need some design review. Please let me know what you think :) 
image

anthonydillon:
Thanks, that is working now but there is a missing overlay behind the mega menu:
image

anthonydillon:
Awesome work @minkyngkm this is looking great. One small (I hope last thing) is that when you open the download menu and click the content overlay it does not close the menu. Is that something you could look into please.

minkyngkm:
Hi @anthonydillon. Thanks for reviewing! :) I had to update the secondary menu(The red area in the screenshot below)'s z-index.
Screenshot 2021-09-15 at 12 35 24 pm
It seems to be working fine but its grayed out a little bit later compared to other components because it just changes z-index. Would there be any better way to fix it other than changing z-indexes? Also, when the secondary menu is open and the search bar is active and then if you click the menu area, the menu is closed with the search bar because of the search bar's z-index I guess... Would that be okay?

anthonydillon:
@minkyngkm I see the styles added to the global nav are causing issues and its still clickable so I would suggest removing the global nav masking and leaving it clickable for now.

minkyngkm:
> @minkyngkm I see the styles added to the global nav are causing issues and its still clickable so I would suggest removing the global nav masking and leaving it clickable for now.

@anthonydillon Hi Ant! Thank you for reviewing 🙌. I fixed the global nav issue. It seems to be fixed - Could you please look once again and then I will remove it if it's still not working! Thank you again 🙏🏻

templates/templates/_navigation.html
```diff
@@ -186,14 +186,24 @@ <h5 class="p-navigation--secondary__logo">
...
+    var banner = document.querySelector(".p-navigation__banner");
+    var dropdownWindow = document.querySelector(".dropdown-window");
+    var navigationItems = document.querySelector('.p-navigation__items');
     var searchButton = document.querySelector('.js-search-button');
     var search = document.querySelector('.p-navigation__search');
     var searchInput = document.querySelector('.p-search-box__input');
     var searchActive = !search.classList.contains('u-hide');
     search.classList.remove('u-hide');
     searchButton.classList.add('u-hide');
+    globalNav.style= "max-width:68rem; opacity: 0.4; transition: opacity 0.5s ease-in-out;"
```
anthonydillon, line 202: Lets not do this as the expected place to set the max-width is the init script of the global nav. I think its better to leave it untouched and active when searching.

anthonydillon (changes requested):

minkyngkm:
Thanks @anthonydillon for reviewing again. Made the global nav back as it was before. Hopefully it's all good now - [demo](https://ubuntu-com-10340.demos.haus)

anthonydillon (approved):
LGTM :+1: thanks for all the work here @minkyngkm

PR #10351, opened 2021-09-13
Add/Edit/Delete account user role

albertkol:
## Done

- Add/Edit/Delete account user role following the spec: https://docs.google.com/document/d/1BwKC43kx-keG3vS5YDYHBRgCN_T0cSfCtsSxeavnhHc/edit#

## QA

- Will be easier when the front-end is ready

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/230

### DIFF
--- a/webapp/advantage/ua_contracts/api.py
+++ b/webapp/advantage/ua_contracts/api.py
@@ -108,6 +108,20 @@ class UAContractsAPI:
 
         return users
 
+    def put_account_user_role(
+        self,
+        account_id: str,
+        user_role_request: dict,
+    ):
+        self._request(
+            method="put",
+            path=f"v1/accounts/{account_id}/user-role",
+            json=user_role_request,
+            error_rules=["default"],
+        )
+
+        return {}
+
     def get_contract_token(self, contract_id: str) -> Optional[str]:
         response = self._request(
             method="post",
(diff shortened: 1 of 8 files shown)


webapp/advantage/ua_contracts/primitives.py
```diff
@@ -126,12 +126,12 @@ def __init__(
         name: str,
         display_name: str,
         email: str,
-        last_login: str,
-        first_login: str,
+        last_login_at: str,
+        first_login_at: str,
```
petermakowski, line 130: Thanks for updating this!

webapp/advantage/ua_contracts/api.py
```diff
@@ -108,6 +108,20 @@ def get_account_users(self, account_id: str):
...
+    def put_account_user_role(
+        self,
+        account_id: str,
+        user_role_request: dict,
+    ):
+        self._request(
+            method="put",
+            path=f"v1/accounts/{account_id}/user-role",
+            json=user_role_request,
+            error_rules=["default"],
```
petermakowski, line 120: If I understand correctly, this will make the endpoint return `UAContractsAPIErrorView`/`UAContractsAPIError` which returns a standard python HTTPError? Does this mean that there's only a single possible error that this endpoint will return?

I'm trying to list different errors and map them to error messages.

petermakowski (approved):
Nice. All very clear, I'd only appreciate some more info about possible error responses for all endpoints to map them to front-end messages.

petermakowski (approved):

PR #10353, opened 2021-09-13
[Certified] - Keep query string when filters are cleared

### LINKED ISSUE #10325: Clearing filters on `/certified?q=...` page takes you back to `/certified`
If you go to https://ubuntu.com/certified?q= and apply some filters, when the page reloads try to clear them and it takes you back to `/certified`. This is not the expected behaviour - clearing filters should keep you on search results page

bethcollins92:
## Done

- Ensure that the "Clear" button clears only the filters applied - not the typed query string
- Add some tests for /certified search results
- Fix console errors when the pagination drop downs don't exist when there are no results 

## QA
1. 
- View page at: https://ubuntu-com-10353.demos.haus/certified
- In the search bar search for something i.e. "dell"
- When you land on the search results page, apply a filter i.e. "Desktop"
- Select "Apply"
- Then select "Clear"
- You should remain on the search results page, with no filters applied but with the query string still present

2.
- Go to the live site: https://ubuntu.com/certified?q=dell&limit=20&vendor=Ampere+Computing%2C+LLC and see the error in the console.
- Go to the demo: https://ubuntu-com-10353.demos.haus/certified?q=dell&limit=20&vendor=Ampere+Computing%2C+LLC and see the error is no longer there

## Issue / Card

Fixes #10325

### DIFF
--- a/static/js/src/certified-search-results.js
+++ b/static/js/src/certified-search-results.js
@@ -2,8 +2,17 @@
 
 function clearFilters() {
   hideDrawerPageReload();
+
   let objUrl = new URL(window.location);
-  objUrl.search = "";
+  const href = window.location.href;
+  if (href.includes("q=") && !href.includes("q=&")) {
+    const startOfQuery = href.indexOf("q");
+    const endOfQuery = href.indexOf("&");
+    const searchQuery = href.substring(startOfQuery, endOfQuery);
+    objUrl.search = searchQuery;
+  } else {
+    objUrl.search = "";
+  }
   window.location.assign(objUrl);
   return false;
 }
--- a/static/js/src/certified-search-results.js
+++ b/static/js/src/certified-search-results.js
@@ -93,17 +93,21 @@ function updateResultsPerPage() {
   const pageSizeTop = document.getElementById("page-size-top");
   const pageSizeBottom = document.getElementById("page-size-bottom");
 
-  pageSizeTop.addEventListener("change", (e) => {
-    // Needs to be set because the other dropdown is a dummy
-    searchResults.submit();
-  });
+  if (pageSizeTop) {
+    pageSizeTop.addEventListener("change", (e) => {
+      // Needs to be set because the other dropdown is a dummy
+      searchResults.submit();
+    });
+  }
 
-  pageSizeBottom.addEventListener("change", (e) => {
-    // Avoids submitting 2 redundant fields
-    let pageSizeTopChange = new Event("change");
-    pageSizeTop.value = e.target.value;
-    pageSizeTop.dispatchEvent(pageSizeTopChange);
-  });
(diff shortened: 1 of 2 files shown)


static/js/src/certified-search-results.js
```diff
@@ -2,8 +2,17 @@
 
 function clearFilters() {
   hideDrawerPageReload();
+
   let objUrl = new URL(window.location);
-  objUrl.search = "";
+  const href = window.location.href;
```
carkod, line 7: you can do `const { href } = window.location;` if you want to be cool, but this is fine.

carkod: I am not sure if we are compiling this "new JS" to "old JS", but appartenly this doesn't work for IE 11. So this comment is optional, just FYI.

static/js/src/certified-search-results.js
```diff
@@ -84,17 +93,21 @@ function updateResultsPerPage() {
...
-  pageSizeBottom.addEventListener("change", (e) => {
-    // Avoids submitting 2 redundant fields
-    let pageSizeTopChange = new Event("change");
-    pageSizeTop.value = e.target.value;
-    pageSizeTop.dispatchEvent(pageSizeTopChange);
-  });
+  if (pageSizeBottom) {
+    pageSizeBottom.addEventListener("change", (e) => {
+      // Avoids submitting 2 redundant fields
+      let pageSizeTopChange = new Event("change");
```
carkod, line 106: I think this can be a `const` it doesn't really get resused?

tests/cypress/certified_spec.js
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it("should take you to seach results page when search bar is used", () => {
+    cy.visit("/certified");
+
+    cy.findByPlaceholderText("Search hardware").type("dell{enter}");
+    cy.findByText(/results/).should("be.visible");
+    cy.findAllByText(/Dell/).should("be.visible");
+  });
+});
```
carkod, line 53: I'm not sure about this, so this is more of a question. Are we running these tests somewhere in the Github actions CI?

bethcollins92: Not yet, I only just [fixed](https://github.com/canonical-web-and-design/ubuntu.com/pull/10191) all the current tests on ubuntu.com to ensure they all pass. I've spoken to Ant about it - he thinks it could be a good idea to run them maybe once a week on production. I'm gonna bring it up in dev catch up today

carkod: but shouldn't this be ran on the PR everytime a dev adds new tests?

static/js/src/certified-search-results.js
bethcollins92: done

carkod (approved):

PR #10362, opened 2021-09-14
Adjust Marketo form payload for IPV6 addresses

caldav:
## Done

- The Marketo rest API is not handling well IPV6 addresses and prevents form submission in some cases, this skips sending the data to the form endpoint when an ipv6 address is used.
- The current behaviour is the cause of many form issues captured by Sentry

## QA

- Test submitting a form (any form)
- Ensure the form submits properly in the console and no error for this form ID appears in sentry

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -869,11 +869,13 @@ def marketo_submit():
     return_url = form_fields.pop("returnURL", None)
 
     visitor_data = {
-        "leadClientIpAddress": flask.request.headers.get(
-            "X-Real-IP", flask.request.remote_addr
-        ),
         "userAgentString": flask.request.headers.get("User-Agent"),
     }
+    client_ip = flask.request.headers.get(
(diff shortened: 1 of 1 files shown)


webapp/views.py
```diff
@@ -869,11 +869,13 @@ def marketo_submit():
...
     visitor_data = {
-        "leadClientIpAddress": flask.request.headers.get(
-            "X-Real-IP", flask.request.remote_addr
-        ),
         "userAgentString": flask.request.headers.get("User-Agent"),
     }
+    client_ip = flask.request.headers.get(
+        "X-Real-IP", flask.request.remote_addr
+    )
+    if client_ip and ":" in client_ip:
```
jpmartinspt, line 877: this should be `if client_ip and not ":" in client_ip:`, I think

jpmartinspt (changes requested):

jpmartinspt (approved):

PR #10363, opened 2021-09-14
Free trial error states

ClementChaumel:
## Done

- Added the error states listed [here](https://miro.com/app/board/o9J_lSky0cQ=/)

## QA

- Go to https://ubuntu-com-10363.demos.haus/advantage/subscribe?test_backend=true
- Log in
- You should not be able to start a trial
- You should see error messages telling you your account is not eligible on the bottom of the wizard and on step 2 of the modal

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/169

## Screenshots
image
image
image

### DIFF
--- a/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
+++ b/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
@@ -6,6 +6,7 @@ import {
   CheckboxInput,
 } from "@canonical/react-components";
 import * as Sentry from "@sentry/react";
+import {useFormikContext} from "formik";
 
 import useStripeCustomerInfo from "../../APICalls/useStripeCustomerInfo";
 import PaymentMethodSummary from "../PaymentMethodSummary";
@@ -22,6 +23,7 @@ import Summary from "../../components/Summary";
 import FreeTrialRadio from "../../components/FreeTrialRadio";
 import { checkoutEvent, purchaseEvent } from "../../../../ecom-events";
 import { getSessionData } from "../../../../../utils/getSessionData";
+import { FormValues } from "../../utils/utils";
 
 type StepOneProps = {
   setStep: React.Dispatch<React.SetStateAction<number>>;
--- a/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
+++ b/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
@@ -36,13 +36,13 @@ type StepOneProps = {
 function StepOne({ setStep, error, setError }: StepOneProps) {
   const { values } = useFormikContext<FormValues>();
   const [areTermsChecked, setTermsChecked] = useState(false);
-  const [isUsingFreeTrial, setIsUsingFreeTrial] = useState(true);
   const {
     data: userInfo,
     isLoading: isUserInfoLoading,
   } = useStripeCustomerInfo();
   const { isLoading: isPreviewLoading } = usePreview();
   const { isLoading: isProductLoading, product, quantity } = useProduct();
+  const [isUsingFreeTrial, setIsUsingFreeTrial] = useState(product?.canBeTrialled);
 
   const purchaseMutation = usePurchase();
   const freeTrialMutation = useFreeTrial();
@@ -226,7 +226,7 @@ function StepOne({ setStep, error, setError }: StepOneProps) {
       >
         <>
           <Summary />
-          {!product?.canBeTrialled && (
+          {product?.canBeTrialled && (
             <FreeTrialRadio
               isUsingFreeTrial={isUsingFreeTrial}
               setIsUsingFreeTrial={setIsUsingFreeTrial}
--- a/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
+++ b/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
@@ -5,8 +5,6 @@ import {
   ActionButton,
   CheckboxInput,
 } from "@canonical/react-components";
-import { useFormikContext } from "formik";
-
 import * as Sentry from "@sentry/react";
 
 import useStripeCustomerInfo from "../../APICalls/useStripeCustomerInfo";
@@ -25,8 +23,6 @@ import FreeTrialRadio from "../../components/FreeTrialRadio";
 import { checkoutEvent, purchaseEvent } from "../../../../ecom-events";
 import { getSessionData } from "../../../../../utils/getSessionData";
 
-import { FormValues } from "../../utils/utils";
-
 type StepOneProps = {
   setStep: React.Dispatch<React.SetStateAction<number>>;
   error: React.ReactNode | null;
@@ -42,7 +38,9 @@ function StepOne({ setStep, error, setError }: StepOneProps) {
   } = useStripeCustomerInfo();
   const { isLoading: isPreviewLoading } = usePreview();
   const { isLoading: isProductLoading, product, quantity } = useProduct();
-  const [isUsingFreeTrial, setIsUsingFreeTrial] = useState(product?.canBeTrialled);
+  const [isUsingFreeTrial, setIsUsingFreeTrial] = useState(
+    product?.canBeTrialled
+  );
 
   const purchaseMutation = usePurchase();
   const freeTrialMutation = useFreeTrial();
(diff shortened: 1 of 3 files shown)


static/js/src/advantage/subscribe/react/APICalls/useFreeTrial.tsx
```diff
@@ -0,0 +1,68 @@
...
+    if (res.errors) {
+      throw new Error(res.errors);
+    }
+    return res;
+  });
+
+  return mutation;
+};
+
+export default FreeTrial;
```
petermakowski, line 68:
```suggestion
export default useFreeTrial;
```

static/js/src/advantage/subscribe/react/APICalls/useFreeTrial.tsx
```diff
@@ -0,0 +1,68 @@
+import { useMutation } from "react-query";
+import {
+  postGuestFreeTrial,
+  postLoggedInFreeTrial,
+} from "../../../contracts-api";
+import useProduct from "./useProduct";
+
+import { FormValues } from "../utils/utils";
+
+const FreeTrial = () => {
```
petermakowski, line 10:
```suggestion
const useFreeTrial = () => {
```

static/js/src/advantage/subscribe/react/APICalls/useFreeTrial.tsx
```diff
@@ -0,0 +1,68 @@
...
+      city: city,
+      country: country,
+      line1: address,
+      postal_code: postalCode,
+      state: country === "US" ? usState : caProvince,
+    };
+
+    let res;
+
+    if (window.isGuest) {
```
petermakowski, line 40: Could you think of a better place/way to handle this? Ideally this should be decoupled from this hook and passed as a parameter. Look at: https://reactjs.org/docs/hooks-custom.html#tip-pass-information-between-hooks

Looks like @huwshimi already has done some work on a hook for window data that you could use:
https://github.com/petermakowski/ubuntu.com/blob/7d99f5bd8749cf1d620151ee85a6a6dd63628440/static/js/src/advantage/react/hooks/useLoadWindowData.ts#L26

static/js/src/advantage/subscribe/react/APICalls/useFreeTrial.tsx
petermakowski: Please start custom hooks names with “use”

> Do I have to name my custom Hooks starting with “use”? Please do. This convention is very important. Without it, we wouldn’t be able to automatically check for violations of rules of Hooks because we couldn’t tell if a certain function contains calls to Hooks inside of it.
https://reactjs.org/docs/hooks-custom.html#using-a-custom-hook

static/js/src/advantage/subscribe/react/components/FreeTrialRadio.tsx
```diff
@@ -0,0 +1,67 @@
...
+        {isUsingFreeTrial && (
+          <>
+            <p>
+              <strong>Your free trial ends: </strong>
+              {format(
+                add(new Date(), {
+                  months: 1,
+                }),
+                DATE_FORMAT
+              )}{" "}
```
petermakowski, line 33: There are some potential edge cases to consider as the end date calculation is handled entirely on the front end. I don't want to make a big deal out of it, just wanted to highlight - would be good to check if we're okay with this.

- What about users in different timezones 
- What about users with an incorrectly set date on their machine (very unlikely, I know)

The above could result in a slightly different date than actual trial end date? If someone really tried (and made a screenshot) this could - in theory - have some legal implications (false/misleading claims).

For that reason I wonder if this value should come from the back-end?

static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
```diff
@@ -168,6 +174,44 @@ function StepOne({ setStep, error, setError }: StepOneProps) {
...
+  const onStartTrialClick = () => {
+    freeTrialMutation.mutate(values, {
+      onSuccess: () => {
+        // The state of the product selector is stored in the local storage
+        // if a purchase is successful we empty it so the customer will see
+        // the default values pre-selected instead of what they just bought.
+        localStorage.removeItem("ua-subscribe-state");
+
+        //redirect
+        if (window.isGuest) {
```
petermakowski, line 186: Would be good to abstract this away to a hook (look at the comment above).

static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
```diff
@@ -168,6 +174,44 @@ function StepOne({ setStep, error, setError }: StepOneProps) {
...
+        }
+      },
+      onError: (error) => {
+        if (
+          error instanceof Error &&
+          error.message.includes("account already had or has access to product")
+        ) {
+          setError(<>You already have trialled this product</>);
+        } else {
+          Sentry.captureException(error);
```
petermakowski, line 201: 👍

static/js/src/advantage/subscribe/react/components/FreeTrialRadio.tsx
```diff
@@ -0,0 +1,67 @@
...
+import usePreview from "../APICalls/usePreview";
+
+const DATE_FORMAT = "dd MMMM yyyy";
+
+type Props = {
+  isUsingFreeTrial: boolean;
+  setIsUsingFreeTrial: React.Dispatch<React.SetStateAction<boolean>>;
+};
+
+const FreeTrialRadio = ({ isUsingFreeTrial, setIsUsingFreeTrial }: Props) => {
```
petermakowski, line 14: Could you add some tests for this?

You can look at the one I did recently for ideas: https://github.com/canonical-web-and-design/ubuntu.com/blob/ea5f3f9f7ebca1162b8d99b10b747b4ad46fe33e/static/js/src/advantage/users/AccountUsers.test.tsx

petermakowski (commented):
Looks good, just a few points to consider below.

static/js/src/advantage/subscribe/react/APICalls/useFreeTrial.tsx
petermakowski: (Not insisting on doing this as part of this PR)

static/js/src/advantage/subscribe/react/components/FreeTrialRadio.tsx
ClementChaumel: Hey! I don't see the issue with users in different time zones what would happen then?
For users with an incorrect date on their machine then yes I guess we would display a wrong date here.
I doubt we risk too much in terms of legal implications as it's easy to doctor a screenshot anyway and I'd argue that if you change the date on your machine then you can't expect everything to behave correctly (Yes I'm a lawyer now :judge: )
@wgx @SirSamTumless What do you think?

static/js/src/advantage/subscribe/react/APICalls/useFreeTrial.tsx
ClementChaumel: I don't know that it should be decoupled from the hook. The hook handles starting a trial and decides which API endpoint to call. I think it's less error-prone to let the hook handle it as it's part of its logic.

I'm not super convinced that we gain something by adding a hook to forward all the data from the window object into a ReactQuery query. I'm not sure ReactQuery should be used as a general store.

static/js/src/advantage/subscribe/react/APICalls/useFreeTrial.tsx
ClementChaumel: Whoops nice catch!

static/js/src/advantage/subscribe/react/APICalls/useFreeTrial.tsx
petermakowski: Fair points.

What I wanted to highlight is that it would be beneficial to  keep the `isGuest` state in React. I think we should move away from using globals like that as much as possible.

petermakowski (approved):

static/js/src/advantage/subscribe/react/components/FreeTrialRadio.tsx
wgx: @ClementChaumel fine to go ahead like this - as you say, nobody will be surprised if they have their clock set incorrectly.

wgx:
@clem i only see the error when monthly is selected - it doesn't appear when annual is selected:

image

ClementChaumel:
> @clem i only see the error when monthly is selected - it doesn't appear when annual is selected:
> 
> image

Oh yeah that's because free trial is enabled on the staging backend. If you check against prod (https://ubuntu-com-10363.demos.haus/advantage/subscribe?test_backend=true) it should appear for every product

PR #10369, opened 2021-09-16
Cancel a subscription via the API.

huwshimi:
## Done

- Add support for fetching the last purchase ids from the API.
- Add support for cancelling a contract via the API.
- Update the cancel component to cancel the contract.

## QA

- Visit [/advantage?test_backend=true](/advantage?test_backend=true) and log in.
- Click on a monthly contract (there is an issue at the moment where you can't select a monthly contract that you also have a yearly subscription for: https://github.com/canonical-web-and-design/commercial-squad/issues/253).
- Click 'Edit subscription'
- Click the 'You can cancel this subscription online or contact us.' link at the bottom.
- Type cancel and then click the red button.
- You subscription should get cancelled.

## Issue / Card

Fixes: https://github.com/canonical-web-and-design/commercial-squad/issues/116.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,8 @@
     "typescript": "4.3.5",
     "url-polyfill": "1.1.12",
     "url-search-params-polyfill": "8.1.0",
-    "vanilla-framework": "2.35.0"
+    "vanilla-framework": "2.35.0",
+    "yup": "0.32.9"
   },
   "resolutions": {
     "lodash": "4.17.21",
--- a/static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.test.tsx
+++ b/static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.test.tsx
@@ -1,14 +1,283 @@
-import { Modal } from "@canonical/react-components";
+import { ActionButton, Modal, Notification } from "@canonical/react-components";
 import React from "react";
-import { shallow } from "enzyme";
+import { mount } from "enzyme";
 
+import * as contracts from "advantage/api/contracts";
 import SubscriptionCancel from "./SubscriptionCancel";
+import { QueryClient, QueryClientProvider } from "react-query";
+import {
+  lastPurchaseIdsFactory,
+  userSubscriptionFactory,
+} from "advantage/tests/factories/api";
+import { LastPurchaseIds, UserSubscription } from "advantage/api/types";
+import { act } from "react-dom/test-utils";
+import { UserSubscriptionPeriod } from "advantage/api/enum";
 
 describe("SubscriptionCancel", () => {
-  it("handles closing the modal", () => {
+  let cancelContractSpy: jest.SpyInstance;
+  let queryClient: QueryClient;
+  let subscription: UserSubscription;
+  let lastPurchaseIds: LastPurchaseIds;
+
+  beforeEach(() => {
+    cancelContractSpy = jest.spyOn(contracts, "cancelContract");
+    cancelContractSpy.mockImplementation(() => Promise.resolve({}));
+    queryClient = new QueryClient();
+    subscription = userSubscriptionFactory.build({
+      period: UserSubscriptionPeriod.Monthly,
+    });
+    lastPurchaseIds = lastPurchaseIdsFactory.build();
+    queryClient.setQueryData("userSubscriptions", [subscription]);
+    queryClient.setQueryData(
+      ["lastPurchaseIds", subscription.account_id],
+      lastPurchaseIds
+    );
+  });
+
+  afterAll(() => {
+    jest.restoreAllMocks();
+  });
+
+  it("handles closing the modal", async () => {
     const onClose = jest.fn();
-    const wrapper = shallow(<SubscriptionCancel onClose={onClose} />);
-    wrapper.find(Modal).invoke("close")!();
+    const wrapper = mount(
+      <QueryClientProvider client={queryClient}>
+        <SubscriptionCancel
+          selectedId={subscription.contract_id}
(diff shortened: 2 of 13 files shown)


static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.test.tsx
```diff
@@ -1,14 +1,283 @@
-import { Modal } from "@canonical/react-components";
+import { ActionButton, Modal, Notification } from "@canonical/react-components";
 import React from "react";
-import { shallow } from "enzyme";
+import { mount } from "enzyme";
```
petermakowski, line 3: would you consider trying out `@testing-library/react`? 🙃

package.json
```diff
@@ -99,7 +99,8 @@
     "typescript": "4.3.5",
     "url-polyfill": "1.1.12",
     "url-search-params-polyfill": "8.1.0",
-    "vanilla-framework": "2.35.0"
+    "vanilla-framework": "2.35.0",
+    "yup": "0.32.9"
```
petermakowski, line 103: Yup is great, but has there been any discussion regarding adding this as a dependency?

Also - is there a reason this couldn't go straight to master?

static/js/src/advantage/api/contracts.test.ts
```diff
@@ -1,25 +1,42 @@
...
 
 describe("contracts", () => {
   afterEach(() => {
     fetch.resetMocks();
   });
 
-  it("can get user subscriptions", () => {
-    const contractData = { data: { test: "contract" } };
+  it("can get user subscriptions", async () => {
+    const contractData = { data: userSubscriptionFactory.buildList(2) };
```
petermakowski, line 20: Just a personal opinion, I don't see a clear benefit of using yet another layer of abstraction here. This line before the change was better at explaining how this code works.

static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.tsx
```diff
@@ -1,57 +1,129 @@
...
             className="u-no-margin--bottom"
+            disabled={!isValid}
+            loading={cancelContract.isLoading}
+            success={cancelContract.isSuccess}
+            onClick={() => {
+              if (formRef.current) {
+                // Because this button appears outside of the form in the DOM
+                // this uses a ref to call the submit handler.
+                formRef.current.handleSubmit();
+              }
```
petermakowski, line 81: You could probably wrap the modal with Formik as well to prevent having to use hacks such as this.

petermakowski (commented):
This PR is very big. Any chance you could break it down please to make it easier to review 🙏

static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.test.tsx
huwshimi: This is not the right point for that, this work is nearly complete and it would intro further fragmentation into our code.

package.json
huwshimi: We tend to use Yup with Formik (it's the recommended validation lib). No reason it couldn't go to master, but I needed it here.

static/js/src/advantage/api/contracts.test.ts
huwshimi: We use factories based on the typescript API types to produce API data as we expect it to. I'm very happy to talk with you about the benefits of using a factory lib, in fact I'm going to do a masterclass which will touch on this topic a bit. However, I'm not sure this PR is the right place to explain that in detail.

static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.tsx
huwshimi: The modal would have to be wrapped in both `Formik` and `form` so that the submit worked, however this results in changing the HTML structure of the modal in a way that affects the flex layout of the modal so this seemed like a simpler solution than having to override the Vanilla css styles.

huwshimi:
> This PR is very big. Any chance you could break it down please to make it easier to review 🙏

I should have explain when I created the PR but I was really hoping to get someone to review this in one go as we're nearing the deadline for me to finish this work.

I've split out the API support into this PR: https://github.com/canonical-web-and-design/ubuntu.com/pull/10385. That won't be able to be QAed until it merges with this branch.

This PR is still pretty large sorry.

package.json
petermakowski: Don't get me wrong, I have nothing against yup (happy to use it, used it in the past), I was just making a point about thinking of merging code straight to master where possible. Such focused PR would also allow for a conversation across the team about adding such dependency.

static/js/src/advantage/react/components/Subscriptions/SubscriptionCancel/SubscriptionCancel.tsx
petermakowski: You'd only need to wrap the modal with `<Formik />`, leaving the `form` as is - no HTML structure changes required.

Then you could use the render function to pass down the handleSubmit:
```jsx
<Formik<SubscriptionCancelValues>>
  {({ handleSubmit }) => (
    <div className="p-subscriptions__sticky-footer-modal">
      <Modal
        buttonRow={
          <ActionButton
            onClick={handleSubmit}
```

huwshimi: Ah yes I see what you're suggesting, that should work nicely.

PR #10370, opened 2021-09-16
Display a spinner for the initial load

huwshimi:
## Done

- Display a spinner for the initial load

## QA

- Visit /advantage?test_backend=true and log in.
- When the card initially appears you should see a nicer loading state.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/265.

## Screenshots

Screen Shot 2021-09-16 at 4 42 25 pm

### DIFF
--- a/static/js/src/advantage/react/components/Subscriptions/Content/Content.test.tsx
+++ b/static/js/src/advantage/react/components/Subscriptions/Content/Content.test.tsx
@@ -16,12 +16,28 @@ describe("Content", () => {
     queryClient.setQueryData("userSubscriptions", [contract]);
   });
 
-  it("renders", () => {
+  it("displays a spinner when loading", () => {
+    // Remove the queries so that the query starts loading.
+    queryClient.removeQueries("userSubscriptions");
     const wrapper = mount(
       <QueryClientProvider client={queryClient}>
         <Content />
       </QueryClientProvider>
     );
+    expect(wrapper.find("[data-test='initial-load'] Spinner").exists()).toBe(
+      true
+    );
+    expect(wrapper.find("SubscriptionList").exists()).toBe(false);
+    expect(wrapper.find("SubscriptionDetails").exists()).toBe(false);
+  });
+
+  it("displays the list and details when loaded", () => {
+    const wrapper = mount(
+      <QueryClientProvider client={queryClient}>
+        <Content />
+      </QueryClientProvider>
+    );
+    expect(wrapper.find("[data-test='initial-load']").exists()).toBe(false);
     expect(wrapper.find("SubscriptionList").exists()).toBe(true);
     expect(wrapper.find("SubscriptionDetails").exists()).toBe(true);
   });
--- a/static/js/src/advantage/react/components/Subscriptions/Content/Content.tsx
+++ b/static/js/src/advantage/react/components/Subscriptions/Content/Content.tsx
@@ -1,4 +1,4 @@
-import { Card } from "@canonical/react-components";
+import { Card, Spinner } from "@canonical/react-components";
 import { useUserSubscriptions } from "advantage/react/hooks";
 import { useScrollIntoView } from "advantage/react/hooks/useScrollIntoView";
 import React, { useCallback, useEffect, useState } from "react";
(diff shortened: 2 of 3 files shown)


static/js/src/advantage/react/components/Subscriptions/Content/Content.test.tsx
```diff
@@ -16,12 +16,28 @@ describe("Content", () => {
...
+    // Remove the queries so that the query starts loading.
+    queryClient.removeQueries("userSubscriptions");
     const wrapper = mount(
       <QueryClientProvider client={queryClient}>
         <Content />
       </QueryClientProvider>
     );
+    expect(wrapper.find("[data-test='initial-load'] Spinner").exists()).toBe(
+      true
+    );
```
petermakowski, line 29: Could I convince you to use `@testing-library/react` for this test instead? 😄

petermakowski (approved):

petermakowski:
image
image

Could you add `u-no-margin--bottom p-card` wrapper to the loader in the HTML loader as well so that it matches the react one? It doesn't look great when the layout jumps after react mounts.

petermakowski:
Also if the `advantage/user-subscriptions` endpoint fails, eventually this is the state of the UI:

image

static/js/src/advantage/react/components/Subscriptions/Content/Content.test.tsx
petermakowski: (totally optional) 😉

petermakowski (changes requested):
Please look at the 2 comments with screenshots above.

static/js/src/advantage/react/components/Subscriptions/Content/Content.tsx
```diff
@@ -38,6 +38,14 @@ const Content = () => {
     }
   }, [selectedId, setSelectedId, allSubscriptions, isLoading]);
 
+  if (isLoading) {
+    return (
+      <Card className="u-no-margin--bottom" data-test="initial-load">
```
petermakowski, line 43: @huwshimi @hatched Should we start using `data-testid` instead of `data-test` regardless of whether a test is written with enzyme or testing-library?

`data-testid` is the convention used for testing-library `getByTestId` matcher.

Arguably it's a better (more explicit) name, plus it's good for consistency and will make switching easier.

https://testing-library.com/docs/queries/bytestid/

hatched: This seams like a reasonable request. I'd just ask that we convert the entire file at once so that we don't have some `test` and some `testid` in the same file.

static/js/src/advantage/react/components/Subscriptions/Content/Content.test.tsx
hatched: This might be out of scope of this PR as we should convert the whole file over vs just one/some of the tests.

huwshimi:
> Also if the `advantage/user-subscriptions` endpoint fails, eventually this is the state of the UI:
> 
> image

Thanks, I've filed that as a new issue: https://github.com/canonical-web-and-design/ubuntu.com/issues/10384.

static/js/src/advantage/react/components/Subscriptions/Content/Content.test.tsx
huwshimi: Yeah, I'm nearing the end of this ua table section of the site which has all been built with Enzyme, so I don't think this would be the right time to introduce a new testing lib.

static/js/src/advantage/react/components/Subscriptions/Content/Content.tsx
huwshimi: Yep, something to introduce across the entire ua-table section of the site, but this is probably not the right time to do that as we're wrapping things up.

huwshimi:
> ### HTML
> image
> 
> ### React
> image
> 
> Could you add `u-no-margin--bottom p-card` wrapper to the loader in the HTML loader as well so that it matches the react one? It doesn't look great when the layout jumps after react mounts.

Done, thanks, I hadn't noticed that.

petermakowski (approved):

PR #10375, opened 2021-09-16
Add update default entitlements endpoint

albertkol:
## Done

- Add endpoint to update default entitlements:
Endpoint: `PUT /advantage/contracts/{contract_id}/entitlements`

Json schema:
```
{
  "entitlements": [
    {
      "type": String(required),
      "is_enabled": Boolean(required),
    }
  ]
}
```

## QA

- Will be easier when the front-end is ready

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/264

### DIFF
--- a/tests/advantage/test_api.py
+++ b/tests/advantage/test_api.py
@@ -1530,9 +1530,7 @@ class TestPutContractEntitlements(unittest.TestCase):
 
         expected_args = {
             "headers": {"Authorization": "Macaroon secret-token"},
-            "json": {
-                "entitlements": [{"type": "fips", "is_enabled": True}]
-            },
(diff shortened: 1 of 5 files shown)


cristinadresch:
@nottrobin can you take a look pls

nottrobin (approved):
I think this is fine.

I'm curious though, did you follow some sort of standard for your `PUT {"object_name": [list]}` scheme? Is it standard practice to repeat the name of the object in the dictionary if you're replacing a list?

Also, I assume that `PUT`ing a list will always *replace* the entire list of entitlements - is that right?

But yeah, I think this is fine.

PR #10381, opened 2021-09-16
200/account users/edit view

petermakowski:
## Done

- handle user editing
  -  refactor TableView to use React components - wrap only a single editable row with Formik
  - fix long email overflow
  - fix column widths
  - refactor to use a single notification
  - display last admin error message
  - display last admin error on failed update

## QA

- Go to https://ubuntu-com-10381.demos.haus/advantage/users?test_backend=true and login
- Add a new user
- Edit a role of that user
- Delete a user
- Make sure error/success messages are displayed on each action

## Issue / Card

https://github.com/canonical-web-and-design/commercial-squad/issues/200

## error messages
image
image
### email overflow fix - delete button still visible
image

### DIFF
--- a/static/js/src/advantage/users/AccountUsers.tsx
+++ b/static/js/src/advantage/users/AccountUsers.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { useQueryClient, useMutation, useQuery } from "react-query";
+import { useQueryClient, useMutation } from "react-query";
 import {
   User,
   Users,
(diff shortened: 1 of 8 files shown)


ClementChaumel:
QA ok just a bit of UX I found weird:
If you create a new user, fill in their name and their email address and then press enter it will close the modal without creating the user.
I would expect the default action when pressing enter to be the create button (Or the cancel button to be highlighted somehow)

ClementChaumel (approved):

petermakowski:
> QA ok just a bit of UX I found weird:
> If you create a new user, fill in their name and their email address and then press enter it will close the modal without creating the user.
> I would expect the default action when pressing enter to be the create button (Or the cancel button to be highlighted somehow)

Thanks! I might've missed type="submit" on the button / messed up onSubmit with Formik. WIll look into 🙏

PR #10387, opened 2021-09-17
Display the renewal settings

huwshimi:
## Done

- Fetch user info from the API.
- Display the renewal settings data.

## QA

- Visit [/advantage?test_backend=true](https://ubuntu-com-10387.demos.haus/advantage?test_backend=true) and log in.
- Click on "Renewal settings" at the top of the list.
- The number of subs, price and renewal date are displayed from the API data.
- The 'Auto-renewal' checkbox should reflect the state of your auto renewal settings (probably on).
- Note: the form doesn't update your settings yet (will be done in https://github.com/canonical-web-and-design/commercial-squad/issues/272).

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/99.

## Screenshots

Screen Shot 2021-09-17 at 1 16 13 pm

### DIFF
--- a/static/js/src/advantage/api/contracts.js
+++ b/static/js/src/advantage/api/contracts.js
@@ -65,6 +65,14 @@ export async function getCustomerInfo(accountId) {
   return data;
 }
 
+export async function getUserInfo() {
+  const queryString = window.location.search; // Pass arguments to the flask backend eg. "test_backend=true"
+  const response = await fetch(`/advantage/user-info${queryString}`, {
+    cache: "no-store",
+  });
+  return await response.json();
+}
+
 export async function getUserSubscriptions() {
   const queryString = window.location.search; // Pass arguments to the flask backend eg. "test_backend=true"
   const response = await fetch(`/advantage/user-subscriptions${queryString}`, {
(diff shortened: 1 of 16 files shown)


ClementChaumel:
The "Renewal settings" button is not displayed when all my monthly subscriptions are expired (which I think is correct) 
But then if I add a new one the number of affected subscriptions will include the expired monthly subs.
Eg. I have 2 expired subs and 1 active one. If I click this it will tell me the settings will affect 3 subscriptions instead of just one

huwshimi:
> The "Renewal settings" button is not displayed when all my monthly subscriptions are expired (which I think is correct)
> But then if I add a new one the number of affected subscriptions will include the expired monthly subs.
> Eg. I have 2 expired subs and 1 active one. If I click this it will tell me the settings will affect 3 subscriptions instead of just one

Thanks @ClementChaumel, I've filtered out any expired or cancelled monthly subscriptions. Does that sound right?

ClementChaumel:
> > The "Renewal settings" button is not displayed when all my monthly subscriptions are expired (which I think is correct)
> > But then if I add a new one the number of affected subscriptions will include the expired monthly subs.
> > Eg. I have 2 expired subs and 1 active one. If I click this it will tell me the settings will affect 3 subscriptions instead of just one
> 
> Thanks @ClementChaumel, I've filtered out any expired or cancelled monthly subscriptions. Does that sound right?

I think it makes more sense that way! Thanks

PR #10396, opened 2021-09-17
Increase the timeout when linkchecking

anthonydillon:
## Done
- Increase the timeout when linkchecking to avoid discourse timeouts

## QA
- Check the code changes look good.
- Run the command locally

### DIFF
--- a/.github/workflows/blog-links.yaml
+++ b/.github/workflows/blog-links.yaml
@@ -17,6 +17,7 @@ jobs:
         run: |
           linkchecker --recursion-level=2 \
           	--check-extern --no-warnings https://ubuntu.com/blog \
+            --timeout 1000 \
           	--no-follow-url '!https:\/\/ubuntu.com\/blog' \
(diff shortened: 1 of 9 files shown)


bethcollins92:
What command should I run locally? Do I need to do for each file e.g. `linkchecker --no-follow-url '!https:\/\/ubuntu.com\/ceph\/docs'` ?

anthonydillon:
Sorry, should have explained in QA. Just take one of the commands and run it like it is. You will need to [linkchecker locally](https://github.com/linkchecker/linkchecker). 

If the command works then just check the syntax of the change is the same for all commands.

If you have any trouble we can jump on a call.

bethcollins92:
When I try and run a couple of the commands locally I get an error:  `no matches found: w_[0-9]*`, am I missing something?

bethcollins92 (approved):
LGTM 👍🏼  - should it have some points / an issue?

PR #10399, opened 2021-09-17
Fix user info page

albertkol:
## Done

- Fix 500 error if auto renewal is off. 
- Turns out the API returns no payment data if the auto renewal is turned off so we cannot show at the moment display the next payment information:

image
 

## QA

- Have a montly subscription
- Turn the auto renewal off
- Go to: /advantage/user-info
- You will not get a 500
- You will see
```
{
  has_mothly_subscriptions: True,
  is_auto_renewing: False
}
```

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/274


huwshimi (approved):
Thanks @albertkol works well now.

PR #10403, opened 2021-09-17
Remove image builder

anthonydillon:
## Done
- Removed image builder and all its dependencies

## QA
- Go to /core/build and see its deleted
- Check site login still works
- Check the /advantage section still looks ok

## Issue / Card
Fixes https://github.com/canonical-web-and-design/web-squad/issues/3088


webapp/views.py
```diff
@@ -1,7 +1,4 @@
 # Standard library
-import hashlib
-import hmac
-import json
```
pmahnke, line 4: I think removing json has broken usn's and cve's?

requirements.txt
```diff
@@ -6,10 +6,8 @@ canonicalwebteam.search==1.0.0
 canonicalwebteam.templatefinder==1.0.0
 canonicalwebteam.image-template==1.3.1
 canonicalwebteam.discourse==4.0.4
-canonicalwebteam.launchpad==0.8.2
```
nottrobin, line 9: Pretty sure other things are gonna need this

webapp/views.py
nottrobin: did it? Where did you see that?

nottrobin (approved):
This is really thorough! Looks good to me. Check @pmahnke agrees before merging, but I don't personally see anything wrong with removing `json` which doesn't seem to be used in `webapp/views.py`.

anthonydillon:
Yeah looks like CVEs and USNs are blocked from loading on demos.

PR #10413, opened 2021-09-20
9 replicas

nottrobin:
Because we can only have 1 per node with our affinity config, and we only have 9 nodes


jkfran (approved):

petermakowski (approved):

carkod:
Maybe that description should be a comment in the code

PR #10421, opened 2021-09-20
Cube shop listing

carkod:
## Done

- Replace `cube.yaml` content with data from UA contracts API.
- Populate modules list with `/cube/microcerts.json`
- No actions yet, it'll make the PR too big and there are other issues opened for that.

## QA

- Check https://ubuntu-com-10421.demos.haus/cube/microcerts to see a new table with content
- Navigate to /cube/microcerts as a guest user, and make sure it redirects to USSO

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/256

## Screenshots

advantage-data


webapp/app.py
```diff
@@ -793,6 +794,9 @@ def takeovers_index():
 app.add_url_rule("/cube", view_func=cube_home)
 app.add_url_rule("/cube/microcerts", view_func=cube_microcerts)
 app.add_url_rule("/cube/study/labs", view_func=cube_study_labs_button)
+app.add_url_rule(
+    "/cube/get-microcerts", view_func=get_microcerts, methods=["GET"]
```
anthonydillon, line 798: Shouldn't endpoint that return json end with a .json extension? And get is implied, right? What about: `/cube/microcerts.json`?

webapp/cube/views.py
```diff
@@ -4,21 +4,30 @@
...
 
 from webapp.cube.api import BadgrAPI, EdxAPI
 from webapp.decorators import login_required
 from webapp.login import user_info
-
+from webapp.advantage.ua_contracts.api import UAContractsAPI
 
 CUBE_CONTENT = yaml.load(
-    Path("webapp/cube/content/cube.yaml").read_text(), Loader=yaml.Loader
+    Path("webapp/cube/content/cube-qa.yaml").read_text(), Loader=yaml.Loader
```
jpmartinspt, line 20: Need to revert this to `cube.yaml`.

webapp/cube/views.py
```diff
@@ -45,6 +54,8 @@
     edx_session,
 )
 
+ua_contracts_session = Session()
```
jpmartinspt, line 57: add `talisker.requests.configure(ua_contracts_session)`

webapp/cube/views.py
```diff
@@ -133,6 +144,149 @@ def cube_microcerts():
...
+    # backend_true requires query string,
+    # therefore we need context
+    # also this avoids loading contracts api where not required
+    is_test_backend = (
+        current_app.config["CONTRACTS_TEST_API_URL"]
+        if flask.request.args.get("test_backend", "false")
+        else current_app.config["CONTRACTS_LIVE_API_URL"]
+    )
+    sso_user = user_info(flask.session)
+    if not sso_user:
```
jpmartinspt, line 159: Since this view is under a @login_required decorator it should be guaranteed that you have an `sso_user`

jpmartinspt: Also kinda linked to this, if this is a JSON view, it should not be redirecting or returning HTML

webapp/cube/views.py
```diff
@@ -133,6 +144,149 @@ def cube_microcerts():
...
+        attempts = []
+        edx_ids = next(
+            (
+                ids
+                for ids in product_list["externalIDs"]
+                if ids["origin"] == "EdX"
+            ),
+            None,
+        )
+        if edx_ids:
```
jpmartinspt, line 223: Slightly confused by this, is there any listing that has more than one EDX id?

carkod: Well that's why I wanted you to take a look at this PR. I get a list of IDs, although currently I just get one in the list. Should I just grab the first always `edx_ids[0]`?

jpmartinspt: I believe so, if a product listing represents a single EDX course it should only have one EDX ID

jpmartinspt: `edx_id = [edx_id for edx_id in product_list["externalIDs"] if ids["origin"] == "EdX"][0]`

carkod: I asked CS team what do they do with the IDs and they said that they just synch Edx with UA contracts using the email. So I trust that they always will give at least one ID?

webapp/cube/views.py
carkod: Can I do instead `talisker.requests.get_session()`

jpmartinspt: No, that will just end up reusing and existing session. You need to configure the created session.

webapp/app.py
carkod: Done

webapp/cube/views.py
carkod: Done

webapp/cube/views.py
carkod: Good point. Done

webapp/cube/views.py
carkod: Done

carkod:
Removed React code at Clem's request

webapp/cube/views.py
```diff
@@ -133,6 +145,135 @@ def cube_microcerts():
...
+        course = {
+            "product_listing_id": product_list["id"],
+            "value": product_list["price"]["value"],
+        }
+        edx_ids = [
+            edx_id
+            for edx_id in product_list["externalIDs"]
+            if edx_id["origin"] == "EdX"
+        ][0]
+        for course_id in edx_ids["IDs"]:
```
jpmartinspt, line 213: I think we can assume that edx_ids["IDs"] only has one value, so I would just do:
```
course_id = [
    edx_id[0]
    for edx_id in product_list["externalIDs"]
    if edx_id["origin"] == "EdX"
][0]
```
And then you can get rid of the for loop.

carkod: mm ok. Let's hope they don't change the structure. There is another list inside, so I changed a bit the code.

ClementChaumel (approved):
LGTM

PR #10422, opened 2021-09-20
Call the delete trial endpoint when the button is clicked

ClementChaumel:
## Done

- Changed the copy to better reflect what the button does
- calls the endpoint on click

## QA

- go to https://ubuntu-com-10422.demos.haus/advantage/subscribe?test_backend=true and add a trial if you don't already have one
- then on https://ubuntu-com-10422.demos.haus/advantage?test_backend=true open the corresponding row
- click the button
- the page should reload and the trial should go away

### DIFF
--- a/static/js/src/advantage/contracts-api.js
+++ b/static/js/src/advantage/contracts-api.js
@@ -334,3 +334,20 @@ export async function postFreeTrial({
   let data = await response.json();
   return data;
 }
+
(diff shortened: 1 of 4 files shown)


carkod:
I can't QA, it says "Free trial not available for this account"

carkod (approved):

PR #10425, opened 2021-09-21
Remove server-side subscription fetching.

huwshimi:
## Done

- Removed server-side fetching of subscription data and passing to the template as the data is now all fetched by the client.

## QA

- Visit [/advantage?test_backend=true](https://ubuntu-com-10425.demos.haus/advantage?test_backend=true) and log in.
- You should get the list of subscriptions.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/279.


huwshimi:
This one might have to wait for @albertkol as I'm not sure if we still need to verify the token on the backend or if we can remove this test: https://github.com/canonical-web-and-design/ubuntu.com/blob/master/tests/test_routes.py#L138.

ClementChaumel (approved):

huwshimi:
> This one might have to wait for @albertkol as I'm not sure if we still need to verify the token on the backend or if we can remove this test: https://github.com/canonical-web-and-design/ubuntu.com/blob/master/tests/test_routes.py#L138.

Just to answer my own question, the redirect happens when a server-side API call fails and as we've removed the server side calls this test is no longer relevant.

PR #10430, opened 2021-09-21
Allow e2e tests for logged-in user journeys - account users e2e tests

petermakowski:
## Done

- add account users e2e test
- add custom Cypress login command **allowing tests for logged-in user journeys** - all you have to do is use `cy.login({ username, password });` in your test
  - handled via puppeteer that then returns login cookies for cypress to use
  - this is a workaround - cypress doesn't allow login flows that involve external domain (cross-domain issues)
- add custom Cypress command `acceptCookiePolicy` for convenience

## QA
verify the cypress test for advantage-users has been successful: https://github.com/canonical-web-and-design/ubuntu.com/runs/3662207566

## Issue / Card

https://github.com/canonical-web-and-design/commercial-squad/issues/267

## Screenshots

image


.github/workflows/cypress.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
       - name: Cypress run
         uses: cypress-io/github-action@v2
         with:
+          env: UBUNTU_USERNAME=peter.makowski+100@canonical.com,UBUNTU_PASSWORD=jibfdq5hmq
```
petermakowski, line 45: Didn't manage to get github secrets being passed correctly with `cypress-io/github-action@v2`, writing this inline for now.

If there's another test user that we should use please let me know.

tests/cypress/support/commands.js
```diff
@@ -1 +1,13 @@
 import "@testing-library/cypress/add-commands";
+
+Cypress.Commands.add("acceptCookiePolicy", () => {
```
bethcollins92, line 3: I added [this](https://github.com/canonical-web-and-design/ubuntu.com/blob/master/tests/cypress/forms_spec.js#L28) to my test but now I'm wondering if this way is better as it's more user centric?

.github/workflows/cypress.yaml
bethcollins92: I'm not too sure about this @anthonydillon might know more

tests/cypress/support/commands.js
petermakowski: I'd say so.

The only reason why I'd opt for your approach would be speed, but we don't have that much to gain here.

It's important to test the cookie banner in all contexts I think. You can sometimes get bugs with z-index etc. that e2e can catch provided we're actually testing the cookie banner, not skipping it entirely.

.github/workflows/cypress.yaml
anthonydillon: Is this supposed to be a secret?

tests/cypress/support/commands.js
bethcollins92: Makes sense, I'll keep that in mind for next time - thanks for clarifying

.github/workflows/cypress.yaml
petermakowski: would be nice if it was, but doesn't really have to be, it's a dummy test user

bethcollins92 (approved):
LGTM 👍🏼

PR #10433, opened 2021-09-21
Fix display date when when lastLoginAt is null

petermakowski:
## Done

- Fix display date when when lastLoginAt is null

## QA
- Go to /advantage/users
- Add a new user
- Make sure last login column value for that user is empty

## Issue / Card

https://github.com/canonical-web-and-design/commercial-squad/issues/281

### Before
image

### After
image

### DIFF
--- a/static/js/src/advantage/users/api.ts
+++ b/static/js/src/advantage/users/api.ts
@@ -5,10 +5,10 @@ type AccountUsersReponse = {
   name: string;
   users: {
     id: string;
-    name: string;
+    name: string | null;
     email: string;
     user_role_on_account: "admin" | "technical" | "billing";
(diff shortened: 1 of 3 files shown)


ClementChaumel (approved):
Oh yeah I caught it but forgot to bring it up last time I reviewed your PR! Nice one.
But maybe it should say "never" instead of being empty? @wgx @elioqoshi opinions?

wgx:
@ClementChaumel Yeah, "Never" would be clearer than just leaving it blank. Nice!

petermakowski:
> @ClementChaumel Yeah, "Never" would be clearer than just leaving it blank. Nice!

Thanks! Updated:
image

PR #10435, opened 2021-09-21
Hide free trial message on live if not eligible

ClementChaumel:
## Done

- Removed free trial messaging on live until we release
- reduced the size of the message

## QA

- go to https://ubuntu-com-10435.demos.haus/advantage/subscribe?test_backend=true
- select a non trialable product (monthly) see that the message is smaller
- go to https://ubuntu-com-10435.demos.haus/advantage/subscribe
- see that the message is not there

## Screenshots

image

### DIFF
--- a/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
+++ b/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
@@ -216,6 +216,20 @@ function StepOne({ setStep, error, setError }: StepOneProps) {
     });
   };
 
+  // Remove this once we want to release free trial
+  const TrialNotAvailable = location.search.includes("test_backend=true") ? (
+    <Row>
+      <Col size={10} emptyLarge={2}>
+        <p>
+          <strong>
+            Free Trial is not available for this account.{" "}
+            <a href="/contact-us">Contact us</a> for further information.
+          </strong>
+        </p>
+      </Col>
+    </Row>
+  ) : null;
+
   return (
     <>
       <ModalHeader title="Your details" />
@@ -232,17 +246,7 @@ function StepOne({ setStep, error, setError }: StepOneProps) {
               setIsUsingFreeTrial={setIsUsingFreeTrial}
             />
           ) : (
-            <Row>
-              <Col size={10} emptyLarge={2}>
-                <p>
-                  <strong>
-                    Free Trial is not available for this account.{" "}
-                    <a href="/contact-us">Contact us</a> for further
-                    information.
-                  </strong>
-                </p>
-              </Col>
-            </Row>
+            { TrialNotAvailable }
           )}
           <PaymentMethodSummary setStep={setStep} />
           <Row className="u-no-padding">
--- a/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
+++ b/static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
@@ -23,7 +23,7 @@ import Summary from "../../components/Summary";
 import FreeTrialRadio from "../../components/FreeTrialRadio";
 import { checkoutEvent, purchaseEvent } from "../../../../ecom-events";
 import { getSessionData } from "../../../../../utils/getSessionData";
-import { FormValues } from "../../utils/utils";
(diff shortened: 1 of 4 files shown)


static/js/src/advantage/subscribe/renderers/form-renderer.js
```diff
@@ -188,7 +188,13 @@ function renderSummary(state) {
       trialUnavailableSection.classList.add("u-hide");
     } else {
       freeTrialLabel.classList.add("u-hide");
-      trialUnavailableSection.classList.remove("u-hide");
+
+      // Remove this once we want to release free trial
```
petermakowski, line 192: I wonder if it would be helpful to introduce something resembling a feature flag, would remove the need for a comment and make it easier to update.

`const getIsFreeTrialEnabled () => location.search.includes("test_backend=true")`

petermakowski (approved):

static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
```diff
@@ -232,17 +246,7 @@ function StepOne({ setStep, error, setError }: StepOneProps) {
...
-                <p>
-                  <strong>
-                    Free Trial is not available for this account.{" "}
-                    <a href="/contact-us">Contact us</a> for further
-                    information.
-                  </strong>
-                </p>
-              </Col>
-            </Row>
+            { TrialNotAvailable }
```
petermakowski, line 249:
```suggestion
           <TrialNotAvailable />
```

petermakowski (commented):
I never see the `TrialNotAvailable ` message  as described in QA steps

ClementChaumel:
> I never see the `TrialNotAvailable ` message as described in QA steps

Do you see the free trial badge telling you you're eligible?
If so try selecting a monthly sub

static/js/src/advantage/subscribe/renderers/form-renderer.js
ClementChaumel: Good idea!

static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
```diff
@@ -216,6 +216,19 @@ function StepOne({ setStep, error, setError }: StepOneProps) {
     });
   };
 
+  const TrialNotAvailable = getIsFreeTrialEnabled() ? (
```
petermakowski, line 219:
```suggestion
  const TrialNotAvailable = () => getIsFreeTrialEnabled() ? (
```

static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
petermakowski: Oh, cause it's not a React component. Probably should be?

petermakowski:
Spacing around the `billing-period` select dropdown looks very strange.

image

static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
ClementChaumel: I mean yes we can't call it as a component because it's not one and I'm happy to change it to one but do we gain anything by doing this instead of just rendering the node like I did?

petermakowski: We gain consistency 🙃  Why use a different pattern than everywhere else all of a sudden? The less you have to wonder "why" while reading code, the better.

petermakowski (approved):

static/js/src/advantage/subscribe/react/components/ModalSteps/StepTwo.tsx
petermakowski: Fair points. But normally PascalCase naming convention indicates a _Component_. I'd rename this to `trialNotAvailable` as it's an _Element_. But it's me just nitpicking, really.

```jsx
const element = <h1>Hello, world</h1>;
```
> https://reactjs.org/docs/rendering-elements.html
https://reactjs.org/docs/components-and-props.html#rendering-a-component

ClementChaumel: Dang! You're right, wrong case!

PR #10440, opened 2021-09-21
Remove form background submit

anthonydillon:
## Done
- Removed the JS to background submit forms
- Removed the JS for the whitepaper which was removed earlier
- Added global form handling for generic success

## QA
- Open the demo
- Check a traditional form like /core/contact-us and check the form submits and goes to the thank you page
- Go to the something like /internet-of-things click contact to show the modal form submit test data and see it returns to the iot page with a success message
- Try a white paper download form at the /download/server, check they work.

## Issue / Card
Fixes https://github.com/canonical-web-and-design/web-squad/issues/4437


bethcollins92:
@anthonydillon do you have any specific QA steps?
After you submit the form now, removing the background submit means instead of the thank you message in the modal it redirects you to `/thank-you` - is this ok for now? not sure what the long term plan is for the change in the submit

anthonydillon:
Sorry I've added some detail to the QA steps. I'm unsure of the long term plan but it's an early step to simplify the forms across the site.

templates/templates/base.html
```diff
@@ -78,6 +78,21 @@
   {% if not(hide_nav == True) %}
     {% include "templates/_navigation.html" %}
   {% endif %}
+  <div class="p-strip is-shallow u-no-padding--bottom" id="success">
+    <div class="row">
+      <div class="p-notification--positive">
+        <p class="p-notification__response">Submittion was successful.</p>
```
nottrobin, line 84: This is fairly neat, but are you sure this is better than sending the user to a default thank-you page? A thank-you page is a bit more straightforward and obvious.

webapp/views.py
```diff
@@ -921,4 +922,4 @@ def marketo_submit():
     if return_url:
         return flask.redirect(return_url)
 
-    return flask.jsonify({"message": "Form submitted."})
```
nottrobin, line 924: Was *every* marketo form previously a background-submit one? That's shocking!

templates/blog/newsletter-form.html
```diff
@@ -77,16 +77,5 @@ <h3 class="p-card--title p-heading--four">Select topics you're <br />interested
   </div>
 </div>
 
-<script src="https://assets.ubuntu.com/v1/5d7e5bbf-jquery-2.2.0.min.js"></script>
-<script src="https://assets.ubuntu.com/v1/d55f58bb-jquery.validate.js"></script>
 <script src="https://www.google.com/recaptcha/api.js?onload=CaptchaCallback&render=explicit" defer>
-<script>
-  $("#mktoForm_1212").validate({
```
nottrobin, line 84: I love the idea of removing this, but why don't we need it any more?

anthonydillon: I am trying to standardised our forms a little. HTML5 validation validates for different types of inputs automatically. This validation method pre-dates HTML5.

templates/templates/base.html
anthonydillon: ReturnURL field is still respected. But this is a fallback for forms without a returnURL set.

webapp/views.py
anthonydillon: Yes, all of them

templates/blog/newsletter-form.html
nottrobin: Are UX cool with HTML5 validation?

anthonydillon: Good question, what do you think @clagom. Happy to demo it if you want to see the difference.

templates/templates/base.html
nottrobin: I understand that, I'm asking about the fall back.

webapp/views.py
nottrobin: :facepalm:

templates/templates/base.html
nottrobin: It's just that your implementation is a little brittle for a couple of reasons:

- you're redirecting them to the "referer" header, which *may* not exist for any number of reasons, including that their browser may deliberately block it.
- changing the hash value will remove the notification - e.g. if someone expands the top nav. This could be confusing.
- even if the notification displays fine, it could still be confusing to end up on a page that has nothing to do with the form you just submitted

So you go to "/blog", you see the newsletter form, you sign up for the newsletter. I find it a little hard to decide whether the user is expecting to be sent back to /blog. I think they probably don't, they probably want to be sure their sign up was successful.

It just seems to me that the more straightforward solution - both technically and for the user's experience - is to send them to a dedicated URL that says "thank you for submitting the form". There's no ambiguity; you're not trying to do two different things with a single page which is usually a bad idea.

If we have a referer value, we could put a link on that page saying "go back where you came from", or if you feel strongly we could automatically redirect the user back after 5 seconds or something (although I'm not crazy about that either).

And if this is just the fallback anyway (as in, well crafted forms will have a better target anyway), then I'd say we should go for the simpler solution that's less likely to break.

huwshimi, line 84:
```suggestion
        <p class="p-notification__response">Submission was successful.</p>
```

nottrobin:
@anthonydillon I think you need to fix the formatting

anthonydillon:
I am adding a generic thank you page as we didn't have one.

webapp/views.py
```diff
@@ -921,4 +922,4 @@ def marketo_submit():
...
+    return flask.redirect("/thank-you")
```
nottrobin, line 925: I know this was my suggestion, but it was just illustrative. `/thank-you` is a 404 :cry:

duynguyen0209 (approved):

templates/templates/base.html
anthonydillon: Your right this is brittle. I had removed the referrer and landed the user on a new generic thank you page.

webapp/views.py
anthonydillon: Sorry, created a thank you page as a separate commit!

templates/thank-you.html
```diff
@@ -0,0 +1,54 @@
...
+{% block head_extra%}<meta name="robots" content="noindex" />{% endblock %}
+{% block meta_copydoc %}https://docs.google.com/document/d/1uNruzpoTYxhrCJDxB0q_FH8z8YFs8w_m02fckevjFrU/edit{% endblock meta_copydoc %}
+{% block canonical_url %}https://ubuntu.com/contact-us/form{% endblock %}
+
+{% block content %}
+<section class="p-strip is-deep is-bordered">
+  <div class="row">
+    <div class="col-7">
+      <h1>Thanks for getting in touch</h1>
+      <h2 class="p-heading--three">Your submittion was successful.</h2>
```
nottrobin, line 13: So you decided not to add a "go back to {referer}" line?

nottrobin:
`webapp/views.py:632:5: F841 local variable 'referrer' is assigned to but never used`

templates/thank-you.html
anthonydillon: I was trying to work out how to pass parameters as a redirect.

nottrobin:
``` python3
def marketo_submit():
    # ...
    referrer = flask.request.referrer
    if referrer:
        return flask.redirect(f"/thank-you?referrer={referrer}")
    else:
        return flask.redirect(f"/thank-you")

def thank_you():
```

Or of course you could avoid the redirect with something like:

``` python3
def submit():
    # ...
    return flask.render_template(
        "submit.html",
        referrer=flask.request.referrer
    )
```

anthonydillon:
Thanks, done.

nottrobin (approved):
LGTM

anthonydillon:
@pmahnke and @clagom would you mind QA'ing this and making sure it works ok and the UX is alright?

clagom:
LGTM @anthonydillon, the only part I couldn't see because I didn't find a white paper to download is this: "Try a white paper download form at the /download/server, check they work."

anthonydillon:
Sorry about that the location of the white paper is at /download/server/s390x.

clagom:
Thanks @anthonydillon, LGTM

pmahnke:
@anthonydillon the QA seems fine and it is faster too.

PR #10467, opened 2021-09-23
add TypeScript check to GitHub actions

petermakowski:
## Done

- add missing FormValues import causing a TS error.
- add TypeScript check to GitHub actions to prevent such errors in the future
 
image
https://github.com/canonical-web-and-design/ubuntu.com/runs/3687925212


ClementChaumel (approved):
Good idea to add it to the PR checks

PR #10509, opened 2021-09-27
202/account users/pagination

petermakowski:
## Done

- Add pagination to Account users page results
  - handle response errors in requestAccountUsers
  - move cypress tests to `tests/cypress/integration`

## QA

- Go to advantage/users and login
- Add more users if you can't see pagination (pagination is enabled from 10 results +)
- Paginate through results and make sure nothing breaks

## Issue / Card

https://github.com/canonical-web-and-design/commercial-squad/issues/202


static/js/src/advantage/users/api.ts
```diff
@@ -55,25 +49,17 @@ const accountUserRequestInit: RequestInit = {
...
     throw new Error(responseJson.error || responseJson.errors);
   }
   return responseJson;
 };
 
-const fetchJSON = async (
-  input: RequestInfo,
-  init?: RequestInit
-): Promise<JSONResponse> => fetch(input, init).then(handleResponse);
+const fetchJSON = async (input: RequestInfo, init?: RequestInit) =>
```
carkod, line 61: Why do we need `async` here? we are not using the `await` afterwards?

carkod (approved):

carkod:
QAed with https://ubuntu-com-10509.demos.haus/advantage/users

static/js/src/advantage/users/api.ts
petermakowski: Good point. Will update 👍

PR #10517, opened 2021-09-27
Purchase preview for microcerts

carkod:
## Done

- Cube preview purchase
- Unified 2 endpoints in 1 by passing parameter `preview=true` if not parameter found it should go to `preview=false`

## QA

- Go to /cube/microcerts and click on purchase
or if you have HTTP client such as Insomnia, POstman, or Curl setup already you can:

POST /cube/microcerts/purchase.json?preview=true&test_backend=true

```
{
  "account_id":"<your account_id>", // This must be your account id on UA contracts
  "value":<price>, // This must be the value provided by GET /cube/microcerts.json
  "product_listing_id":",
  "previous_purchase_id":""
}
```
If successful it should provide data for the modal to do the complete purchase
```
{
  "amountDue": 1200, 
  "currency": "usd", 
  "customerAddress": {
    "city": "London", 
    "country": "GB", 
    "line1": "Bluefin", 
    "line2": "", 
    "postal_code": "12123", 
    "state": ""
  },
  ... 
}
```

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/262


webapp/cube/views.py
```diff
@@ -153,7 +153,7 @@ def get_microcerts():
     # also this avoids loading contracts api where not required
     is_test_backend = (
         current_app.config["CONTRACTS_TEST_API_URL"]
-        if flask.request.args.get("test_backend", "false")
+        if flask.request.args.get("test_backend", "true")
```
nottrobin, line 156: Why did you switch this around?

webapp/cube/views.py
```diff
@@ -275,6 +275,53 @@ def get_microcerts():
...
+    Purchase preview for CUBE microcertifications
+    """
+    preview = flask.request.args.get("preview", False)
+    account_id = flask.request.json["account_id"]
+    # Only purchase of one item allowed at a time
+    product_listing_id = flask.request.json["product_listing_id"]
+    previous_purchase_id = flask.request.json["previous_purchase_id"]
+    value = flask.request.json["value"]
+
+    is_test_backend = (
```
nottrobin, line 290: This is a really confusing block. So you set the `is_test_backend` variable, which I would expect to contain a boolean, to the value of either the `CONTRACTS_TEST_API_URL` or the `CONTRACTS_LIVE_API_URL`, which I'm pretty sure is a string?

I imagine you've copied this pattern from somewhere else, but at least for this function could you please rename this variable to something that makes sense to contain a URL?

webapp/cube/views.py
carkod: Because I assume that production is never going to have `test_backend=true`? so not setting it means it wiill be false?

webapp/cube/views.py
carkod: Yeah, I am basically using the same system as advantage to make it consistent, but removed the complexity of using the global store `flask.g.api` which I find it even worse (because we can use the current context in this case, no need for the global context). 
If we want to change the way we handle these 2 URLs, I think we should do it in both advantage and here too.

webapp/cube/views.py
carkod: Reversed. Just checked and it seems to work correctly.

webapp/cube/views.py
carkod: Changed it to your liking :). But yeah I agree about the variable name, didn't really notice.

webapp/cube/views.py
```diff
@@ -275,6 +275,53 @@ def get_microcerts():
     )
 
 
+@login_required
+def post_microcerts_purchase():
+    """
+    Purchase preview for CUBE microcertifications
+    """
+    preview = flask.request.args.get("preview", False)
```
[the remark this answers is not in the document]
carkod: I made it briefier, as we don't really need the preview False

nottrobin (approved):
Code looks good.

Looking at it now, I realise it's a POST endpoint. It's a little odd to be submitting parameters both on the URL and as POST data. You could maybe consider receiving `preview` as POST data too. But it's your call.

PR #10528, opened 2021-09-29
UA table redesign

ClementChaumel:
## Done

- Feature branch for the UA table redesign
- the PR is up for demo and QA but not ready to be merged yet


anthonydillon:
@wgx should auto renewing monthly contracts say they are expiring?

wgx:
> @wgx should auto renewing monthly contracts say they are expiring?

No, anything set to auto-renew should not show the user an expiry warning.

PR #10604, opened 2021-10-12
Prototype for demo of the proposed navigation

anthonydillon:
## Done
Prototype for demo of the proposed navigation

Currently only large screens only.


lyubomir-popov:
@waynecrosby @anthonydillon the text are actually vertically cventered, because if we align byu baseline, the menu items would sink a bit too much, which is noticeable when inside such a narrow strip:
image

I suggest making them vertically centered - all of them, e.g. by making them all have the same line-height, which automatically centers the text. (The menu padding will need to be factored in).

anthonydillon:
Done thanks @lyubomir-popov

anthonydillon:
Closing as we now have Scotts: https://github.com/canonical-web-and-design/ubuntu.com/pull/10996

PR #10619, opened 2021-10-13
Adapt user subscriptions to multiple marketplaces

albertkol:
## Done

- Adapt /advantage/user-subscriptions to return `blender` subscriptions

## QA

- Will be easier to QA when we can make blender purchases.

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/318


petermakowski:
QA all good, the endpoint returns correct data - but wasn't able to test blender obviously.

PR #10647, opened 2021-10-18
Render the plan name as HTML

ClementChaumel:
## Done

- use `dangerouslySetInnerHTML` to render the product name in the payment modal

If there is a better way to do this, let me know!

## QA

- https://ubuntu-com-10647.demos.haus//advantage/subscribe?test_backend=true
- select infra + apps
- check that the & in the name is rendered correctly

## Screenshots

image


anthonydillon:
Would this be better on the new UA table branch or you happy to resolve this conflict?

PR #10716, opened 2021-10-25
Add support for annual subscription renewals

alnvdl:
This is the adaptation of the renewal settings menu to support annual subscriptions. It changes a few important things:
1. The user-info endpoint no longer exists; instead, we compile subscription information from individual user subscriptions (contract items). This means there's a reduction in the number of calls made to ua-contracts to get information.
2. The renewal settings form and its associated endpoint now handle multiple subscriptions of different periods, even from different marketplaces if needed.
3. The subscription expiry notification is now conditioned on the fact that the subscription is not set to auto-renew. If it's configured to be renewed automatically, there's no notification.

This is my first time using TypeScript, Formik, react-query and the new setup of the Flask application (with webargs and all). Please question any of my choices here that you think may not be ideal.

I'm seeing a few tests failing in my machine, and Python linting is not working for me for some reason. I think I'll need some help fixing this.

To test this, go to https://ubuntu-com-10716.demos.haus/advantage?test_backend=true and checkout your renewals settings: if you have any active yearly/monthly subscription, it should now allow you to toggle that auto-renewal.

@frankban suggested that we only show the renewal toggle for annual subscriptions in the last few months. We should discuss how to signal that condition and hide the toggle if needed.


ClementChaumel (approved):
Thanks for the changes this looks good to me @alnvdl !

PR #10760, opened 2021-10-29
Make test_backend persist

albertkol:
## Done

- CUBE is terrible for testing at the moment because the testing flag gets dropped with every chance.
- This PR should patch at least the common user journeys

## QA

0. Ideally have an account without Edx user. DO NOT CREATE ONE UNTIL YOU REACH THE LAST STEP!!!
1. Test the /cube page
- Go to https://ubuntu-com-10760.demos.haus/cube
- Check the URL of the microcerts green button - DO NO CLICK ON IT YET! Hover over to see the link or inspect element.
- You should see /cube/microcerts
- Go to https://ubuntu-com-10760.demos.haus/cube?test_backend=true
- You should see /cube/microcerts?test_backend=true so the flag is there
2. Test the login redirect - because only logged in users can reach /cube/microcerts
- Go to https://ubuntu-com-10760.demos.haus/cube - BE LOGGED OUT!!!
- Click the microcerts link and login
- You should be redirected to /cube/microcerts
- LOG OUT!!!
- Go to https://ubuntu-com-10760.demos.haus/cube?test_backend=true - BE LOGGED OUT!!!
- Click the microcerts link and login
- You should be redirected to /cube/microcerts?test_backend=true
3. Test the Edx account creation redirect - Ideally you'd QA with a Ubuntu One account without an Edx account
- When you reach /cube/microcerts, if you have no Edx account you will get a pop-up asking you to create one
- If you click the create Edx account button in the pop-up and create the account you should be sent back to /cube/microcerts without losing the session
- If you already have an account copy-paste this URL and go there (this is the Edx account creating url, you can manually use it):
https://qa.cube.ubuntu.com/auth/login/tpa-saml/?auth_entry=login&idp=ubuntuone&next=https://ubuntu-com-10760.demos.haus/cube/microcerts?test_backend=true

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/356

### DIFF
--- a/webapp/cube/views.py
+++ b/webapp/cube/views.py
@@ -193,12 +193,16 @@ def cube_microcerts():
 
             courses.append(course)
 
+    edx_register_url = f"{edx_url}{flask.request.base_url}"
+    if flask.request.args.get("test_backend"):
+        edx_register_url = edx_register_url + "?test_backend=true"
+
     return flask.render_template(
         "cube/microcerts.html",
         **{
             "account_id": account["id"] if account else None,
             "edx_user": edx_user,
-            "edx_register_url": f"{edx_url}%2F",
+            "edx_register_url": edx_register_url,
             "sso_user": sso_user,
             "certified_badge": certified_badge or None,
             "modules": courses,
--- a/templates/cube/index.html
+++ b/templates/cube/index.html
@@ -13,7 +13,7 @@
       <h1>You&rsquo;re a professional.</br>Prove it with CUBE.</h1>
       <p class="p-heading--4">The Certified Ubuntu Engineer (CUBE) is a professional endorsement indicating mastery of Ubuntu. You need to pass 15 microcerts to earn your CUBE qualification.</p>
       <p>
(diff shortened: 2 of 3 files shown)


webapp/cube/views.py
```diff
@@ -193,12 +193,16 @@ def cube_microcerts():
 
             courses.append(course)
 
+    edx_register_url = f"{edx_url}{flask.request.base_url}"
+    if flask.request.args.get("test_backend"):
```
carkod, line 197: This will also work with `flask.request.args.get("test_backend") == "false"`? Anyway, if only devs use it should be fine, as long as people know that they need to drop it entirely

albertkol: Valid question. It would not work correctly. If users do `test_backend=false`, the env would be still set to testing. 

It was just a quick thing. I am not very happy I had to check for the `test_backend` like this inside the view function either, but I did not know a better way around it.

I will take a look to see if I can move it inside the decorator somehow.

PR #10763, opened 2021-10-29
Use the new GetAccountForMarketplace endpoint from ua-contracts

alnvdl:
This new endpoint behaves exactly like the old one, it's just a different path. The change in ua-contracts happened because of early account management developments, and some requirements we anticipate will arrive in the future.

The same migration was done to the new `EnsureAccountForMarketplace` last week, when the captcha for guest purchases was introduced. This PR is just finishing the work.

To test it, go to https://ubuntu-com-10763.demos.haus/advantage/subscribe?test_backend=true and view your payment methods. If it works, it works :)

### DIFF
--- a/tests/advantage/test_api.py
+++ b/tests/advantage/test_api.py
@@ -1002,7 +1002,7 @@ class TestGetPurchaseAccount(unittest.TestCase):
             client = make_client(session, is_for_view=is_for_view)
 
             with self.assertRaises(expected_error) as error:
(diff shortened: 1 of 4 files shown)


albertkol (approved):
:+1: It's working. 

If in the future the logic for getting the purchase account for different marketplaces diverge, we will have to revisit which marketplace we choose to use for our /account pages. @alnvdl

PR #10787, opened 2021-11-02
Fix blender cancel subscription

albertkol:
## Done

- Add marketplace parameter to the cancel subscription endpoint

## QA

- Have a monthly blender purchase
- If you do not have one use this demo and purchase one: https://ubuntu-com-10711.demos.haus/advantage/subscribe/blender?test_backend=true
- After you finished purchasing the blender product go to this demo: https://ubuntu-com-10787.demos.haus/advantage?test_backend=true
- You will see you blender subscription you purchased
- Hit edit subscription and then try to cancel the subscription
- It should work and after that the edit subscription button will disappear because the status of the subscription is `is_cancelled`. You can confirm that by checking the raw data ttps://ubuntu-com-10787.demos.haus/advantage/user-subscriptions?test_backend=true

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/378

### DIFF
--- a/static/js/src/advantage/api/contracts.js
+++ b/static/js/src/advantage/api/contracts.js
@@ -186,7 +186,12 @@ export async function postPurchaseData(
   return data;
 }
 
-export async function cancelContract(accountId, previousPurchaseId, productId) {
+export async function cancelContract(
+  accountId,
+  previousPurchaseId,
+  productId,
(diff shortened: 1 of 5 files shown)


ClementChaumel (approved):
Nice one!

PR #10788, opened 2021-11-02
Microcerts fix envs

carkod:
## Done

Fix issues between environments mentioned in https://github.com/canonical-web-and-design/ubuntu.com/pull/10743#issuecomment-957316184

## QA

- I believe best way to QA this is to use [Morgan's PR](https://github.com/canonical-web-and-design/ubuntu.com/pull/10743#issuecomment-957316184)
- Purchase a course and make sure that the buttons change to "Take" now
- Fail a course and check that it's not accessible anymore
- Header should activate the correct course you are taking (no changes from ubuntu.com)

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/366


konf/site.yaml
```diff
@@ -60,6 +60,16 @@ env:
     secretKeyRef:
       key: password
       name: badgr
+  
+  - name: BAGDR_QA_USER
```
albertkol, line 64: You define this, but you never you it because you hard-code it in the decorator.

webapp/advantage/decorators.py
```diff
@@ -144,18 +168,57 @@ def decorated_function(*args, **kwargs):
...
+                if not test_backend
+                else os.getenv("CUBE_EDX_CLIENT_QA_SECRET"),
+                edx_session,
+            )
+
+            if response == "html":
+                ua_api.set_is_for_view(True)
+
+            return func(
+                badgr_issuer,
```
albertkol, line 214: Up to you, but I'd add `badgr_issuer`, `certified_badge`, `ua_api`, `edx_api` in `kwargs`, if you want to take this approach to provide these values to the view functions. That way you do not have to define these 5 arguments for every future view function definition.

albertkol (changes requested):

albertkol:
QA results:
The env switch does not work at the moment, your condition in the decorator is `if not test_backend`. You have to change it to `if not is_test_backend`. `test_backend` is set to `false` as a string which is always `true` as a bool.

Please add the modules sorting done on /microcert.json to mircocerts view function. 

```
"modules": sorted(
    courses, key=lambda c: MODULES_ORDER.index(c["id"])
),
```

webapp/advantage/decorators.py
carkod: I think it's much clear this way, because you know exactly which are the parameters passed and the function will act as a guard if there are missing arguments.

konf/site.yaml
carkod: Yeah, I was thinking maybe we need to add a different user? But maybe that's not up to us. So removed.

carkod:
> QA results: The env switch does not work at the moment, your condition in the decorator is `if not test_backend`. You have to change it to `if not is_test_backend`. `test_backend` is set to `false` as a string which is always `true` as a bool.
> 
> Please add the modules sorting done on /microcert.json to mircocerts view function.
> 
> ```
> "modules": sorted(
>     courses, key=lambda c: MODULES_ORDER.index(c["id"])
> ),
> ```

👍 Fixed

carkod:
> More QA results: @mrgnr
> 
> The un-enrollement seems to be done correctly on submitting the exam. And the repurchase re-enrolls correctly.
> 
> What I am worried about is that we show `Failed` immediately after the exam was taken, even if you passed, because all we do is we look at the `attempts`. Not sure at this point how to fix this. We would need to investigate.

I can't replicate this. Don't you get "pass" after you are awarded a badge?

albertkol:
> I can't replicate this. Don't you get "pass" after you are awarded a badge? If you take an exam after you have enrolled, it will not show pass or fail, it will just allow you to continue. We can maybe catch up to check this.

We can solve that in a separate PR.

But we show `Passed` to people only if a Badge is awarded. 

But from the moment you take an exam and until you are awarded the badge. Your status will show `Failed`. Even if you got 100/100 from the exam. Unless the exam pass triggers automatically the badge assignment, but I don't think it happens, from what I tested.

albertkol (approved):
:+1:

carkod:
Let's catch up after planning week before we merge this to figure out the exam issue

albertkol:
It's not exactly related to this env work. I do not think it should block this from being merged.

The problem should exist on prod already.

carkod:
> It's not exactly related to this env work. I do not think it should block this from being merged.
> 
> The problem should exist on prod already.

Oh sure, in that case yes, it shouldn't block. We should create a new issue about this.

PR #10808, opened 2021-11-05
Fix issue that caused renewal settings to be displayed empty

alnvdl:
This is fixing a small issue that caused the renewal settings button and menu to be displayed empty if the user had any active subscription (monthly/yearly), even if it was not time to display it yet (nothing was displayed though, the menu would be empty).

Tests were also updated to catch this.


ClementChaumel (approved):
LGTM

PR #10851, opened 2021-11-13
release-cycle: updated text

nmav:
## Done

- Updated text not in copy doc

The text replaced (end-of-life) was picked by google when searching for "ubuntu 16.04 lifecycle", but it was incorrect. It is not visible in the page itself. If the text is coming from a supported spreadsheet let me know, but the spreadsheet I checked did not use these terms for LTS releases, so I'm updating the page directly.

### DIFF
--- a/templates/about/release-cycle.html
+++ b/templates/about/release-cycle.html
@@ -68,8 +68,8 @@
           <tr>
             <td>&nbsp;</td>
             <th scope="col">Released</th>
+            <th scope="col">End of Standard Support</th>
             <th scope="col">End of Life</th>
-            <th scope="col">Extended security maintenance</th>
           </tr>
           <tbody>
             <tr>


sowasred2012:
@nmav it looks like these terms come from this spreadsheet: https://docs.google.com/spreadsheets/d/1KIuOAa620yseE0uhIl67u2SlS0bWlPI_d98bvYM2Yl8/edit#gid=1902971094 (Desktop table sheet), so I don't think we should make this change.

nmav:
@sowasred2012 this is not a Desktop page but an Ubuntu lifecycle page, so I do not think it should get information from the Desktop tab. I'm happy to change them in the Desktop tab if that would help you, but it is not the right place.

nmav:
@pmahnke would it be possible to merge this? We already had a customer pointing this to us.

pmahnke:
The information is from the spreadsheet as Scott says.  And it based on the desktop.  The table is for mobile, so if you look at it there, you will see it.  I should parallel the chart.

Is the issue the term 'end of life'? Normally I would ask @lechsandecki  his opinion on the phrasing.

Also, Scott, the table and sheet are not out of date with the work that removed the 'Ubuntu' and added the release nickname.

sowasred2012:
> Also, Scott, the table and sheet are not out of date with the work that removed the 'Ubuntu' and added the release nickname.

Ah - will update that shortly :+1:

nmav:
Yes, the issue is the term "end of life" because we are now using "end of standard support". I've updated the spreadsheet, we are already in sync with @lechsandecki  about the terms.

sowasred2012 (approved):
LGTM

PR #10882, opened 2021-11-17
Update user management role description

ClementChaumel:
## Done

- Update user management role description

## QA

- The page is under maintenance but it's a copy change. How bad can I mess it up?

## Issue / Card

Fixes canonical-web-and-design/commercial-squad#400


albertkol (approved):
:+1: I mean... it's just some copy changes. What's the worst that can happen?

PR #10916, opened 2021-11-22
Display new price when resizing

ClementChaumel:
## Done

- Added a preview of what will happen if the user decides to resize
- Moved the formatter to utils

## QA

- go to 
- select a subscription (works best with monthly ones because they can go up and down)
- click on Edit subscription
- change the number
- see that it previews how the change will affect your subscription.

## Issue / Card

Fixes canonical-web-and-design/commercial-squad#387

## Screenshots

image

image

image


wgx:
@ClementChaumel looks good to me!

albertkol (approved):
:+1:

PR #10922, opened 2021-11-23
Disable the SSL check on live link checking

anthonydillon:
## Done
- Migrated the live-links check from linkchecker to linkcheckerrc
- Disable SSL certificate checking to hide external cert issues

## QA
- Check the code


sowasred2012 (approved):
LGTM

PR #10966, opened 2021-11-30
Fix entitlement problem

albertkol:
## Done

- If `obligations.enabledByDefault` was missing from the entitlement payload the code would crash. This change prevents that, defaulting to `False`.

## QA

- Go to /advantage. Does it error out?

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/420


ClementChaumel (approved):
:fire: :leaves:

PR #11011, opened 2021-12-08
Add endpoint to surface users purchase account linked offers

jpmartinspt:
## Done

Add endpoint to surface users purchase account linked offers

## QA

- Create an offer using the contracts API
- Go to http://0.0.0.0:8001/advatage/offers and check the offer

### DIFF
--- a/webapp/advantage/ua_contracts/api.py
+++ b/webapp/advantage/ua_contracts/api.py
@@ -109,6 +109,15 @@ class UAContractsAPI:
 
         return contract
 
+    def get_account_offers(self, account_id: str):
+        response = self._request(
+            method="get",
+            path=f"v1/accounts/{account_id}/offers",
+            error_rules=["default", "no-found", "user-role"],
+        )
+
+        return response.json()
+
     def get_account_users(self, account_id: str):
         response = self._request(
             method="get",
--- a/webapp/advantage/views.py
+++ b/webapp/advantage/views.py
@@ -770,6 +770,22 @@ def accept_renewal(renewal_id):
     return g.api.accept_renewal(renewal_id)
 
 
+@advantage_decorator(permission="user", response="json")
+def get_account_offers():
+    try:
+        account = g.api.get_purchase_account("canonical-ua")
+    except UAContractsUserHasNoAccount:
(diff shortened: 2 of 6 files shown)


webapp/advantage/views.py
```diff
@@ -757,6 +757,17 @@ def accept_renewal(renewal_id):
     return g.api.accept_renewal(renewal_id)
 
 
+@advantage_decorator(permission="user", response="json")
+def get_account_offers():
+    try:
+        account = g.api.get_purchase_account("canonical-ua")
+    except UAContractsUserHasNoAccount:
```
albertkol, line 764: If the user has `technical` role `get_purchase_account` throws a `AccessForbiddenError`. We have to `except` that too and throw a `403`. 

You can see the error rules of an endpoint at `error_rules=["default", "no-found", "user-role"]` in `api.py`, not saying it's the smartest thing in the world, but it's attempting to gather most of the error handling logic in one place.

webapp/app.py
```diff
@@ -407,6 +408,12 @@ def utility_processor():
     methods=["PUT"],
 )
 
+app.add_url_rule(
+    "/advantage/offers",
```
albertkol, line 412: The endpoint path conflicts with the one Clem needs for the HTML template. No?

We do not have a convention for our endpoints to avoid these conflicts. Any suggestions are welcome:
- we did `.json` suffix at some point
- perhaps doing `/api/v1` might become handy too

Would be worth discussing.

webapp/advantage/ua_contracts/api.py
```diff
@@ -109,6 +109,15 @@ def get_contract(self, contract_id: str):
 
         return contract
 
+    def get_account_offers(self, account_id: str):
```
albertkol, line 112: You can write unit tests for the endpoint.

You can see examples at: `/tests/advantage/test_api.py`. I think all our other api functions are covered.

albertkol (approved):
Small things mentioned but it's clean and it's working.

jpmartinspt:
@albertkol I think I addressed all your comments, if all is good feel free to merge. Otherwise leave feedback and I'll look tomorrow.

PR #11040, opened 2021-12-14
Second step of the modal is reached even if setting customer info fails

ClementChaumel:
## Done

- Remove the explicit instruction to ignore VAT validation. :thinking: 

I can't remember why this was done. I tried most journeys, they still work as expected.

## QA

- go to https://ubuntu-com-11040.demos.haus/advantage/subscribe?test_backend=true as a guest or with a brand new SSO account
- select a product
- Fill in the details
- Select a country where VAT is applicable (eg. UK)
- input `GB123` as the VAT number
- Click next
- it should fail with an error telling you to fix the VAT number

## Screenshots

image

### DIFF
--- a/static/js/src/PurchaseModal/PurchaseModal.tsx
+++ b/static/js/src/PurchaseModal/PurchaseModal.tsx
@@ -15,6 +15,7 @@ import {
 import StepOne from "./components/ModalSteps/StepOne";
 import StepTwo from "./components/ModalSteps/StepTwo";
 import { BuyButtonProps } from "./utils/utils";
+
 type Props = {
   accountId?: string;
   termsLabel: React.ReactNode;
@@ -97,6 +98,11 @@ const PurchaseModal = ({
                   <a href="/login">sign in</a> to your account first.
                 </>
               );
+            } else if (error.message === "tax_id_invalid") {
+              actions.setErrors({
+                VATNumber:
+                  "That VAT number is invalid. Check the number and try again.",
+              });
             } else {
               const knownErrorMessage = getErrorMessage({
                 message: "",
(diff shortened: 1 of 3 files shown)


albertkol:
You will hate me for this but... the VAT field does not get highlighted with a red border on error. 

Look at city for example.. you get the highlight with the message under it..
image

Looking at logged in step 1, is it difficult to have the logged out VAT validation work the same?
image

The message does show up indeed at the top.

Up to @wgx if he is happy with how it is right now, then I am happy to leave it.

ClementChaumel:
> You will hate me for this but... the VAT field does not get highlighted with a red border on error.
> 
> Look at city for example.. you get the highlight with the message under it.. image
> 
> Looking at logged in step 1, is it difficult to have the logged out VAT validation work the same? image
> 
> The message does show up indeed at the top.
> 
> Up to @wgx if he is happy with how it is right now, then I am happy to leave it.

I fixed it!

I did find another issue though, to register the payment method for guest users we first need to post to `/advantage/purchase-account` to get the accountID needed for `/advantage/customer-info`

Meaning, if someone tries to make a purchase as a guest and gets the VAT wrong they will see the message above. But when they will try to correct it they will see the "An account with this email address already exists" error.

albertkol:
Can you not set the `account_id` as soon as the the `POST /advantage/purchase-account` happens? And prevent the request to happen again after?

albertkol (approved):
The validation works, but we are falling in another problem where after the VAT is corrected the user can still not submit the form anymore because the `purchase-account` with the email address already exists. 

After `purchase-account` endpoint is called and an `account_id` is returned, we should store it and prevent further `purchase-account` calls.

ClementChaumel:
> The validation works, but we are falling in another problem where after the VAT is corrected. The user can still not submit the form anymore because the `purchase-account` with the email address already exists.
> 
> After `purchase-account` endpoint is called and an `account_id` is returned, we should store it and prevent further `purchase-account` calls.

It should work correctly now!

albertkol:
@ClementChaumel it works! Awesome stuff!

PR #11041, opened 2021-12-14
increase cypress page load timeout

petermakowski:
## Done

- increase cypress page load timeout
- remove invalid formFolder configuration option


minkyngkm (approved):
LGTM! Thanks Peter 👍

PR #11056, opened 2021-12-15
UA Shop: Add label to subscriptions that are not editable

jpmartinspt:
## Done
Show a label saying a subscription is not editable when they are not.

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/advantage?test_backend=true
- Check a subscription that was purchased on offer and make sure it displays `This subscription cannot be edited online.`

### DIFF
--- a/static/js/src/advantage/react/components/Subscriptions/SubscriptionDetails/SubscriptionDetails.tsx
+++ b/static/js/src/advantage/react/components/Subscriptions/SubscriptionDetails/SubscriptionDetails.tsx
@@ -136,7 +136,14 @@ export const SubscriptionDetails = forwardRef<HTMLDivElement, Props>(
                   >
                     Edit subscription&hellip;
                   </Button>
-                ) : null}
+                ) : (
+                  <h5 className="u-no-padding--top p-subscriptions__details-small-title">
+                    <span style={{ fontWeight: 300 }} className="u-text--muted">
+                      {" "}
+                      This subscription cannot be edited online.
+                    </span>
+                  </h5>
+                )}
               </>
             )}
           </div>


static/js/src/advantage/react/components/Subscriptions/SubscriptionDetails/SubscriptionDetails.tsx
```diff
@@ -136,7 +136,14 @@ export const SubscriptionDetails = forwardRef<HTMLDivElement, Props>(
                   >
                     Edit subscription&hellip;
                   </Button>
-                ) : null}
+                ) : (
```
albertkol, line 139: Will this not make it so that any subscription that is not `isResizable` will show the message: "This subscription cannot be edited online"? One problem is that `legacy` user subscriptions have `isResizable` always `false` because we do not allow them to be resized. If we want this to show up only for `Offers`, we need to condition it by user subscription type.

albertkol: Or did we drop the idea of having user subscriptions of type `offer`?

jpmartinspt: Sorry I left some comments on the ticket but not here. Since we can't differentiate an offer from a legacy without extra API calls, I asked Will if we could just do a generic message whenever we are dealing with something we can't edit/resize. The messaging was updated on [miro](https://miro.com/app/board/o9J_lydRtns=/?moveToWidget=3458764514679240808&cot=14) now too.

wgx:
@albertkol the string of text looks right to me on the right-hand panel.

static/js/src/advantage/react/components/Subscriptions/SubscriptionDetails/SubscriptionDetails.tsx
albertkol: I see.. but my first comment still stands. If we cannot differentiate between `legacy` and `offers` then it mean we cannot correctly condition for the message to show only for `offers`. 

The condition for displaying this should still not be based on the `else` side of `isResizable`. 

Can CS not help us with distinguishing `legacy` from `offers`?

albertkol (approved):
:+1:

PR #11106, opened 2022-01-05
Sharable cart referral link - get the data

ClementChaumel:
## Done

- get and display the user's offers on the /offers page

## QA

- log in on https://ubuntu-com-11106.demos.haus/advantage?test_backend=true
- go to https://ubuntu-com-11106.demos.haus/advantage/offers?test_backend=true (with some offers on your account)
- check that they display correctly

## Issue / Card

Fixes #

## Screenshots

image

## Help


wgx:
LGTM! 👍🏽

static/js/src/advantage/offers/Offers.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import { Row, Col } from "@canonical/react-components";
-import OffersList from "./components/OffersList";
+import OffersList from "./components/OffersList/OffersList";
```
anthonydillon, line 3: You could just the index file to load this to keep it tidy

static/js/src/advantage/offers/components/Offer/Offer.tsx
```diff
@@ -0,0 +1,58 @@
...
+        );
+      })}
+      <div className="advantage-offer-row">
+        <p className="p-text--x-small-capitalised allowance">
+          Total before taxes
+        </p>
+        <p className="price">{currencyFormatter.format(offer.total / 100)}</p>
+      </div>
+      <div
+        style={{ width: "100%", display: "flex", justifyContent: "flex-end" }}
```
anthonydillon, line 41: Can you use vanilla utilities instead of inline styles?

static/js/src/advantage/offers/components/Offer/Offer.tsx
```diff
@@ -0,0 +1,58 @@
...
+          Contract item
+        </p>
+        <p className="p-text--x-small-capitalised allowance">Allowance</p>
+        <p className="p-text--x-small-capitalised price">Price</p>
+      </div>
+      <hr />
+      {offer.items.map((item: Item) => {
+        return (
+          <div key={item.id} className="advantage-offer-row">
+            <p className="contract-item">
```
anthonydillon, line 24: Why do each of these items have a unique class name?

static/js/src/advantage/offers/components/Offer/Offers.test.tsx
```diff
@@ -0,0 +1,31 @@
+import React from "react";
```
anthonydillon, line 1: This file name should be `Offer` not `Offers`

static/js/src/advantage/offers/components/Offer/Offer.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <div
+        style={{ width: "100%", display: "flex", justifyContent: "flex-end" }}
+      >
+        <ActionButton
+          appearance="positive"
+          className="u-no-margin--bottom"
+          onClick={() => {
+            alert("nope.");
+          }}
+          disabled={!offer.actionable}
```
anthonydillon, line 49: Could you add a comment please. Seems strange to have an offer that is not actionable from the code.

static/js/src/advantage/offers/components/OffersList/OffersList.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { OfferFactory } from "../../tests/factories/offers";
+
+describe("OffersList", () => {
+  let queryClient: QueryClient;
+
+  beforeEach(async () => {
+    queryClient = new QueryClient();
+  });
+
+  it("can display an offer", () => {
```
anthonydillon, line 15: Can you add a test for no offers?

static/js/src/advantage/offers/components/OffersList/OffersList.tsx
```diff
@@ -0,0 +1,39 @@
...
+import useGetOffersList from "../../hooks/useGetOffersList";
+import Offer from "../Offer";
+import { Offer as OfferType } from "../../types";
+
+const OffersList = () => {
+  const { isLoading, isError, data: offersList, error } = useGetOffersList();
+
+  if (isError) {
+    Sentry.captureException(error);
+    return <p>Something went wrong while trying to retrieve your offers.</p>;
```
anthonydillon, line 12: Maybe offer the user since advise?

static/sass/styles.scss
```diff
@@ -1026,3 +1026,27 @@ html {
...
   }
 }
+
+// Advantage offers
+
+.advantage-offer-row {
+  align-items: baseline;
+  display: grid;
+  grid-template-areas: "contract-item allowance price";
+  grid-template-columns: 2fr 1fr 1fr;
```
anthonydillon, line 1036: Could this not use grid classes?

templates/advantage/offers/index.html
```diff
@@ -15,9 +15,11 @@ <h1>Your Ubuntu Advantage offers</h1>
...
         <div class="row">
             <div class="col-12">
-                <div class="u-no-margin--bottom p-card">
-                    <div class="p-card__content"><span class="p-text--default"><i class="p-icon--spinner u-animation--spin"></i></span>
-                        Loading…</div>
+                <div class="p-card__content">
+                    <span class="p-text--default">
+                    <i class="p-icon--spinner u-animation--spin"></i>
+                    </span>
+                    <p>Loading offers...</p>
```
anthonydillon, line 22: This loading is duplicated

anthonydillon (commented):
Few quick comments @ClementChaumel

static/js/src/advantage/offers/Offers.tsx
ClementChaumel: VSCode tried to be smart here

static/js/src/advantage/offers/components/Offer/Offer.tsx
ClementChaumel: I'm not sure this is an actual use case but there is an `actionable` attribute so I figured out we should at least disable the button if it's not actionnable

static/js/src/advantage/offers/components/OffersList/OffersList.tsx
ClementChaumel: @wgx should we ask them to reach out to the saleperson who added the offer to their account?

templates/advantage/offers/index.html
ClementChaumel: What do you mean? 
The idea is that the loading state in the template is the same as the react App's so the mounting of the app is seamless. Do you think it's redundant?

static/js/src/advantage/offers/components/Offer/Offer.tsx
ClementChaumel: I tried first with Vanilla utilites but couldn't get it to work the way I wanted. Could you point me in the right direction?

static/sass/styles.scss
ClementChaumel: I'll try!

static/js/src/advantage/offers/components/Offer/Offer.tsx
ClementChaumel: I think I got it!

templates/advantage/offers/index.html
anthonydillon: I think its good at first but once changed is likely to lead to off jumps and so I would recommend just having it within the JS component.

static/js/src/advantage/offers/components/Offer/Offer.tsx
anthonydillon: Would be interesting to find out why that is in the payload and add a note as this is likely to confuse the next person to touch this code

static/js/src/advantage/offers/components/OffersList/OffersList.tsx
anthonydillon: When the user is not logged in they get this error when it would be good to let them know they need to sign in.

templates/advantage/offers/index.html
anthonydillon: Could you try and put the Loading text next the icon not under?

anthonydillon:
I get a `Error: User has no purchase account` when QA'ing

static/js/src/advantage/offers/components/OffersList/OffersList.tsx
ClementChaumel: Or maybe we should redirect to /advantage if the user is logged out?

static/js/src/advantage/offers/components/Offer/Offer.tsx
ClementChaumel: @albertkol do you know if it comes from the CS API?

ClementChaumel:
> I get a `Error: User has no purchase account` error in the console when QA'ing

Do you have offers on your account? Did you log in with the `?test_backend=true` parameter?

templates/advantage/offers/index.html
ClementChaumel: I udpated it, PTAL

anthonydillon:
I assume I don't buy would expect the length to be 0 then. I did login with test backend.

static/js/src/advantage/offers/components/Offer/Offer.tsx
albertkol: We can filter them out. It's not a problem. I will check to see if we can fetch only actionable ones.

albertkol:
@anthonydillon @ClementChaumel 

> `Error: User has no purchase account`

All the purchase activities are tied to a purchase account. It gets initialised on the 1st purchase a user makes. Offers are also tied to a purchase account.

In real life, with offers, Sales ops will be the ones initialising the purchase account for the user when they create the offer. So clients won't end up seeing this error message.

anthonydillon:
> @anthonydillon @ClementChaumel
> 
> > `Error: User has no purchase account`
> 
> All the purchase activities are tied to a purchase account. It gets initialised on the 1st purchase a user makes. Offers are also tied to a purchase account.
> 
> In real life, with offers, Sales ops will be the ones initialising the purchase account for the user when they create the offer. So clients won't end up seeing this error message.

I think display an error of "Something went wrong while trying to retrieve your offers." in that case is disingenuous and likely to lead to confusing if the user logs into the wrong account by mistake. I would suggest displaying the "No offers available" message in both cases. And only display the "Something went wrong..." message when is a genuine error with or from the API

static/js/src/advantage/offers/components/OffersList/OffersList.tsx
anthonydillon: The is what we do for user accounts. Not a huge fan of that behaviour but it would be consistent.

static/js/src/advantage/offers/components/Offer/Offer.tsx
anthonydillon: Hiding unactionable offers could lead to confusion for users, depending on the reason why. If the user is expecting to see an offer and it expires it would be a better UX to display the offer with a message of why this is now unactionable.

anthonydillon:
@ClementChaumel I still get the same message with the latest change

anthonydillon (approved):
Code and QA (from an account without offices) :+1:

PR #11110, opened 2022-01-05
Surface actual number of machines

albertkol:
## Done

- Add `current_number_of_machines` as a property of the `user_subscription`. This value is the actual number of machines the user will have starting with the next billing cycle. It should also be used as the pre-filled value in the edit number of machines field.

## QA

- Go to https://ubuntu-com-11110.demos.haus or check out this feature branch 
- Do you have a stripe subscription that you can down-sell? If no, buy one.
- Down-sell the user subscription; you might have to manually refresh the page, as the front-end is not fully supporting the new kind of resizing
-  On /advantage you will still see the old value of machine numbers (as until the end of the billing period it won't change), but on /advantage/user-subscriptions you will see a new property called `current_number_of_machines` which will show you the next billing period's number of machines

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/445

### DIFF
--- a/tests/advantage/test_helpers.py
+++ b/tests/advantage/test_helpers.py
@@ -15,6 +15,7 @@ from tests.advantage.helpers import (
 )
 from webapp.advantage.models import Entitlement
 from webapp.advantage.ua_contracts.helpers import (
+    get_current_number_of_machines,
     get_items_aggregated_values,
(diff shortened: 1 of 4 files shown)


ClementChaumel (approved):
:snake:

PR #11119, opened 2022-01-06
Add Modal demo using react-aria

mrgnr:
## Done

- Implemented an accessible modal using Vanilla and [react-aria](https://react-spectrum.adobe.com/react-aria/)

## QA

- Run the server with `dotrun`
- Go to http://localhost:8001/demo
- Click the button to open the modal
  - You should be able to tab through focusable elements in the modal
  - You should be able to close the modal by pressing Esc

## Screenshots

react-aria-modal


bartaz:
Would that work with our Vanilla `Modal` component from react-components? https://canonical-web-and-design.github.io/react-components/?path=/story/modal--default-story

mrgnr:
> Would that work with our Vanilla `Modal` component from react-components? https://canonical-web-and-design.github.io/react-components/?path=/story/modal--default-story

@bartaz Yes, looking at the code I think it would work. I'll try to give it a shot this week and hopefully have a PR up so we can decide whether this is a good direction to go in.

PR #11136, opened 2022-01-11
Replace deprecated p-button--neutral class with p-button

sowasred2012:
## Done

This is part of the work being done for the Vanilla 3.0 upgrade (see: https://github.com/canonical-web-and-design/ubuntu.com/pull/11130).

3.0 includes the removal of a lot of classes which were simply deprecated in earlier versions, however ubuntu.com is still using these old classes, in many places. Addressing those changes in the renovate PR would lead to a massive changeset that would be a nightmare to review, so I'm addressing the changes that can be done without actually updating to 3.0 in separate PRs, each with a single change.

This PR replaces all uses of `p-button--neutral` with `p-button`.

## QA

- Browse the site, look for any instances of the neutral button (a lot of the contact us modals have them), see that they look the same as they do on live.
- Search the codebase for instances of "p-button--neutral", there should be none.


bethcollins92 (approved):
Great idea with the separate PRs 😅. LGTM

PR #11137, opened 2022-01-11
Replace deprecated heading class names

sowasred2012:
## Done

This is part of the work being done for the Vanilla 3.0 upgrade (see: https://github.com/canonical-web-and-design/ubuntu.com/pull/11130).

3.0 includes the removal of a lot of classes which were simply deprecated in earlier versions, however ubuntu.com is still using these old classes, in many places. Addressing those changes in the renovate PR would lead to a massive changeset that would be a nightmare to review, so I'm addressing the changes that can be done without actually updating to 3.0 in separate PRs, each with a single change.

This PR replaces all uses of deprecated heading-level-as-word classes with heading-level-as-number classes, e.g. `p-heading--four` with `p-heading--4`.

## QA

- Browse the site, make sure any headings look the same as they do on live.
- Search the codebase for instances of "p-heading--one"/"two"/"three"/"four"/"five"/"six", there should be none.


minkyngkm (approved):
LGTM👍

PR #11139, opened 2022-01-12
Change cancellation feature

albertkol:
## Done

- Change cancellation feature. Cancellations are technically a downsize to 0. Before it used to have an extra step of logic where if the user cancelled the last stripe user subscription then we cancel the subscription completely. With the new downsize rule, the stripe subscription cancellation is no longer necessary

## QA

- Buy a monthly subscription
- Upsize 
- Downsize
- Upsize
- Cancel
- Did it work?

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/444

### DIFF
--- a/tests/advantage/test_api.py
+++ b/tests/advantage/test_api.py
@@ -6,7 +6,6 @@ from webapp.advantage.ua_contracts.api import (
     UAContractsAPIError,
     UAContractsAPIErrorView,
     UAContractsUserHasNoAccount,
-    CannotCancelLastContractError,
     UnauthorizedError,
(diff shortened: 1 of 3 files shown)


jpmartinspt (approved):
I can do what you suggested on the QA steps but there seems to be no permanent messaging when you downzise/cancel something, is that supposed to be done after? If so feel free to merge.

albertkol:
@jpmartinspt the messages will be done in the front-end separately. This can be merged without any worry because only monthly products can be cancelled and no monthly products are on production at the moment.

PR #11161, opened 2022-01-18
Shareable cart link, buy offers

ClementChaumel:
## Done

- Open the payment modal and allow users to purchase offers when they click buy

## QA

- go to https://ubuntu-com-11161.demos.haus/advantage?test_backend=true 
- login with an account that has offers available
- go to https://ubuntu-com-11161.demos.haus/advantage/offers?test_backend=true
- try to buy an offer
- it should work correctly and redirect you to /advantage

## Screenshots

image

### DIFF
--- a/static/js/src/PurchaseModal/PurchaseModal.tsx
+++ b/static/js/src/PurchaseModal/PurchaseModal.tsx
@@ -22,7 +22,7 @@ type Props = {
   marketingLabel: React.ReactNode;
   product?: any;
   preview?: any;
-  quantity: number;
+  quantity?: number;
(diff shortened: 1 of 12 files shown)


albertkol:
@ClementChaumel from QA-ing it. The purchase seems to work. 

In edge scenarios of users buying with:
- a card that requires 3DS checks (4000002760003184) results in the form freezing
- a card that has the payment failing (4000000000000341) results in the form freezing

Also the table does not look good on small screens.

albertkol (approved):
:+1:

PR #11170, opened 2022-01-20
Adjust offer redirect

albertkol:
## Done

- Send people to login if hitting /advantage/offers logged out

## QA

- use demo
- go to /advantage/offers?test_backend=true
- it will redirect you to login with staging account and then send you back to /advantage/offers?test_backend=true
- do the same but without the test flag, it should work the same

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/471


jpmartinspt (approved):

jpmartinspt:
Seems good, except for the linting.

PR #11171, opened 2022-01-20
Fix email keyerror in marketo forms

carkod:
## Done

Fix `KeyError: 'email'`

## QA

- Sentry error points to this engage page https://ubuntu.com/engage/wellcome-sanger-case-study as the referrer, but it probably will not fail even without this check.

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/11143

## Help

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -679,8 +679,12 @@ def marketo_submit():
         ],
     }
 
-    # Enrichment data for global enrichment form (id:4198)
-    enrichment_fields = {"email": form_fields["email"]}
+    enrichment_fields = None
+
+    if "email" in form_fields:
+        # Enrichment data for global enrichment form (id:4198)
+        enrichment_fields = {"email": form_fields["email"]}
+
     try:
         ip_location = ip_reader.get(client_ip)
         if ip_location and "country" in ip_location:


webapp/views.py
```diff
@@ -679,19 +679,22 @@ def marketo_submit():
...
-        if ip_location and "country" in ip_location:
-            enrichment_fields["country"] = ip_location["country"]["iso_code"]
-    except Exception:
-        pass
-
-    enriched_payload = {
-        "formId": "4198",
-        "input": [{"leadFormFields": enrichment_fields}],
-    }
+    if "email" in form_fields:
```
anthonydillon, line 682: We just want this condition around line 684 as we would expect line 736 to still submit all passed fields.

anthonydillon (changes requested):

webapp/views.py
carkod: Ok. In that case I initiated the variable with `None` otherwise we'll get variable not defined if it fails the check

sowasred2012 (approved):
LGTM - I signed up on live and locally, and then checked Marketo. In both cases, I could see my submission, so I'm not sure how the original error came about, but it hasn't broken anything. I think this is OK to merge?

PR #11174, opened 2022-01-21
Tidy up security routing

nottrobin:
- Remove unused /security/api/notices endpoints
- Tidy up routing of /security section
  - Only route /security/notices and /security/cves to `ubuntu-com-security` service - this is because I want to isolate specifically the security database logic in this service, so problems with the database can be seen very clearly and won't effect other areas.
  - Route /security/certifications/docs to `ubuntu-com-discourse` (like other docs)
- change /security/cve URL to /security/cves
  - This is to match the API (/security/cves.json) and because all our other URLs follow this plural scheme, as well as this being the recommendation in [Web API Design](https://robinwinslow.uk/api-design-ebook-2012-03.pdf).

## QA

Can't really QA the k8s routing stuff, we have to put that on a k8s platform first. But you can QA the /security/cve -> /security/cves redirects are working ok.

### DIFF
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -515,7 +515,7 @@ staging:
         - name: SENTRY_DSN
           value: https://0293bb7fc3104e56bafd2422e155790c@sentry.is.canonical.com//13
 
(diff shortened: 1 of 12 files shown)


mtruj013 (approved):
LGTM!

PR #11189, opened 2022-01-25
Fix "Pay now" display for resizes

ClementChaumel:
## Done

- Only display "You will pay now $X" if the user will actually pay as a result of the resize
- Display the correct amount in the pay now paragraph if the user has downsized their subscription during the same cycle

## QA

- Go to [/advantage](https://ubuntu-com-11189.demos.haus/advantage?test_backend=true)
- log in with an account that has a monthly subscription
- downsize it
- the resize it again but up this time
- check that it states that you will pay when you click on resize only for an amount of machines greater than what you had initially

## Screenshots

resize


albertkol (approved):
:+1: Remarkably managed! :1st_place_medal:

PR #11205, opened 2022-01-28
Disabe email field if it was prepopulated

ClementChaumel:
## Done

- If we can't fetch Stripe data, fall back on `account.json` to prepopulate the name and email fields
- disable email field if it was prepopulated 

## QA

- Go to [/advantage/subscribe](https://ubuntu-com-11205.demos.haus/advantage/subscribe?test_backend=true)
- Log in with a brand new SSO account
- your name and email should be prepopulated, email field should be disabled
- Log in with an account that already has made a purchase
- the same should happen

## Issue / Card

Fixes https://github.com/canonical-web-and-design/commercial-squad/issues/299

## Screenshots

image


albertkol:
It does not do the magic for the renewal modal. Test with: /advantage?email=albert.kolozsvari%2Btester83@canonical.com&test_backend=true

albertkol (approved):
:+1: One in a million, Clem!

PR #11210, opened 2022-01-31
Move common view function outisde

albertkol:
## Done

- Move common endpoints of advantage and CUBE in /shop/views
- These are mostly account related endpoints used by the modal 

## QA

- Make a UA and a CUBE purchase. Does it work? 

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/481

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -911,3 +911,12 @@ nvidia: /engage/ubuntu-at-nvidia-gtc-2021
 # Move /securtiy/cve to /security/cves
 security/cve: /security/cves
 security/cve/sitemap(?P<suffix>.*).xml: /security/cves/sitemap{suffix}.xml
+
+# Shop redirects
+/advantage/purchase-account: "/account/purchase-account"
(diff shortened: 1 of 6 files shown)


redirects.yaml
```diff
@@ -911,3 +911,12 @@ nvidia: /engage/ubuntu-at-nvidia-gtc-2021
 # Move /securtiy/cve to /security/cves
 security/cve: /security/cves
 security/cve/sitemap(?P<suffix>.*).xml: /security/cves/sitemap{suffix}.xml
+
+# Shop redirects
+/advantage/purchase-account: "/account/purchase-account"
```
jpmartinspt, line 916: I don't think the redirects on "POST" endpoints will work.

jpmartinspt:
Seems to be working locally, but I see some cypress failures, so might have missed something?

redirects.yaml
albertkol: I can re-QA it. But it think it worked.

jpmartinspt (approved):

PR #11230, opened 2022-02-07
Russian engage pages

sowasred2012:
## Done

- Enabled display of cyrillic characters
- Added Russian ("Русский") as an option in the /engage index

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/engage
- Select "Русский" from the language dropdown, there should be four results
- Click any of them, the font on the engage page should look a lot better than it does on the live site, and the font family being used should be "UbuntuFullCharSet".
- Go to a non Russian engage page, inspect the text, the font family should be "Ubuntu"

## Screenshots

Before:

Screenshot from 2022-02-07 16-59-54

After:

Screenshot from 2022-02-07 17-00-06

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -10,23 +10,10 @@
 // import cookie policy
 @import "@canonical/cookie-policy/build/css/cookie-policy";
 
-// Font display Vanilla override
-@font-face {
-  font-display: optional;
-  font-family: "UbtuOverride";
-  font-style: normal;
-  font-weight: 300;
-  src: url("https://assets.ubuntu.com/v1/e8c07df6-Ubuntu-L_W.woff2")
-      format("woff2"),
-    url("https://assets.ubuntu.com/v1/8619add2-Ubuntu-L_W.woff") format("woff");
-}
-
 // import vanilla-framework
 @import "vanilla-framework/scss/build";
 @import "vanilla-placeholders";
 
-// Vanilla framework overrides
-$font-base-family: '"UbtuOverride", -apple-system, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif' !default;
 $color-shadow: rgba(0, 0, 0, 0.5) !default;
 
 // import site specific patterns and overrides
--- a/static/sass/_global-settings.scss
+++ b/static/sass/_global-settings.scss
@@ -10,4 +10,5 @@ $contrast-ratio: 5%;
 $color-whitepaper-accent: #772a54;
 $color-mid-x-light: #e5e5e5;
 $font-use-subset-latin: true;
+$font-allow-cyrillic-greek-latin: true;
 $font-display-option: fallback;
--- a/templates/engage/index.html
+++ b/templates/engage/index.html
@@ -37,6 +37,7 @@
           <option value="fr">Fran&ccedil;ais</option>
           <option value="it">Italiano</option>
           <option value="pt">Portugu&ecirc;s</option>
+          <option value="ru">Русс&#1082;&#1080;&#1081;</option>
         </select>
       </div>
     </div>


templates/engage/base.html
```diff
@@ -16,12 +16,37 @@
   {% if document['metadata']['path'] == "/engage/ShellOnboarding" %}
     <meta name="robots" content="noindex, nofollow" />
   {% endif %}
+
+  {# The Ubuntu font used acrros the site doesn't contain Russian characters, #}
```
anthonydillon, line 20:
```suggestion
  {# The Ubuntu font used across the site doesn't contain Russian characters, #}
```

templates/engage/base.html
```diff
@@ -16,12 +16,37 @@
...
+  {# import the full char set for any Russian engage pages #}
```
anthonydillon, line 21:
```suggestion
  {# if the page content is Russian then import the full character set #}
```

anthonydillon (changes requested):
It's a shame that we need to load so many font's. Would need to explore some ideas to work out what might be better but this works.

My concern is that the text on this page would not render nicely? https://ubuntu-com-11230.demos.haus/engage?language=ru

anthonydillon:
In fact I think you can just use:
```
@font-face {
          font-display: $font-display-option;
          font-family: 'Ubuntu';
          font-style: normal;
          font-weight: $font-weight-regular-text;
          src: url('#{$assets-path}5bea8279-Ubuntu-L-cyrillic-subset.woff2') format('woff2'), url('#{$assets-path}b8058442-Ubuntu-L-cyrillic-subset.woff') format('woff');
          unicode-range: U+0400-045F, U+0490-0491, U+04B0-04B1, U+2116;
        }
```

anthonydillon:
Or just set the Vanilla setting `$font-allow-cyrillic-greek-latin` to true.

anthonydillon:
The `Ubuntu-L-cyrillic-ext-subset.woff2` file is 26kb where as the `Ubuntu-L_W.woff2` you are using is 127.3kb,

sowasred2012:
@anthonydillon I just tried that subset, it _mostly_ works, but the heading in the hero strip uses `font-weight: 100;`, which that subset doesn't seem to cover - @pmahnke is going to create a new subset for that font-weight to use here.

pmahnke:
I created these two subsets:

- [light](https://assets.ubuntu.com/v1/3026c62c-Ubuntu-L-subset-cyrillic.woff2)
- [regular](https://assets.ubuntu.com/v1/d72414f8-Ubuntu-R-subset-cyrillic.woff2)

anthonydillon:
> My concern is that the text on this page would not render nicely? https://ubuntu-com-11230.demos.haus/engage?language=ru

This is still a concern

anthonydillon:
Sorry @sowasred2012 I miss read the code in Vanilla. You can use both settings together:
```sass
$font-use-subset-latin: true;
$font-allow-cyrillic-greek-latin: true;
```
Looks like we just need to add light font to the set provided by Vanilla. Which @pmahnke is doing, right?

Then you can remove the `UbtuOverride` stuff. As we `$font-display-option` variable locally on ubuntu.com. This slightly increases (138kb to 197kb) the font size across the site but if we have Russian on the site we should include the font globally.

carkod:
Wow looks nice, no difference between the two alphabets, but you may also want to include it in the /index page.

index-engage-cyrllic

sowasred2012:
> Update: Oh I suppose that is the issue with the font weight variations?
Yep, will be looking at this again today.

static/sass/styles.scss
```diff
@@ -10,23 +10,11 @@
...
-    url("https://assets.ubuntu.com/v1/8619add2-Ubuntu-L_W.woff") format("woff");
-}
-
 // import vanilla-framework
 @import "vanilla-framework/scss/build";
 @import "vanilla-placeholders";
 
-// Vanilla framework overrides
-$font-base-family: '"UbtuOverride", -apple-system, "Segoe UI", "Roboto", "Oxygen", "Cantarell", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif' !default;
+// // Vanilla framework overrides
```
anthonydillon, line 17: Do you need this comment anymore? It's double commented too.

anthonydillon (approved):
Piece of art @sowasred2012, perfect :+1:

static/sass/styles.scss
sowasred2012: Nope!

PR #11261, opened 2022-02-18
Set MarkupSafe to version 2.0.1

mtruj013:
## Done

- Set MarkupSafe to version 2.0.1

## QA

- See that the site runs correctly 

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/11260

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+MarkupSafe==2.0.1
 canonicalwebteam.flask-base==0.9.3
 alembic==1.7.5
 canonicalwebteam.http==1.0.3


requirements.txt
```diff
@@ -1,3 +1,4 @@
+MarkupSafe==2.0.1
 canonicalwebteam.flask-base==0.9.3
```
carkod, line 2: Can you move it below flask-base? we had issues before, whereas the module on top overrode the Python version of the rest of the modules. This way we ensure that the flask-base Python version is the one we use.

requirements.txt
```diff
@@ -1,3 +1,4 @@
+MarkupSafe==2.0.1
```
carkod, line 1: Can you move it below flask-base? we had issues before, whereas the module on top overrode the Python version of the rest of the modules. This way we ensure that the flask-base Python version is the one we use.

nottrobin:
Fixed in https://github.com/canonical-web-and-design/ubuntu.com/pull/11259 instead

PR #11284, opened 2022-02-24
Fix vendor test to request a vendor that currently exists

sowasred2012:
## Done

- Updated vendor test to make a request to /certified/vendors/HP, rather than /certified/vendors/Dell (Dell seems to have been removed from the partners API)

## QA

- See that all the checks pass


anthonydillon (approved):
LGTM :+1:

PR #11305, opened 2022-03-02
Fix security error pages

mtruj013:
## Done

- Render error page based on error code
- create a 404 security error page
- fix date bug for cves that do not have a published date included in the api payload

## QA

- Visit http://0.0.0.0:8001/security/CVE-2003-0854 and see that a 404 error page is returned which includes the id you tried to query and a message explaining that a cve with that id does not exist
- Visit http://0.0.0.0:8001/security/CVE-2007-0822 and see that the cve template renders without error (this cve does not have much in its payload, if you notice missing sections it's not an error, they just don't exist)

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/11290


webapp/app.py
```diff
@@ -198,7 +198,11 @@ def bad_request_error(error):
 
 @app.errorhandler(SecurityAPIError)
 def security_api_error(error):
-    return flask.render_template("security-error.html"), 500
+    message = error.response.json()
+    if error.response.status_code == 404:
+        return flask.render_template("security-error-404.html", message=message), 404
```
nottrobin, line 203: You're certainly right that we need to catch 404s here. But I don't see the need to create a separate template, this just means we have more different templates to maintain, when the actual 404 page you've added is pretty similar.

I'm not sure if it's helpful to show the "CVE-xxxx does not exist" message, or if a normal 404 page will suffice. It's up to you. But if you do still want to pass through that message, you can still do it in a generic way without creating an extra template. Change this code to:

```suggestion
    if error.response.status_code == 404:
        flask.abort(404, message)
```

And then change the general `404.html` template to:

``` html
<p class="p-heading--4">{{ message or 'Sorry, we couldn't find that page.' }}</p>
```

webapp/app.py
```diff
@@ -198,7 +198,11 @@ def bad_request_error(error):
 
 @app.errorhandler(SecurityAPIError)
 def security_api_error(error):
-    return flask.render_template("security-error.html"), 500
+    message = error.response.json()
```
nottrobin, line 201: Rather than doing `message.message` in the template, which is kinda clunky, why not extract the  message here instead:

```suggestion
    message = error.response.json().get("message")
```

nottrobin (changes requested):
Just a couple of suggestios

webapp/app.py
```diff
@@ -198,7 +198,16 @@ def bad_request_error(error):
 
 @app.errorhandler(SecurityAPIError)
 def security_api_error(error):
-    return flask.render_template("security-error.html"), 500
+
+    message = error.response.json().get("message")
+
+    if error.response.status_code == 404:
+        return flask.render_template("404.html", message=message), 404
```
nottrobin, line 205: Did you prefer this syntax to my suggestion of `flask.abort()`? Or did `abort` not work?

nottrobin, line 205: I did write a message about "why not use `flask.abort`?", then I realised it's because that doesn't work when you're already within an error handler. This is a good solution.

I'm wondering, though, if this highlights that we're doing the wrong thing a bit? If what we really want to say is that something is just missing, maybe raising is as a `SecurityAPIError` is fundamentally misleading.

I think I'm going to merge this so we can get it out and then file a follow-up PR to try a restructure and then you can review it and see what you think. Sorry for all the back and forth.

nottrobin (approved):
LGTM, QAed etc.

PR #11343, opened 2022-03-12
Improve handling of API errors

nottrobin:
This is an attempt to provide a better structure for handling CVE 404 errors following my comment here:

https://github.com/canonical-web-and-design/ubuntu.com/pull/11305#discussion_r824645422

Basically, we ended up in a situation where we were trying to raise a 404 error from within the `security_api_error` error handler, but we had to do it manually with `flask.render_template` rather than the more proper `flask.abort` because you can't `abort` from within an active `errorhandler`.

I think this was pointing to a sort of code smell, that we shouldn't really have got as far as raising a `SecurityAPIError` if really all that happened is that a CVE is missing. In fact, I think it would be much more reasonable for `SecurityAPI.get_cve` to simply return `None`, rather than an error, if the CVE in question doesn't exist. This is reinforced by the fact that the code to turn that `None` response into a `flask.abort(404)` actually [already existed](https://github.com/canonical-web-and-design/ubuntu.com/blob/b001f4e0a14af6ac586ab4de8c318f6ff99b0a8c/webapp/security/views.py#L725-L726) in the `cve` view, even though it was not not being used.

So I've restructured the code to do the logic of reading the `404` code from the API's response inside the `SecurityAPIError.get_cve` method itself, and return `None` in that case. The `SecurityAPIError._get` method now simply does `raise_for_status()`, which seems reasonable as it is simply a generic HTTP request method and shouldn't have any greater knowledge than that, and then it's up to the public methods (currently `get_cve` and `get_releases`) to turn those `HTTPError` exceptions into `SecurityAPIError`s or `None` as appropriate. I hope this makes sense.

To facilitate this branch I've also [proposed a change](https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/pull/56) to canonicalwebteam.flask-base. This branch uses that new version. A review of this branch is effectively a review of that one. Please go and approve that one once this is reviewed, but don't merge this until that PR is merged and published and this PR's `requirements.txt` is updated.

QA
--

Go to https://ubuntu-com-11343.demos.haus/security/CVE-1234-5678, see a beautiful 404 page, with a custom error message.

Go to https://ubuntu-com-11343.demos.haus/zfgzdfsd check the normal 404 page still works.

Go to https://ubuntu-com-11343.demos.haus/security/CVE-2022-23639, check a normal CVE still displays properly.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-canonicalwebteam.flask-base==0.9.3
+canonicalwebteam.flask-base==1.0.2
 alembic==1.7.5
 canonicalwebteam.http==1.0.3
 canonicalwebteam.blog==6.4.0
@@ -27,5 +27,3 @@ macaroonbakery==1.3.1
 sortedcontainers==2.4.0
 vcrpy-unittest==0.1.7
 webargs==7.0.1
-markupsafe==2.0.1
-itsdangerous==2.0.1
--- a/templates/404.html
+++ b/templates/404.html
@@ -20,11 +20,13 @@
     <div class="col-6 u-vertically-center">
       <div>
         <h1>404: Page not found</h1>
-        {% if message %}
-          <p class="p-heading--4">{{ message }}</p>
-        {% else %}
-          <p class="p-heading--4">Sorry, we couldn't find that page.</p>
-        {% endif %}
+        <p class="p-heading--4">
+          {% if message %}
+            {{ message }}
+          {% else %}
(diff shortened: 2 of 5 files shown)


mtruj013 (approved):
QA'd successfully, [#56](https://github.com/canonical-web-and-design/canonicalwebteam.flask-base/pull/56) has been merged so I believe all that's needed is to update `requirements.txt` :)

PR #11360, opened 2022-03-18
WIP - Update cve_list view so that it consumes security api

mtruj013:
## Done

- Add function to api class for getting all cves
- Update cve_list view to api instead of querying the db directly 
- Refactor any broken filtering because of the above change
**NOTE: Refactoring is still in progress**

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/security/cves
- Make sure cves load successfully and that filtering works as expected 

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/11359


webapp/security/views.py
```diff
@@ -507,31 +507,49 @@ def cve_index():
...
         return flask.redirect(f"/security/{query.lower()}")
 
-    all_releases = (
-        db_session.query(Release)
-        .order_by(desc(Release.release_date))
-        .filter(Release.codename != "upstream")
-        .all()
-    )
+    # releases in desc order
+    releases_json = security_api.get_releases().get("releases")
```
nottrobin, line 523: I'd suggest that you move the `.get("releases")` part into the `security_api.get_releases()` function, so that `get_releases` literally returns a list of releases. Makes things less surprising.

``` python3
def get_releases():
    # ...
    return response.json().get("releases")
```

templates/security/cve/_cve-table.html
```diff
@@ -47,71 +57,66 @@
...
+          <td>{{ package.name }}</td>
+          {% if version_releases %}
+          {% for version in version_releases  %}
+            {% set status = package.release_statuses[version.codename] %}
+  
+            {% if status %}
+              <td class="cve-table-cell status-{{ status.slug }}">
+                <div class="cve-color-strip--{{ status.slug.lower() }}"></div>
+                <div class="icon-container__text">
+                  {{ status.name or "&mdash;" }}
```
nottrobin, line 69: Did this work? I wrote it and expected it to fail...

templates/security/cve/_cve-table.html
```diff
@@ -2,20 +2,30 @@
...
+        {% endfor %}
+      {% else%}
+        {% for release in releases %}
+          <th style="width: 8em;">
+            <div class="icon-container__text">
+              {{ release.version }} {{ release.support_tag }}
+            </div>
+          </th>
+        {% endfor %}
+      {% endif %}
```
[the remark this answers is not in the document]
mtruj013: Releases are the 7 default filtered releases that show on the table when you don't have a version query, version_releases (which I should prob rename now that you mention it) are any releases that match the version you are querying

nottrobin, line 23: Here, these two loops are literally identical right? except built off different lists. Could we rename `version_releases` to `selected_releases` and always build this block off that?

templates/security/cve/_cve-table.html
```diff
@@ -47,71 +57,66 @@
...
+                  {% endif %}
+                </div>
+              </td>
             {% else %}
               <td class="cve-table-cell--muted">
                 <div class="cve-color-strip--dne"></div>
                 &mdash;
               </td>
             {% endif %}
           {% endfor %}
```
nottrobin, line 118: Are there any HTML differences between this block and the previous block? If not, again, we should build this once off a single `selected_releases` list, which should contain all releases if there's no filter releases specified.

templates/security/cve/_cve-table.html
nottrobin: (and then rename `releases` to `all_releases`, for building the form, to make it clear that it's never the filtered set)

webapp/security/views.py
```diff
@@ -499,202 +498,102 @@ def cve_index():
...
-        versions and statuses and len(versions) == len(statuses)
+    # get cves and total results
+    cves_response = security_api.get_cves(
+        query=query,
+        priority=priority,
+        package=package,
+        limit=limit,
+        offset=offset,
+        component=component,
+        version=version,
```
andesol, line 519: @mtruj013 Didn't debug the API response very well, but I'm gonna take a guess:

If you select two Ubuntu versions the url is like  `...&version=jammy&status=&version=focal&status=...`

Since `version` is only the first one (`flask.request.args.get("version", default="", type=str)` only returns one) could it be that you're requesting only one version from the API and that's why the table is empty for the others?

nottrobin (approved):

PR #11367, opened 2022-03-18
Add new purchase endpoint and get purchase v2

albertkol:
## Done

- Add new purchase endpoint which does:
  - Get purchase account ensure a purchase account
  - PUT customer info
  - Preview or make purchase
  - Endpoint info: `POST /advantage/purchase/{preview}` Returns an `Invoice` if preview or `Purchase` if not preview.
  - Payload sample:
```
{
    "customer_info": {
        "payment_method_id": "",
        "email": "albert.kolozsvari+testerabcd340@canonical.com",
        "name": "abc",
        "address": {
            "city" : "London2",
            "country": "GB",
            "line1": "",
            "postal_code": "123123",
            "state": ""
        },
        "tax_id": {
            "type": "",
            "value": ""
        }
    },
    "products": [
        {
            "product_listing_id": "lAObiUcHYzxOh_EVHCLGHoc8eARipfgy7itnjK1xSOiE",
            "quantity": 1
        }
    ],
    "marketplace": "canonical-ua",
    "action": "purchase",  // can be resize or trial too
    "previous_purchase_id": "",
    "captcha_value": "123"
}
```
- Add get purchase v2 endpoint mapped version of the old purchase endpoint

## QA

- Will be easier to test with the front-end

## Issue / Card

Fixes #https://github.com/canonical-web-and-design/commercial-squad/issues/508

### DIFF
--- a/webapp/shop/api/ua_contracts/advantage_mapper.py
+++ b/webapp/shop/api/ua_contracts/advantage_mapper.py
@@ -1,8 +1,9 @@
-from typing import List, Dict, Optional
+from typing import List, Dict, Optional, Union
 
 from webapp.shop.api.ua_contracts.api import UAContractsAPI
 from webapp.shop.api.ua_contracts.models import (
     Purchase,
+    Invoice,
     UserSubscription,
     Offer,
     Listing,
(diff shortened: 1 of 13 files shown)


webapp/shop/api/ua_contracts/advantage_mapper.py
```diff
@@ -186,3 +209,123 @@ def get_user_subscriptions(self, email: str) -> List[UserSubscription]:
...
+
+            account = self.ensure_purchase_account(
+                marketplace=marketplace,
+                email=email,
+                account_name=name,
+                captcha_value=captcha_value,
+            )
+
+            self.ua_contracts_api.set_authentication_token(account.token)
+            self.ua_contracts_api.set_token_type("Bearer")
```
albertkol, line 230: I know this is nasty, but I was unsure how to change the state of the API object for the remaining calls to the API (to use the token received when we created the user's purchase account).

Basically as a guest we use `Beader <token>` authentication and for logged in users its `Macaroon <long-cookie-thing>`.

webapp/shop/api/ua_contracts/advantage_mapper.py
```diff
@@ -186,3 +209,123 @@ def get_user_subscriptions(self, email: str) -> List[UserSubscription]:
             )
 
         return build_user_subscriptions(user_summary, listings)
+
+    def get_or_create_user_account(
+        self, marketplace, customer_info, captcha_value
+    ) -> Account:
+        try:
+            return self.get_purchase_account(marketplace)
```
jpmartinspt, line 217: Does this call not need authentication?

albertkol: It does. If the user is logged in and user has a purchase account the `get_purchase_account` would return thier `Account` instance.

If user is not logged in or the user is logged in but has no purchase account, it will fall down the `except` path and we ensure an `Account` for them.

jpmartinspt: Hmm so this relies on having set an attribute on the API class somewhere else?

jpmartinspt (approved):

PR #11369, opened 2022-03-21
Update notice views to consume security api

mtruj013:
## Done

- add get notice function to api class
- refactor notice view to comply with new payload structure
- update usn template to comply with new variable names 
**NOTE: Must be merged AFTER releases endpoint has been updated on security api as base_url needs will need to be updated**

## QA

- Click on any notice and see that it renders correctly
- Try out any and all links on the template and make sure that they work

## Issue / Card

Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/11272

### DIFF
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -60,7 +60,6 @@ def notice(notice_id):
     releases = {
         release["codename"]: release["version"] for release in all_releases
     }
-    print(releases)
 
     if notice["release_packages"]:
         for codename, pkgs in notice["release_packages"].items():
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -43,7 +43,7 @@ def notice(notice_id):
 
     security_api = SecurityAPI(
         session=session,
-        base_url="http://192.168.64.6:8030/security/",
+        base_url="http://ubuntu.com/security/",
     )
 
     notice = security_api.get_notice(notice_id)
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -30,6 +30,11 @@ markdown_parser = Markdown(
 )
 session = talisker.requests.get_session()
 
+security_api = SecurityAPI(
+    session=session,
+    base_url="https://ubuntu.com/security/",
+)
+
 
 def get_processed_details(notice):
     pattern = re.compile(r"(cve|CVE-)\d{4}-\d{4,7}", re.MULTILINE)
@@ -41,11 +46,6 @@ def get_processed_details(notice):
 
 def notice(notice_id):
 
-    security_api = SecurityAPI(
-        session=session,
-        base_url="http://ubuntu.com/security/",
-    )
-
     notice = security_api.get_notice(notice_id)
 
     if not notice:
@@ -55,10 +55,9 @@ def notice(notice_id):
     package_versions = {}
     release_packages = SortedDict()
 
-    all_releases = security_api.get_releases().get("releases")
-
     releases = {
-        release["codename"]: release["version"] for release in all_releases
+        release["codename"]: release["version"]
+        for release in notice["releases"]
     }
 
     if notice["release_packages"]:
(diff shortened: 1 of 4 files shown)


webapp/security/views.py
```diff
@@ -31,8 +31,22 @@
...
+        pattern, r'<a href="/security/\g<0>">\g<0></a>', notice["description"]
+    )
+
+
 def notice(notice_id):
-    notice = db_session.query(Notice).get(notice_id)
+
+    security_api = SecurityAPI(
+        session=session,
+        base_url="http://ubuntu.com/security/",
```
nottrobin, line 46: Same point, `https` please

webapp/security/views.py
```diff
@@ -41,13 +55,14 @@ def notice(notice_id):
     package_versions = {}
     release_packages = SortedDict()
 
+    all_releases = security_api.get_releases().get("releases")
```
nottrobin, line 58: Weren't we going to do this outside of the view, so it only gets loaded once?

webapp/security/views.py
```diff
@@ -31,8 +31,22 @@
...
+
+    return re.sub(
+        pattern, r'<a href="/security/\g<0>">\g<0></a>', notice["description"]
+    )
+
+
 def notice(notice_id):
-    notice = db_session.query(Notice).get(notice_id)
+
+    security_api = SecurityAPI(
```
nottrobin, line 44: Could you please move the `security_api` from the `cve` function to the global part of this file, so it can be used by both views?

webapp/security/views.py
```diff
@@ -41,13 +55,14 @@ def notice(notice_id):
...
-        release.codename: release.version
-        for release in db_session.query(Release).all()
+        release["codename"]: release["version"] for release in all_releases
     }
 
-    if notice.release_packages:
-        for codename, pkgs in notice.release_packages.items():
+    if notice["release_packages"]:
+        for codename, pkgs in notice["release_packages"].items():
             release_version = releases[codename]
```
nottrobin, line 66: Hmm. Looking into this, I don't think you need the `releases` endpoint at all.

So, we only need the `releases` data for getting the `release_version` for every release codename mentioned in `release_packages`.

It looks to me like every notice JSON response (e.g. https://ubuntu.com/security/notices/USN-5340-1.json or https://ubuntu.com/security/notices/LSN-0085-1.json) already has a `releases` property, which contains exactly the set of releases mentioned in the `release_packages` list.

So we really just need to do the dictionary you're making in line 44 from `notice["releases"]` rather than `all_releases` and that should be it.

nottrobin (changes requested):
- Remove use of releases endpoint
- Switch to `https://ubunntu.com`

webapp/security/views.py
```diff
@@ -41,13 +55,14 @@ def notice(notice_id):
     package_versions = {}
     release_packages = SortedDict()
 
+    all_releases = security_api.get_releases().get("releases")
+
     releases = {
-        release.codename: release.version
-        for release in db_session.query(Release).all()
+        release["codename"]: release["version"] for release in all_releases
```
nottrobin, line 61: Hmm. Looking into this, I don't think you need the `releases` endpoint at all.

So, we only need the `releases` data for getting the `release_version` for every release codename mentioned in `release_packages`.

It looks to me like every notice JSON response (e.g. https://ubuntu.com/security/notices/USN-5340-1.json or https://ubuntu.com/security/notices/LSN-0085-1.json) already has a `releases` property, which contains exactly the set of releases mentioned in the `release_packages` list.

So we really just need to do the dictionary you're making in line 44 from `notice["releases"]` rather than `all_releases` and that should be it.

```suggestion
        release["codename"]: release["version"] for release in notice["releases"]
```

webapp/security/views.py
nottrobin, line 58:
```suggestion
```

PR #11375, opened 2022-03-21
Add unlisted engage pages

sowasred2012:
## Done

We had a complaint recently that an engage page went live, and it should not have been visible in the index or "You may be interested in" section on engage page thank you pages. We didn't have the functionality to support that, so this PR adds it, with two pages that should not appear on /engage, or in the "You may be interested in" section.

## QA

  - https://ubuntu-com-11375.demos.haus/engage/unlisted/custom-ubuntu-pro-image-on-azure
  - https://ubuntu-com-11375.demos.haus/engage/unlisted/shell-onboarding
- You should see two engage pages
- Attempt to visit:
  - https://ubuntu-com-11375.demos.haus/engage/custom-ubuntu-pro-image-on-azure
  - https://ubuntu-com-11375.demos.haus/engage/ShellOnboarding
- You should be redirected to their unlisted counterparts

### DIFF
--- a/templates/engage/unlisted/custom-ubuntu-pro-image-on-azure.html
+++ b/templates/engage/unlisted/custom-ubuntu-pro-image-on-azure.html
@@ -1,0 +1,65 @@
+{% extends "engage/base_engage.html" %}
+
+{% block title %}Using Shell's Ubuntu Pro image on Azure{% endblock %}
+{% block meta_description %}Get started with custom images of Ubuntu Pro with Support on Azure{% endblock %}
+
+{% block head_extra %}
+<meta name="robots" content="noindex, nofollow" />
+{% endblock %}
+
+{% block meta_copydoc %}{% endblock meta_copydoc %}
+
+{% block content %}
+<section class="p-strip p-engage-banner--grad">
+  <div class="u-fixed-width navigation-logo-engage">
+    <a href="/">
+      {{
+        image(
+          url="https://assets.ubuntu.com/v1/f263d9c4-logo-ubuntu-white.svg",
+          alt="Ubuntu",
+          height="32",
+          width="143",
+          hi_def=True,
+          loading="auto"
+        ) | safe
+      }}
+    </a>
+  </div>
+
+  <div class="row">
+    <div class="col-7 u-vertically-center">
+      <h1>Using Shell&#39;s Ubuntu Pro image on Azure</h1>
+      <p class="u-no-padding--top p-heading--4">
+        Get started with custom images of Ubuntu Pro with Support on Azure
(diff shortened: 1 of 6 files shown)


templates/engage/unlisted/custom-ubuntu-pro-image-on-azure.html
```diff
@@ -0,0 +1,65 @@
+{% extends "engage/base_engage.html" %}
+
+{% block title %}Using Shell's Ubuntu Pro image on Azure{% endblock %}
+{% block meta_description %}Get started with custom images of Ubuntu Pro with Support on Azure{% endblock %}
+
+{% block head_extra %}
+<meta name="robots" content="noindex, nofollow" />
```
anthonydillon, line 7: I am slightly worried that someone might create a page here in the future and omit this piece which is the important contract with the unlisted page. I wonder if it's better to create a new base_engage that enforces this?

templates/engage/unlisted/custom-ubuntu-pro-image-on-azure.html
```diff
@@ -0,0 +1,65 @@
+{% extends "engage/base_engage.html" %}
+
+{% block title %}Using Shell's Ubuntu Pro image on Azure{% endblock %}
+{% block meta_description %}Get started with custom images of Ubuntu Pro with Support on Azure{% endblock %}
```
anthonydillon, line 4: Probably don't need this as its unlisted.

templates/engage/unlisted/custom-ubuntu-pro-image-on-azure.html
```diff
@@ -0,0 +1,65 @@
...
+
+{% block head_extra %}
+<meta name="robots" content="noindex, nofollow" />
+{% endblock %}
+
+{% block meta_copydoc %}{% endblock meta_copydoc %}
```
anthonydillon, line 10: If you have a new base template you don't need this as empty noise

anthonydillon (changes requested):
Some comments in the code for dicussion.

templates/engage/unlisted/custom-ubuntu-pro-image-on-azure.html
sowasred2012: That's a good point/something I wondered - I originally went with adding this to both templates, thinking in future people would just use them as references for other pages and copy that block over, and we then wouldn't need to create a new base, but maybe the trade-off of another template is worth it if it makes it less error-prone.

anthonydillon (approved):
Much nicer, thanks. Just a reminder to remove posts from discourse too.

PR #11377, opened 2022-03-21
UA selector: Set billing period to monthly when selecting apps.

jpmartinspt:
## Done
UA selector: Set billing period to monthly when selecting apps. 
This will allow users who have UA apps in their offerings to purchase them.

## QA

- Login with an account that has been offered an UA apps listing.
- Go to `/advantage/subscribe`
- Try to buy `UA Apps`


albertkol (approved):
:+1: Cypress needs fixing or just rerunning.

PR #11398, opened 2022-03-24
Reduce the number of pods across ubuntu.com

nottrobin:
This is so that we don't max out database connections.

See https://chat.canonical.com/canonical/channels/web--design/yzw9p348j7g83xpw4nad4n4ymr


jpmartinspt:
This looks fine, but we might not need to drop it more, as the issue we had does not seem related to db connections this time.

PR #11399, opened 2022-03-24
Add a regex match on USN routes.

jpmartinspt:
## Done

Match the USN IDs are valid so we don't match potentially problematic strings.

## QA

- Make sure you can still view USN and LSN detail pages

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,3 +27,5 @@ macaroonbakery==1.3.1
 sortedcontainers==2.4.0
 vcrpy-unittest==0.1.7
 webargs==7.0.1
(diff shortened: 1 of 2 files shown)


nottrobin (approved):

jpmartinspt:
@nottrobin something is up with the run-image step, I think somehow when building the image it's pulling a version of jinja2 that is incompatible with the version of flask we use. jinja2 was released yesterday but I could not pinpoint what is the other library that is pulling jinja2 and overriding flask's setup.py. If you want to merge this in the morning you might have to do some digging. Also, I don't seem to be able to replicate with dotrun..

PR #11405, opened 2022-03-28
Add 'New' label to 'Account users' navigation link

jpmartinspt:
## Done

When user has not dismissed the notification show on /advantage about
account management show the 'New' label on 'Account users' navigation link.

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/advantage?test_backend=true
- Make sure you see the `New` label on the secondary navigation before `Account Users`
- Dismiss the notification about managing users on the page
- Refresh and make sure the label and the notification are gone.


albertkol:
The label does not show up.

`localStorage.getItem("dismissedOnboardingNotification")` does not get initialised to `false` so it can be also `null` your condition should be something like this to display the label:

`if (localStorage.getItem("dismissedOnboardingNotification") === null ||  localStorage.getItem("dismissedOnboardingNotification") === "false") {
`

jpmartinspt:
Fixed, I was testing on the page that sets the the property to `false` when it renders, so probably run into some race condition.
Added the null check now and fixed the linting.

albertkol (approved):
:+1:

PR #11424, opened 2022-03-30
Update to site restricted custom search

carkod:
## Done

Update all u.com search to site restricted search which avoids daily query limits.

## QA

- Check https://ubuntu-com-11424.demos.haus/tutorials?q=packer
- Check https://ubuntu-com-11424.demos.haus/search?q=ubuntu
- Check https://ubuntu-com-11424.demos.haus/server/docs/search?q=ubuntu
- Check https://ubuntu-com-11424.demos.haus/ceph/docs/search?q=ubuntu
- Check https://ubuntu-com-11424.demos.haus/core/docs/search?q=ubuntu
- Check https://ubuntu-com-11424.demos.haus/openstack/docs/search?q=ubuntu
- Check https://ubuntu-com-11424.demos.haus/security/livepatch/docs/search?q=ubuntu
- Check https://ubuntu-com-11424.demos.haus/security/certifications/docs/search?q=ubuntu

Make sure you find results

## Issue / Card

Fixes #

## Screenshots

## Help


webapp/views.py
```diff
@@ -289,7 +289,7 @@ def tutorials_index():
         """
 
         # The webteam's default custom search ID
-        search_engine_id = "009048213575199080868:i3zoqdwqk8o"
+        search_engine_id = "adb2397a224a1fe55"
```
nottrobin, line 292: Where can I see the search engine with this ID? I can't see it in https://cse.google.co.uk/cse/all

carkod: If you click on the specific search engine e.g. https://cse.google.co.uk/cse/setup/basic?cx=adb2397a224a1fe55 you will see it

carkod: I am thinking that maybe a reasonable way to separate them is by tribe?

nottrobin: It's fine to merge this now to get this live - is it ready?

But I don't think we want to split it by squad, we want to split it by what the sites are supposed to show search results for. We should discuss it.

anthonydillon:
Does this change need to be made for each site with search?

nottrobin:
I believe the existing search engine & id hasn't changed, right Carlos?

carkod:
> I believe the existing search engine & id hasn't changed, right Carlos?

No, for u.com the old search engine falls out of the 10 limit, so I had to create a new one.

anthonydillon:
When does the rate limit refresh for all the other sites?

carkod:
> When does the rate limit refresh for all the other sites?

from what I've read

> When does the rate limit refresh for all the other sites?

The dashboard says "Daily quotas reset at midnight Pacific Time (PT)."

nottrobin:
All sites used the same engine and engine id, which is presumably the problem. As in, they all fall under a single rate limit. They will have all broken and all been refreshed together.

Under this PR, ubuntu.com would change to a new search engine ID, but everything else would remain on the old one, so no other sites should be any more broken than they currently are.

I would suggest we merge this PR now which will split our search traffic at least two ways, avoiding the rate limit today.

Then we need to think of the proper strategy for splitting up our search into multiple engines based on solid principles.

nottrobin (approved):

carkod:
> > When does the rate limit refresh for all the other sites?
> 
> The dashboard says "Daily quotas reset at midnight Pacific Time (PT)."



> All sites used the same engine and engine id, which is presumably the problem. As in, they all fall under a single rate limit. They will have all broken and all been refreshed together.
> 
> Under this PR, ubuntu.com would change to a new search engine ID, but everything else would remain on the old one, so no other sites should be any more broken than they currently are.
> 
> I would suggest we merge this PR now which will split our search traffic at least two ways, avoiding the rate limit today.
> 
> Then we need to think of the proper strategy for splitting up our search into multiple engines based on solid principles.

I actually went ahead, and started to do some of this. I separated them by "Tribe"

PR #11488, opened 2022-04-19
make sure takeover text appears if image is not present

sowasred2012:
## Done

- Fixed an issue where the takeover text wouldn't become visible if an image isn't present

## QA

- Visit the [demo](https://ubuntu-com-11488.demos.haus/)
- Refresh the page until you see the Jammy Jellyfish takeover
- See that the text is visible


petesfrench (approved):
LGTM

PR #11496, opened 2022-04-19
Fix bug where takeovers would show an empty anchor element if not CTA data was present

sowasred2012:
## Done

Prevented an empty link appearing when a takeover doesn't have a primary CTA:

Screenshot from 2022-04-19 17-26-12

## QA

- Visit the [demo](https://ubuntu-com-11496.demos.haus/)
- Refresh the page until you see the Jammy takeover
- See that the empty green `a` element is not there


steverydz (approved):
LGTM 👍

PR #11540, opened 2022-04-25
Fix and schedule cypress test for forms

minkyngkm:
## Done

- Removed duplicated data test id that causes error on `/core/services` and `/robotics`
- Removed duplicated cypress test for `/robotics` page
- Scheduled cypress test for forms so it runs only one time a day

## Issue / Card

Fixes #


ClementChaumel (approved):
LGTM

PR #11552, opened 2022-04-27
Remove test backend flag

albertkol:
## Done

- In the previous PR we made `test_backend` to no longer switch environments. Now we remove all traces of this little bugger.

## QA

- Does cyress pass?


ClementChaumel (approved):
So long `test_backend` you will not be missed :wave:

PR #11688, opened 2022-06-07
Format dates in UTC

ClementChaumel:
## Done

- Always format UA dates in UTC

## QA

- Check that the dates are correct

## Issue / Card

Fixes canonical-web-and-design/commercial-squad#614

### DIFF
--- a/static/js/src/advantage/react/utils/formatDate.test.ts
+++ b/static/js/src/advantage/react/utils/formatDate.test.ts
@@ -6,6 +6,6 @@ describe("formatDate", () => {
   });
 
   it("returns the unformatted date if it can't be parsed", () => {
-    expect(formatDate("2021-07-09")).toBe("2021-07-09");
+    expect(formatDate("abcd")).toBe("abcd");
   });
 });


alnvdl (approved):
LGTM, I think this should do the trick. But you can achieve the same with just vanilla JS:
```js
function formatDate(date) {
    return (new Date(date)).toLocaleDateString("en-UK", {
        day: "numeric",
        month: "short",
        year: "numeric",
        timeZone: "UTC"
    });
}

formatDate("9999-12-31T00:00:00Z"); // should return "31 Dec 9999"
formatDate("9999-12-31T23:59:59Z"); // should return "31 Dec 9999"
```

Location would be hardcoded to UK, but our entire website is already hardcoded to English anyway, we probably want to be coherent in that aspect :)

Or if you want to keep using date-fns, you could use `intlFormat` instead (https://date-fns.org/v2.28.0/docs/intlFormat), which accepts a `timeZone` parameter (it's just a façade for what I did above with vanilla JS).

ClementChaumel:
> LGTM, I think this should do the trick. But you can achieve the same with just vanilla JS:

True but it would leave us open to a date that has been instantiated outside of the function and would then be using the user's timezone. I feel like explicitly casting it to UTC is a tiny bit safer!

PR #11695, opened 2022-06-09
Update the Third Party and Geolocation Policy

anthonydillon:
## Done
Update the Third Party and Geolocation Policy to the updated 2022 version.

## QA
- Check the new link works and the document has 2022 in it.


andesol (approved):
:+1:

PR #11710, opened 2022-06-13
Disable site restricted search

nottrobin:
It appears to be costing us a lot.

## QA

Go to the demo, run searches on each of the effected search pages.


carkod (approved):
LGTM now

PR #11775, opened 2022-06-27
Fix height issue

### LINKED ISSUE #10365: Subscription list hides selected subscription when the content height changes
Small unrelated issue: when you select different contracts, the height of the app jumps around to accommodate the content but sometimes it messes up with the scroll-into-view thing

scrooll

_Originally posted by @ClementChaumel in https://github.com/canonical-web-and-design/ubuntu.com/issues/10357#issuecomment-919122506_

abhigyanghosh30:
## Done

- Made the subscriptions card height constant so that the height does not change when changing types of subscriptions.

## QA

- Visit https://ubuntu-com-11775.demos.haus/advantage
- Buy different types of subscriptions
- Height of the column on the right should be the same irrespective of change in type of subscription.
## Issue / Card

Fixes #10365

### DIFF
--- a/static/js/src/advantage/react/components/Subscriptions/SubscriptionList/SubscriptionList.tsx
+++ b/static/js/src/advantage/react/components/Subscriptions/SubscriptionList/SubscriptionList.tsx
@@ -87,49 +87,47 @@ const SubscriptionList = ({ selectedId, onSetActive }: Props) => {
   );
 
   return (
-    <div className="p-subscriptions__list">
-      <div className="p-subscriptions__list-scroll">
-        {sortedUASubscriptions.length ? (
-          <ListGroup
-            data-test="ua-subscriptions-group"
-            title="Ubuntu Advantage"
-            marketplace={UserSubscriptionMarketplace.CanonicalUA}
-          >
-            {uaSubscriptions}
-          </ListGroup>
-        ) : null}
+    <div className="p-subscriptions__list p-card">
+      {sortedUASubscriptions.length ? (
+        <ListGroup
+          data-test="ua-subscriptions-group"
+          title="Ubuntu Advantage"
+          marketplace={UserSubscriptionMarketplace.CanonicalUA}
+        >
+          {uaSubscriptions}
+        </ListGroup>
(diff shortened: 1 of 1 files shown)


jpmartinspt:
I might have misunderstood the issue, but I think it still persists. @huwshimi do you mind having a look?

huwshimi:
> I might have misunderstood the issue, but I think it still persists. @huwshimi do you mind having a look?

I think so, it now happens if the data is over the new `min-height`, though I had to modify the content a bit to make it happen (at least with the data I could see in my account):

Screen Recording 2022-06-29 at 9 36 39 am

See in that gif that when I click on the free sub the content height changes and so the item I've just clicked on gets cut off.

The other issue now is that if the content is less than `68rem` there is a blank area at the bottom:

Screen Shot 2022-06-29 at 9 39 05 am

Fixing this might be tricky as it needs to watch for changes to the selected sub that result in content height changes and then scrolling the sub into view.

wgx:
@huwshimi @abhigyanghosh30 @jpmartinspt 

Would it be possible to have this page so that;

1. There's a min height to stop it looking weird, but
2. There's no max height so the "left panel" never scrolls - it's just simply a list of all subscriptions. 

This would mean users with *loads* would have a weird jump if they picked a sub lower down.. not sure how else to solve this though? 

Rationale: only a handful of users have *loads* of subscriptions, and if they do - they're probably managing them another way (through sales, through scripts that use `ua` on the CLI, etc). 

WDYT?

wgx:
This works for me :)

jpmartinspt:
This might need a git commit tidy up!

ClementChaumel:
Currently if you select a subscription far down below it will scroll back up to show you the details but you won't be able to see on the side which subscription it was that you selected
heightprob


If we set the sidebar to sticky then we don't need to scroll back up when switching subscription and it'll be automatically in view
heightfix

To achieve that you'll need to set the position to sticky for the column and unset the overflow values form its parents
image

ClementChaumel (approved):
LGTM!

PR #11803, opened 2022-07-04
ubuntu.com: Sentry schism

albertkol:
## Done

- Send /advantage and /account to separate sentry dashboards
- Send /security to separate sentry dashboard too
  - /security rule updated because single CVE pages urls structure is /security/<CVE_id> which right now would not be sent to the security service


mtruj013 (approved):
👍🏻

PR #11830, opened 2022-07-13
Use release name instead of 'slug'

anthonywong:
'slug' seems to be an internal ID and not something to be used on the web page.

Signed-off-by: Anthony Wong <anthony.wong@canonical.com>


anthonydillon (approved):
Perfect, thank you. 👍

PR #11877, opened 2022-07-25
Expand to 2 gunicorn workers

nottrobin:
Fixes https://github.com/canonical/Web-design-systems-squad/issues/56 :

> This should help with stability of ubuntu.com. We do need to run more than one worker, as even with gevent multithreading, the worker still sometimes restarts and while it's restarting, requests can't be handled at all.
> 
> Should help with https://portal.admin.canonical.com/C136508/.

## QA

`dotrun`. Check the site runs.

### DIFF
--- a/entrypoint
+++ b/entrypoint
@@ -2,10 +2,14 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


albertkol (approved):
:+1:

PR #11878, opened 2022-07-25
Fix the global navigation click

anthonydillon:
## Done
Fix the global navigation click

## QA
Click products in the global nav and see it works.


minkyngkm (approved):
LGTM👍

PR #11904, opened 2022-07-28
Blog - 500 when article has no group

jpmartinspt:
## Done
Check if the article has a group set before checking redirect condition

## QA

- Find a blog article with no group set
- View the site locally in your web browser at: http://0.0.0.0:8001/blog/<atricle_slug>
- Make sure it does not 500

## Issues
Fixes https://github.com/canonical-web-and-design/ubuntu.com/issues/11908#event-7087743195

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -623,7 +623,8 @@ class BlogRedirects(BlogView):
         )
 
         # Redirect canonical annoucements
-        if article["group"]["id"] == 2100:
+        group = article.get("group")
+        if isinstance(group, dict) and group["id"] == 2100:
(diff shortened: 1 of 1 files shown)


jpmartinspt:
Some blogs pages were 500, because the articles did not have a group. This PR should fix the issue, but I realised while fixing this, that some articles have more than one group and the way the blog module transforms articles we end up disregarding some of them since the transformer only picks the first group of the array, this means that at the point we do the redirect check some group information might have been lost and we can have articles that we wanted to be redirected not being so because we lost group information while transforming the article. Hope it makes sense, if it does not you know where to find me!

carkod (approved):

carkod:
Thanks!

PR #11912, opened 2022-07-29
fix server section of certified home page

### LINKED ISSUE #11907: ubuntu.com/certified home page server block is missing Dell and has incorrect counts for HPE and Lenovo
## Summary

The listing of Vendors in the server block does not list at least the top three server vendors.  In addition it shows incorrect numbers.

First, it only shows HPE and Lenovo. It should, at the very least, also be showing Dell. I am pretty sure Dell used to be there, and has disappeared.

Additionally, all the vendor listings show an incorrect number of certificates for Server (I did not sanity check anything else)
Lenovo is listed as "Lenovo (219)" but if you click on Lenovo there are only 79 certificates in the results.
HPE is listed as "HPE (90)" but if you click through, there are only 58 results.

Doing some digging on the back end and I suspect those counts include certificates marked "Private" and they should NOT.  Those numbers should only reflect the number of public certificates (i.e. the ones that appear when you click through to a particular vendor search such as 

https://ubuntu.com/certified/servers?vendor=Lenovo

I realize that last is probably an API issue so I'll point skay at this too.

## Process

Go to https://ubuntu.com/certified and look at the Servers section.
Note that Dell is no longer listed
Also note that Lenovo shows 219 certificates and HPE shows 90, both of which are not correct.

Current: Server block only only shows HPE and Lenovo
Expected: Server block at least shows Dell, HPE and Lenovo and if we must show counts, the counts reflect public certificates only and do not include private ones in the count.

## Screenshot

Screenshot from 2022-07-28 21-32-51

codersquid:
The certified home page server section was displaying
incorrect numbers and was missing the Dell Technologies
vendor.

## Done

This change updates the home view to make the logic that builds up the server section match the logic that builds up all of the other form factor types. It also updates the template to use the same pattern as well as changes the match string to _Dell Technologies_ since _Dell_ is not used.

This updates the servers section of the home page so that it includes Dell and also fixes the counts.

## QA

View https://ubuntu-com-11912.demos.haus/certified  and verify that a link for Dell Technologies shows up in the server section of the page with a count of 89.

## Issue / Card

Fixes #11907

### DIFF
--- a/templates/certified/index.html
+++ b/templates/certified/index.html
@@ -125,9 +125,9 @@
       <p class="u-text--muted u-no-margin--bottom">Vendor:</p>
       <ul class="p-inline-list--midline" style="margin-bottom: 0.5rem;">
         {% for server_vendor in server_vendors %}
-          {% if server_vendor.vendor in ['Dell', 'Lenovo', 'HPE'] %}
+          {% if server_vendor.make in ['Dell Technologies', 'Lenovo', 'HPE'] %}
             <li class="p-inline-list__item">
-              <a href="/certified/servers?vendor={{ server_vendor.vendor }}">{{ server_vendor.vendor }}&nbsp;({{ server_vendor.total }})</a>
+              <a href="/certified/servers?vendor={{ server_vendor.make }}">{{ server_vendor.make }}&nbsp;({{ server_vendor.total }})</a>
             </li>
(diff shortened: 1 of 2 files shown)


mtruj013 (approved):
LGTM!

PR #11915, opened 2022-08-01
Remove mistune

carkod:
## Done

Looking [at this old changes](https://github.com/canonical-web-and-design/ubuntu.com/commit/68efa93bf5c52f32ac793ec80ef3e68a5b6ff738) it seems that we no longer need mistune? At least not pinned? As we already removed alembic

## QA

- View the demo https://ubuntu-com-11915.demos.haus/ and check any /security pages that could be broken with this removal.

## Help

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,6 @@ Flask-WTF==0.15.1
 requests==2.27.0
 sqlalchemy==1.3.23
 marshmallow==3.14.1
-mistune==0.8.4
 psycopg2-binary==2.9.3
 pyyaml==5.4.1
 ubuntu-release-info==21.1


anthonydillon:
Ah that is why I couldn't find it with `npm ls mistune`.

nottrobin:
I don't think mistune has anything to do with alembic. If you look at that commit you link to, you can see alembic was already in dependencies before that. I'll do a bit more digging...

carkod:
So I think this may need to be updated with `template_finder` then. Bartek raised [an issue](https://github.com/canonical/canonicalwebteam.templatefinder/issues)

nottrobin:
yeah I think we should close this. Mistune is used [directly in ubuntu.com](https://github.com/canonical/ubuntu.com/blob/ac3a3ef95dc9af50e5a7128150de026a1ee11bd8/webapp/security/views.py#L12) as well as being used in templatefinder. So let's leave this as it is until templatefinder is update, which might be some time.

Issue #11584, opened 2022-05-06
Alternate Arch download link is incorrect on certificates.


bladernr:
While looking up some stuff for a different issue I filed, I discovered that download links for PowerPC and s390x systems point to https://ubuntu.com/download/server which is incorrect... ARM certificates have a download button that points correctly to their arch specific landing page...

s390x: https://ubuntu.com/certified/201909-27363
ppc64le: https://ubuntu.com/certified/201805-26248

Those are examples and in each case they point to the AMD64 server landing page.

Each of those, like ARM, should point to the appropriate arch specific pages instead:

s390x: https://ubuntu.com/download/server/s390x
ppc64le: https://ubuntu.com/download/server/power

nottrobin:
The model details pages like are generic, and at present they simply use the "category" to display the download link:

https://github.com/canonical-web-and-design/ubuntu.com/blob/f822b9ee491539a9b5d9a7548ab2a7ebf66762ca/templates/certified/model-details.html#L88-L101

Is the solution here to add categories to the certifications API for "s390x" and "power" (or "ppc64le") and extend our template to provide those download links too? The other option, I suppose, is to ask the certifications API to always provide a download URL with these items?

@codersquid do you have any input on this?

codersquid:
Category is different than architecture, and we don't use architecture for the categories. We don't store download urls for certificates, I guess that's something you could handle in the template by checking for the architecture or something we could hardcode on our side. Since you already have the logic to construct the download links, would you consider keeping it on your side and checking for the arch for server pages? The logic could live in the views instead of the template. I think there used to be some utility code that did that, but it's gone now.

carkod:
> Category is different than architecture, and we don't use architecture for the categories. We don't store download urls for certificates, I guess that's something you could handle in the template by checking for the architecture or something we could hardcode on our side. Since you already have the logic to construct the download links, would you consider keeping it on your side and checking for the arch for server pages? The logic could live in the views instead of the template. I think there used to be some utility code that did that, but it's gone now.

I'll create a PR with that utility code. I'll let you guys know when it's ready for review

PR #11920, opened 2022-08-02
Fix certified download links

carkod:
## Done

Restore the utility function to create the download links

## QA

Check that the download button of these two pages:
s390x: https://ubuntu-com-11920.demos.haus/certified/201909-27363
ppc64le: https://ubuntu-com-11920.demos.haus/certified/201805-26248

Go to these
s390x: https://ubuntu.com/download/server/s390x
ppc64le: https://ubuntu.com/download/server/power

## Issue / Card

Fixes https://github.com/canonical/ubuntu.com/issues/11584

## Help

### DIFF
--- a/webapp/certified/helpers.py
+++ b/webapp/certified/helpers.py
@@ -1,0 +1,35 @@
+def get_download_url(model_details):
+    """
+    Return the appropriate ubuntu models.com/download url for the model
+    :param model: a certifiedmodel resource
+    :param model_details: a certifiedmodeldetails resource
+    :return: appropriate ubuntu.com/download url
+    """
+    platform_category = model_details.get("category", "").lower()
+    architecture = model_details.get("architecture", "").lower()
+
+    if model_details.get("level") == "Enabled":
+        # Enabled systems use oem images without download links.
+        return
+
+    if platform_category in ["desktop", "laptop"]:
+        return "https://ubuntu.com/download/desktop"
+
+    if "core" in platform_category:
+        return "https://ubuntu.com/download/iot/"
+
+    if "server" in platform_category:
+        # Server platforms have special landing pages based on architecture.
+        arch = ""
+        if "arm" in architecture:
+            arch = "arm"
+        elif "ppc" in architecture:
+            arch = "power"
+        elif "s390x" in architecture:
+            arch = "s390x"
+        else:
+            return "https://ubuntu.com/download/server/"
+
+        return f"https://ubuntu.com/download/server/{arch}"
+
+    return "https://ubuntu.com/download"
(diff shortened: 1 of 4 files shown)


webapp/certified/views.py
```diff
@@ -125,6 +127,8 @@ def certified_model_details(canonical_id):
         "objects"
     ]
 
+    models = api.certified_models(canonical_id=canonical_id)["objects"]
+
```
codersquid, line 131: I would move this above the  component summaries call and add a 404 check. It is unlikely to be an empty list since there was a result on certified model details, but technically it's possible.

If it does 404, then you save an extra step of calling the component summaries api call when you didn't need to.

webapp/certified/helpers.py
```diff
@@ -0,0 +1,32 @@
...
+    if "server" in platform_category:
+        # Server platforms have special landing pages based on architecture.
+        arch = ""
+        if "arm" in architecture:
+            arch = "arm"
+        elif "ppc" in architecture:
+            arch = "power"
+        elif "s390x" in architecture:
+            arch = "s390x"
+        return f"https://ubuntu.com/download/server/{arch}"
```
codersquid, line 30: you should have a fallthrough case where you just return https://ubuntu.com/download/server.

codersquid (changes requested):

webapp/certified/views.py
carkod: Done, thanks

webapp/certified/helpers.py
carkod: Done.

webapp/certified/helpers.py
```diff
@@ -0,0 +1,35 @@
+def get_download_url(model, model_details):
+    """
+    Return the appropriate ubuntu models.com/download url for the model
+    :param model: a certifiedmodel resource
+    :param model_details: a certifiedmodeldetails resource
+    :return: appropriate ubuntu.com/download url
+    """
+    platform_category = model.get("category", "").lower()
```
codersquid, line 8: I missed this at first. You can get category from model_details and don't need to pass model in.

webapp/certified/views.py
```diff
@@ -114,6 +116,10 @@ def certified_hardware_details(canonical_id, release):
 
 def certified_model_details(canonical_id):
 
+    models = api.certified_models(canonical_id=canonical_id)["objects"]
```
codersquid, line 119: I missed this at first. You don't actually need to call the certifiedmodels api. The `category` info is available from certifiedmodeldetails now. A while back I saw that it should be in there so that people wouldn't need to make two api calls to get category information.

codersquid (changes requested):
Sorry for missing this the first round of review. You can remove the api call you added. It's no longer needed to get `category` information.

codersquid (approved):
lgtm

PR #11928, opened 2022-08-04
[Copy Update] Download Intel IoT G Contact Us form

ClementChaumel:
## Done

- Updated the copy according to the [copydoc](https://docs.google.com/document/d/12HsccVGTbjOxRX0go2m7ywFFoqrtU2JIizwIFtVWnWs/edit?disco=AAAAcF8vgUE&usp=comment_email_document&ts=62e97c6d&usp_dm=false)

## QA

- go on [/download/iot/intel-iotg](https://ubuntu-com-11928.demos.haus/download/iot/intel-iotg)
- click "contact us" 
- check that the form matches the copydoc
- try to progress to the last page of the form without filling in any details
- you should see a validation message on the "Tell us more about your project" field
- fill out the field and try to progress again
- you should be able to progress

## Issue / Card

Fixes canonical/commercial-squad#687

### DIFF
--- a/static/js/src/dynamic-forms.js
+++ b/static/js/src/dynamic-forms.js
@@ -244,15 +244,25 @@
               window.location.pathname
             );
           } else {
-            index = index + 1;
-            setState(index);
-            ga(
-              "send",
-              "event",
-              "interactive-forms",
-              "goto:" + index,
-              window.location.pathname
-            );
+            var valid = true;
+
+            if (button.classList.contains("js-validate-form")) {
+              var form = button.closest("form");
+
+              valid = validateForm(form);
+            }
+
+            if (valid) {
+              index = index + 1;
+              setState(index);
+              ga(
+                "send",
+                "event",
+                "interactive-forms",
+                "goto:" + index,
+                window.location.pathname
+              );
+            }
           }
         });
       });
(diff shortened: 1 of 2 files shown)


ClementChaumel:
@sowasred2012 I added a `required` attribute on the "tell us more about your project" textarea but it doesn't seem to work. Do you know how I can specify that a field is required before pagination?

leiaru-c (commented):
This looks good, it does tell you that "The tell me more about your project" field is needed to progress onto the next page. Small detail though, the style used for the "Tell us more about your project" is not the same weight as other questions. It should follow the same style choices to make it clear it's a different section from the previous question. Is it also possible to make it explicit that only the "Tell us more" field is required (thinking of an asterisk or something along those lines...)?

leiaru-c (approved):

static/js/src/dynamic-forms.js
```diff
@@ -273,6 +283,49 @@
...
+              inputValid = true;
+            }
+
+            if (input.type === "radio" && input.checked) {
+              inputValid = true;
+            }
+
+            if (input.type === "text" && input.value) {
+              inputValid = true;
+            }
```
minkyngkm, line 310: It seems only the textarea is required in this form but is this for future use? just wondering!

minkyngkm (commented):
Looking good! but the label(question) "Tell us more about your project" is not printed in the `comments_from_lead__c` 

This is the payload. 
image

Please disregard if it's not necessary! :)

static/js/src/dynamic-forms.js
sowasred2012: ah yes - yeah, this function should make it possible for us to add required fields to other forms in future.

sowasred2012:
> Looking good! but the label(question) "Tell us more about your project" is not printed in the `comments_from_lead__c`
> 
> This is the payload. image
> 
> Please disregard if it's not necessary! :)

Good point! I'll update this morning

minkyngkm:
Screenshot 2022-08-09 at 11 54 35 am

Thanks @sowasred2012 I see all the the labels and values now but this two things in the red box seems to be the values of the other two questions below and these values are printed twice. Should it be fixed or do you think it's okay?

minkyngkm (approved):
LGTM - Thanks @sowasred2012 🏅 🙌

Issue #11979, opened 2022-08-30
Add Cloudinary to the blog images


anthonydillon:
This Cloudinary part of the image links has been removed at some point. We should add them back as the thumbnails are adding up to ~10MB on the blog listing page.

PR #11980, opened 2022-08-30
Add Cloudinary to the blog images

anthonydillon:
## Done
- Used the blog modules Cloudinary image to render the blog image
- Added some medium screen grid classes to improve that view
- Increased the thumbnail size to accommodate the size of the image on small screens

## QA
- Open the demo
- Go to /blog
- Check all viewports render well
- Check the Network tab for images is 2MB instead of >10MB in production

## Issue / Card

Fixes https://github.com/canonical/ubuntu.com/issues/11979

### Large
image

### Medium
image

### Small
image

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -522,7 +522,7 @@ app.add_url_rule(
 # blog section
 
 blog_views = BlogViews(
-    api=BlogAPI(session=session),
+    api=BlogAPI(session=session, thumbnail_width=555, thumbnail_height=311),
     excluded_tags=[3184, 3265, 3408, 3960],
     per_page=11,
     blog_title="Ubuntu blog",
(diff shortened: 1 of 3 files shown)


webapp/app.py
```diff
@@ -522,7 +522,7 @@ def utility_processor():
 # blog section
 
 blog_views = BlogViews(
-    api=BlogAPI(session=session),
+    api=BlogAPI(session=session, thumbnail_width=555, thumbnail_height=311),
```
sparkiegeek, line 525: where do these numbers come from? in my limited testing I only see images at width 333

anthonydillon: It's from the breakpoint of 620 which is when the UI switches to stacked which results in the image rendering at 555px.

image

carkod (approved):
LGTM

PR #12093, opened 2022-10-03
Get hidden listings fix

albertkol:
## Done

- Loading /advantage when listings are hidden throws an error. This fixes it.

## QA

- Go to /advantage when having already only products bought.
- Does it work?


abhigyanghosh30 (approved):
/advantage loads without erros

PR #12105, opened 2022-10-04
fix scss lints

sowasred2012:
## Done

- Fixed SCSS issues raised by Prettier

## QA

- The checks should all pass


ClementChaumel (approved):
TYSM but I'm just about to create more issues! :smiling_imp:

PR #12122, opened 2022-10-05
Hide support portal for trials

ClementChaumel:
## Done

Hide support portal button for free trials


albertkol (approved):
:+1: Test needs fixing. Rest is impeccable.

PR #12123, opened 2022-10-05
update pro subnav to read "Pro"

sowasred2012:
## Done

- previously the subnav heading on /pro read "Support", now it reads "Pro"

## QA

- Go to /pro
- See "Pro" in the subnav
- Yay!


wgx (approved):
looks good

PR #12125, opened 2022-10-05
fix blurry logos; fix author and h2 spacing

lyubomir-popov:
## Done

- fix blurry logos; 
- fix testimonial author column count 
- fix spacing of links inside h2s (links were outside with an h2 class applied to them, which inserted extra space we don't need)

## QA

- check the above looks right


sowasred2012 (approved):
LGTM

PR #12130, opened 2022-10-05
Update pricing table

sowasred2012:
## Done

- switched the columns in the first table of /pricing/pro to better align with the tables below
- updated the "Purchase a subscription" link to point to /pro/subscribe

## QA

- visit https://ubuntu-com-12130.demos.haus/pricing/pro
- see that the "UBUNTU PRO (INFRA-ONLY)" column appears before "UBUNTU PRO" in the first table, and that the values in each column are correct
- click "Purchase a subscription" and see that you're taken to /pro/subscribe


lyubomir-popov (approved):
LGTM

PR #12153, opened 2022-10-11
Add JetBrains .idea folder to .gitignore

yurii-vasyliev:
## Done

- Added .idea folder to .gitignore, because we don't track local env config

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ Thumbs.db
 .LSOverride
 .spelling
 .vscode
+.idea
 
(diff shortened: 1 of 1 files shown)


nottrobin:
I've never heard of a `.idea` folder. What is it?

nottrobin:
Found it: https://rider-support.jetbrains.com/hc/en-us/articles/207097529-What-is-the-idea-folder-

nottrobin (approved):

PR #12214, opened 2022-10-20
22 10 checksums

sowasred2012:
## Done

- Updated releases.yaml with 22.10 checksums

## QA

- Check the checksums match the ones given in the original issue (several of the RISC-V checksums aren't included, we've been asked not to publish the 22.10 update for that page until the 25th, we're checking in with Felicia later today to clarify why that is)
- Visit /download/desktop and /download/server, find the 22.10 download CTAs, click them. They won't work yet because the images aren't live.

## Issue / Card

Fixes https://github.com/canonical/ubuntu.com/issues/12212

### DIFF
--- a/releases.yaml
+++ b/releases.yaml
@@ -31,27 +31,33 @@ openstack_lts:
 
 checksums:
   desktop:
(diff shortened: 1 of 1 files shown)


minkyngkm (approved):
LGTM👍

Just wondering this image can be hidden in medium/small screens  on [/download/desktop](https://ubuntu-com-12214.demos.haus/download/desktop) as the image is not aligned in the center.
image

PR #12248, opened 2022-11-02
Require SSO login for /pro beta signup form

sowasred2012:
## Done

- Added a check to see if the user is logged in on /pro. If they're logged in, they can trigger the contact modal and the fields will be pre-populated with their SSO info (and the email field will be disabled). If they're not logged in, they're directed to SSO, and then returned to the page with the modal already open.

## QA

- Visit https://ubuntu-com-12248.demos.haus/pro
- Make sure you're logged out of SSO
- Find the "Login to request beta access" CTA and click it
- You should be taken to SSO and asked to login
- Login
- When you've successfully logged in, you should be returned to /pro with the contact modal already populated (including a disabled email address field)
- Visit /pro again (without the #get-in-touch hash)
- See that the "Login to request beta access" CTA has been updated to read "Request beta access at no extra cost", and triggers the modal when clicked

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -957,6 +957,7 @@ security/cve/sitemap(?P<suffix>.*).xml: /security/cves/sitemap{suffix}.xml
 
 /advantage: "/pro"
 /advantage/(?P<path>.*): "/pro/{path}"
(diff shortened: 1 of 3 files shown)


anthonydillon (approved):
Code and QA :+1:

PR #12251, opened 2022-11-03
Add remaining meganav sections

sowasred2012:
## Done

- Added Products, Use Case, Community and Support sections to the meganav

## QA

- Visit https://ubuntu-com-12251.demos.haus/
- Open all of the dropdowns in the mega nav
- See that they work as expected:
  - Dropdowns should animate open and closed (they will switch instantly when navigating from one dropdown to another)
  - Some sections contain a side navigation, which should work
- Check the mobile view, see that it works

## Issue / Card

Fixes https://github.com/canonical/web-squad/issues/5983, https://github.com/canonical/web-squad/issues/5985, https://github.com/canonical/web-squad/issues/5986, https://github.com/canonical/web-squad/issues/5987

### DIFF
--- a/meganav.yaml
+++ b/meganav.yaml
@@ -438,7 +438,7 @@ use_case:
           description: Performant open source
           url: /managed/apps
         - title: Managed infrastructure
-          description: Infrastructre support
+          description: Infrastructure support
           url: /pricing/infra
         - title: Observability
           description: Monitoring tools
--- a/meganav.yaml
+++ b/meganav.yaml
@@ -36,6 +36,7 @@
 #             - title:
 #               url:
 #       section_footer: (optional)
+#         copy:
 #         cta_title:
 #         cta_url:
 
@@ -83,6 +84,10 @@ products:
             - title: Ubuntu Certified Hardware
               description: Machines you can trust
               url: /certified
+      section_footer:
+        copy: Not sure where to start, need some guidance or just ready to learn more? We're here to help you.
+        cta_title: Contact us
+        cta_url: /contact-us
 
     - title: Ubuntu OS
       primary_links:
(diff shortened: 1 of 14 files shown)


fitojb:
Blue links on a dark background doesn’t seem very readable to me…

sowasred2012:
@fitojb we took that into consideration when designing this, the blue we typically use in Vanilla was updated to make this meet WCAG level AA standards:

image

fitojb:
@sowasred2012 The color contrast shouldn’t be the only criterion to consider when designing readable text. WCAG doesn’t consider the font weight, and it is _anæmic_. Using regular instead of light would help.

sowasred2012:
> @sowasred2012 The color contrast shouldn’t be the only criterion to consider when designing readable text. WCAG doesn’t consider the font weight, and it is _anæmic_. Using regular instead of light would help.

I didn't say it was, I was addressing the original comment. Your point about font weight is valid though, we'll take that into consideration.

carkod:
It's not a nuance for me, but I've noticed that if you have both "All canonical" opened and one of the navs on the left opened, you need to click twice to close the one you clicked. Here is a video:
[Screencast from 29-11-22 10:29:27.webm](https://user-images.githubusercontent.com/14939793/204505908-8f5edebc-77ac-4163-b71e-4d9ab2d1ca8f.webm)

carkod (approved):
Looking good

sowasred2012:
> It's not a nuance for me, but I've noticed that if you have both "All canonical" opened and one of the navs on the left opened, you need to click twice to close the one you clicked.

Ah good point - I've got an issue to fix remaining bugs on this before merging the feature branch, I'll add this to that.

PR #12281, opened 2022-11-15
Ubuntu Pro Upgrade form - Duplicated hidden email field

caldav:
## Issue
On the https://ubuntu.com/pro#get-in-touch form, the email field is disabled to avoid modification. A side effect is the creation of leads without email addresses as disabled form fields are not sent on form submission.

## Done

- Added a hidden duplicate email field when the user email is available from SSO.

## QA

- When submitting the /ubuntu.com/pro#get-in-touch form after SSO login, ensure the form submission data contains the email address.


minkyngkm (approved):
LGTM! 

image

PR #12283, opened 2022-11-15
Display Card form when there is no saved card 

minkyngkm:
## Done
Display Card form when there is no saved card 

The view when there's no saved card

- Before fix 
image

- After fix

image

The view when there's saved card  
image

## QA

- Create an UO account and login
- Go to /pro/subscribe
- Check a form is displaying when there's no card saved in the payment method field.
- Also check saved card info is displaying when saved card exists. 

## Issue / Card

Fixes #https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/801?modal=detail&selectedIssue=WD-394


minkyngkm:
Question: Billing address: City, or Postal code also need to be a form when there's no saved value?
image

abhigyanghosh30:
When I enter something which exceeds the card field length, it resets it to empty
I think the default behaviour should be to ignore the extra characters

static/js/src/PurchaseModal/components/UserInfoForm/UserInfoForm.tsx
```diff
@@ -69,10 +69,72 @@ const UserInfoForm = ({ setCardValid, isGuest }: Props) => {
     return errorMessage;
   };
 
+  const { data: userInfo } = useStripeCustomerInfo();
+
+  const defaultPaymentMethod = userInfo?.customerInfo?.defaultPaymentMethod;
+
   const displayMode = (
```
jpmartinspt, line 76: I think the solution here is too change the condition that decides if we are on `displayMode` or `editMode` rather than changing the `displayMode` component to also allow edits.

I think doing something like the following around [line 38](https://github.com/canonical/ubuntu.com/pull/12283/files#diff-ccac169c2330178b4a0ae8fbe0d3eb832a79869af42630925e71d651d2ae2786R38) should be enough to solve this issue:
```js
  const { data: userInfo } = useStripeCustomerInfo();
  const defaultPaymentMethod = userInfo?.customerInfo?.defaultPaymentMethod;
  const [isEditing, setIsEditing] = useState(isGuest || !defaultPaymentMethod);
```

minkyngkm: Ah, thanks @jpmartinspt! I thought we wanted to update displayMode. Fixed! 👍

jpmartinspt (approved):
Seems to work but there seem to be a few of the checks failing. You can merge once the checks are passing(cypress can be red for now)

PR #12300, opened 2022-11-21
Ignores free contracts outside of personal accounts

alnvdl-work:
## Done

We had a bug in the backend between Nov 2020 and May 2021 that resulted in the creation of free contracts in some paid accounts. That is wrong, and we are cleaning this up in the backend already.

However, in order to be able to direct customers to get their correct free personal token, we need to make sure `/pro/dashboard` does not display these incorrect contracts.

As a general measure, this is a good idea regardless of this bug. As a kind of necessary pre-condition, this PR makes the account type available to the frontend code in general. The backend has had this attribute exposed for a long time now.

## QA

## Issue / Card

Fixes CSS-2568.


albertkol (approved):
:+1: Looking good.

PR #12312, opened 2022-11-24
Fix test-js & lint-ts

minkyngkm:
## Done

- Fix code so `test-js` & `lint-ts` pass.

## QA

- `dotrun lint-ts`
- `dotrun test-js`

## Issue / Card

Fixes https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/801?modal=detail&selectedIssue=WD-464

### DIFF
--- a/static/js/src/PurchaseModal/PurchaseModal.tsx
+++ b/static/js/src/PurchaseModal/PurchaseModal.tsx
@@ -229,6 +229,7 @@ const PurchaseModal = ({
                 product={product}
                 quantity={quantity}
                 action={action}
+                isCardValid={isCardValid}
               ></BuyButton>
             </ModalFooter>
           </>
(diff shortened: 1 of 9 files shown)


minkyngkm:
Hi @albertkol ! line 234 this test "/pro/subscribe logged in purchase" keeps failing. but ` cy.login();` is not working at all in my machine. Could you please take a look at this test by any chance?

static/js/src/PurchaseModal/PurchaseModal.tsx
```diff
@@ -229,6 +229,7 @@ const PurchaseModal = ({
                 product={product}
                 quantity={quantity}
                 action={action}
+                isCardValid={isCardValid}
```
albertkol, line 232: The cypress test seems to be failing due the buy button being disabled despite a valid card being added. 

For as much as I understand. `isCardValid` is set to `false` and I cannot see that status changed. And a result the button never gets enabled.

albertkol: `isCardValid` has to check to see if the user has a card already set as part of his customer info and if it exists then have it initialised to true.

At the moment `isCardValid` status get's changed from its default `false` status only if the user edits his details.

albertkol (changes requested):

albertkol (approved):
:+1:

PR #12350, opened 2022-12-07
Instant provisioning for TA exams

mrgnr:
## Done

- Implement instant provisioning for TA exams
  - On `GET /credentials/provision`, the session is checked for a reservation UUID.
    - If reservation UUID does not exist, check the API for a reservation and store its UUID if found.
    - If reservation UUID exists, the reservation and assessment are fetched and the status is displayed to the user.
  - On `POST /credentials/provision`, if there is no known reservation:
    - An assessment is scheduled for 70 seconds in the future and the reservation UUID is stored as part of the session.
  - The page automatically refreshes every 30 seconds to display the provisioning status.
- Determine country code based on user's time zone

## QA

- Go to https://ubuntu-com-12350.demos.haus/credentials/provision
  - Click the Start button
  - Wait for provisioning to finish
  - Click the Take Exam button and start the exam

## Screenshots

cred-provision-start

cred-provision-wait

cred-provision

cred-provision-take

## Help

### DIFF
--- a/templates/credentials/assessments.html
+++ b/templates/credentials/assessments.html
@@ -17,22 +17,20 @@
       <thead>
         <tr>
           <th>Exam</th>
-          <th>Date</th>
+          <th>Status</th>
           <th>Time</th>
           <th>Timezone</th>
-          <th>Status</th>
-          <th>UUID</th>
+          <th>Email</th>
         </tr>
       </thead>
       <tbody>
         {% for exam in exams %}
         <tr>
-          <td><a href="/credentials/exam?id={{ exam['id'] }}">{{ exam["name"] }}</a></td>
-          <td>{{ exam["date"] }}</td>
-          <td>{{ exam["time"] }}</td>
-          <td>{{ exam["timezone"] }}</td>
+          <td><a href="/credentials/exam?id={{ exam['id'] }}">{{ exam["name"] }}</a><br />{{ exam["id"] }}</td>
           <td>{{ exam["state"] }}</td>
-          <td>{{ exam["uuid"] }}</td>
+          <td>{{ exam["date"] }}<br />{{ exam["time"] }}</td>
+          <td>{{ exam["timezone"] }}</td>
+          <td>{{ exam["user_email"] }}</td>
         </tr>
         {% endfor %}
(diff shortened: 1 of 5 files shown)


linux-in-flux:
We need to deemphasize the "up to 20 minutes" thing, as that shouldn't be true outside of the sandbox and publicly accessible. That should only be potentially true for a handful of Canonical people while we have limited provisioning.  I would suggest:

1. Smaller font (potential info, not equal to "Click start" or "Your exam...")
2. Rephrase: from Note... to "During this limited testing cycle, exams could potentially take up to 20 minutes under heavy system load."

We will also need to remove this info when we push TrueAbility to their prod environment.

mrgnr:
@linux-in-flux Here are a couple options...

Small text for 20 minute warning:
cred-provision-provisioning-small

Info box for 20 minute warning:
cred-provision-provisioning-info

IMO the small text doesn't do much to de-emphasize. The info box is more prominent, but I think helps make clear that it's a temporary issue.

For the sandbox environment, having to wait for a full provisioning will probably be the norm for testers, since we only have single-machine pools in two regions.

linux-in-flux (approved):
I like the second option with the info box.

jpmartinspt (approved):
Looks good. Do you think it would be worth making that page refresh a partial refresh of the page only, in a different PR?

jpmartinspt:
Actually, this seems to work on my local environment, but on the demo it went straight into telling me `Your exam environment has been scheduled and is awaiting provisioning. ` without me starting or pressing anything.

mrgnr:
@jpmartinspt 

> Looks good. Do you think it would be worth making that page refresh a partial refresh of the page only, in a different PR?

Agree, we should probably reactify the page and just fetch data as needed instead of doing a full refresh.

> Actually, this seems to work on my local environment, but on the demo it went straight into telling me Your exam environment has been scheduled and is awaiting provisioning. without me starting or pressing anything.

If you started provisioning already, that should be shown whether you are on localhost or the demo server (we check the TrueAbility API for reservations and don't allow another exam to be scheduled if one is found).

jpmartinspt (approved):

PR #12356, opened 2022-12-09
Aligning text table with SVG rendering of same, dropped 10.04/12.04

desrod:
The current template misses the 22.10 release and still includes 10.04 and 12.04. This PR aligns this to match what is visibly rendered in the SVG on the release page (https://ubuntu.com/about/release-cycle). Ideally this should be auto-generated, but currently is not.


petesfrench (approved):
Thank your work on this, LGTM.

PR #12366, opened 2022-12-13
Call calculate endpoint on step 1 only for guest users

minkyngkm:
## Done

- Call calculate endpoint on step 1 only for guest users

## QA

- Go to /pro/subscribe
- QA when logged in and when guest
- Change country and VAT and click "Save" button 
- Check calculate endpoint is called only when guest 
- Also check tax changes in summary properly

## Issue / Card

Fixes #https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/801?modal=detail&selectedIssue=WD-911

[WD-911]: https://warthogs.atlassian.net/browse/WD-911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/js/src/PurchaseModal/components/Taxes/Taxes.tsx
+++ b/static/js/src/PurchaseModal/components/Taxes/Taxes.tsx
@@ -61,7 +61,7 @@ const Taxes = ({ product, quantity }: TaxesProps) => {
 
   const onSaveClick = () => {
     setIsEditing(false);
-    if (isGuest) {
+    if (isGuest || !window.accountId) {
       taxMutation.mutate();
     } else {
       genericPurchaseMutation.mutate(
--- a/static/js/src/PurchaseModal/components/Taxes/Taxes.tsx
+++ b/static/js/src/PurchaseModal/components/Taxes/Taxes.tsx
@@ -7,7 +7,7 @@ import {
   Input,
   ActionButton,
 } from "@canonical/react-components";
-import { FormValues, getLabel } from "advantage/subscribe/react/utils/utils";
+import { getLabel } from "advantage/subscribe/react/utils/utils";
 import {
   caProvinces,
   countries,
@@ -16,6 +16,8 @@ import {
 } from "advantage/countries-and-states";
 import useCalculateTaxes from "PurchaseModal/hooks/useCalculateTaxes";
 import useStripeCustomerInfo from "PurchaseModal/hooks/useStripeCustomerInfo";
+import useMakePurchase from "PurchaseModal/hooks/useMakePurchase";
+import { FormValues } from "../../utils/utils";
 
 type TaxesProps = {
   product: any;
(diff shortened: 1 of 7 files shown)


static/js/src/PurchaseModal/components/Taxes/Taxes.tsx
```diff
@@ -48,10 +57,36 @@ const Taxes = ({ product, quantity }: TaxesProps) => {
     quantity,
     VATNumber: values.VATNumber,
   });
+  const genericPurchaseMutation = useMakePurchase();
 
   const onSaveClick = () => {
     setIsEditing(false);
-    taxMutation.mutate();
+    if (isGuest) {
```
albertkol, line 64: `isGuest` is not enough to decide which option to go for. We must also check if there is an existing `accountId` for the user. We call `preview` only when user not `isGuest` and there is and account ID associated with the user logged in. I think the account ID gets stored in the window.

static/js/src/PurchaseModal/components/Taxes/Taxes.tsx
```diff
@@ -48,10 +57,36 @@ const Taxes = ({ product, quantity }: TaxesProps) => {
...
+          product,
+          quantity,
+          action: buyAction,
+          preview: true,
+        },
+        {
+          onSuccess: () => {
+            queryClient.invalidateQueries("preview");
+          },
+          onError: (error) => {
```
albertkol, line 79: If preview errors for example due to do not show up. If I insert random string in the VAT number field and hit save. `preview` errors due to invalid VAT but nothing surfaces to the user.

albertkol (changes requested):
Everything else seems to work nicely.

static/js/src/PurchaseModal/components/Taxes/Taxes.tsx
minkyngkm: Fixed :) 

Screenshot 2023-01-04 at 4 21 41 pm

minkyngkm:
@albertkol Thanks for reviewing! All addressed! 🙌

One thing I am wondering though is country and vat number changes only work when recaptcha is checked and the recpatcha is down in step 6. (It looks like its working but its not actually working. Error is made without recaptcha checked in the console). What do you think?

albertkol:
> @albertkol Thanks for reviewing! All addressed! raised_hands
> 
> One thing I am wondering though is country and vat number changes only work when recaptcha is checked and the recaptcha is down in step 6. (It looks like its working but its not actually working. Error is made without recaptcha checked in the console). What do you think? Ah I remember I asked you about this before! What was the solution?

I thought the back-end errors out because the captcha is not check but I checked quickly and I no longer think that. I think there is somewhere a front-end validation happening that fails due to the captcha not being checked. It requires some investigation. 

Similarly for the case one has a freshly logged in Ubuntu One account and attempts a purchase, the step one save forces the user to insert a credit card. That also require an investigation, but I think the answer is in the sample place.

We should fix these 2 problems as part of this PR.

minkyngkm:
Thanks @albertkol 👍 The recaptcha thing is fixed and the issue has been made at https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/801?modal=detail&selectedIssue=WD-1049

albertkol (approved):
:+1:

PR #12436, opened 2023-01-19
Keep URL parameters after SSO authentication

jpmartinspt:
## Done
As the title says, with this change URL parameters will be preserved during the login flow for shop pages. This is going to be handy for Ubuntu Pro pages where users login in the middle a flows and it's important to keep the URL parameters that are set before logging in.

## QA
- Go to https://staging.ubuntu.com/pro/dashboard?params=true
- Check that after SSO the parameters are gone.
- Go to https://ubuntu-com-12436.demos.haus/pro/dashboard?params=true
- Check that after SSO the parameters are kept.

### DIFF
--- a/webapp/shop/decorators.py
+++ b/webapp/shop/decorators.py
@@ -104,7 +104,7 @@ def shop_decorator(area=None, permission=None, response="json", redirect=None):
 
             if permission == "user" and response == "html":
                 if not user_token:
-                    redirect_path = redirect or flask.request.path
+                    redirect_path = redirect or flask.request.full_path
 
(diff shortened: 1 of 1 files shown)


nottrobin:
This works, but it also seems to work on live. E.g. if I go to https://ubuntu.com/?hello=world and login, I'm returned with the parameters. Same as with the demo.

jpmartinspt:
@nottrobin I changed this. Turns out this was only happening in Ubuntu Pro pages. I changed the appropriate decorator and updated the QA steps.

PR #12466, opened 2023-01-26
Fix a11y cypress error

albertkol:
## Done

- Fix /pro page accessibility error

## QA

- Are cypress checks passing?

### DIFF
--- a/templates/pro/index.html
+++ b/templates/pro/index.html
@@ -644,7 +644,7 @@ https://docs.google.com/document/d/1In4NSnckAtL51_7D_AZCHa6TPnPGXqctXo-5rE77_aY/
       </div>
     </div>
     <div class="row">
-      <pre><code>
+      <pre><code tabindex="0">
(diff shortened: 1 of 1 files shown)


bartaz:
Is making the code focusable intentional?

Screenshot 2023-01-26 at 16 57 19

What a11y issue is this fixing?

albertkol:
@bartaz the Cypress test failed for this reason:

image

And looking into it this is the information I found:
https://accessibilityinsights.io/info-examples/web/scrollable-region-focusable/

TL;DR: 
image

I am not even noob level on the accessibility topic. This is what I have discovered today. If this is not the right fix, I am happy to discuss a better solution to this problem.

bartaz:
Thanks for context @albertkol.

Would it work to put the tabindex on `pre` instead? I think the `pre` is the "scrollable" element in this case, and it would look a bit less weird when focused:

Screenshot 2023-01-26 at 17 04 59

PR #12471, opened 2023-01-27
Fix the live links check

anthonydillon:
## Done

- Remove the erroneous wildcard from the live link check. 

## QA

- Run the script local and check it runs and works


petesfrench (approved):
LGTM

PR #12518, opened 2023-02-08
Use CS endpoints for instant provisioning

mrgnr:
## Done

- Instantly-provisioned exams are scheduled through the CS contracts service rather than directly using the TA API

~~**Note:** Instantly-provisioned exams are currently scheduled for 10 mins into the future to due to TA platform limitations. We will remove this window once TA has implemented the necessary patch.~~

## QA

- Obtain a CUE activation key using the Backoffice tool (or get one from Morgan)
- Visit https://ubuntu-com-12518.demos.haus/credentials/redeem/[activation-key] to redeem your key
- Visit https://ubuntu-com-12518.demos.haus/credentials/your-exams
  - Verify that your exam is listed
  - Click the "Take now" button
  - Verify that the exam is built and that you are redirected to the exam environment

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/CRED-289

## Screenshots

cred-cs-provision

cred-beta-provisioning

## Help

### DIFF
--- a/webapp/shop/cred/views.py
+++ b/webapp/shop/cred/views.py
@@ -330,9 +330,7 @@ def cred_provision(ua_contracts_api, trueability_api, **_):
     else:
         error = "Exam not found"
 
-    if flask.request.method == "POST" and not reservation_uuid:
-        data = flask.request.form
-        contract_item_id = data["contractItemID"]
+    if not reservation_uuid:
         tz_info = pytz.timezone("UTC")
         starts_at = tz_info.localize(datetime.utcnow() + timedelta(seconds=70))
         first_name, last_name = sso_user["fullname"].rsplit(" ", maxsplit=1)
@@ -359,9 +357,7 @@ def cred_provision(ua_contracts_api, trueability_api, **_):
         response = trueability_api.get_assessment_reservation(reservation_uuid)
 
         if "error" in response:
-            error = response.get(
-                "message", "An error occurred while fetching your exam."
-            )
+            error = response.get("message", "No exam booking could be found.")
         else:
             reservation = response["assessment_reservation"]
             assessment = reservation["assessment"]
--- a/webapp/shop/cred/views.py
+++ b/webapp/shop/cred/views.py
@@ -205,13 +205,13 @@ def cred_your_exams(ua_contracts_api, trueability_api, **_):
                         "text": "Schedule",
                         "href": "/credentials/schedule?"
                         f"contractItemID={contract_item_id}",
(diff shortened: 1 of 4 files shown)


webapp/shop/cred/views.py
```diff
@@ -293,76 +300,71 @@ def cred_exam(trueability_api, **_):
 
 @shop_decorator(area="cred", permission="user", response="html")
 @canonical_staff()
-def cred_provision(trueability_api, **_):
+def cred_provision(ua_contracts_api, trueability_api, **_):
+    contract_item_id = flask.request.args.get("contractItemID")
```
jpmartinspt, line 304: If you do `contract_item_id = flask.request.args.get("contractItemID", type=int)` you can remove the cast on line `321`

mrgnr: Done :+1:

webapp/shop/cred/views.py
```diff
@@ -293,76 +301,78 @@ def cred_exam(trueability_api, **_):
...
+def cred_provision(ua_contracts_api, trueability_api, **_):
+    contract_item_id = flask.request.args.get("contractItemID", type=int)
+
+    if contract_item_id is None:
+        return flask.redirect("/credentials/your-exams")
+
     sso_user = user_info(flask.session)
-    sso_user_email = sso_user["email"]
-    ability_screen_id = 4194
+    country_code = get_user_country_by_ip().json["country_code"] or "DE"
```
jpmartinspt, line 311: Do we want to default to "DE"?

mrgnr: @jpmartinspt get_user_country_by_ip() may set `country_code` to `None` if the lookup failed. The country code is a mandatory field to create an exam reservation (and is used to determine the deployment region). Would it be better show an error, or use a default?

jpmartinspt: Hmm I think it's fine to default to a specific region, the question was more if "DE" was indeed the optimal default.

jpmartinspt: If there is no drawback in doing so maybe we could let the user pick the region.(We don't need to block this PR for that now)

mrgnr: Cool, we can do another PR to let the user select their country if we can't look it up.

For this PR, maybe "GB" is a more suitable default since Canonical is UK-based? (Also, one of our Pools is configured to use europe-west2 (London).)

albertkol (approved):
:+1: It seemed like you made Joao happy in the comments.

PR #12520, opened 2023-02-08
Capture referrer during a shop purchase

minkyngkm:
## Done

-Capture referrer during a shop purchase

## QA

- Go to `/pro/subscribe`
- 
## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-1790


webapp/shop/decorators.py
```diff
@@ -73,6 +74,10 @@ def decorated_function(*args, **kwargs):
                     flask.session.pop(metadata_key, None)
                     value = flask.request.args.get(query_parameter)
                     flask.session[metadata_key] = value
+                else:
```
jpmartinspt, line 77: What is the purpose of this else? Does it not remove the referrer from session straight after we set it?

minkyngkm: When setting a query string, removing it (without making any purchase) and trying to make a purchase, it still has the previous value in session. I thought if there's no query string in the url it should be removed in session as well.

jpmartinspt:
@minkyngkm seems like we need to re-open against main as the checkout branch was deleted.

webapp/shop/decorators.py
jpmartinspt: I think this else could be removed along with the changes to the JS files and we could just `flask.session.pop("referrer")`once a successful purchase as been made. That way you don't need to worry about propagating the parameter through the URL. What do you think?

PR #12530, opened 2023-02-10
Community member no of machines

### LINKED ISSUE #12453: Free Pro personal token incorrect count for community members
Hello, our documentation for Ubuntu Pro says that there's two free tiers, one for everybody with five machines, and one for Ubuntu members with fifty machines:

> Anyone can use Ubuntu Pro for free on up to 5 machines, or 50 if you are an official Ubuntu Community member.

https://ubuntu.com/pro

You can see from my launchpad page https://launchpad.net/~seth-arnold that I'm a member of https://launchpad.net/~ubuntumembers but this isn't reflected in my own dashboard:

count

This was reported to me by https://launchpad.net/~jdstrand

Thanks

---
*Reported from: https://ubuntu.com/pro/dashboard*

abhigyanghosh30:
## Done

- Change number of machines for community members to 50.
- @abhigyanghosh30 I moved the launchpad call to check for teams as part of the login. I noticed that the subscriptions took considerably longer to load if we do it when we call `/pro/user-subscriptions`. Even this with method login time is much longer. @jpmartinspt there should be ways to use `open_id` to query for team membership. But I did not want to spend much more time on this, did what we do for security api. 

Update: switched to check team membership as a OpenID exptension with the help of the logic from the django-openid-auth package we used for XYZ. Login is quick again.

## QA

- Fetch branch
- `dotrun`
- login you should see 5 (or 3) machines if you are not a community member
- change team to anything you are part of that's a public launchpad team fx: `xyz-admins`
- logout
- login
- you should see 50 machines now

## Issue / Card

Fixes #12453


webapp/shop/advantage/views.py
```diff
@@ -89,6 +90,16 @@ def advantage_view(advantage_mapper, **kwargs):
...
+    launchpad = Launchpad.login_anonymously(
+        "ubuntu.com/pro", "production", version="devel"
+    )
+    sso_user = user_info(flask.session)
+    lp_user = launchpad.people.getByEmail(email=sso_user["email"])
+    community_members = launchpad.people("ubuntumembers").members
+    if lp_user in community_members:
+        for subscription in user_subscriptions:
+            if subscription.type == "free":
+                subscription.number_of_machines = 50
```
albertkol, line 102: Solution seems to work. However, please change it a bit. Don't override the `subscription` here.

Instead please save `lp_user in community_members` as a variable and send it to the `advantage_mapper.get_user_subscriptions`. Please override the number of machines at the end of that function. Still not ideal, but it's a temporary thing anyway.

`user_subscriptions = advantage_mapper.get_user_subscriptions(email)` must return the final shape of the subscriptions.

albertkol (changes requested):

albertkol (dismissed):

jpmartinspt (approved):

PR #12656, opened 2023-03-08
fix for peer score coming as string

abhigyanghosh30:
## Done

- Fixes NPS scores coming in as string on sheets

## QA

- fill the form in /credentials/exit-survey
- Check the sheet to check if the NPS scores land as integers instead of string

## Issue / Card

Fixes #[CRED-333](https://warthogs.atlassian.net/browse/CRED-333?atlOrigin=eyJpIjoiZWQ4OTA4Yjk2OThlNDg3MmEzY2E3ZDEwYjBiYmNkYzciLCJwIjoiaiJ9)

## Screenshots

## Help

[CRED-333]: https://warthogs.atlassian.net/browse/CRED-333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webapp/shop/cred/views.py
```diff
@@ -496,6 +496,12 @@ def cred_submit_form(**_):
         value = ", ".join(values)
         if value:
```
jpmartinspt, line 497: Do we need the if? The value already seems to default to `''` anyway. You could actually change the if condition to be something like:
```
if key in ["ExitSurveyPromoterManager", "ExitSurveyPromoterPeer"]:
    form_fields[key] = int(value)
else:
    form_fields[key] = value
``` 
Another thing to consider is that if "ExitSurveyPromoterManager" or "ExitSurveyPromoterPeer" are not convertable to int, the form submission will fail.

abhigyanghosh30: It is a radio button with values from 1 to 10. It has to be a number.

abhigyanghosh30: I think the default should also be changed to 0 or 5. The form will not submit without this anyways as it is a required field

jpmartinspt: We should not have the `if` because we should at least write the `''` to the spreadsheet, or we risk getting the fields out of order.

jpmartinspt: I don't think we should worry particularly about the defaults for required fields, but we should make sure the rows we write are always the same length.

abhigyanghosh30: The rows will always be of the same length because all the keys are already in the dictionary and all have a default value. The value in the dictionary gets updated only if there is a non null value replacing the current value. That is why the if condition is there.
Otherwise if there is an empty field, None value will be added to the row which will give errors while writing to the sheet

jpmartinspt: Got it, i misread something and thought we were building the row in the for loop.

jpmartinspt (approved):

PR #12657, opened 2023-03-08
WD-2509 Replace the intel logo on the automotive page

anthonydillon:
## Done

- Replace the intel logo on the automotive page

## QA

- Check the new logo matches the one in [the copy doc](https://docs.google.com/document/d/10jSLdwz2Ks7gnEMXlKWzrpsY6TQFxISUgnQo4y3yOrA/edit#)

### DIFF
--- a/templates/automotive/index.html
+++ b/templates/automotive/index.html
@@ -357,13 +357,12 @@
           </div>
           <div class="p-logo-section__item">
             {{ image (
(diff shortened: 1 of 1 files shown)


carkod:
What about removing the two other logos (BMW and the Daimler logos)? Are we doing that?

anthonydillon:
> What about removing the two other logos (BMW and the Daimler logos)? Are we doing that?

Why not, done :+1:

carkod (approved):
LGTM

PR #12685, opened 2023-03-16
Fix broken links

anthonydillon:
## Done

- Fixed the broken links reported here: https://github.com/canonical/ubuntu.com/actions/runs/4434564485/jobs/7780674156


mtruj013 (approved):
LGTM, but we should probably also update or remove this redirect https://github.com/canonical/ubuntu.com/blob/7ab63379e6e91f23b796195b2c39b276af89d3d3/redirects.yaml#L619

PR #12695, opened 2023-03-17
Schedule shop for maintenance

albertkol:
## Done

- Change to the full maintenance feature:
  - `MAINTENANCE_URLS` is the list of urls that will show the maintenance view when maintenance is on.
  - un-hardcode `is_in_maintenance` we make use of this value to hide specific features(fx: resizing, or buy buttons)
- Schedule it


minkyngkm (approved):
Looks great! 🙌

Issue #11916, opened 2022-08-01
Able to select disabled sections


anthonydillon:
When tabbing through the UA selector the disabled areas are accessible.

PR #12730, opened 2023-03-29
Accessibility for disabled area on /pro/subscribe

minkyngkm:
## Done

- Disable tab key on disabled area

## QA
- Go to /pro/subscribe
- Tab through the page and check tab is not working on disabled area

## Issue / Card

Fixes https://github.com/canonical/ubuntu.com/issues/11916

### DIFF
--- a/static/js/src/advantage/subscribe/react/components/Form/Feature/Feature.tsx
+++ b/static/js/src/advantage/subscribe/react/components/Form/Feature/Feature.tsx
@@ -7,6 +7,7 @@ import {
   isIoTDevice,
   isPublicCloud,
   ProductTypes,
+  Support as SupportEnum,
 } from "advantage/subscribe/react/utils/utils";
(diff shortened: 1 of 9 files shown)


static/js/src/advantage/subscribe/react/components/Form/Feature/Feature.tsx
```diff
@@ -25,14 +25,35 @@ const Feature = () => {
 
   const infraOnlyDisabled = ProductTypes.desktop === productType;
 
+  const isDisabled = (support?: string) => {
```
jpmartinspt, line 28: Same comments that I left for the other component, use `SupportEnum` and make the function argument required.

jpmartinspt:
I think we should definitely do something about it, either disable it or not show it at all. Let's create a separate ticket for that, so we can decide what's the best approach first.

jpmartinspt (approved):
LGTM

PR #12739, opened 2023-03-31
Fix renewal message for trials

minkyngkm:
## Done

- Enable auto-renewal by default for Trials.

## QA

- Go to `/pro/dashboard`

## Issue / Card

Fixes #https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/801?modal=detail&selectedIssue=WD-2659

### DIFF
--- a/webapp/shop/api/ua_contracts/helpers.py
+++ b/webapp/shop/api/ua_contracts/helpers.py
@@ -252,8 +252,11 @@ def get_user_subscription_statuses(
         ]
 
         statuses["is_trialled"] = True if active_trial else False
-        if not statuses["is_cancelled"]:
-            statuses["is_subscription_auto_renewing"] = True
+        statuses[
+            "is_subscription_auto_renewing"
+        ] = is_billing_subscription_auto_renewing(
+            subscriptions, subscription_id
+        )
 
     if type in ["yearly", "monthly"]:
         statuses["is_subscription_active"] = is_billing_subscription_active(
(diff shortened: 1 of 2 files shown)


albertkol:
Asked for a few more questions on this one to get a clarification on how this error can occur to begin with.
https://chat.canonical.com/canonical/pl/mcdfmmt1epy9tdt6ukhd4a3wbe

Trials, unlike Monthly and Yearly subscriptions do not display the Auto-renewal settings. So the message is still misleading. 

Trials should naturally auto-renew, unless they were cancelled. There are no actions a user can currently take to prevent a Trial from expiring. 

But one of the potential fixes is to enable auto-renewal settings for Trails too, depending on the conclusion of the conversation.

webapp/shop/api/ua_contracts/helpers.py
```diff
@@ -252,6 +252,8 @@ def get_user_subscription_statuses(
         ]
 
         statuses["is_trialled"] = True if active_trial else False
+        if not statuses["is_cancelled"]:
```
albertkol, line 255: The condition by which a `trial` is auto-renewing is not determined by its cancelled status.

When we call the UAContract API for an account's subscriptions, the subscription payload has a property to tells us if it's auto-renewing or not, we just need to look at that. That property is parsed and converted to the `subscription.is_auto_renewing`.

But that's details. If you replace this `if` with it should fix it:
```
        statuses[
            "is_subscription_auto_renewing"
        ] = is_billing_subscription_auto_renewing(
            subscriptions, subscription_id
        )
```
We do this just below for the `yearly` and `monthly`. It seems that for trials we don't do it, so the default value `False` remains unchanged even if the subscription is auto-renewing.

albertkol (changes requested):

albertkol:
We do this as part of https://github.com/canonical/ubuntu.com/pull/12942

Too much headache to rebase to `main`.

PR #12745, opened 2023-04-03
Update list activation keys for contract endpoint path

pandrey2003:
## Done

Commercial Systems has updated the path to list activation keys for a contract. The changes on our side will be deployed on Wed, Apr 5th. Given this page is Canonical-only, we don't see much point in `try` / `except` dances when back-porting from an older to a newer endpoint.


pandrey2003:
Deployed on our side - ready to be merged & deployed here.

albertkol (approved):
:+1:

PR #12759, opened 2023-04-05
Hide error message when value changes

minkyngkm:
## Done

- Hide error message when value changes

## QA
- Go to /pro/subscribe 
- Click "Buy now" button
- See the global error message displays at the top and validate message on the fields
- Update some of the fields that show validate message
- See the global error message disappears right after updating field
- Type invalid vat number and check global error message for the vat number displays properly
- Type invalid card number and  check global error message displays properly
- QA with the other error cases as well 

## Issue / Card

Fixes #https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/801?modal=detail&selectedIssue=WD-2941

### DIFF
--- a/static/js/src/advantage/subscribe/checkout/components/BuyButton/BuyButton.tsx
+++ b/static/js/src/advantage/subscribe/checkout/components/BuyButton/BuyButton.tsx
@@ -17,9 +17,10 @@ type Props = {
   quantity: number;
   product: Product;
   action: Action;
+  error: any;
 };
 
-const BuyButton = ({ setError, quantity, product, action }: Props) => {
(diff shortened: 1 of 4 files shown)


albertkol (commented):
I don't like this implementation. It touches too many components. 

I was hoping Formik has a onChange handler. I googled a bit and found this: https://plainenglish.io/blog/how-to-listen-to-formik-onchange-event-in-react
But the `FormObserver`, did not work as expected.

But rather than a component, perhaps having just a simple JS event listener for all `input` and `select` HTML elements that would hide the notification once it is displayed is enough. 

Either way.. I would rather always have the notification up than add all these new changes to the components.

minkyngkm:
Closing this PR as the ticket has been moved into ice box.

PR #12806, opened 2023-04-19
(feat) Data Fabric Docs landing page take 3 

grobbie:
## Done

- Add initial /data/docs landing page
- Add integration with charmhub discourse for charmed spark, mongodb and postgres documentation
- (driveby) fix Spark page layout
- Fix branding on Kafka page to align with /data/docs and current Data Fabric branding

### DIFF
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -103,6 +103,16 @@ env:
       key: api-key
       name: trueability
 
+  - name: CHARMHUB_DISCOURSE_API_KEY
+    secretKeyRef:
+      key: charmhub-api-key
+      name: discourse-api
+
+  - name: CHARMHUB_DISCOURSE_API_USERNAME
+    secretKeyRef:
+      key: charmhub-api-username
+      name: discourse-api
+
 memoryLimit: 512Mi
 
 extraHosts:
@@ -522,6 +532,16 @@ staging:
     - name: CANONICAL_LOGIN_URL
       value: https://login.ubuntu.com/
 
+    - name: CHARMHUB_DISCOURSE_API_KEY
+      secretKeyRef:
+        key: charmhub-api-key
+        name: discourse-api
+
+    - name: CHARMHUB_DISCOURSE_API_USERNAME
+      secretKeyRef:
+        key: charmhub-api-username
+        name: discourse-api
+
   routes:
     - paths: [/blog]
       name: ubuntu-com-blog
@@ -866,3 +886,13 @@ demo:
 
     - name: CANONICAL_LOGIN_URL
       value: https://login.ubuntu.com/
+
+    - name: CHARMHUB_DISCOURSE_API_KEY
+      secretKeyRef:
+        key: charmhub-api-key
+        name: discourse-api
+
+    - name: CHARMHUB_DISCOURSE_API_USERNAME
+      secretKeyRef:
+        key: charmhub-api-username
+        name: discourse-api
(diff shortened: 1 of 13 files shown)


mtruj013:
Hi @grobbie , quick question about this one, I see you added a new page [/data/docs/index.html](https://ubuntu-com-12806.demos.haus/data/docs), did UX get a chance to look at it beforehand? A visual designer would also need to take a look, I imagine it'd need a redesign as we're not supposed to be merging pages in the old style anymore 

cc: @juanruitina, @lyubomir-popov

grobbie:
Ah, no it's not been past UX, sorry. Let me know what changes to make and I'll make 'em

mtruj013:
Waiting on UX and Visual design, marking as blocked for now

mtruj013:
I think it'd be better to do:
 
/data/docs/kafka
/data/docs/kafka/iaas
/data/docs/kafka/k8s
/data/docs/mongodb
/data/docs/postgresql

This is more in line with URL structure we have for other docs pages, like [k8s docs](https://ubuntu.com/kubernetes/docs) for example. Unrelated to the URLs, but we should also make sure the docs module is able to render the design as I'm not sure if it has the functionality to build this currently: 
image

I could be wrong about this though, maybe you got this design from an existing docs page. If so, ignore me :)

grobbie:
About the language being quite marketing oriented - I was just modelling it
based on the Server docs page.

I'm a bit worried about how navigation between K8s documentation and IAAS
documentation will work, could you have a think about how that would work?
I fear it'll mean another page - so a landing page to a landing page to the
actual documentation, which I feel is putting barriers and frustration in
the way of the user. It would be nice if the user could select the docs for
their target platform directly from the landing page.

juanruitina:
@mtruj013 Sure, we can use that URL structure then. The two column layout seems to be doable, we use it in the [Server docs](https://ubuntu.com/server/docs).

@grobbie We can do something about presenting both options to users in the landing page, let me give it some thought. One question: to what extent will any of the docs overlap with the docs of their respective snaps?

grobbie:
The docs are for complete solutions. The snap package is just one component
of the overall IAAS solution. Obviously for Kubernetes there's no snap
package in general, although we provide a client side tools collection in a
snap for Spark on K8s.

Actually I feel this is a broader problem in our documentation systems
which are not really geared towards documenting solutions but more towards
components ("X charm", "Y Snap", "Z rock"). Obviously users have interest
in both - the macro solution view and the micro component view of the
components that compose that solution, but at the moment it feels like we
don't generally have easy ways to help the user navigate to the macro
solution view.

Charmhub and snapcraft very much focus on offering hooks for documenting
individual charms, but solutions are composed of several Charms, many Rocks
images and Snaps, and maybe other stuff too like Terraform plans, and it
can be hard for the documentation author to tie it all together so that
wherever the user enters the documentation from, they're easily able to
find a view of the complete solution. This is also compounded by multiple
solutions using the same charms, like the Traefik Charm or the
tls-certificates-operator Charm.

HTH
Rob

anthonydillon:
@juanruitina I don't understand, there are only three docs sections:
1. /data/docs/postgresql-k8s
2. /data/docs/mongodb-iaas
3. /data/docs/spark-k8s

Is there due to be more? @grobbie 

If that is the case, I believe the structure in this PR is good.

grobbie:
Hi Anthony
This PR can be disregarded for everything except the little bit of custom session logic for charmhub discourse / ubuntu discourse to stop it getting confused and crashing.
There should be just one /docs page and that shall integrate the landing page we've made on charmhub discourse here - https://discourse.charmhub.io/t/canonical-data-fabric-documentation/10863 that landing page has a heirarchy for all the  product docs below it.

PR #12818, opened 2023-04-21
Actions: Force node 16 on test-js job

jpmartinspt:
## Done

Latest actions image started defaulting to node 18 and broke some tests. Forcing node 16 as before to get everything back to green. We should look at upgrading node version whenever possible.

## QA

- Check `test-js` is passing.


albertkol (approved):
:+1:

Side note:
Begs the question.. why do we use `node:16` for tests and local development too, but we use `node:18` for production. Would be nice if we brought them all in line.

I see the Dockerfile was updated to use jammy, with python3.10 and node:18, which if I know right are not in line with `dotrun`. I am all for using the latest versions, but you'd think important decisions like this one would be made as a team.

Another wish of mine would be for flask-base to be updated with the lasted flask version. 

But all these points are beside the point. Thanks for the help on this one, Joao! Well done!

PR #12865, opened 2023-05-12
Fix page title

mtruj013:
## Done

- Fix page title "Install Landscape" -> "Managed Landscape"

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/landscape/managed
- Check that title has been updated

### DIFF
--- a/templates/landscape/managed.html
+++ b/templates/landscape/managed.html
@@ -1,6 +1,6 @@
 {% extends "landscape/base_landscape.html" %}
 
-{% block title %}Install Landscape{% endblock %}
+{% block title %} Managed Landscape {% endblock %}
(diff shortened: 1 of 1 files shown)


templates/landscape/managed.html
```diff
@@ -1,6 +1,6 @@
 {% extends "landscape/base_landscape.html" %}
 
-{% block title %}Install Landscape{% endblock %}
+{% block title %} Managed Landscape {% endblock %}
```
tbille, line 3: is there a reason why you added an extra space before and after?

mtruj013: Nah, just habit, I was trying to get this up quickly. For the copydoc block I do it intentionally so that the link is clickable

tbille (approved):

PR #12911, opened 2023-05-30
Improve performance of engage pages submit

carkod:
## Done

Improve performance of engage pages.

I put some timestamps in various places before and after the submit, I found about 5 seconds difference between the start of the marketo endpoint and then return to the thank-you page. Another 5 secs roughly for rendering the thank-you page, so I focused on improving the rendering part of the thank-you page. Changing marketo/submit endpoint could bring breaking changes, so I tried not to do that, but the other 5 seconds delay is contained within this endpoint.

So this PR should decrease by half the time it takes to submit.

## QA

- Go to https://ubuntu-com-12911.demos.haus/engage, pick a whitepaper
- Fill in the form with test data and submit. Compare it with prod, it should be faster now

## Help

### DIFF
--- a/templates/engage/thank-you.html
+++ b/templates/engage/thank-you.html
@@ -72,23 +72,4 @@
   </div>
 </section>
 
-{% if related | length > 0 %}
-<section class="p-strip--light">
-  <div class="row">
-    <div class="col-8">
-      <h2 class="p-heading--3">You may also be interested in &hellip;</h2>
-    </div>
-  </div>
-  <div class="row p-divider">
-    {% for page in related %}
-    <div class="col-4 p-divider__block">
-      <!-- THREE ADDITIONAL CTAs -->
-      <h4>{{page["topic_name"]}}</h4>
-      <p>{{page["subtitle"]}}</p>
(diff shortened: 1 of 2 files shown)


petesfrench (approved):
LGTM, I tested a few pages and it definitely shaves a few seconds off. However, i think we should look into this more as it is still taking a considerable amount of time.

As for removing the 'related articles', do we plan to implement this in another, more efficient manor at some point?

carkod:
> LGTM, I tested a few pages and it definitely shaves a few seconds off. However, i think we should look into this more as it is still taking a considerable amount of time.
> 
> As for removing the 'related articles', do we plan to implement this in another, more efficient manor at some point?

Thanks. Yes the other side of things is the marketo endpoint, I got consistently 5 seconds from start of the marketo/submit function to the return to the thank-you page, but if I start refactoring that can be a breaking change. For now, I think this is the best course of action.

The related pages was buggy before, and no Pdm or anybody reported it, makes me think we don't really use it. But yes, definitely need a better way to do it. Currently we are doing it on our side (looping through all engage pages and matching their tags, which obviously is super expensive), ideally discourse should spit out somehow already for us to display it (maybe with data-explorer plugin?)

PR #12920, opened 2023-06-01
Pass accountID when activating key

pandrey2003:
## Done

Some keys can be activated only in personal accounts, others - only in paid. By providing `accountID`, we eliminate ambiguity regarding the account getting the entitlement.

### DIFF
--- a/static/js/src/advantage/credentials/api/keys.js
+++ b/static/js/src/advantage/credentials/api/keys.js
@@ -1,4 +1,4 @@
-export async function listAllKeys(contractId) {
+export async function listAllKeys() {
   let response = await fetch(`/credentials/keys/list`, {
     method: "GET",
     headers: {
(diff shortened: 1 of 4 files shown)


static/js/src/advantage/credentials/api/keys.js
```diff
@@ -25,7 +25,7 @@ export async function rotateKey(activationKey) {
   return data;
 }
 
-export async function activateKey(activationKey) {
+export async function activateKey(activationKey, accountId) {
```
pandrey2003, line 28: The idea is that we can pass `window.accountId` for the current account.

pandrey2003:
The CI failures:
```
ImportError: 
NOTE: In order to use the FileCache you must have
filelock installed. You can install it via pip:
  pip install filelock
```
...do not seem to result from changes in this PR.

Please review, Albert(@albertkol), Min(@minkyngkm), Leia(@leiaru-c), Osama(@Osama-Kassem) and Allan(@alnvdl-work).

PR #12930, opened 2023-06-06
use different value for level in certified template

codersquid:
The api now returns Certified Pre-Install instead of Enabled. This changes the template to reflect that.

## Done

This change updates the template so that it checks for `Certified Pre-Install` instead of enabled. The api changed the values stored in the database so that the string changed.

## QA

- Go to the https://ubuntu-com-12930.demos.haus//certified/202012-28554 endpoint.  You should see the string about pre-installed systems rather than no string about the image.

### DIFF
--- a/templates/certified/model-details.html
+++ b/templates/certified/model-details.html
@@ -81,7 +81,7 @@
             {% else %}
             <section>
             {% endif %}
-            {% if release.level == "Enabled" %}
+            {% if release.level == "Certified Pre-Install" %}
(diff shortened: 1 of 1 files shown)


codersquid:
Jenkins shows that the demo timed out when trying to build, is there a way to kick off another demo?

nottrobin:
OMG the demo finally built successfully! https://ubuntu-com-12930.demos.haus/ 🎉

PR #12943, opened 2023-06-09
Cannot buy in trial

albertkol:
## Done

- If user attempts to buy a product while trialling, it will throw a general error.
- This change will make it clearer why a purchase fails.

## QA

- Buy a trial subscription
- Attempt to buy another subscription
- Do you see the error message?

## Issue / Card

Fixes #https://warthogs.atlassian.net/browse/WD-4249


minkyngkm:
It works well! 

As a personal preference, I would like the error message to display as soon as the user goes to the checkout page but for me it showed during reCaptcha verification with image selection in confirm and buy step. I thought it would be great if checkout loading spinner or something shows until error message shows. It's a small issue and it still blocks users from purchasing though.

albertkol:
@minkyngkm I reworked it a bit. Ready for re-review. The error should show up immediately.

minkyngkm (approved):
LGTM! 👍

PR #12948, opened 2023-06-12
Fix cancelled subscription bug

albertkol:
## Done

- Fix cancelled subscription bug
- This happens because the interface assumes that the users cancels a subscription using the interface, so down-sizing to 0. The condition based on which we'd set a subscription to cancelled is if the number of machines for next billing cycle == 0
- Using the backoffice tool changes the status of a subscription to `deactivated`, but leaves the number of machines for next billing cycle untouched, therefore subscription would still show as active on the u.c.
- Changed so we also take into account the status of the subscription when determining the if the subscription is cancelled or not
- Added Cancelled status indicator

## QA

- Have a fresh account
- Buy a product
- Cancel it using the backoffice tool
- Go to /pro/dashboard on staging, the Edit subscription button will show up
- Got to /pro/dashboard on the demo, the Edit subscription button will not show up

## Issue / Card

Fixes #https://warthogs.atlassian.net/browse/WD-2040

## Screenshot
image

### DIFF
--- a/static/js/src/advantage/react/components/Subscriptions/SubscriptionDetails/SubscriptionDetails.test.tsx
+++ b/static/js/src/advantage/react/components/Subscriptions/SubscriptionDetails/SubscriptionDetails.test.tsx
@@ -508,6 +508,28 @@ describe("SubscriptionDetails", () => {
     expect(wrapper.find(".p-chip__value").text()).toBe("Cancelled");
   });
 
+  it("it does display the cancelled label for trial shop purchases", () => {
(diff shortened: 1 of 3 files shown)


minkyngkm (commented):
It works well according to QA steps 👍 but can we have "cancelled" or something instead of "Auto-renewal on" it still looks very confused for me to know if its cancelled or not. maybe in the near future? 
image

minkyngkm (approved):
Looks great! As we talked, it will be great if "auto-renewal on" label is added on the trial card as well by other PR! 👍👍

PR #12954, opened 2023-06-13
Fix trial status

albertkol:
## Done

- Make the user trial status consistent between the product selector and the checkout
- The current rules they both abide are:
  - user cannot have active / locked subscriptions
  - user cannot have any subscriptions that have started as a trial
  - list must have more that 0 day of trial available
If these conditions are fulfilled the the trial setting should show up and the user will be able to start a trial.

## QA

- Buy a subscription (not a free trial)
- Cancel subscription from the backoffice tool
- Go to product selector you will see the free trial, you will see the setting on the checkout too

- Buy a trial
- Cancel trial
- Trail will no longer be available

## Issue / Card

Fixes #https://warthogs.atlassian.net/browse/WD-4364

### DIFF
--- a/webapp/shop/views.py
+++ b/webapp/shop/views.py
@@ -55,26 +55,28 @@ def get_purchase_account_status(advantage_mapper: AdvantageMapper, **kwargs):
         return flask.jsonify({"error": "access forbidden"}), 403
 
     last_purchase_ids = {}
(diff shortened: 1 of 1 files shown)


minkyngkm (approved):
LGTM 👍

PR #12974, opened 2023-06-20
K8s docs deletions and redirects

evilnick:
There are a number on non-maintained docs pages for specific charms which are now maintained in charmhub. Until now these pages have been left alone as historic version docs pointed to them for reference, but this is no longer regarded as safe

## Done

- delete the relevant charm pages
- add redirects pointing to the current maintained docs on charmhub
 (N.B. the 'kubernetes-master' charm was renamed 'kubernetes-control-plane' which is why there are two redirects to the same charmhub page)


anthonydillon (approved):
LGTM

PR #12985, opened 2023-06-23
Fix cypress

albertkol:
## Done

- Some bizarre `TypeError` causing problems. Muting these problems for Cypress.

## QA

- Are checks passing?

### DIFF
--- a/tests/cypress/support/pro/commands.js
+++ b/tests/cypress/support/pro/commands.js
@@ -34,7 +34,9 @@ Cypress.Commands.add(
 
 Cypress.on("uncaught:exception", (err, runnable) => {
   // returning false here prevents Cypress from failing the test
-  return false;
+  if (err.message.includes("window.lintrk is not a function")) {
+    return false;
+  }
 });
 
 Cypress.Commands.add(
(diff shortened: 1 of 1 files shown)


tests/cypress/support/pro/commands.js
```diff
@@ -32,6 +32,11 @@ Cypress.Commands.add(
   }
 );
 
+Cypress.on("uncaught:exception", (err, runnable) => {
+  // returning false here prevents Cypress from failing the test
+  return false;
+});
+
```
petermakowski, line 39: This will make cypress ignore all uncaught JS runtime exceptions, are we sure this is what we want? If possible, it might be better to identify and fix the actual error that's thrown instead.

petermakowski: If we want to workaround a specific error, do you think you could narrow it down as explained here: https://docs.cypress.io/api/cypress-api/catalog-of-events#To-conditionally-turn-off-uncaught-exception-handling-for-a-certain-error

albertkol: Thanks @petermakowski 

I will narrow it down to a specific error, then. I am not sure why this typing error happens on cypress. It does not pop up on prod/demos.

petermakowski (approved):

PR #13006, opened 2023-06-30
Adjust footer size on small screens

Jim-Dillon:
## Done

Some suggested improvements for the Ubuntu.com footer on small screens: 

- Reduced the size (font and line-height) of footer accordions to take up less space on small devices 
- Increased the size of 'back to top' button - to more closely resemble the other accordions and improve accessibility of this [touch target](https://accessibility.digital.gov/ux/touch-targets/#:~:text=Make%20touch%20targets%20at%20least,extend%20padding%20past%20the%20icon)
- Changed colour of 'back to top' button to help differentiate between itself and newly resized accordions. It might look better if the arrow image was also orange, but I couldn't access the svg to change it 
- Added left-side padding to secondary footer content, as previously it started against the border
- Adjusted all breakpoints so the above changes only apply to small screens
 

## QA

- Open demo 
- In demo, inspect Ubuntu.com homepage in live server on small screen (<629px) and scroll down to footer
- Open live Ubuntu.com website in separate tab for comparison 
- In demo, see the above changes that have been made

## Screenshots

Before:

Ubuntu footer #1

Ubuntu footer #2

After: 

Ubuntu footer #3

Ubuntu footer #4

## Help

### DIFF
--- a/static/sass/_pattern_footer.scss
+++ b/static/sass/_pattern_footer.scss
@@ -1,12 +1,16 @@
 // Footer pattern styling
 @mixin footer-title {
   color: $color-mid-dark;
-  line-height: 1.5;
+  line-height: 1.25;
   margin-bottom: 0;
 
   @media only screen and (max-width: $breakpoint-small) {
     cursor: pointer;
   }
+
+  @media only screen and (min-width: $breakpoint-small) {
+    line-height: 1.5;
+  } 
 }
 
 @mixin ubuntu-p-footer {
@@ -97,7 +101,7 @@
     background-size: 13px 13px;
     border-top: 1px solid $color-mid-light;
     color: $color-dark;
-    font-size: 1.2rem;
+    font-size: 1rem;
     font-weight: 300;
     margin-top: 0;
     max-width: inherit;
@@ -172,6 +176,7 @@
     padding: $sp-small 0 $sp-small $sp-x-large;
     position: relative;
     z-index: 2;
+    line-height: 2.3;
 
     @media only screen and (min-width: $breakpoint-small) {
       display: none;
@@ -183,7 +188,6 @@
   }
 
   .p-footer--secondary {
-    padding: $sp-large $sp-small 0;
 
     @media only screen and (min-width: $breakpoint-small) {
       padding: 0;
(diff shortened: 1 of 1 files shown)


anthonydillon:
Demo running at https://ubuntu-com-13006.demos.haus

static/sass/_pattern_footer.scss
```diff
@@ -172,25 +176,26 @@
...
     z-index: 2;
+    line-height: 2.3;
 
     @media only screen and (min-width: $breakpoint-small) {
       display: none;
     }
 
     a {
-      color: $color-mid-dark;
+      color: $color-brand;
```
lyubomir-popov, line 186: why brand instead of regular text colour ($color-x-dark)?

lyubomir-popov:
Something strange is happening with nested accordion items - they are a different font-weight. Could it be because the top level accordion is using a heading and a heading weight appears? 

Despite that, it is a clear improvement, so I've given a +1.

lyubomir-popov (approved):

juanruitina:
Much needed indeed! Thank you, @Jim-Dillon.

A few comments:

- "Link to top" orange doesn't have enough colour contrast so it's not accessible, I suggest to revert this change. Making it more noticeable is good, but placement and the use of an icon already provides a bit of that.
- We need to add some spacing between the links at the bottom, so they are more easily tappable. I suggest to add `margin-top: .5rem` to `.p-footer--secondary .p-inline-list__item` on mobile as a quick fix.
- This is a bit out of scope but would like to get it fixed: the top-level items ("OpenStack", "Ceph") used to be bold on desktop and aren't anymore (see below). Can we tweak them so they become bold?

Screenshot 2023-07-03 at 10 23 03

Jim-Dillon:
Thanks for the feedback on those points @lyubomir-popov @juanruitina !

I was conscious that my changing the size of the accordion items and back to top button could’ve made the two less distinguishable, which is why I chose the brand orange to differentiate them. I do agree though about the contrast not being strong enough so I’m happy to revert it back, or change it to another colour?

I’m also happy to look at those other things you both raised - I’m actually away for a few days but I can pick this up when I return home on Friday.

juanruitina:
No hurries! Yes, please, change it to `$color-x-dark` as per @lyubomir-popov's comment.

juanruitina:
This looks good to me, thank you Jim!

PR #13023, opened 2023-07-06
Update views.py - fix utm_content field

laszlokajtar:
the API name of utm_content in Marketo is utmcontent and that cannot be changed

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -811,6 +811,8 @@ def marketo_submit():
         value = ", ".join(values)
         if value:
             form_fields[key] = value
+            if 'utm_content' in form_fields:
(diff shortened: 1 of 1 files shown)


webapp/views.py
```diff
@@ -811,6 +811,8 @@ def marketo_submit():
         value = ", ".join(values)
         if value:
             form_fields[key] = value
+            if 'utm_content' in form_fields:
+                form_fields['utmcontent'] = form_fields.pop('utm_content')
```
carkod, line 815: Can you change it to double quotes (") for consistency? Thanks

laszlokajtar: @carkod changed, thank you for catching!

carkod (approved):

PR #13033, opened 2023-07-11
Change active footer accordion colour - small screens

Jim-Dillon:
## Done

Currently when a footer accordion header is clicked, it's a little tricky to distinguish between the accordion header, its second-level dropdown contents and the rest of the footer. I've suggested some separate improvements to the footer in [another PR](https://github.com/canonical/ubuntu.com/pull/13006) which could outdate my below screenshot, but those still have this same issue. 

Proposing an improvement for the footer on small screens:

- When a footer accordion header is clicked (active), its nested items change background colour (`$color-mid-x-light`). This grey is light enough for you to make out the divider between each item.
- When the header is clicked, it also increases font-weight to distinguish it from the other unclicked headers above it.

## QA

- Open demo and view Ubuntu.com footer on small screen (620px>). Open separate tab of live website for comparison. 
- In demo, open accordions to see changes made. Drag screen-size to see they're only made to small-screens.

## Screenshots

Before:
Screenshot 2023-07-11 at 10 20 56

After:
Screenshot 2023-07-11 at 10 21 11

Please feel free to provide any tweaks or feedback on this! Thanks :)

### DIFF
--- a/static/sass/_pattern_footer.scss
+++ b/static/sass/_pattern_footer.scss
@@ -216,7 +216,7 @@
   .p-footer__title.active {
     background-color: $colors--light-theme--background-hover;
 
-    a, 
+    a,
     span {
       &::before {
         transform: none;
--- a/static/sass/_pattern_footer.scss
+++ b/static/sass/_pattern_footer.scss
@@ -257,7 +257,7 @@
     display: block;
 
     @media only screen and (max-width: $breakpoint-small) {
-      background-color: #e5e5e5;
+      background-color: $color-mid-x-light;
     }    
   }
 
--- a/static/sass/_pattern_footer.scss
+++ b/static/sass/_pattern_footer.scss
@@ -245,7 +245,7 @@
 
   .p-footer__title.active {
     background-image: url("#{$assets-path}43e2b367-arrow_up_9fa097.png");
-    font-weight: 800;
+    background-color: $colors--light-theme--background-hover;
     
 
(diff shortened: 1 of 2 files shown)


anthonydillon:
@juanruitina @ClementChaumel @lyubomir-popov would you mind taking a look at this please?

juanruitina:
I would agree with reusing the accordion component. Is it possible to use it just on mobile, so design on desktop remains unchanged?

Jim-Dillon:
Hi all, I've made changes to the footer so that it looks and acts like the vf-accordion on small screens only. 

I initially tried to use the accordion HTML to create a second footer, which could then be hidden on larger screens. However, there were some issues as the dummy accordion content is static in the HTML file and the live site footer content is dynamically rendered from a separate file with its own style classes, which were difficult to override. I thought it could be simpler to just apply styling so the current footer looks like the accordion. 

Notes: 

- The example [accordion component ](https://vanillaframework.io/docs/examples/patterns/accordion/default) doesn't cover instances with multiple items inside an accordion. It does show the border removed between the header and first child though, so I've just extended that to the rest of the children. 
- There are 6 headers at the bottom of the footer without a dropdown icon. With the indent it looks a little odd but it might be best to leave all headers with level indent.

anthonydillon:
@akbarkz this has been sitting up address for months now. Can we assign someone to review this please?

akbarkz:
@anthonydillon with the new redesigned footer this PR is not relevant anymore

PR #13076, opened 2023-08-08
Change cost to use currect number of machines

albertkol:
### Problem 1: Cost column on Subscription Details
image
Inconsistencies that happen in the cost we display on right side, **Subscription Details**, **Cost column**. 

- In the image the user had **6** machines. And down-sized to **5**. The cost still shows **$3000** which is for **6** machines. 
- The total cost of the renewal reflects the down-sized cost, **$2500**, which is for **5** machines.
- The user was confused why the cost in the **Subscription Details** still showed **$3000**, unsure if the down-size was successful (despite the notification)
#### Note:
- The cost in the **Subscription Details** is **without VAT**. It does not say it, but it is without. Similar to the renewal settings.

### Problem 2: Resizing message
image
- Resizing message telling users how much they Yearly subscription is increasing / decreasing is bugged, as it's calculating incorrectly by how much it increases
- They user would be miss-informed about their subscription cost update
#### Note:
- the cost showing by how much they will be changed on up-sizing is correct. **It includes VAT.**
- when they downsize the message will say that they **are not charged** 

## Solution:
- Remove **Cost column** from **Subscription Details**
- Change resize message to inform the user correctly on their subscription cost update with a message letting them know that it's not including VAT
image

## QA

- Buy a subscription
- Downsize it
- Check that the Solution works

## Issue / Card

Fixes #https://warthogs.atlassian.net/browse/WD-2823


minkyngkm (approved):
LGTM! 👍 It looks so much better!

Code also looks good. One thing I wanted to mention though is `SubscriptionEdit.tsx` seems to contain so many things in one file. Might look better to separate `ResizeSummary` into a different file or something in the future?

albertkol:
> LGTM! +1 It looks so much better!
> 
> Code also looks good. One thing I wanted to mention though is `SubscriptionEdit.tsx` seems to contain so many things in one file. Might look better to separate `ResizeSummary` into a different file or something in the future?

@minkyngkm I was very confused for a few minutes while working.. because I kept trying to switch files between the `ResizeSummary` and `SubscriptionEdit` component. I restarted VSCode because I thought it's bugged. Then I noticed they are the same file. :rofl: You are absolutely right!

albertkol:
Merged this one, so we don't delay the bug fixes. If UX review comes, happy to address them in new PRs.

PR #13077, opened 2023-08-08
Improve not renewed label logic

albertkol:
## Done

- Some renewals are added to legacy purchases earlier than expected. Display the `Not renewed` label only if the renewal is also ready to be purchased : `statuses.is_renewable == true`

## QA

- Hard to QA, you need to View As user from a local environment. 
- But look, label "Not Renewed" disappears:
Prod: 
image
After the change:
image
 

## Issue / Card

Fixes #https://warthogs.atlassian.net/browse/WD-5368


minkyngkm (approved):
👍

PR #13108, opened 2023-08-24
Require completion of confidentiality agreement to take exams

jpmartinspt:
Fixes https://warthogs.atlassian.net/browse/WD-3763

## QA

### DIFF
--- a/webapp/shop/api/datastore.py
+++ b/webapp/shop/api/datastore.py
@@ -1,0 +1,36 @@
+import os
+from typing import Callable
+
+from google.cloud import datastore
+
+
+def get_datastore_client() -> datastore.Client:
+    return datastore.Client.from_service_account_info(
+        {
+            "token_uri": "https://oauth2.googleapis.com/token",
+            "project_id": os.getenv("GOOGLE_DATASTORE_PROJECT_ID"),
+            "private_key": os.getenv("GOOGLE_DATASTORE_PRIVATE_KEY").replace(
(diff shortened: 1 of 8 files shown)


webapp/shop/api/datastore.py
```diff
@@ -0,0 +1,36 @@
...
+                "\\n", "\n"
+            ),
+            "client_email": os.getenv("GOOGLE_DATASTORE_EMAIL"),
+        }
+    )
+
+
+def handle_confidentiality_agreement_submission(
+    email: str,
+    client_factory: Callable[[], datastore.Client] = get_datastore_client,
```
albertkol, line 22: I would have rather not provided the `client_factory` as a parameter, making the two functions also responsible for creating the `client`.

I would rather pass the `client` directly, even if it makes it a bit annoying to create the `client` in each view.

webapp/shop/cred/views.py
```diff
@@ -351,6 +373,13 @@ def cred_assessments(trueability_api, **_):
 
 @shop_decorator(area="cred", permission="user", response="html")
 def cred_exam(trueability_api, **_):
+    email = flask.session["openid"]["email"].lower()
+    if (
+        not flask.current_app.debug
```
albertkol, line 378: Would not have left `not flask.current_app.debug` after I finished developing/testing.

albertkol (approved):
:+1: It works.

I left a couple of comments, but they're just opinions.

PR #13171, opened 2023-09-20
Update the automotive buyers guide pdf link

anthonydillon:
## Done

- simply update the automotive buyers guide pdf link

## QA

- Open the demo
- Go to /automotive/buyers-guide
- full in the form and see it lands you on the PDF


huwshimi (approved):
Works as described.

PR #13186, opened 2023-09-29
Set show down in October 2nd

albertkol:
## Done

- Set shop to come down in Oct 2nd from 3am to 6am.

## QA

- Go to https://ubuntu-com-13186.demos.haus/pro/status you should see it scheduled.


andesol (approved):
:clap:

Issue #13093, opened 2023-08-17
Pressing "Enter" after filling out a field doesn't start the search


axinojolais:
Hi,

On https://ubuntu.com/security/cves, if I put some text in the `CVE ID or description contains` text input and press `Enter`, the search does not start. Instead, I need to click on the `Search` button.

I would expect pressing `Enter` would start the search.

Thanks

---
*Reported from: https://ubuntu.com/security/cves

Harsh0707005:
Please assign me #13093

juanruitina:
Duplicate of #12478

mtruj013:
Fixed in https://github.com/canonical/ubuntu.com/pull/13188, closing

PR #13188, opened 2023-09-30
Issue #13093 & #12478 resolved

Harsh0707005:
### UI Issue Resolved
Issue #13093 & #12478 have been resolved and search works on enter press

### DIFF
--- a/static/js/src/cve/cve.js
+++ b/static/js/src/cve/cve.js
@@ -64,10 +64,10 @@ function handleCveIdInput(value) {
 handleCveIdInput(searchInput.value);
 
 function handleSearchInput(event) {
-  if (event.key === "Enter"){
+  if (event.key === "Enter") {
     searchForm.submit();
-  }else{
-  handleCveIdInput(event.target.value);
+  } else {
+    handleCveIdInput(event.target.value);
   }
 }
 
--- a/static/js/src/cve/cve.js
+++ b/static/js/src/cve/cve.js
@@ -8,6 +8,7 @@ import {
 } from "./cve-search.js";
 
 const searchInput = document.querySelector("#q");
+const searchForm = document.querySelector("#searchForm");
 
 function handleCveIdInput(value) {
   const packageInput = document.querySelector("#package");
@@ -63,15 +64,14 @@ function handleCveIdInput(value) {
 handleCveIdInput(searchInput.value);
 
 function handleSearchInput(event) {
+  if (event.key === "Enter"){
+    searchForm.submit();
+  }else{
   handleCveIdInput(event.target.value);
+  }
 }
 
 searchInput.addEventListener("keyup", handleSearchInput);
-
-searchForm.addEventListener("submit", function (event) {
-  event.preventDefault(); // Prevent the default form submission
-  handleCveIdInput(searchInput.value);
-});
 
 attachEvents();
 handleButtons();
--- a/static/js/src/cve/cve.js
+++ b/static/js/src/cve/cve.js
@@ -68,6 +68,11 @@ function handleSearchInput(event) {
 
 searchInput.addEventListener("keyup", handleSearchInput);
 
+searchForm.addEventListener("submit", function (event) {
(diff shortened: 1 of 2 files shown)


Saurav101010:
@mtruj013 The PR seems to work well

mtruj013:
Hi @Harsh0707005, thanks for helping out with this! Unfortunately, this doesn't work as expected, I tried it on the [demo](https://ubuntu-com-13188.demos.haus/security/cves) and locally. Could you take another look? The JS lint error looks like it could be related

Harsh0707005:
I'll look into it

Harsh0707005:
@mtruj013 , I guess have figured it out there's a minor mistake, so to check it may you please guide me how to set it up locally, as i am encountering errors

mtruj013:
@Harsh0707005 sure thing, we have instructions for how to run the project locally [here](https://github.com/canonical/ubuntu.com#local-development). You'll need on install `dotrun` first if you haven't already, instructions for how to do that [here](https://github.com/canonical/dotrun#installation)

Harsh0707005:
@mtruj013 Thanks a lot for the help
I guess, there is some problem with canonicalwebteam.flask-base, as whenever it is tried to installed, it tries to install an older version of gevent which in turn conflicts and doesnt install canonicalwebteam.flask-base.
Can you check for the error, to verify it exists?
image_2023-10-10_221553952

mtruj013:
@Harsh0707005 sure, happy to help. Could you confirm what command you're getting this error from (eg, while trying to install `dotrun` following X command, or maybe `dotrun` installed correctly but you get this when you try to use the command to run the project)? It'd also be helpful if you could let me know which OS you're running. :)

Harsh0707005:
sure, i ran "dotrun serve" and it was executing well, but when trying to load the webpage, i saw that js files like cve.js and css files like style.css are not found (404), then checking around i realised the canonicalwebteam.flask-base is failing to install as it is trying to install and build wheels for older versions of gevent (21.12.0) and greenlet, which is failing
i am running it on linux.
image_2023-10-11_131317842

mtruj013:
@Harsh0707005 Ah ok, so two things to note here. `dotrun serve` only runs the python app so it's expected that the site will look broken (no styles/JS). The command you want is just `dotrun` as stated in the [read me](https://github.com/canonical/ubuntu.com#local-development), this will install all the needed dependencies. Usage documentation for all dotrun commands [here](https://github.com/canonical/dotrun#usage). 

The gevent error that you get after running `dotrun serve` is strange though, I'm not able to recreate it on my machine. I'm not sure if running just `dotrun` will also produce this issue for you, could you try it and confirm if it works or if you get the same error?

Harsh0707005:
@mtruj013 Thanks a lot for helping out, after trying to host it on VPS i realised some file and modules were broken while downloading locally on my device so had to clone and load the project again and it resolved the error. 

I have made the changes and committed, also I checked the project locally and it works well..

static/js/src/cve/cve.js
```diff
@@ -63,7 +64,11 @@ function handleCveIdInput(value) {
 handleCveIdInput(searchInput.value);
 
 function handleSearchInput(event) {
+  if (event.key === "Enter"){
+    searchForm.submit();
+  }else{
   handleCveIdInput(event.target.value);
```
mtruj013, line 70:
```suggestion
  if (event.key === "Enter") {
    searchForm.submit();
  } else {
    handleCveIdInput(event.target.value);
```

mtruj013: This will fix the linting error :)

mtruj013 (approved):
Thanks @Harsh0707005, works great! Pleas either commit the suggested code, or if you prefer, you can also run `dotrun format-prettier` locally on your machine and push the change to fix the [linting error](https://github.com/canonical/ubuntu.com/actions/runs/6491779556/job/17669785782?pr=13188). Once that's done this is good to go!

Harsh0707005:
@mtruj013 It has been committed. :)

PR #13189, opened 2023-10-02
[WD-5489] replace an image with HTML table on Ceph page

akbarkz:
## Done

- Replaced an image that showed a table with HTML table
- Created a SCSS pattern styles for HTML bordered table for future similar use cases (and to use SCSS variables from Vanilla)

## QA

- Naviage to `/ceph` page and scroll down to `Benefits of Ceph on Ubuntu` and test the table

## Issue / Card

Closes https://warthogs.atlassian.net/browse/WD-5489

## Help


lyubomir-popov:
Looks good @akbarkz. I would confine it to 6 columns:
image

Does the highlighting (the white cells) match what was required to be highlighted? If so, good to go.

akbarkz:
> Looks good @akbarkz. I would confine it to 6 columns: image
> 
> Does the highlighting (the white cells) match what was required to be highlighted? If so, good to go.

Confined to 6 columns @lyubomir-popov. And yes, the highlighted areas are accurate with the requirements.

juanruitina:
The contrast between white and gray backgrounds is too low, it will not be perceivable for some users and on some screens. One quick fix is using bold text only on the cells that are blue in the pdf [attached on the ticket](https://warthogs.atlassian.net/browse/WD-5489).

Also, to match the prominence there, i'd suggest to invert the use of white and gray: white is more prominent in this case given the background, so i'd set the cells that are blue in the PDF in white here, and the others in the background's light gray (or transparent).

One final question: why are we using divs instead of a HTML table? I'd assume a well constructed HTML table would be easier to parse by a screen reader than a CSS grid-based one.

lyubomir-popov:
@juanruitina what pdf? I don't see a link here.

Originally I had used black in the previous version - but I'm still not clear why some things are in bold, some in different background.

akbarkz:
> One final question: why are we using divs instead of a HTML table? I'd assume a well constructed HTML table (with colspan) would be easier to parse by a screen reader than a CSS grid-based one.

We discussed this with @lyubomir-popov and afaiu an HTML table needs to have a header section.

lyubomir-popov:
yes, the header row is semantically required.

juanruitina:
> @juanruitina Originally I had used black in the previous version - but I'm still not clear why some things are in bold, some on different background. What is most important and highlighted? How many levels of highlighting are needed?

Looking at [this](https://docs.google.com/drawings/d/1CI8lyfB13YsNSH6TaO8bPYagVuIBhSw5TlzydPr8v-k/edit):

Screenshot 2023-10-04 at 12 29 44

There's two levels: 1) the Ubuntu releases + Ceph version available with each Ubuntu release, 2) further Ceph upgrades for each Ubuntu release. I think your original one, where 1) was black and 2) transparent/light gray (perhaps without using any bold), does the trick, also contrast wise.

About using tables vs grid, given this is markup-related, I defer to @akbarkz.

juanruitina:
Elaborating a bit on the table vs div thing: the issue that i have with this is that, by not using a table because we don't have a heading, we are using something as vague as it gets semantically (divs). With divs, screen reader users won't have a hint of which Ceph versions are available on which Ubuntu release.

It's still tricky, because complex tables are complicated on screen readers apparently (see [this](https://www.accessibility-developer-guide.com/examples/tables/spanning-rows-cols/), users need to infer colspan fromd from column number). But I would rather  have semantics that are slightly off than getting rid of them altogether.

lyubomir-popov:
@juanruitina I've done very [little research](https://stackoverflow.com/questions/43995169/is-table-header-th-mandatory-for-table-if-i-provide-aria-label-to-the-cell-e) on this,

lyubomir-popov:
@juanruitina I don't want to fail an accessibilty requirement by omitting a mandatory element - see here for example https://stackoverflow.com/questions/43995169/is-table-header-th-mandatory-for-table-if-i-provide-aria-label-to-the-cell-e

maybe we can add th cells with column headers? Or if you have any other suggestion that solves both the problems you mention and fits the table requirements?

juanruitina:
I'd suggest to add a header _column_ that is visible only for screen readers in line with [this](https://www.w3.org/WAI/tutorials/tables/one-header/#table-with-header-cells-in-the-first-column-only). So: prepend a `th` cell to the first row that reads "Ceph releases", and another to the second row that reads "Ubuntu releases". These cells can be [hidden](https://vanillaframework.io/docs/utilities/off-screen) with the `u-off-screen` utility.

(Interestingly, in using divs we are failing the requirement to [use table markup to present tabular information](https://www.w3.org/WAI/WCAG21/Techniques/html/H51))

lyubomir-popov:
@juanruitina happy to leave it to you to decide, as long as it is semantically correct. I am still not convinced this is tabulaer data - if it were, we wouldn't need to be hiding heeader rows/columns. BUt if your research indicates a table is semantically ok, thne happy to go with that. Sorry @akbarkz for asking you to do this twice, we should have included @juanruitina in the implementation discussion to begin with...

akbarkz:
@juanruitina @lyubomir-popov I replaced css grid table with the HTML table and added hidden `th` headers and accessibility tags. Checked it with the screen reader as well, it looks good to me now. Please check.

lyubomir-popov:
Looks good to me, thanks @akbarkz

juanruitina:
We are still failing to differentiate the available version (light gray background) for each release and the further updates (white) with enough contrast. Removing `is-bold` would make that distinction more perceivable.

UX+1 this as to not block this, meaning is implied anyway.

anthonydillon (approved):
LGTM :+1:

PR #13202, opened 2023-10-04
increase timeouts in Cypress tests to fix the failing pipeline issue

akbarkz:
## Done

- Try different changes in Cypress settings to check if it helps with fixing failing CI pipelines

## QA

- Check the Cypress related CI pipelines

## Help

### DIFF
--- a/tests/cypress/integration/pro/subscriptions_spec.js
+++ b/tests/cypress/integration/pro/subscriptions_spec.js
@@ -1,5 +1,7 @@
 /// <reference types="cypress" />
 
+Cypress.config("defaultCommandTimeout", 10000);
+
 context("/pro/dashboard", () => {
   it("header links work correctly", () => {
     const links = [
@@ -12,6 +14,8 @@ context("/pro/dashboard", () => {
     ];
 
     cy.login();
+    cy.wait(2000);
+
     cy.visit("/pro/dashboard");
     cy.acceptCookiePolicy();
 
@@ -24,6 +28,8 @@ context("/pro/dashboard", () => {
 
   it("free personal token is at the bottom of the list", () => {
     cy.login();
+    cy.wait(2000);
+
     cy.visit("/pro/dashboard");
     cy.acceptCookiePolicy();
 
@@ -34,6 +40,8 @@ context("/pro/dashboard", () => {
 
   it.skip("sends a correct request when resizing a subscription", () => {
     cy.login();
+    cy.wait(2000);
+
     cy.visit("/pro");
     cy.acceptCookiePolicy();
 
@@ -69,6 +77,7 @@ context("/pro/dashboard", () => {
 
   it.skip("prevents a user from creating disallowed combinations of features", () => {
     cy.login();
+    cy.wait(2000);
 
     cy.visit("/pro");
     cy.acceptCookiePolicy();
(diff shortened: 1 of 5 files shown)


tests/cypress/integration/pro/subscriptions_spec.js
```diff
@@ -12,6 +14,8 @@ context("/pro/dashboard", () => {
     ];
 
     cy.login();
+    cy.wait(2000);
```
petermakowski, line 17: Using `cy.wait()` with arbitrary time values is generally not recommended and should be used as a last resort (not reliable, can lead to flaky tests).

Do you know why we need to wait here, are we waiting for a specific network request to complete or UI element to become visible?

We should be able to assert those instead, e.g. use `cy.intercept()` to wait for a specific network request.

Also, any such check should ideally go to the login command to reduce repetition, which you can safely modify.

tests/cypress/integration/pro/subscriptions_spec.js
```diff
@@ -1,5 +1,7 @@
 /// <reference types="cypress" />
 
+Cypress.config("defaultCommandTimeout", 10000);
```
petermakowski, line 3: [Cypress.config() changes the configuration for the remaining execution of the current spec file](https://docs.cypress.io/guides/references/configuration#Cypressconfig), but it's not immediately obvious (I actually had to refer to the documentation to double check this).

For that reason I'd recommend to avoid it. If you need to customise the timeout for a specific test/group of tests, it might be better to pass it as a configuration option, as explained in the docs. It is more self-explainatory as it's tied to a particular function.
 https://docs.cypress.io/guides/references/configuration#Syntax

petermakowski (commented):
Glad you managed to resolve this. I suggested a few tweaks to follow best practices and make this easier to maintain.

tests/cypress/integration/pro/subscriptions_spec.js
akbarkz: The reason I added this wait statement was because this was the only test that was failing right after the loging was happening. And I noticed that in other similar tests there is a wait statement after the login. But good point with adding a wait statement to login itself 👍

tests/cypress/integration/pro/subscriptions_spec.js
akbarkz: From the docs I see that there is nothing wrong with using this statement. And since it's used in all other test files in the project, it's consistent. We can add a comment on top of this statement to make it more clear what it does. Wdyt?

tests/cypress/integration/pro/subscriptions_spec.js
petermakowski: It might be useful to run cypress in the UI mode (`yarn cypress:open`)  where you'll be able to see what's actually happening as the runner goes through each step.
https://docs.cypress.io/guides/core-concepts/cypress-app

petermakowski: Some suggestions for alternatives to timeout wait: https://filiphric.com/waiting-in-cypress-and-how-to-avoid-it

petermakowski: Let's try and pin this down if we can and use alternatives. Otherwise, let's at least make sure `wait` it's only added in a single place (login util).

akbarkz: I removed all the `cy.wait()` statements after logins. Instead slightly increased the timeout in the `login` task.

petermakowski (approved):

PR #13233, opened 2023-10-11
inclusive identities

eslerm:
Please update Ubuntu's Disclosure Policy to be more inclusive.

Currently, the policy states that only "real names" will be credited for the discovery of vulnerabilities.

The term "real name" is ambiguous. It can be interpreted as meaning someones legal name. Requiring legal name can be used for discrimination. For this reason, [the Kernel recently changed their DCO requirements with](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d4563201f33a022fc0353033d9dfeb1606a88330): 
```
-using your real name (sorry, no pseudonyms or anonymous contributions.)
+using a known identity (sorry, no anonymous contributions.)
```

People around the world use Romanized identities on the internet. Individuals may associate *real name* to a *dead name*. There are many reasons to use another identity, including personal safety. We should be inclusive of a discoverer's requested identity.

As a recent example, the Security team participated as judges at ZDI's Pwn2Own 2023 event. When a vulnerability was discovered in Ubuntu we attributed the CVE to the discoverer's requested common name out of respect.

Please [join LXD](https://github.com/canonical/lxd/commit/4dd79c063e6b0d719b6acfa8a3ee08aea55c50a9) to make our policies more inclusive, by giving security research credit to a person's requested identity.

### DIFF
--- a/templates/security/disclosure-policy.html
+++ b/templates/security/disclosure-policy.html
@@ -15,6 +15,7 @@
       </h1>
       <p>
         <i>Valid since: October 2020</i>
+        <i>Last updated: October 2023</i>
       </p>
       <p>
         Canonical and the Ubuntu Security Team participate in responsible disclosure and collaborate with the wider community on security issues. This describes how to contact the Ubuntu Security Team, what you can expect when you contact us, and what we expect from you.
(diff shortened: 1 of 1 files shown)


pfsmorigo:
LGTM

jhenstridge:
This seems like a worthwhile change.

I do wonder how this change interacts with the "with no affiliations" clause of that sentence? When read in the context of the current language, it sounds like a desire to credit individuals rather than employers (presumably so that credit can follow them if they change jobs or want to put it on their CV). With the changed language, is it a problem if the reporter gives their company name as their "known identity"?

Maybe the answer is that the ambiguity is fine, and it can be handled on a case by case basis if it ever arises.

eslerm:
Thanks James!

The intent is not to change the "no affiliations" clause. The Security team does not attribute groups a researcher is associated with on our CVE pages or USNs.

As I understand the terms, known identity could be written as *established common name*.

setharnold (approved):
Looks good to me, thanks.

anthonydillon (approved):

eslerm:
Thank you so much everyone! ❤️

PR #13259, opened 2023-10-19
Update example.com wildcard in livelinks check

anthonydillon:
## Done

- Update the regex wildcard in livelink workflow to work with all forms of example.com links 

## QA

- Go to https://regex101.com/
- Enter `http://sunbeam01.example.com` into the test string
- Use `http://*.example.com` as the regular expression and see that it does not match and in fact fails
- Enter `http:\/\/[a-z0-9]*\.example\.com` as the regex and see it matches the sunbeam URL

### DIFF
--- a/.github/workflows/live-links.yaml
+++ b/.github/workflows/live-links.yaml
@@ -39,7 +39,7 @@ jobs:
             http://www.adlinktech.com/en/index
             /flags.png
             /flags@2x.png
-            http://*.example.com
+            http:\/\/[a-z0-9]*\.example\.com
 
           [output]
           status=0


.github/workflows/live-links.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
             http://www.adlinktech.com/en/index
             /flags.png
             /flags@2x.png
-            http://*.example.com
+            http:\/\/[a-z0-9]*\.example\.com
```
bartaz, line 42: Do we care if it matches https?

.github/workflows/live-links.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
             http://www.adlinktech.com/en/index
```
bartaz, line 39: There seem to be other plain text URLs on the ignore list, that are also not valid regexp - do they work just fine?

.github/workflows/live-links.yaml
aaryanporwal, line 42: should it be case insensitive as well? 
```suggestion
            http:\/\/[a-z0-9]*\.example\.com/i
```

anthonydillon: We don't have an example of a secure link but could add it

.github/workflows/live-links.yaml
anthonydillon: Yeah, this is to specify ignore that URL but in this case, we have sunbeam01, sunbeam02 and sunbeam03 but all example.com URLs should be ignored.

.github/workflows/live-links.yaml
anthonydillon: Could be I guess

.github/workflows/live-links.yaml
bartaz: OK. I guess these full URLs work a bit by accident, because they are valid Python regexp. I guess only `.` in these can is interpreted as any char, not as a `.`.  But for the purpose of ignoring it should be fine.

anthonydillon:
Removed the `/i`

bartaz (approved):
LGTM

aaryanporwal (approved):

PR #13361, opened 2023-11-22
Fix notice prefix error

mtruj013:
## Done

- Fixed false 500 when user tries to visit a notice with an incomplete id which matches an existing template name

## Rationale

Notices can have one of two prefixes: usn or lsn. When you visit `/security/notices` and click on any given notice in the list, the template that gets triggered is `/notices/lsn.html` or `/notices/usn.html`, respectively. This bug came about because we had some users try to visit `/security/notices/usn` which, because we actually do have a template with that name and we did not have a rule in place for it, was causing jinja errors. This PR changes the behavior so that when a user tries to visit `/security/notices/<prefix>` it correctly shows 404. 

## QA

- View the site locally in your web browser at: https://ubuntu-com-13361.demos.haus/security/notices/lsn
  - See that you get 404
- View the site locally in your web browser at: https://ubuntu-com-13361.demos.haus/security/notices/usn
  - See that you get 404

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-6886

### DIFF
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -28,6 +28,7 @@ from webapp.security.views import (
     cve_index,
     cves_sitemap,
     notice,
+    notice_redirect,
     notices,
     notices_feed,
     notices_sitemap,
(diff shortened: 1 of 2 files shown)


webapp/app.py
```diff
@@ -451,6 +451,11 @@
 # usn section
 app.add_url_rule("/security/notices", view_func=notices)
 
+app.add_url_rule(
+    "/security/notices/<notice_id>",
```
carkod, line 455: Doesn't this id need some kind of regex filtering?

carkod:
Getting 404 with QAs, it's not redirecting anywhere?

mtruj013:
@carkod reaching 404 is what we want. I explained this in the rationale, but I've also updated the qa instructions

carkod (approved):

PR #13381, opened 2023-11-29
Pass along UTMs when doing a blog redirect

caldav:
## Done

- Updated the blog redirection logic to append URL params.

## QA

- Go to https://ubuntu-com-13381.demos.haus/blog/chiselled-ubuntu-ga?utm_source=marketo&utm_medium=email&utm_campaign=newsletter-2023-48
- Ensure you are redirected to the Canonical blog
- Ensure UTMs from the original URL are still there

### DIFF
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -736,10 +736,17 @@ class BlogRedirects(BlogView):
         if "article" not in context:
             return flask.abort(404)
 
-        # Redirect canonical annoucements
+        # Capture the original query string
+        original_query_string = flask.request.query_string.decode("utf-8")
+
+        # Redirect canonical announcements
         group = context["article"].get("group")
         if isinstance(group, dict) and group["id"] == 2100:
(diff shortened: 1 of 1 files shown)


caldav:
Somehow, this doesn't work in the demo, but it works locally...

lorumic (approved):
I confirm that it works locally for me as well. Not sure why it fails on the demo server though. Let's maybe wait for a second reviewer from sites, who might know which could be the reason.

webapp/views.py
```diff
@@ -736,10 +736,17 @@ def dispatch_request(self, slug):
         if "article" not in context:
             return flask.abort(404)
 
-        # Redirect canonical annoucements
+        # Capture the original query string
+        original_query_string = flask.request.query_string.decode("utf-8")
```
akbarkz, line 740: if this variable is used only inside the `if` block below, can you declare it there?

PR #13425, opened 2024-01-03
Update release links

mtruj013:
## Done

- Added links to filtered releases

## QA

- View the site locally in your web browser at: https://ubuntu-com-13425.demos.haus/security/cves
- Click each "By Ubuntu release" link and see that each shows the expected filtered table (styling updates to this list to come in a later PR)
 Note: "Other releases" should just be the unfiltered CVE list, the table should match the current version with exception of 24.04

## Issue / Card

Fixes #https://warthogs.atlassian.net/browse/WD-7802

### DIFF
--- a/static/js/src/cve/cve.js
+++ b/static/js/src/cve/cve.js
@@ -153,11 +153,11 @@ tooltipIconList.forEach(function (tooltipIcon) {
   );
 });
 
-function toggleCVEList() {
+function showCVEList() {
   if (urlParams.has("status")) {
     recentCves.classList.add("u-hide");
     cveList.classList.remove("u-hide");
   }
 }
 
-toggleCVEList();
+showCVEList();
--- a/static/js/src/cve/cve.js
+++ b/static/js/src/cve/cve.js
@@ -157,7 +157,7 @@ function toggleCVEList() {
   if (urlParams.has("status")) {
     recentCves.classList.add("u-hide");
     cveList.classList.remove("u-hide");
-  } 
+  }
 }
 
 toggleCVEList();
--- a/static/js/src/cve/cve.js
+++ b/static/js/src/cve/cve.js
@@ -9,6 +9,10 @@ import {
 
 const searchInput = document.querySelector("#q");
 const searchForm = document.querySelector("#searchForm");
+const cveList = document.querySelector("#cve-list");
+const recentCves = document.querySelector("#recent-cves");
+const url = new URL(window.location.href);
+const urlParams = new URLSearchParams(url.search);
 
 function handleCveIdInput(value) {
   const packageInput = document.querySelector("#package");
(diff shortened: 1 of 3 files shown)


static/js/src/cve/cve.js
```diff
@@ -148,3 +152,12 @@ tooltipIconList.forEach(function (tooltipIcon) {
     false
   );
 });
+
+function toggleCVEList() {
```
akbarkz, line 156: To me the word toggle means switching from one state to another and back. In this case though there is only one switch in case there is a parameter `status` present. Should we still be calling this function `toggle` or will the actual switching back come later?

webapp/security/views.py
```diff
@@ -425,6 +432,8 @@ def cve_index():
         versions=versions,
         statuses=statuses,
         selected_releases=selected_releases,
+        lts_releases=lts_releases,
+        maintained_releases=maintained_releases,
```
akbarkz, line 436: these two new variables are always empty arrays, what are they for?

templates/security/cve/index.html
```diff
@@ -68,21 +68,22 @@ <h3 class="p-heading--5">By Ubuntu release</h3>
...
         <div class="col-7">
           <ul class="p-inline-list" style="margin-top: .3rem;">
-            <li class="p-inline-list__item"><a>All maintained releases</a></li>
-            <li class="p-inline-list__item"><a>All LTS</a></li>
-            <li class="p-inline-list__item"><a>22.04 LTS</a></li>
-            <li class="p-inline-list__item"><a>20.04 LTS</a></li>
-            <li class="p-inline-list__item"><a>18.04 LTS</a></li>
-            <li class="p-inline-list__item"><a>23.10</a></li>
-            <li class="p-inline-list__item"><a>Other releases</a></li>
+            <!-- Once the search results page is finished and the cve list is not tied to the current table cols, selected releases, maintained releases, and lts releases will be accounted for separately -->
```
akbarkz, line 71: Sorry, I couldn't understand this comment. Can you please add more details here?

static/js/src/cve/cve.js
mtruj013: I added this mostly for testing as the cve cards only exist for recent cves in this current branch (ie, I still needed to use the table to show filtered releases), but yes I think you're right about toggle maybe not being the best verb here. I'll update!

templates/security/cve/index.html
mtruj013: This and the above comment are related, currently the cve table loops through `selected_releases` to decide which releases are going to be column headers, but in the new design for these links we really need `maintained_releases` (which is basically the same as `selected_releases`) and `lts_releases`. I didn't think it was worth it to spend time changing the way the current table works just to test that these links are correct, given that that table isn't going to be used anymore, so I added the comment explaining why lts release links are hard coded but maintained are not

PR #13439, opened 2024-01-05
Limit width of list items

carkod:
## Done

TL:DR;
Make paragraphs full width. Following [this discussion](https://chat.canonical.com/canonical/pl/77oyc3561ff5um3gjuiksgyash)

## Reasoning/Argument behind this task
We already have the option to take [full width in Vanilla](https://vanillaframework.io/docs/base/typography#line-length) so design-wise this is ok to do, also the lists already take the full width, so there is no point keeping the 40em default width for paragraphs only, both stakeholders and I agree that it looks like a lot of useless space on the right, when there is no form. But in the case of engage pages, because the paragraphs are generated in Discourse, I can't add the classname to every single `p`. So I extended the same classname for wide use in engage pages. I'm all ears if you can think of a better option.

## QA

- Check an engage page that doesn't have a form on the right side e.g. https://ubuntu-com-13439.demos.haus/engage/simplifying-mongodb-operations
- The text should expand full width.

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-7419

## Screenshots

Screenshot 2024-01-05 at 16 58 03

## Help

### DIFF
--- a/templates/engage/base.html
+++ b/templates/engage/base.html
@@ -86,7 +86,7 @@
   {% endif %}
 {% endif %}
 
-<section class="p-strip is-shallow is-bordered">
+<section class="p-strip is-shallow is-bordered u-text-max-width">
   <div class="row">
     <div class="{% if 'form_id' in metadata and metadata['form_id'] != '' %}col-7{% else %}col-12{% endif %}">
       {{ metadata["body_html"] | safe }}


bartaz:
I think the opposite should be applied - limit lists to text with, not making paraghaphs full width.

The Vanilla documentation states that removing max width is OK only in specific cases:

image


Please use `u-text-max-width` on lists (or parent container) to make lists appear in the width of the paragraphs.
https://vanillaframework.io/docs/utilities/text-max-width

lyubomir-popov:
This is not ok from a design perspective - longer paragraphs are harder to read. I have to put a design -1 on this, sorry!

lyubomir-popov:
@bartaz beat me to it, thanks @bartaz!

lyubomir-popov:
In terms of how to best addres the feeling of unused space - why don't you apply the 50/50 split, title on the left, everything else on the right @carkod?

bartaz:
> In terms of how to best addres the feeling of unused space - why don't you apply the 50/50 split, title on the left, everything else on the right @carkod?

I believe content comes from discourse (it's an engage page), so we have little control over layout here.

carkod:
> > In terms of how to best addres the feeling of unused space - why don't you apply the 50/50 split, title on the left, everything else on the right @carkod?
> 
> I believe content comes from discourse (it's an engage page), so we have little control over layout here.

Yes, we don't have much control over the formatting of the headings, and the engage page authors will try to change the system to try to fill up that empty space on the left...

carkod:
> I think the opposite should be applied - limit lists to text with, not making paraghaphs full width.
> 
> The Vanilla documentation states that removing max width is OK only in specific cases:
> 
> image
> Please use `u-text-max-width` on lists (or parent container) to make lists appear in the width of the paragraphs. https://vanillaframework.io/docs/utilities/text-max-width

Do you consider this something that could be upstreamed to Vanilla?

bartaz:
> Do you consider this something that could be upstreamed to Vanilla?

@carkod I'm not sure what do you mean to be upstreamed. The `u-text-max-width` I mentioned is in Vanilla (https://vanillaframework.io/docs/utilities/text-max-width). You can put it on lists or on parent container (in this case that would be needed) to make all text elements (including lists) to keep same max width.

This is used in docs, like here: https://microk8s.io/docs/release-notes

carkod:
> > Do you consider this something that could be upstreamed to Vanilla?
> 
> @carkod I'm not sure what do you mean to be upstreamed. The `u-text-max-width` I mentioned is in Vanilla (https://vanillaframework.io/docs/utilities/text-max-width). You can put it on lists or on parent container (in this case that would be needed) to make all text elements (including lists) to keep same max width.
> 
> This is used in docs, like here: https://microk8s.io/docs/release-notes

@bartaz I mean, if `p`s are by default max-width 40em, why are lists not like that by default?

bartaz:
> @bartaz I mean, if `p`s are by default max-width 40em, why are lists not like that by default?

That would be ideal, unfortunately lists (ul/ol/li) are used for semantics in many other contexts: navigations, tabs, stepped lists, etc. It would be a hugely breaking change to limit their max width by default. That's why we had to introduce this utility that allows to target the part of the page where text lists are used.

bartaz (approved):
LGTM, thanks!

PR #13494, opened 2024-01-23
CVE search results filters

mtruj013:
## Done

- Added filtering functionality

### Affected packages
Will be handled in separate PR.   

### Status filter
There is a bug when trying to filter with an uneven number of releases and status. This comes from the security API itself, it groups a single version and a single status together and throws 500 if it can't make a pair. The fix needs to be implemented on that codebase first and will be handled in a separate PR. 

### No results view
Currently not implemented, will be taken care of in separate ticket. 

### Show maintained releases link
This link (found under the releases filter) is currently not implemented. Waiting on a design to explain the correct functionality for this and will implement this in a separate PR.

## QA

- View the site locally in your web browser at: https://ubuntu-com-13494.demos.haus/security/cves
- Check that each release link routes to the search results with the correct filters applies
- Click the clear all filters button and see that it works as expected
- Play with the filters (please keep in mind the status filter bug) and see that there are no abnormalities. It might help to cross reference the total results for any given query against the live site. 

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-8029


juanruitina:
This is a partial review.

- [ ] An empty query (https://ubuntu-com-13494.demos.haus/security/cves?q=) should show the list with all results, not the home page.
- [ ] Can we make the list of releases programmatic? I just noticed the designs still featured first 22.10 kinetic (it's mantic now, just updated on designs), and probably this shouldn't be hardcoded, given we have [releases.json](https://ubuntu.com/security/releases.json)
- [ ] The "All maintained releases" checkbox should mark all checkboxes for currently maintained releases according to the YAML file. That can include interim releases. (That's also how the "All maintained releases" link in the home should work). I suggest that we do all this programatically, if we don't yet.
- [ ] Chrevrons don't toggle the filters yet.
- [ ] https://ubuntu-com-13494.demos.haus/security/cves?priority=high&priority=critical seems to show results only for the first priority parameter. Couldn't verify if this behaviour applies to other filters too.
- [ ] Please make the release codenames ("mantic") gray as per the design

PR #13503, opened 2024-01-26
Fix IoT and Server SoC filter

carkod:
## QA

- 

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-8457


webapp/certified/views.py
```diff
@@ -714,9 +713,8 @@ def create_category_views(category, template_path):
     """
     if len(request.args.getlist("category")) > 1:
         url = f"/certified?{request.query_string.decode()}&category={category}"
-        clean_url = bleach.clean(url, tags=[], strip=True)
```
carkod, line 717: This additional clean up is useless, as we are already matching every category.

webapp/certified/views.py
```diff
@@ -102,12 +101,12 @@ def _parse_query_params(all_releases, all_vendors):
             "servers",
             "iot",
             "socs",
```
akbarkz, line 103: If you add a new `Server SoC` category, aren't existing ones (`socs` and `SoC`) become redundant?

webapp/certified/views.py
```diff
@@ -102,12 +101,12 @@ def _parse_query_params(all_releases, all_vendors):
...
+            "Ubuntu Core",
```
akbarkz, line 104: and same here, aren't `iot` and `IoT` redundant?

webapp/certified/views.py
carkod: No, because the `socs` are coming from our JS, and the `Server SoC` are from the Python code because the API only accepts `Server SoC`, it doesn't understand what is `socs`. So this is something I missed in the security PR.

webapp/certified/views.py
carkod: Same answer as previous comment. We could additionally just do lowercase and have only `iot` I guess, but we need to accomodate for both due to UX vs Certification API reuirements.

carkod: ok, looks like we can't do that either, because it messes up with Server SoC, you can't match server soc with Server SoC, because if you make the initial letter uppercase it's not enough. It's mainly because of the UX requirement this needs to be passed this way.

akbarkz (approved):

PR #13636, opened 2024-03-01
Switched from ip to tz for geolocation

samhotep:
## Done

- Switched from using ip address for geolocation to using timezone

## QA

- Open https://ubuntu-com-13636.demos.haus/contact-us/form?product=generic-contact-us 
  - Check that the phone number input shows the correct country flag for your location.  
  
## Issue / Card

Fixes [WD-1173](https://warthogs.atlassian.net/browse/WD-1173)

## Screenshots

image

## Help

## Follow up
Should we also auto-fill the country input? 

[WD-1173]: https://warthogs.atlassian.net/browse/WD-1173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webapp/views.py
```diff
@@ -1,9 +1,10 @@
 # Standard library
+import datetime
+import html
```
carkod, line 3: This doesn't seem to be used anywhere?

samhotep: Removing!

carkod: Sorry, it's not used by your code, but it seems we are using it here https://github.com/canonical/ubuntu.com/pull/13636/files#diff-eb0c47c22a26e963ac57d9840cff927d3d40d2f0889900c17e6ed717459a2db6L911

This file is becoming too big...

carkod (changes requested):
This works for your example with telephone numbers, but because /user-country.json needs the query arg, this is breaking takeovers geolocation https://github.com/canonical/ubuntu.com/blob/8ac389f3332ed038af3a91173ee5990835534b7d/templates/base_index.html#L1417

You may need to do the `Intl.DateTimeFormat().resolvedOptions().timeZone;` for takeovers too

webapp/app.py
```diff
@@ -628,7 +628,7 @@ def takeovers_index():
 core_services_guide.init_app(app)
 
 
-app.add_url_rule("/user-country.json", view_func=get_user_country_by_ip)
+app.add_url_rule("/user-country.json", view_func=get_user_country_by_tz)
```
carkod, line 631: If you are replacing this completely, then you might as well delete `get_user_country_by_ip` function

webapp/views.py
```diff
@@ -1071,6 +1072,27 @@ def thank_you():
     )
 
 
+def get_user_country_by_tz():
```
carkod, line 1075: Can you also write a comment here for this function explaining how it works? I guess the key of this system is to use `Intl.DateTimeFormat().resolvedOptions().timeZone;`

carkod:
Also can you remove his `Keyerror`, this was put there to check which IPs are not detected, which is not needed anymore with this PR.

webapp/app.py
samhotep: Ok, but will do separately when updating the takeovers as well

carkod (approved):
LGTM, thanks.

PR #13637, opened 2024-03-01
K8s-docs 1.29-updates

evilnick:
## Done

- Added 1.29 section
- Updated release-notes
- Updated upgrading
- Added install method for existing machines
- Updated LXD instructions

## QA


kwmonroe:
@evilnick folks are starting to use 1.29 with the old 1.28 docs :(  any heat we can put on this?  the woke stuff is coming because we link to branch names upstream that are named `master`.

Do we need an exception for those?

evilnick:
@kwmonroe The woke check always fails - there is a way to disable it inline, but comically this then causes the markdown not to render

akbarkz (approved):

PR #13652, opened 2024-03-07
add an instruction to robots.txt to exclude docs search page

akbarkz:
## Done

- Added an instruction to robots.txt to exclude /server/docs/search* page since it's dynamic and doesn't need to be indexed by Google.

## QA

- Can be QA-ed after deployment by checking logs in Google Search Console. Currently there are a lot of 5xx and 403 errors coming from the page as Google is trying to index all the different search queries.

## Help

### DIFF
--- a/static/files/robots.txt
+++ b/static/files/robots.txt
@@ -1,4 +1,5 @@
 User-Agent: *
 Disallow: /search
 Disallow: /search*
+Disallow: /server/docs/search*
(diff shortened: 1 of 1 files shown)


lorumic (approved):

carkod:
We may need to do this to every search endpoint if what was there `Disallow: /search/*` wasn't working. But I think we could use some kind of wildcard i.e. `*/docs/search/*` (I made it up may need to be tested).

PR #13715, opened 2024-04-03
WD-9806: Send X-Forwarded-For to UA-Contracts

jpmartinspt:
## Done

- Send X-Forwarded-For in UA-Contracts API requests.

## QA

- Run this branch
- Go to `/pro/dashboard`
- Check the logs and see that after a UA-Contracts request your machine's IP is logged.

## Issue / Card

Fixes #WD-9806

### DIFF
--- a/webapp/shop/decorators.py
+++ b/webapp/shop/decorators.py
@@ -1,11 +1,12 @@
-from distutils.util import strtobool
 import os
+
+from distutils.util import strtobool
 from functools import wraps
 from datetime import datetime
 from dateutil.parser import parse
-import pytz
 
 import flask
+import pytz
 import talisker.requests
 
 from webapp.shop.api.ua_contracts.api import UAContractsAPI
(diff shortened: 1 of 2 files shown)


minkyngkm (approved):
LGTM. Every UA-Contracts request, remote address is logged.

webapp/shop/decorators.py
```diff
@@ -117,7 +118,7 @@ def decorated_function(*args, **kwargs):
                     return flask.redirect(f"/login?next={redirect_path}")
 
             ua_contracts_api = get_ua_contracts_api_instance(
-                user_token, response, session
+                user_token, response, session, flask.request.remote_addr
```
alnvdl-work, line 121: You may want to look at a different field instead: https://stackoverflow.com/questions/12770950/flask-request-remote-addr-is-wrong-on-webfaction-and-not-showing-real-user-ip

But it will all depend on how the content-cache/K8s ingress will be configured in front of it.

PR #13718, opened 2024-04-08
Upgrade vanilla and global nav versions

mtruj013:
## Done

- Upgrade Vanilla version to 4.9.1
- Upgrade global nav to 3.6.4

## QA

- View the site locally in your web browser at: https://ubuntu-com-13718.demos.haus/
- Check as many pages as possible and see that nothing looks broken

## Issue / Card

Fixes #

## Screenshots

## Help


static/sass/_pattern_navigation.scss
```diff
@@ -57,7 +57,12 @@ $meganav-height: 3rem;
     }
 
     .p-link--inverted {
+      color: $color-x-light;
```
mtruj013, line 60: I added this as AFAIK the design for things like the meganav hasn't changed, the links should not be blue as they are when using the is-dark class. If they are indeed supposed to be blue now, please let me know and I'll remove

static/sass/styles.scss
```diff
@@ -1588,3 +1588,11 @@ ol.p-list--divided .p-list__item::before {
     }
   }
 }
+
+.p-link--inverted {
```
mtruj013, line 1592: I added this because using the is-dark class as the docs suggest results in blue links on the meganav and AFAIK these should still be white. I'll remove if that's not the case anymore

bartaz:
Demo seems to be up: https://ubuntu-com-13718.demos.haus/

static/sass/styles.scss
bartaz: As far as I can see the meganav is not a child of `p-navigation`. So the `is-dark` (dark theme) of navigation has no effect on meganav.

I believe the proper fix would be to set `is-dark` on parent of the meganav (`dropdown-window`?)

image

mtruj013: That seems to have done the trick!

mtruj013:
Thanks @bartaz, would you mind taking another look?

static/sass/_pattern_buttons.scss
```diff
@@ -5,7 +5,7 @@
   }
 
   .p-button--tag {
-    @extend %p-button-light;
+    @extend %p-button-light !optional;
```
bartaz, line 8: Not sure how much it is in scope of this PR.

What is this supposed to extend? It doesn't seem like `p-button-light` exists anywhere (not in Vanilla or in ubuntu.com).

As far as I can see this `p-button--tag` is used only on security pages.

image

That almost looks like a broken button style, as it doesn't look like button at all.
Is this intended? Would it be fine to just change this to standard p-button, or p-button--base?
This would allow to get rid of this (possibly broken) custom styling.

bartaz (approved):
LGTM.

I commented on a button issue, but it's likely not related to this PR and can be addressed separately.

static/sass/_pattern_buttons.scss
mtruj013: It was throwing errors without the `!optional` bit whenever I ran the project locally, it was just a quick fix for that. But I'm not sure why that class was made in the first place tbh. I'll make a follow up pr

PR #13849, opened 2024-05-08
Update to Vanilla 4.11 (and fix equal heights component)

bartaz:
## Done

- Updates to latest Vanilla 4.11
- Updates the divider class name in equal heights row component from `has-1st-divider` to `has-divider-1` (as per change in Vanilla 4.11)
- As a drive-by updates the HTML structure of the equal heights row component

**NOTE:** The reason for the drive-by HTML changes was combining multiple component class names on single element. This is a bad practice. Even if it works now, may potentially break in future if both components affect the same CSS properties. As a rule of thumb a single HTML element should not have class names coming from 2 different components. Ideally there should be only one `p-…` class name on element, if there are multiple, they should be from same component `p-the-same-…`.

## QA

- Go to desktop page: https://ubuntu-com-13849.demos.haus/desktop#:~:text=Enterprise%2Dgrade%20support%20and%20integration (Enterprise grade support)
- Check if the equal heights row component looks as expected

image

## Help

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "typescript": "4.5.4",
     "url-polyfill": "1.1.12",
     "url-search-params-polyfill": "8.1.1",
-    "vanilla-framework": "4.10.0",
+    "vanilla-framework": "4.11.0",
     "yup": "0.32.11"
   },
   "resolutions": {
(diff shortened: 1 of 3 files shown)


mtruj013 (approved):
LGTM

PR #13903, opened 2024-05-30
[Fix] Handle deprecated products on checkout

pandrey2003:
## Description

When a customer tries to renew a subscription for now a deprecated product, the `ubuntu.com` website gets a similar error from the Ubuntu Pro backend: `missing one-off product listing for renewal product: "uaia-essential-virtual": product listing not found` The current front-end code does not handle this error message correctly, so the opaque `Sorry, there was an unknown error with your credit card` error message is shown on UI. This PR handles the case of a missing offering correctly and proposes the customer choose another product.

## TODO / considerations

It is too late to show such an error at checkout. I believe some validation should be added before going to checkout to ensure that the product is among `ValidProducts`. However, even despite this consideration, this PR is still relevant to handle integration errors on all levels.

### DIFF
--- a/static/js/src/advantage/subscribe/checkout/components/BuyButton/BuyButton.tsx
+++ b/static/js/src/advantage/subscribe/checkout/components/BuyButton/BuyButton.tsx
@@ -198,6 +198,18 @@ const BuyButton = ({ setError, quantity, product, action }: Props) => {
           problem persists.
         </>
       );
+    } else if (
+      error.message.includes(
+        "missing one-off product listing for renewal product"
+      )
+    ) {
+      setError(
+        <>
+          The chosen product cannot be renewed as it has been deprecated.
+          Contact <a href="https://ubuntu.com/contact-us">Canonical sales</a>
+          to choose a substitute offering.
+        </>
+      );
     } else {
       const knownErrorMessage = getErrorMessage({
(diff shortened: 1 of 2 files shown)


pandrey2003:
@minkyngkm / @abhigyanghosh30, I'd appreciate your reviews on this PR. Wanted to flex some front-end skills not to overwhelm my colleagues with such a simple change :wink: Thank you!

abhigyanghosh30:
@pandrey2003 Could you provide details of how we could QA this stuff?

minkyngkm:
@abhigyanghosh30 This PR is to handle this error in Sentry: https://sentry.is.canonical.com/canonical/ubuntu-com/issues/66482/?query=checkout (Andrew doesn't seem to have access to sentry so I'm adding this link for more reference)

I don't think we can reproduce the error again, but since sentry provides the error message, we can use that to handle the error.

minkyngkm (dismissed):
LGTM 👍

minkyngkm:
I guess in `Summary.tsx` the preview api (which causes the error) is also used. Could we try applying this fix in `/static/js/src/advantage/subscribe/checkout/components/Summary/Summary.tsx` as well?

minkyngkm (approved):
👍

PR #13990, opened 2024-06-25
WD-11695 - redirect users to different takeovers

lizzochek:
## Done

Added switching between 2 takeover types for different user groups. PUT SWITCH TO FALSE BEFORE MERGE

## QA

- [demo link](https://ubuntu-com-13990.demos.haus/)
- View the site on demo
- Run through the following [QA steps](https://discourse.canonical.com/t/qa-steps/152)
- Find control_or_variant cookie, delete it a couple times to see takeover change between options:
(may be different content, mind layout only)
Screenshot 2024-06-27 at 11 55 42

Screenshot 2024-06-27 at 10 40 23

- Check that the takeover changes correctly according to the user type

## Issue / Card
[WD-11695](https://warthogs.atlassian.net/browse/WD-11695)

[WD-11695]: https://warthogs.atlassian.net/browse/WD-11695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


petesfrench:
@lizzochek Can you add some more details to the QA steps please

petesfrench:
It isn't switching for me based on the cookie:
image
image

lizzochek:
currently on demo it does change correctly for me. Please make sure to reload the page. And the current takeover style is. also in black - maybe that is also confusing. The new one is ONLY 'A CTO's guide...', all other ones are from current page 
Screenshot 2024-06-27 at 11 49 55
Screenshot 2024-06-27 at 11 50 17

templates/base_index.html
```diff
@@ -22,49 +22,54 @@
...
-    <!-- Default Takeover: Download the latest Ubuntu -->
-    <section data-lang="all" id="takeover" class="p-section--hero">
-      <div class="row p-takeover-animation" id="takeover-animaition">
-        <div class="col-6">
-          <div class="p-section--shallow">
-            <h1 id="takeover-title" class="u-no-margin--bottom">What's new in Ubuntu {{ releases.latest.short_version }}?</h1>
-            <p class="p-heading--2" id="takeover-subtitle">
-              Introducing the Ubuntu Desktop for Raspberry Pi, the latest desktop features and micro clouds.
-            </p>
+    {% if request.cookies.get("control_or_variant") == 'variant' %}
```
petesfrench, line 28: Just a thought, but as logic is server side, if it is a user's first visit to the site they wont be in either group when the takeovers are rendered

lizzochek: as far as I can see from the code and from my trial - they will as the script for setting a cookie is run on loading the page

petesfrench: image

petesfrench:
Hey, the problem is still persisting, here we have a 'control' takeover, while the cookie is set to 'variant'. The reason for this is because the check for the cookie type is done server-side, before the JS has run. You need to move the takeover rendering into some JS that runs after the JS responsible for assigning the control/variant cookie.
image

akbarkz:
@lizzochek Why when the cookie value is set to "variant" I'm only seeing one takeover. It is supposed to show different takeovers but with a different layout.

lizzochek:
@akbarkz as far as I understood from my discussion, for the test template we are not fetching the data from discourse, therefore there is one takeover. And please don't remove QA -1, I have just fixed the issue and it needs to be QAd again

templates/base_index.html
```diff
@@ -1104,7 +1127,41 @@ <h2 class="p-heading--1 u-no-margin--bottom">Multi-cloud applications&nbsp;&ndas
...
-    <script>
+    <script type="module">
+      const testTakeover = document.getElementById('test-takeover');
+      const mainTakeover = document.getElementById('takeover');
+
+      const getCookie = () => document.cookie.match(new RegExp('(^| )' + "control_or_variant" + '=([^;]+)'));
+      // Switch for turning takeover switching on and off on demand
+      const takeoverSwitch = true;
+      // check if user doesn't already have a group
+      if (!getCookie()) {
```
akbarkz, line 1138: I have a question. From looking at this condition it seems like the "takeoverSwitch" will only work for people who don't have the "control_or_variant" cookie. What about those who have the cookie and we at some point we decide we do not want to split people into two different groups and show them different layouts?

akbarkz (approved):

PR #13998, opened 2024-06-26
GROWENG-3933 refactor dataLayer.push for GA4 structure

laszlokajtar:
## Done

- Refactored sendContributionFormAnalytics function not to send multiple GAEvent pushes to the dataLayer but to send one push with the data as key-value pairs within that push in order to respect GA4 event parameter data model

## QA

- Visit https://ubuntu.com/download/desktop/thank-you?version=24.04&architecture=amd64&lts=true
- Click on contribution with various contribution amount settings (including 0)
- Check the value of the dataLayer array in the console whether the "contribution" event object in the array appears and whether it has the correct data

## Help

### DIFF
--- a/static/js/src/contributions.js
+++ b/static/js/src/contributions.js
@@ -143,29 +143,22 @@ function sendContributionFormAnalytics() {
   const inputs = document.querySelectorAll(".p-slider__input");
   const total = document.querySelector(".js-total-amount").innerText;
 
+  let dataLayerObject = {
+    event: "contribution",
+    total: total
+  };
+
   // add the individual items
   inputs.forEach((amountElement) => {
     const name = amountElement.parentNode.parentNode.id;
     let value = amountElement.value || 0;
 
     if (value > 0) {
-      dataLayer.push({
-        event: "GAEvent",
-        eventCategory: "Contributions",
-        eventAction: "Item",
-        eventLabel: name,
-        eventValue: value,
-      });
+      dataLayerObject.name =  value;
     }
   });
 
-  dataLayer.push({
-    event: "GAEvent",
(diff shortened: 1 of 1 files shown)


static/js/src/contributions.js
```diff
@@ -143,29 +143,22 @@ function sendContributionFormAnalytics() {
...
-      dataLayer.push({
-        event: "GAEvent",
-        eventCategory: "Contributions",
-        eventAction: "Item",
-        eventLabel: name,
-        eventValue: value,
-      });
+      dataLayerObject.name =  value;
     }
   });
```
samhotep, line 159: Isn't this pushing only the last object in `inputs` to the `dataLayer` since `dataLayerObject.name` is being overwritten on every pass?

`dataLayerObject[name] = value;` could work, as it adds the values for each unique key, but it would also overwrite values with the same key name

laszlokajtar: maybe i'm wrong but i suppose each iteration of the forEach will give a different value to "name" and the dataLayerObject.name would be replaced with the value of "name". so dataLayerObject will end up with a new key-value pair at each iteration.

samhotep: Right now the iteration is directly referencing the `dataLayerObject.name` attribute, so each pass rewrites the value on the same attribute rather than storing the keys

samhotep (approved):
LGTM!

samhotep:
A few checks are failing, but otherwise seems good! 
@laszlokajtar

laszlokajtar:
@samhotep prettier fails to run for me so i can't really correct formatting. (i'm not sure what playwright is trying to do...) any way this can be merged?

PR #14035, opened 2024-07-04
Engage pages pagination

carkod:
## Done

- I've made some filters fixed, for example language filters have to be fixed, because if an author adds a new language, we need a thank-you template translated into the language, so it can't be used right after immediately, this forces authors to go through us first. Same with resource types, there should be a limited amount of them, otherwise authors may create things in the wrong place, case in point blogs, blog should be in blogs not in engage pages.
- Very few options around querying strings inside markdown (since this is not separate Postgres database field), make it very difficult to get a list of tags. So for now I just reused the query engage pages param data explorer query to create a list, which is not very performant, could be improved if anybody has better ideas.
- As a result, I also had to limit the filters to search one at a time, this is probably the fastest. If users ping us to ask for hierarchical search, this should be defined more clearly and we can write another data explorer query.
- I didn't add pagination for takeovers.json, as it is usually pretty low numbers, and it can be a bit more logic that needs to be discussed e.g. how many do we limit? how will stakeholders want to show this priority etc. But it is already supported, just need to change the view function to do it.

## QA

1. Go to https://ubuntu-com-14035.demos.haus/engage
2. Scroll down and click next page and/or the numbers to check that next page doesn't repeat the same engage pages.
3. Select language, and check that the correct language of engage pages shows
4. Same for resource types and tags

Check https://ubuntu-com-14035.demos.haus/tutorials work and check https://ubuntu-com-14035.demos.haus/takeovers work.

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-12662

## Help


carkod:
Also, I'd advise deploying this on Monday instead of Friday.

webapp/security/views.py
```diff
@@ -245,9 +245,9 @@ def single_notices_sitemap(offset):
         links.append(
             {
                 "url": f"https://ubuntu.com/security/notices/{notice_id}",
-                "last_updated": notice["published"]
-                if notice["published"]
-                else "",
+                "last_updated": (
+                    notice["published"] if notice["published"] else ""
+                ),
```
carkod, line 250: Python linting

webapp/shop/advantage/views.py
```diff
@@ -523,10 +523,10 @@ def post_auto_renewal_settings(ua_contracts_api, **kwargs):
...
     account_id = kwargs.get("account_id")
 
-    subscriptions: List[
-        Subscription
-    ] = advantage_mapper.get_account_subscriptions(
-        account_id=account_id, marketplace="canonical-ua"
+    subscriptions: List[Subscription] = (
+        advantage_mapper.get_account_subscriptions(
+            account_id=account_id, marketplace="canonical-ua"
+        )
```
carkod, line 529: Python linting

webapp/shop/api/ua_contracts/api.py
```diff
@@ -44,9 +44,9 @@ def _request(
         error_rules=None,
         headers={},
     ):
-        headers[
-            "Authorization"
-        ] = f"{self.token_type} {self.authentication_token}"
+        headers["Authorization"] = (
+            f"{self.token_type} {self.authentication_token}"
+        )
```
carkod, line 49: Python linting

webapp/shop/api/ua_contracts/helpers.py
```diff
@@ -255,10 +255,10 @@ def get_user_subscription_statuses(
...
         statuses["is_trialled"] = True if active_trial else False
 
-        statuses[
-            "is_subscription_auto_renewing"
-        ] = is_billing_subscription_auto_renewing(
-            subscriptions, subscription_id
+        statuses["is_subscription_auto_renewing"] = (
+            is_billing_subscription_auto_renewing(
+                subscriptions, subscription_id
+            )
```
carkod, line 261: Python linting

webapp/shop/cred/views.py
```diff
@@ -526,10 +526,12 @@ def cred_your_exams(ua_contracts_api, trueability_api, **kwargs):
...
-                                    "text": "Continue exam"
-                                    if state
-                                    == RESERVATION_STATES["in_progress"]
-                                    else "Take exam",
+                                    "text": (
+                                        "Continue exam"
+                                        if state
+                                        == RESERVATION_STATES["in_progress"]
+                                        else "Take exam"
+                                    ),
```
carkod, line 534: Everything on this file is Python linting. The linting Github action didn't pass if I didn't fix this.

britneywwc: Thanks for fixing the Python linting issue!

britneywwc (approved):
Thanks @carkod, looks great! Will keep note of the filters fix and search filters

PR #14041, opened 2024-07-05
Use smaller notices on notices page

samhotep:
## Note
- Should **only** be reviewed after https://github.com/canonical/ubuntu-com-security-api/pull/163 has been merged

## Done
- Load notices on the `/security/notices` page from a new datasource `get_page_notice`
- This returns a smaller notice, _without_ the `cves` and `packages` fields which are not required on this page, and have a significant effect on page speed, especially when a `notice` has hundreds of `cves`.

## QA

- Check that the demo runs
- Open https://ubuntu-com-14041.demos.haus/security/notices and click through the pagination
- Pages should open quickly without timeouts

### DIFF
--- a/webapp/security/api.py
+++ b/webapp/security/api.py
@@ -118,6 +118,42 @@ class SecurityAPI:
 
         return notices_response.json()
 
+    def get_page_notices(
+        self,
+        limit: int,
+        offset: int,
+        details: str,
+        release: str,
+        order: str,
+    ):
+        """
+        Makes request for all notices with ongoing support,
+        returns json object if found.
+        Note that these notices are missing the 'cves' fields.
+        """
+
+        parameters = {
+            "limit": limit,
+            "offset": offset,
+            "details": details,
+            "release": release,
+            "order": order,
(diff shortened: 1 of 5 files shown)


webapp/security/api.py
```diff
@@ -113,6 +113,42 @@ def get_notices(
 
         return notices_response.json()
 
+    def get_page_notices(
```
mtruj013, line 116: If we switch the remaining views `single_notices_sitemap`, `notices_sitemap`, and `notices_feed` to consume this new endpoint we can remove the old `get_notices` function completely

samhotep: Alright, I'll switch them over

samhotep: Done!

mtruj013 (approved):
LGTM! Sitemaps and rss tested fine as well. I think we're safe to remove the old `get_notices`

samhotep:
@mtruj013 I've taken it out, just waiting on the checks to pass now :)

PR #14064, opened 2024-07-11
Cleanup notices

samhotep:
## Done

- Removed unused get_notices method

## QA

- Check that the demo works
- Open /security page and check it's working as expected


mtruj013 (approved):
LGTM!

PR #14085, opened 2024-07-18
Refactor summarized status logic

mtruj013:
## Done

- Refactored summarized statuses logic

## QA

- View the site locally in your web browser at: https://ubuntu-com-14085.demos.haus/security/cves
- See that the summarized statuses are displaying as expected
- Click on any of the "By Ubuntu release" links 
  - See that summarized statuses are correct here as well and that they are hidden when the detailed view switch is toggled

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-8197


juanruitina:
Nice. I made some tests and this seems to work OK. It is lacking one feature, though: as we discussed at some point, summarised status should reflect the numbers accounting for the filters, not for all the available data. So, for instance, if user just selects one release, we should account for every package status but for that release only. Can we do this either now or in a separate task?

One visual bug: let's please align statuses to the baseline of heading and priority.

Screenshot 2024-07-19 at 09 49 21

Also leaving here a note for the individual CVE page: we should show there the reasoning for the "Ignored" status, I don't think we are now (see how it works on the live website, e.g. [here](https://ubuntu.com/security/CVE-2024-26925)).

webapp/security/helpers.py
```diff
@@ -21,16 +23,23 @@ def get_summarized_status(
     for package in cve["packages"]:
         package_count += len(package["statuses"])
```
juanruitina, line 24: I see `package_count` being used in `total_fixable` below, I think we need to tweak this so we filter out unselected releases.

juanruitina:
Besides my comment above, please:

- [ ] "Some fixes available" is still not aligned, can you tweak?
- [ ] Add a notification like this when there one or more releases selected and the detailed view is _not_ enabled ([design](https://www.figma.com/design/nUJEBpOPTW74RqDRuMQXxB/23.04-U.com---CVEs?node-id=1621-21477&t=Ynb6AV2j9k4Y2ZKL-0)).

Screenshot 2024-07-19 at 17 05 31

mtruj013:
Thanks @juanruitina , could you have another look?

juanruitina:
I noticed something wrong in the logic, this should be "Vulnerable". I couldn't find in the code where the issue might live. 

Screenshot 2024-07-23 at 16 26 12
Screenshot 2024-07-23 at 16 26 08

static/js/src/cve/cve.js
```diff
@@ -70,7 +70,7 @@ handlePackageInput();
 // Adds event listeners to all filter checkboxes
 function handleFilters() {
   releaseCheckboxes.forEach(function (checkbox) {
-    checkbox.addEventListener("change", function (event) {
+    checkbox.addEventListener("change", function (event) {
```
britneywwc, line 73:
```suggestion
    checkbox.addEventListener("change", function (event) {
```

webapp/security/views.py
```diff
@@ -323,7 +324,7 @@ def cve_index():
         statuses=statuses,
         order=order,
     )
-
+
```
britneywwc, line 327: Extra space here, should be removed once prettier/formatting is run

webapp/security/views.py
```diff
@@ -435,13 +423,29 @@ def cve_index():
         "pending": {"name": "Vulnerable", "icon": "warning"},
         "ignored": {"name": "Ignored", "icon": "error-grey"},
         "released": {"name": "Fixed", "icon": "success"},
+        "vulnerable": {"name": "Vulnerable", "icon": "warning"},
     }
 
+    for cve in high_priority_cves:
+        cve["summarized_status"] = {}
+        get_summarized_status(
```
britneywwc, line 431: I see that `get_summarized_status()` sets the `cve["summarized_status"]` within the function. Can we modify it so that it returns the value of the summarized_status and we set it where the function is being called? Since it is a getter function I think it should be more explicit that the function returns the value of the status rather than setting it implicitly.

Something like this:

```suggestion
    for cve in high_priority_cves:
        cve["summarized_status"] = get_summarized_status(...)
```

webapp/security/helpers.py
```diff
@@ -1,105 +1,104 @@
...
 
-        elif any(
-            d["status"] in vulnerable_indicators for d in package["statuses"]
-        ):
-            cve["summarized_status"] = {
-                "name": "Vulnerable",
-            }
+    if count == 1:
+        if key_with_non_zero_value == "ignored-high":
+            cve["summarized_status"] = friendly_names["ignored"]
```
britneywwc, line 67: As per the comment in `views.py`, we can return the value of the status rather than setting it such as:
```suggestion
            return friendly_names["ignored"]
```

If there is a specific reason that you'd prefer to set it within the function please let me know, perhaps we could change the function name to `set_summarized_status` instead if that's the case?

templates/security/cves/_cve-card.html
```diff
@@ -28,40 +28,24 @@ <h2 class="p-heading--5">
...
-              <span>Not in release</span>
-            </p>
-          {% else %}
-            {{ status }}
-          {% endif %}
-        {% endif %}
-      </div>
+      {% if cve.summarized_status %}
+        {% set status = cve.summarized_status %}
+        <div class="col-3 cve-summary col-medium-2" {% if status.icon %}style="padding-top: 0.4rem;"{% endif %}>
```
britneywwc, line 33: The missing of icon in "Not in release" excludes top padding, making it look inconsistent compared to the other statuses with icons
Screenshot 2024-08-07 at 4 22 52 PM

templates/security/cves/_cve-search-results.html
```diff
@@ -40,6 +41,15 @@
           </div>
         </div>
         <div class="p-section--shallow">
+          {% if versions %}
+          <div class="p-notification--information cve-summary is-light">
+            <div class="p-notification__content">
+              <p class="p-notification__message">
+                Status is adjusted based on your filters.
```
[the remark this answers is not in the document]
mtruj013: Yup, I did it the way as per Juan's [comment](https://github.com/canonical/ubuntu.com/pull/14085#issuecomment-2239426855) above. It makes sense too, we hide summarized statuses when detailed view is enabled and it's these statuses that change based on the applied filters, not the ones in the detail view table 🙂

webapp/security/views.py
mtruj013: Nice, this is better indeed!

webapp/security/helpers.py
mtruj013: No specific reason, update pushed :)

templates/security/cves/_cve-card.html
mtruj013: Could've sworn I'd set it to check for the icon, fixed!

britneywwc (approved):
Thanks for making the changes, looks amazing! 🎉

mtruj013:
Merging without UX as agreed as Juan is OOO this week. UX review to be done separately

PR #14146, opened 2024-08-05
Use the security notices api service

goulinkh:
## Done

- Use the service `ubuntu-com-security-api-notices` for `/security/notices.json`
- Use the service `ubuntu-com-security-api` for the rest of `/security` endpoints


mtruj013 (approved):
LGTM!

PR #14209, opened 2024-08-20
Vanilla Macro loader example (DO NOT MERGE)

pastelcyborg:
This PR exists solely to demonstrate examples and experiments for loading Vanilla Macros from `node_modules/` using Flask/Jinja. A permanent solution will likely be integrated into Flask Base directly.

DO NOT MERGE.

### DIFF
--- a/templates/desktop/index.html
+++ b/templates/desktop/index.html
@@ -1,4 +1,5 @@
 {% extends "desktop/base_desktop.html" %}
+{% from "_macros/tiered-list.jinja" import tiered_list %}
 
 {% block title %}Ubuntu Desktop PC operating system{% endblock %}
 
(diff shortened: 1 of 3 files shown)


templates/desktop/index.html
```diff
@@ -16,6 +17,82 @@
 
 {% block content %}
 
+  {%- call(slot) tiered_list(is_description_full_width_on_desktop=false, is_list_full_width_on_tablet=true) -%}
+    {%- if slot == 'title' -%}
+      <h2>H2 - up to two lines; ideally one.</h2>
```
jmuzina, line 22: If we're loading the macros from node modules, why do we need to duplicate them here? Is this just to test/POC this?

jmuzina (commented):
This is much simpler than I thought it would be!

britneywwc:
This is great, thank you! :)

I tried running the example on demo and it's throwing 500, running it locally shows that the macros are not found. 
Screenshot 2024-08-22 at 10 32 23 AM

Upon checking, there are no `templates/` files under `node_modules/vanilla_framework`. Is this soon to be added from Vanilla's side?

webapp/app.py
```diff
@@ -166,6 +166,17 @@
...
+# Vanilla Macro loader example
+
+from jinja2 import ChoiceLoader, FileSystemLoader
+
+# ChoiceLoader attempts loading templates from each given path in successive
+# order
+loader = ChoiceLoader([
+    FileSystemLoader('templates'),
+    FileSystemLoader('node_modules/vanilla-framework/templates/')
+])
```
jmuzina, line 178: What happens if a project has a file in `templates` with the exact same file path as a macro we publish that they'd like to use?

My assumption is the project's template will be used since it's listed first here. However this means we are slightly restricting the file structure of projects that use flask base. 

This isn't a bad thing, but maybe we should add some logic that outputs warnings if any identical file-paths are found between these two directories?

jmuzina: > Yep, I'd been wondering the same thing - maybe the filenames of our macros should be namespaced somehow (`vf.tiered_list.jinja` or whatever)?

I'd support that.

PR #14215, opened 2024-08-21
WD-13861 - create a function for form execution

lizzochek:
## Done

Created a function for form execution

REMOVE FORM FROM BASE INDEX AND AWS AFTER TESTING

## QA

- Check the form on the demo: [static](https://ubuntu-com-14215.demos.haus/form-template) and [dynamic](https://ubuntu-com-14215.demos.haus/aws#get-in-touch)
- Make sure that the form opens correctly and has all needed fields (you can check expected field sets in form-data.json)
- Check that the form submits necessary data
- Check that the form has correct required fields and does not submit if a required field is empty
- Ignore GH checks for now, the testing template will be removed before merge and the checks will be successful

## Issue / Card
[WD-13861](https://warthogs.atlassian.net/browse/WD-13861)

[WD-13861]: https://warthogs.atlassian.net/browse/WD-13861?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/shared/forms/form-template.html
```diff
@@ -0,0 +1,50 @@
+{% if isModal %}
+  <div class="p-modal" id="{{ modalId }}">
```
petesfrench, line 2:
```suggestion
  <div class="p-modal js-modal-ready" id="{{ modalId }}">
```
You need this js-hook so we can check the modal has loaded in dynamicForm.js

webapp/app.py
```diff
@@ -1161,3 +1163,26 @@ def render_supermicro_blogs():
...
+    @wraps(render_form)
+    def wrapper_func():
+        resultForm = {}
+        with app.app_context() and app.test_request_context():
+            filename = os.path.join(app.static_folder, 'files', 'forms-data.json')
+            with open(filename) as form_file:
+                data = json.load(form_file)
+                for form in data["forms"]:
+                    if form["formData"]["formId"] == formId:
+                        resultForm = form
```
petesfrench, line 1177:
```suggestion
              resultForm = next(
                    (
                        form
                        for form in data["forms"]
                        if form["formData"]["formId"] == formId
                    ),
                    None,
                )
```
Looping through all the forms will get expensive as the number of forms grows. You can use the [next function](https://www.w3schools.com/python/ref_func_next.asp)  to only loop through the minimum amount of forms.

petesfrench: Alternatively, you can create a hash map within the JSON and instead of looping through, use a unique identifier to directly fetch the relevant  form data. See my comment here

static/files/forms-data.json
```diff
@@ -0,0 +1,266 @@
+{
+    "forms": [
+        {
+            "formData": {
+                "title": "Test form",
+                "introText": "Intro text",
+                "formId": "1266",
+                "returnUrl": "/appliance/thank-you",
+                "product": ""
+            },
```
petesfrench, line 10:
```suggestion
    "forms": {
       "/path/to/page": {
            "formData": {
                "title": "Test form",
                "introText": "Intro text",
                "formId": "1266",
                "returnUrl": "/appliance/thank-you",
                "product": ""
            },
```
You could use a sort of hash map here to access the form data. This could be the page path as there is usually on one form per page

petesfrench (commented):
I have left a couple comments which should help with the performance as the datasets grow

webapp/app.py
petesfrench: The above comment will improve performance but require a little refactoring of the code here

petesfrench:
@lizzochek Can you get the checks passing and make sure it is re-based onto main

petesfrench:
'how-many-machines-do-you-have' was submitting to the payload. This will cause the form to fail.
image
Also, comments for lead was empty:
image

lizzochek:
The form is submitted for me, the comments field is correct and 'how many devices" is not blocking the form submission. Can you share more details on what and where you submitted?

Screenshot 2024-09-03 at 19 12 55

petesfrench:
This was for /form-template. I checked in [marketo](https://engage-sj.marketo.com/?munchkinId=066-EOV-335#/mktform/1266/overview/details) and couldn't see the submission

static/files/forms-data.json
```diff
@@ -0,0 +1,534 @@
...
+            {
+              "type": "long-text",
+              "id": "comments"
+            }
+          ]
+        },
+        {
+          "title": "How many devices?",
+          "id": "how-many-machines",
+          "inputName": "how-many-machines-do-you-have",
```
petesfrench, line 48: I think you can just get rid of this

static/files/forms-data.json
```diff
@@ -0,0 +1,534 @@
...
+            {
+              "type": "radio",
+              "id": "greater-than-100",
+              "value": "greater than 100",
+              "label": "> 100 machines"
+            }
+          ]
+        },
+        {
+          "title": "How should we get in touch?",
```
petesfrench, line 84: I think you could add a check here like 'no comment for lead', as you don't want these being attached to the comments for lead. I believe all fields in the 'about you' sections have a 'name' attribute, meaning they are attached to the payload

templates/shared/forms/form-fields.html
```diff
@@ -0,0 +1,206 @@
+<section class="p-section">
+  <form {% if isModal %}class="js-modal-form"{% endif %}
+        action="/marketo/submit"
+        method="post"
+        id="mktoForm_{{ formData.formId }}">
+    {% for fieldset in fieldsets %}
+      <div class="p-section">
+        <hr class="is-fixed-width" />
+        <fieldset class="p-fieldset-section {% if fieldset.id=='ubuntu-versions' %} js-toggle-checkbox-visibility {% elif fieldset.isRequired and fieldset.fields[0].type=='checkbox' %} js-required-checkbox {% endif %}" id="{{ fieldset.id }}" aria-labelledby="{{ fieldset.id }}">
+        <div class="row--50-50 js-formfield">
```
petesfrench, line 10:
```suggestion
        <div class="row--50-50 {% if not fieldset.no-js %}js-formfield{% endif%}">
```
this is what I am talking about in [this comment](https://github.com/canonical/ubuntu.com/pull/14215#discussion_r1742497272), fyi, this is psudo code

petesfrench:
The thank-you page url looks like `/thank-you?referrer=http://0.0.0.0:8001/aws`, usually we just redirect back to the original page? Unless I am missing something

petesfrench:
I tested the `/form-template` form again and it worked :raised_hands:

petesfrench:
Looking good! Are you planning to address the return url issue in another pr? Can you link the issue here please

lizzochek:
@petesfrench thank you pages will be removed and we will keep banners only, this will be tackled here: https://warthogs.atlassian.net/browse/WD-13873

petesfrench (approved):
LGTM, just need to fix the lint-python check

PR #14252, opened 2024-09-02
Add optional CTA button for engage pages

britneywwc:
## Done

- Add optional CTA button on engage page banners

## QA

- Go to [engage page template](https://discourse.canonical.com/t/engage-pages-and-takeovers-v2/358) on discourse
- Check that the `show_cta_on_banner` doc is added 
- Create a "New Topic" on discourse engage pages [here](https://discourse.ubuntu.com/c/design/engage-pages/51)
- See that `show_cta_on_banner` shows up on the template
- Go to https://ubuntu-com-14252.demos.haus/engage/microk8s-nvidia-tech-stack
- See that the primary CTA button shows up on banner and that the [original discourse post](https://discourse.ubuntu.com/t/simplifying-kubernetes-across-the-clouds/22260) sets `show_cta_on_banner` to true
- Go to https://ubuntu-com-14252.demos.haus/engage/redhat-openstack-comparison-whitepaper
- See that the primary CTA button doesnt show up and that the [original discourse post](https://discourse.ubuntu.com/t/red-hat-openstack-platform-vs-canonicals-charmed-openstack/18159) sets `show_cta_on_banner` to false
- Go to https://ubuntu-com-14252.demos.haus/engage/cra-yocto-core
- See that the primary CTA button doesnt show up and that the [original discourse post](https://discourse.ubuntu.com/t/whitepaper-yocto-vs-core-refresh-cra/47610) does not have `show_cta_on_banner` set

## Issue / Card

Fixes [WD-7421](https://warthogs.atlassian.net/browse/WD-7421)

## Screenshots

## Help

[WD-7421]: https://warthogs.atlassian.net/browse/WD-7421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/engage/base.html
+++ b/templates/engage/base.html
@@ -56,7 +56,7 @@
         {{ metadata['subtitle'] }}
       </p>
       {% if ("primary_cta" in metadata) and (metadata["primary_cta"] != "") %}
-      <p class="{% if ('show_cta_on_banner' in metadata) and (metadata["show_cta_on_banner"] == "true" ) %}u-show--large{% else %}u-hide--large{% endif %}">
+      <p class="{% if ('show_cta_on_banner' in metadata) and (metadata["show_cta_on_banner"] == "true" ) %}u-show--large{% else %}u-hide--large{% endif %} u-hide--medium u-hide--small">
           <a href="{{ metadata['primary_link'] }}" class="p-button--positive">
             {{ metadata["primary_cta"] }}
           </a>
--- a/templates/engage/base.html
+++ b/templates/engage/base.html
@@ -56,7 +56,7 @@
         {{ metadata['subtitle'] }}
       </p>
       {% if ("primary_cta" in metadata) and (metadata["primary_cta"] != "") %}
-      <p class="{% if ('show_cta_on_large_banner' in metadata) and (metadata["show_cta_on_large_banner"] == "true" ) %}u-show--large{% else %}u-hide--large{% endif %} u-hide--medium u-hide--small">
+      <p class="{% if ('show_cta_on_banner' in metadata) and (metadata["show_cta_on_banner"] == "true" ) %}u-show{% else %}u-hide{% endif %}">
           <a href="{{ metadata['primary_link'] }}" class="p-button--positive">
             {{ metadata["primary_cta"] }}
           </a>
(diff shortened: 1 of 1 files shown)


immortalcodes:
>See that the primary CTA button doesnt show up and that the [original discourse post](https://discourse.ubuntu.com/t/whitepaper-yocto-vs-core-refresh-cra/47610) does not have show_cta_on_banner set

So this one does not have a show_cta_on_banner flag, if that is intended then it's fine. Even if it's not set should it show up as empty so may be it could be set later by author if needed?

immortalcodes:
Another question I had was is this parameter meant for large screens only since the CTA would show up on medium-small screens.
If that is the case shouldn't we name the flag a bit differently?

britneywwc:
> So this one does not have a show_cta_on_banner flag, if that is intended then it's fine. Even if it's not set should it show up as empty so may be it could be set later by author if needed?

Yes that is right, because the older engage pages were created without the new template with the "show_cta_on_banner" flag. So the default behaviour would be to not show the CTA buttons.

britneywwc:
> Another question I had was is this parameter meant for large screens only since the CTA would show up on medium-small screens. If that is the case shouldn't we name the flag a bit differently?

That's a good point. I changed the flag to `show_cta_on_large_banner` and updated it on the discourse docs and templates. I've also pushed a quick fix to only show CTA on large screens and hide on small/medium by default. Thanks for the suggestion!

immortalcodes:
About the CTA on large screen, what was earlier being done is that the if CTA flag was true it was shown on all screens but if it was false it was shown on small and medium screens but not the large screens.
Reading the issue I feel why would someone not want to have CTA on small/medium screens?

So I meant that if the CTA is strictly meant for large screen then we should name the parameter accordingly otherwise if we show CTA, we show on all screens and if we not, we do not show it anywhere.

britneywwc:
I see, thanks for clarifying. I pushed up a change where it shows consistently on all screen sizes if `show_cta_on_banner` is set to true/false.

immortalcodes (approved):

PR #13412, opened 2023-12-15
CVE overhaul

akbarkz:
## Done

- This is a feature branch for CVE Overhaul story

## QA

- Check out this feature branch
- Run the site using the command `./run serve` or `dotrun`
- View the site locally in your web browser at: http://0.0.0.0:8001/
    - Be sure to test on mobile, tablet and desktop screen sizes

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-162, https://github.com/canonical/ubuntu.com/issues/11216, https://github.com/canonical/ubuntu.com/issues/11591, https://github.com/canonical/ubuntu.com/issues/12056, https://github.com/canonical/ubuntu.com/issues/12055, https://github.com/canonical/ubuntu.com/issues/12054, https://github.com/canonical/ubuntu.com/issues/13024, https://github.com/canonical/ubuntu.com/issues/13386

## Help

[QA steps](https://discourse.canonical.com/t/qa-steps/152) - [Commit guidelines](https://discourse.canonical.com/t/commit-guidelines/148)

PR #14270, opened 2024-09-05
Final CVE fixes

mtruj013:
## Done

- Addressed the requested changes in [this comment](https://github.com/canonical/ubuntu.com/pull/13412#issuecomment-2321341401)
- Drive by:
  - Removed the "no data" status filters as statuses are required for any given package on the api and so there will never be a package with an empty status field

## QA

- View the site locally in your web browser at the following and see that the changes have been addressed: 
  - https://ubuntu-com-14270.demos.haus/security/cves
  - https://ubuntu-com-14270.demos.haus/security/cves?q=
  - https://ubuntu-com-14270.demos.haus/security/CVE-2024-6984
  - https://ubuntu-com-14270.demos.haus/security/CVE-2024-6387
  - https://ubuntu-com-14270.demos.haus/security/cves/about

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-14630


juanruitina:
- [ ] Can you please move [here](https://github.com/canonical/ubuntu.com/blob/f775af248e9a174ab53de22d6365dca8d66e050f/templates/security/cves/cve.html#L202-L204) the "Read the notes from the security team" out of the `{% if cve.ubuntu_description %}`? Please do check if there's notes other than "Priority reason: " ones before showing it though.
- [ ] Please do exclude the "Priority reason: " note from the list of notes, but keep the others.

Screenshot 2024-09-05 at 17 57 31

- [ ] Reduce the space below the "Status" heading in the individual CVE page, it's huge right now.

>  I accidentally stumbled upon this CVE, which doesn't seem to affect any release. Should we even list these? I feel tempted to change the notification to "No releases are affected by this CVE." and hide the table, but I worry it will be misleading.

- [ ] About the above, please let's change the error message to "No information for any release." , a bit more conservative.
- [ ] Vulnerable icons still not changed in the about page.

And that would be it :)

webapp/security/helpers.py
```diff
@@ -1,13 +1,19 @@
+# Standard library
```
mtruj013, line 1: Everything in here has already been reviewed, I've only moved the existing functions from views.py to this file to make the `cve_index` view dryer and more readable

webapp/security/helpers.py
```diff
@@ -100,3 +106,108 @@ def get_summarized_status(
             return friendly_names["not-affected"]
         elif status_count["DNE"] > 0:
             return friendly_names["DNE"]
+
+
+"""
+Checks if there is only an upstream status for a CVEs with only 1 package
```
petesfrench, line 112:
```suggestion
Checks if there is only an upstream status for a CVE with only 1 package
```
I am not familiar with the terminology, but does it need the 's' here?

templates/security/cves/cve.html
```diff
@@ -527,32 +535,72 @@ <h3>Other references</h3>
...
     function handleHiddenReleasesSwitch() {
       const maintainedCount = {{ maintained_count | tojson | safe }};
+      const isOnlyUpstream = {{ only_upstream | tojson | safe }};
       const maintainedReleasesSwitch = document.querySelector(".js-toggle-maintained");
-      let hiddenCells = document.querySelectorAll(".js-unmaintained-cell");
+      const hiddenCells = document.querySelectorAll(".js-unmaintained-cell");
 
       // When the CVE does not include statuses for maintained releases, the switch is disabled and checked
-      if (maintainedCount === 0) {
+      if (maintainedCount === 0 || isOnlyUpstream) {
```
petesfrench, line 548: Do you need to update the above comment to reflect this addition?

petesfrench (approved):
LGTM, code looks good. Just a couple comments about comments. I haven't QA'd as I wasn't sure how to test it exactly

petesfrench:
When I click the 'Toggle filters' button on medium and small screens on https://ubuntu-com-14270.demos.haus/security/cves?version=noble&version=jammy&version=focal#drawer the doesn't open. This is also happening on the feature branch.
[Screencast from 06-09-24 16:42:38.webm](https://github.com/user-attachments/assets/cee0a808-d606-4a37-8115-e83053e6d3bb)

webapp/security/helpers.py
mtruj013: Nope, typo!

templates/security/cves/cve.html
mtruj013: Yes, thanks for catching!

static/sass/_pattern_cve.scss
```diff
@@ -145,10 +145,6 @@
       }
 
       .p-accordion__tab {
-        z-index: -1;
```
mtruj013, line 148: This fixes the side nav categories not toggling when clicked eg:
[Screencast from 09.09.2024 17:01:50.webm](https://github.com/user-attachments/assets/16636a8d-20ce-4180-975c-9d3571dbcf00)

webapp/security/helpers.py
```diff
@@ -100,3 +106,144 @@ def get_summarized_status(
             return friendly_names["not-affected"]
         elif status_count["DNE"] > 0:
             return friendly_names["DNE"]
+
+
+"""
```
mtruj013, line 111: This has been refactored more clearly differentiate what is meant by "maintained" releases. These are the combination of lts releases (which are flagged as LTS in the UI), esm releases (which are also flagged as LST in the UI), and interim releases (which do not have a support tag). There aren't any interim releases right now so this can be qa'd as follows (on [landing](https://ubuntu-com-14270.demos.haus/security/cves)):
- "All maintained" should apply 24.04, 22.04, 20.04, 18.04, 16.04, and 14.04 filters when clicked
- "All LTS" should only apply 18.04, 16.04 and 14.04 filters when clicked

webapp/security/helpers.py
```diff
@@ -100,3 +106,144 @@ def get_summarized_status(
...
+            return True
+    return False
+
+
+"""
+Checks if references link is from a default reference URL
+"""
+
+
+def does_not_include_base_url(link):
```
mtruj013, line 238: Already reviewed, just moved from views.py to this file

petesfrench:
Given the response time of the CVE search, I think the search button would benifit a lot from the [vanilla loading spinner](https://vanillaframework.io/docs/patterns/icons)
image
Currently it feels like the button isn't working

petesfrench:
On the right hand filters, the focus-outline appears to be cut off by something. Not sure if that is a big deal, but thought I would flag it:
image

petesfrench:
In the 'sort by:' dropdown, if you select 'newest' the text overlaps the chevron a little 
image

mtruj013:
Thanks @petesfrench! 

I addressed 2 points, but didn't address the focus overflow as it would require breaking out of of the grid. I could maybe find way to make it work if I spend more time on it but it doesn't seem worth delaying the pr to do so.

As for the loading spinner, I added it but it does introduce a bit of a bug. When you click the search button on the landing page the spinner will load as expected and all is well, but when you click the back button after landing on the search results the spinner hangs as if it's still loading the request. This seems to be the case in production for static forms as well (opened ticket [here](https://warthogs.atlassian.net/browse/WD-14853)). I've tried using a number of event listeners, (popstate, popstate with settimeout, hashchange, load, etc) but none seem to work. I know I could reroute to the url dynamically but that would trigger a new round of calls to the api which would be expensive.

So that leaves me with the following 3 qs:
1. How detrimental of a bug is it to have the loading spinner hang on back click?
2. How detrimental is it to _not_ have a loading spinner? (context here being that sometimes the search query takes a while to return)
3. If neither of the above situations are acceptable, does a solution exist wherein we don't have to fully refresh the page on back?

cc: @juanruitina as you're probably the person who should make the call on this 🙂

juanruitina:
Mmm. Spinner makes total sense when submitting a regular form, because it is the _submitting_ that takes time (writing in the database), and there's the potential of data loss if user does something to stop the process. Here I understand what takes time is not the submission of the query, but the _fetching_ of the results, the reads in the database.

I don't think I've seen spinners in search submission buttons, but in the listings of search suggestions or results (see Jira, Mattermost). If results were fetched dynamically on the page, then yes, we could be showing a "Loading results" or such wherever results will be shown, perhaps with a spinner. It would more accurately convey that the submission was immediate, it's the searching that's taking time.

Here I think the loading indicators of the browser might be enough. It's not ideal, and Pete is absolutely right in pointing out that some users might be confused. But if implementing the spinner brings its own problems, I'd suggest to just drop it. In my opinion, the problem to fix is that _search is slow_.

PR #14343, opened 2024-09-25
Decouple cve pagination from certified pages

mtruj013:
## Done

- Refactored the certified search results pages to use the default pagination instead of sharing the cve pagination  

## QA

- View the site locally in your web browser at: http://0.0.0.0:8001/certified?q=
- Check that pagination works as expected 

## Issue / Card

Fixes #


andrejvelichkovski:
Thanks for the quick follow-up @mtruj013. I went through the certified pages and the changes look good to me. We don't really have a lot of complex behaviour on the certified pages regarding pagination, so this should work well.

lizzochek (approved):
LGTM!

PR #14440, opened 2024-10-21
Implement user_id and pass it in marketo submission

mtruj013:
## Done

- Implemented “user_id” feature to send universally unique identifiers to Google Analytics as described [here](https://warthogs.atlassian.net/browse/WD-14908)
- Added functionality to pass value on form submission

## QA

- View the site locally in your web browser at: https://ubuntu-com-14440.demos.haus/ (or any page)
- If you have not set cookie preferences:
  - Before accepting any cookies open dev tools, log `document.cookie` and see that `user_id` does not exist
  - Accept either all or performance cookies and do the same as above, this time ensuring that `user_id` is now present
- Go to any form and fill it out (https://ubuntu-com-14440.demos.haus/ai#get-in-touch for example)
- Go to the enrichment form on maketo: https://engage-sj.marketo.com/?munchkinId=066-EOV-335#/mktform/4198/overview/details -> "Lead enrichment details" -> "People"
- Search for the email you used and the click on the ID link, eg:
image
- Click on the "SFDC Custom Fields" tab and see that `Google Analytics User ID` has you user id as a value:
image

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-14942

### DIFF
--- a/static/js/src/cookie-policy-with-callback.js
+++ b/static/js/src/cookie-policy-with-callback.js
@@ -1,0 +1,20 @@
+import { v4 as uuidv4 } from "https://jspm.dev/uuid";
+// Set user_id after cookie acceptance
+function setUserId() {
+  const getCookie = (targetCookie) =>
+    document.cookie.match(new RegExp("(^| )" + targetCookie + "=([^;]+)"));
+  const cookieAcceptanceValue = getCookie("_cookies_accepted");
+
+  if (
+    cookieAcceptanceValue?.[2] === "all" ||
+    cookieAcceptanceValue?.[2] === "performance"
+  ) {
+    // Check if user doesn't already have a user_id
+    if (!getCookie("user_id")) {
+      // Generate a universally unique identifier
+      const user_id = uuidv4();
+      document.cookie = "user_id=" + user_id + ";max-age=31536000;";
+    }
+  }
+}
+cpNs.cookiePolicy(setUserId);
--- a/static/js/src/cookie-policy-with-callback.js
+++ b/static/js/src/cookie-policy-with-callback.js
@@ -9,7 +9,7 @@ const getCookie = (targetCookie) =>
   document.cookie.match(new RegExp("(^| )" + targetCookie + "=([^;]+)"));
 let cookieAcceptanceValue = getCookie("_cookies_accepted");
 
-if (cookieAcceptanceValue === null) {
+if (!cookieAcceptanceValue) {
   cpNs.cookiePolicy(setUserId);
 } else {
   setUserId();
--- a/static/js/src/cookie-policy-with-callback.js
+++ b/static/js/src/cookie-policy-with-callback.js
@@ -1,10 +1,21 @@
-import { v4 as uuidv4 } from "https://jspm.dev/uuid";
-// Set user_id after cookie acceptance
-function setUserId() {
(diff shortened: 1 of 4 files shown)


eabashidze:
Hi team! I could not catch user_id in Marketo after testing. Here are the results:
As the first step, I added the hidden field to marketo enrichment form, and set it as non-mandatory. 
On the website end, user_id does not become visible in document.cookies after cookies are accepted, only after I refresh it. Same for dataLayer, user_id becomes visible after accepting cookies + the refresh. When I navigate to the form (submitted 2 contact us forms), it's visible neither in data layer, nor in the payload. I can see user_id in the cookie part of the network tab (checking the submitting activity), but still not passed to mkto. Please advise on next steps.

eabashidze:
Hi @mtruj013 @laszlokajtar ,
I confirm that all the issues are solved and test went through.
Thanks!
Screenshot from 2024-10-24 14-01-57
Screenshot from 2024-10-24 13-58-31

laszlokajtar:
@mtruj013 @eabashidze i've had multiple sessions now where user_id wasn't generated for me. is this an exception?

laszlokajtar:
@mtruj013 @eabashidze also, in a new incognito session, while the user_id was in the cookies, it disappears or doesn't get pushed into the dataLayer on page reload for me

static/js/src/cookie-policy-with-callback.js
```diff
@@ -0,0 +1,35 @@
+/*
+  Set a unique user_id if the user has accepted cookies.
+  If the user has not set their cookie preference, wait for the cookie policy to run first. 
+*/
+
+import { v4 as uuidv4 } from "https://jspm.dev/uuid";
```
akbarkz, line 6: I think it would be better to install and import `uuidv4` npm package instead. This import looks a bit unusual to me. Is it written somewhere that this is the way it should be done?

static/js/src/cookie-policy-with-callback.js
```diff
@@ -0,0 +1,35 @@
...
+
+const getCookie = (targetCookie) =>
+  document.cookie.match(new RegExp("(^| )" + targetCookie + "=([^;]+)"));
+let cookieAcceptanceValue = getCookie("_cookies_accepted");
+
+if (cookieAcceptanceValue === null) {
```
akbarkz, line 12: what if the value is None?

static/js/src/cookie-policy-with-callback.js
mtruj013: It won't be available client side without a bundler, so I've used a cdn as suggested [here](https://stackoverflow.com/a/75202002)

static/js/src/cookie-policy-with-callback.js
mtruj013: The value of that cookie when not set is `null`, but updated to catch more falsey values

mtruj013:
@laszlokajtar I fixed the `dataLayer` not persisting but, I couldn't replicate any session wherein `user_id` wasn't generated on chrome or ff, incognito or otherwise. Please flag if you're still having this issue. I've tested on multiple pages and checked on marketo and everything seems to be working as expected

akbarkz (approved):

laszlokajtar:
@mtruj013 @akbarkz thank you!

PR #14638, opened 2025-01-10
chore(links): Convert all local links to relative links

anthonydillon:
## Done

- Replace all local links with relative ones to clean things up and prevent local and demo links from jumping out into the real world.

## QA

- Check the code changes are correct and will not cause any issues


muhammad-ali-pk (approved):
LGTM! Thanks.

PR #14672, opened 2025-01-23
add compression of js/css assets

akbarkz:
## Done

- add `Flask-Squeeze` package to compress js and css
- defer one of the JS files that is not needed immediately
- lazy load all images on the home page that are not visible initially

## QA

- Compare JS and CSS files sizes of the demo instance and production and check that it has significantly decreased.


muhammad-ali-pk (approved):
LGTM!

PR #14681, opened 2025-01-25
upgrade to flask-base 2.1.0 that does the assets comression

akbarkz:
## Done

- This is the next step after [this PR](https://github.com/canonical/ubuntu.com/pull/14672), doing the assets compression on the flask-base level.

## QA

- Check that the resources on the demo and production are the same. You can do that by using Network tab of the dev console. For example, compare the sizes of `styles.css` on the demo and on production, and check that it didn't become bigger.


muhammad-ali-pk (approved):
LGTM!

PR #14704, opened 2025-02-04
Add redirects for static case studies

britneywwc:
## Done

- Add redirect rules for static case studies that have been migrated from u.com to c.com

## QA

- Go to https://ubuntu-com-14704.demos.haus/case-study/esa
- See that it redirects to the case study on c.com
- Repeat for 
  - https://ubuntu-com-14704.demos.haus/case-study/sbi-bits
  - https://ubuntu-com-14704.demos.haus/case-study/grundium-ubuntu-pro-for-devices
  - https://ubuntu-com-14704.demos.haus/case-study/ubuntu-pro-support-for-games-developer

## Issue / Card

Fixes [WD-18974](https://warthogs.atlassian.net/browse/WD-18974)

## Screenshots

## Help

[WD-18974]: https://warthogs.atlassian.net/browse/WD-18974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -124,7 +124,11 @@ category/product/landscape/?: /landscape
 category/product/(server|ubuntu-server-edition)/?: /server
 category/product/ubuntu-advantage/?: /support
 category/product/ubuntu-light/?: http://lubuntu.net/
-case-study: "https://canonical.com/case-study"
+case-study/?: "https://canonical.com/case-study"
+case-study/esa/?: "https://canonical.com/case-study/esa/"
+case-study/sbi-bits/?: "https://canonical.com/case-study/sbi-bits/"
+case-study/grundium-ubuntu-pro-for-devices/?: "https://canonical.com/case-study/grundium-ubuntu-pro-for-devices/"
+case-study/ubuntu-pro-support-for-games-developer/?: "https://canonical.com/case-study/ubuntu-pro-support-for-games-developer/"
 # old certification.ubuntu.com redirects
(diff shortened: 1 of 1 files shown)


redirects.yaml
```diff
@@ -124,7 +124,11 @@ category/product/landscape/?: /landscape
 category/product/(server|ubuntu-server-edition)/?: /server
 category/product/ubuntu-advantage/?: /support
 category/product/ubuntu-light/?: http://lubuntu.net/
-case-study: "https://canonical.com/case-study"
+case-study/?: "https://canonical.com/case-study"
+case-study/esa/?: "https://canonical.com/case-study/esa/"
```
akbarkz, line 128: the line above should be enough for every case study to be redirected. You don't need to list all the individual ones.

akbarkz (changes requested):

redirects.yaml
britneywwc: `?` only checks for if the link is `/case-study` or `/case-study/` so the other individual case studies would not fall under the redirect rule.
If it did redirect with just one rule, it would've redirected users to the index page rather than the specific case study. Since there are only 5 case studies I thought it would be better to add redirect rules to the exact path in c.com instead of the index page.

akbarkz: you're right, sorry

PR #14740, opened 2025-02-13
feat: Add multiple filters functionality on engage pages

britneywwc:
## Done

- Add multiple filters functionality on engage page index
- Users can now search up to three filters simultaneously. Previously this was set at one

## QA

- Go to https://ubuntu-com-14740.demos.haus/engage
- Select one filter from the dropdown
- See that page loads, select another filter. 
- Repeat and see that page loads according to selected filters.

## Issue / Card

Fixes [WD-16129](https://warthogs.atlassian.net/browse/WD-16129)

## Screenshots

## Help

[WD-16129]: https://warthogs.atlassian.net/browse/WD-16129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/engage/index.html
+++ b/templates/engage/index.html
@@ -143,10 +143,6 @@ function clearFilters() {
   setFiltersDropdown("tag", tagSelector);  
 
   function setFilters(key, selector) {
-    //if (urlObj.searchParams.has(key)) {
-    //  selector.value = urlObj.searchParams.get(key);
-    //}
-    
     if (selector.value != "all") {
       urlObj.searchParams.set(key, selector.value);
     } else {
@@ -161,6 +157,7 @@ function clearFilters() {
     setFilters("resource", resourceSelector);
     setFilters("tag", tagSelector);
 
+    urlObj.searchParams.delete("page");
     window.location = urlObj.href;
 }
   applyFiltersButton.addEventListener("click", applyFilters);
--- a/templates/engage/index.html
+++ b/templates/engage/index.html
@@ -126,21 +126,6 @@ function clearFilters() {
     clearFiltersButton.removeAttribute("disabled");
   }
 
-  // Disable other fields if a filter is already applied
-  const searchParams = new URLSearchParams(urlObj.search);
-  for (const [key, value] of searchParams) {
-    if (key === "language") {
-      resourceSelector.setAttribute("disabled", true);
-      tagSelector.setAttribute("disabled", true);
-    } else if (key === "resource") {
-      languageSelector.setAttribute("disabled", true);
-      tagSelector.setAttribute("disabled", true);
-    } else if (key === "tag") {
-      languageSelector.setAttribute("disabled", true);
-      resourceSelector.setAttribute("disabled", true);
-    }
-  } 
-
   function handleFilter(key, el, url) {
     if (!el) {
       return;
@@ -152,14 +137,11 @@ function clearFilters() {
 
     el.addEventListener("change", function(e) {
       const value = e.target.value;
-      const selectFields = document.querySelectorAll("select:checked")
-
-      // Reset filters, data explorer query can't search multiple fields
-      // at the same time
-      url.search = "";
 
       if (value !== "all") {
-        url.searchParams.append(key, value);
+        url.searchParams.set(key, value);
+      } else {
+        url.searchParams.delete(key);
       }
       window.location = url;
     });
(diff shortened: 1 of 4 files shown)


Sophie-32:
Add 'Apply filters' button, see for example: https://ubuntu.com/security/cves?version=noble&version=jammy&version=focal&version=bionic&version=xenial&version=trusty&version=oracular

petesfrench:
Some of the tag's are not capitalized (such as 'events'):
image

For some posts it seems there is no specific 'resource type' and the only way to find them is the use 'All resource types' (e.g. the 'Meet us at GITEX 2023' post):
image

There is a 'blank' tag that functions the same as 'All tags':
image

Do we want to add a fallback for posts that don't have a resource type like this one:
image

The 'max' posts per page should be a multiple of 3 to so there isn't an empty space at the bottom of the page like this:
image

britneywwc:
> Some of the tag's are not capitalized (such as 'events')

There was a decision made to query all tags from discourse posts and have them in the dropdown. I think this was because there were missing tags/inconsistencies in the past.

> For some posts it seems there is no specific 'resource type' and the only way to find them is the use 'All resource types' (e.g. the 'Meet us at GITEX 2023' post). But I guess the resource type could be 'Event'...

Thanks, I'll address this together with the posts with missing 'resource type'

> Do we want to add a fallback for posts that don't have a resource type like this one

Good idea, I'll add this in!

> The 'max' posts per page should be a multiple of 3 to so there isn't an empty space at the bottom of the page...

I did not make any changes to the pagination function on the index page, it's similar to [prod](https://ubuntu.com/engage) at the moment. The index page on [c.com/case-study](https://canonical.com/case-study) also has an empty space at the bottom of the page. Could design @Sophie-32 have a look and see if we should change these pages to remove the empty space at the bottom?

petesfrench:
This is most likely caused by improper language settings on Wordpress, but the second result when you set the language as 'English' is not English:
https://ubuntu-com-14740.demos.haus/engage?language=en

@Sophie-32 Would it be worth putting the tag's in alphabetical order? Right now it is very hard to find anything

Sophie-32:
@petesfrench - would definitely be worth it, especially since I think I see a whole lot of dupes. Where is this list pulling from and I can review it?

britneywwc:
Thanks for the suggestion, I will create a separate ticket for this because it is out of scope. 
@Sophie-32 The tags list is created by querying all engage pages and getting all their tags from existing pages. There is currently no single source where tags are being set, which is why there are a lot of inconsistencies/duplicates in the list.

britneywwc:
> This is most likely caused by improper language settings on Wordpress, but the second result when you set the language as 'English' is not English: https://ubuntu-com-14740.demos.haus/engage?language=en

This is because the PM for that engage page has set the language to "en" on Discourse, it should be fixed on Discourse. Then again, they have asked about adding a "TK" filter for the page and is currently setting it to "en" for visibility purposes.

templates/engage/index.html
```diff
@@ -63,7 +63,7 @@ <h1>Engage pages</h1>
         <select name="tag-selector" id="tag-selector" class="u-no-margin--bottom">
           <option value="all">All tags</option>
           {% for tag in tags %}
-            <option value="{{ tag }}">{{ tag }}</option>
+            <option value="{{ tag }}">{% if tag == "" %}-{% else %}{{ tag }}{% endif %}</option>
```
petesfrench, line 66:
```suggestion
            <option value="{{ tag }}">{% if tag == "" %}-{% else %}{{ tag }}{% endif %}</option>
```
Can you add the `disabled` attribute if the value is `-`

templates/engage/index.html
```diff
@@ -119,55 +120,47 @@ <h3>No results &mdash; why not try widening your search?</h3>
   const languageSelector = document.getElementById("language-selector");
   const resourceSelector = document.getElementById("resource-selector");
   const tagSelector = document.getElementById("tag-selector");
+
+  const applyFiltersButton = document.getElementById("apply-filters");
```
petesfrench, line 124: A little of of scope so feel free to ignore, but the 'best practice' for querying HTML elements in JS is to use a class such as `js-name`.

petesfrench (approved):
LGTM, just one minor change

templates/engage/index.html
britneywwc: Thanks for the suggestion, I'll apply the changes

PR #14774, opened 2025-02-21
fix(ci): Perf checks weekly cron report

anthonydillon:
## Done

- Set the performance checks to execute weekly and report into the channel if they fail.

## QA

- Check the code looks good


bartaz (approved):
LGTM

PR #14799, opened 2025-02-26
Prevent undefined values from being sent for the address

minkyngkm:
## Done
- `userInfo` always exists for both first-time buyers and existing users, so first-time buyers will always get undefined in the address on line 303 if checking `userInfo`
- Get the address from form values always instead of checking `userInfo` and prevent sending `undefined` values

## QA
- Go into the local branch and console.log address and check the values (for both first-time buyers and existing users)

## Issue / Card

Fixes #https://warthogs.atlassian.net/browse/WD-18214

### DIFF
--- a/static/js/src/advantage/subscribe/checkout/components/BuyButton/BuyButton.tsx
+++ b/static/js/src/advantage/subscribe/checkout/components/BuyButton/BuyButton.tsx
@@ -300,9 +300,16 @@ const BuyButton = ({ setError, products, action, coupon }: Props) => {
         localStorage.removeItem(key);
       });
 
-      const address = userInfo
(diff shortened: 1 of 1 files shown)


abhigyanghosh30 (approved):

static/js/src/advantage/subscribe/checkout/components/BuyButton/BuyButton.tsx
```diff
@@ -300,9 +300,16 @@ const BuyButton = ({ setError, products, action, coupon }: Props) => {
...
+      const address = [
+        values?.address,
+        values?.postalCode,
+        values?.city,
+        values?.usState,
+        values?.caProvince,
+        values?.country,
+      ]
+        .filter(Boolean)
+        .join(" ");
```
alnvdl-work, line 312: Down below, shouldn't you also change it to:
```
if (address) {
    formData.append("street", address);
}
```

That way you only set the field if it's present with a valid value.

minkyngkm: Thanks Allan, yep true! Updated 👍

PR #14976, opened 2025-04-10
pro: edit billing details from invoices page

andesol:
## Done

- `/account/invoices`: let users update their billing details (customer info)

See [design](https://www.figma.com/design/GXrVXEGISyJVJbGk4V7PoD/25.04-Pro-Dashboard-Updates?node-id=501-6579&m=dev)

## QA

- Go to `account/invoices`
- Click `Edit billing details`
- Modify the form
- Click save
- [x] Check that the data is updated if you click the button again

---
- [x] Leave one or more fields empty
- Click save
- [x] Check that an error is shown below the empty field(s) and the data is not saved

---

- [x] Write an invalid VAT
- [x] Check that an error is shown below the VAT field and the data is not saved

---

- [x] If the country you used for purchasing doesn't require a VAT, no VAT field should show up.

---

Check that the overall behaviour looks alright (the modal behaves as it should, etc.)

## Issue

https://warthogs.atlassian.net/browse/WD-20699

### DIFF
--- a/build.js
+++ b/build.js
@@ -15,6 +15,7 @@ let entries = {
   appliance: "./static/js/src/appliance.js",
   costCalculator: "./static/js/src/openstack/react/app.jsx",
   "ua-payment-methods": "./static/js/src/ua-payment-methods.js",
+  "account-billing": "./static/js/src/account-billing.js",
   "sticky-nav": "./static/js/src/sticky-nav.js",
   chassisAnimation: "./static/js/src/chassis-animation.js",
   cve: "./static/js/src/cve/cve.js",
--- a/static/js/src/account-billing.js
+++ b/static/js/src/account-billing.js
@@ -1,0 +1,196 @@
+import {
+  getCustomerInfo,
+  postCustomerInfoToStripeAccount,
+} from "./advantage/api/contracts.js";
+import { countries } from "./advantage/countries-and-states.js";
+
+const accountId = window.accountId;
+
+// This partially replicates the functionality of the modal.js file.
+// We do it so we can control when the modal opens in a predicatable way,
+// i.e. after the fetch the data and populate the content.
+function Modal({ modalSelector }) {
+  let currentDialog = null;
+  let lastFocus = null;
+  let ignoreFocusChanges = false;
+  let focusAfterClose = null;
+  const modal = document.getElementById(modalSelector);
+
+  function trapFocus(e) {
+    if (ignoreFocusChanges) return;
+    if (currentDialog.contains(e.target)) {
+      lastFocus = e.target;
+    } else {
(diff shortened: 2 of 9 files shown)


abhigyanghosh30:
[ ] Cannot save details
[ ] Cannot change country

andesol:
@abhigyanghosh30 Thanks for the review!

> * [ ]  Cannot save details

Definitely a bug. Could you share what is the behaviour? Notice that the modal doesn't close after clicking save, that's expected ([see comment in Figma](https://www.figma.com/design/GXrVXEGISyJVJbGk4V7PoD?node-id=691-6869&m=dev#1215829680)). The change should be saved if you close and reopen again, though.


> * [ ]  Cannot change country

That's intended. cc: @polgarp

abhigyanghosh30:
> @abhigyanghosh30 Thanks for the review!
> 
> > * [ ]  Cannot save details
> 
> Definitely a bug. Could you share what is the behaviour? Notice that the modal doesn't close after clicking save, that's expected ([see comment in Figma](https://www.figma.com/design/GXrVXEGISyJVJbGk4V7PoD?node-id=691-6869&m=dev#1215829680)). The change should be saved if you close and reopen again, though.
> 
> 
> > * [ ]  Cannot change country
> 
> That's intended. cc: @polgarp 
> 
> 

The request was definitely failing but it was not showing any error on the frontend

andesol:
@abhigyanghosh30 Strange, I tried with a country with VAT (UK) and a country without (India) and the request seems to be going through just fine. At the same time, we have this error that _in theory_ should show up

image

(and it did show up when I introduced an arbitrary error in the back-end).

Can you share a screenshot when you have time? Or maybe we can have a look next week.

polgarp:
@abhigyanghosh30 
> * [ ]  Cannot change country

In short if country is changed, that should be a new business entity, with possibly new contract or term, so that is a special case. It could be that there are some edge cases, like mergers or acquisitions or transfer of duties between sub entities, but we will handle those cases on a case-by-case basis manually.

abhigyanghosh30 (dismissed):

andesol:
@abhigyanghosh30 Pushed a last commit to make the VAT field not required, for consistency with the checkout (as discussed with @polgarp). Please have a look at the last commit and let me know if you have any concern.

abhigyanghosh30:
> @abhigyanghosh30 Pushed a last commit to make the VAT field not required, for consistency with the checkout (as discussed with @polgarp). Please have a look at the last commit and let me know if you have any concern.

I think it should be alright because it will tell the user that the VAT is invalid in case they try to purchase something with an unverified and invalid VAT

abhigyanghosh30 (approved):

PR #14992, opened 2025-04-15
Set up core docs redirect and remove related stale code

mtruj013:
## Done

- Applied redirects as per ticket 
- Removed stale templates and logic related to these docs 
- Updated old redirects so that those paths don't have to go through 4 layers of redirects to get to the target path

## QA

- View the site locally in your web browser at: https://ubuntu-com-14992.demos.haus/core/docs and see that you are rerouted correctly 
- Test remaining redirects and see that all work as expected

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-21341


immortalcodes (commented):
These changes look good to me, there are some mentions for /core/docs in architechture.md , konf/site.yaml , performance check workflow in github.
Some html pages also have /core/docs links, shall we make changes at those places as well?

mtruj013:
> there are some mentions for /core/docs in architechture.md , konf/site.yaml , performance check workflow in github.
Some html pages also have /core/docs links, shall we make changes at those places as well?

I removed additional mentions to this path, but have left the nginx configurations in site.yaml as there might be a dependency there with IS. Since stakeholders are chasing for this work and redirects will still work as expected, I think we should leave that for another pr

petesfrench (approved):
LGTM

PR #15052, opened 2025-04-28
WD-21511 WD-21512: show correct exam names when scheduling

usamabinnadeem-10:
## Done

- When scheduling CUE.02 Desktop exam, the schedule confirmation showed CUE.01 Linux (incorrect name)
- This PR fixes that & also prevents beta exams from being proctored

## QA

- Buy CUE.02 Desktop exam and schedule it
    - You should see CUE.02 Desktop in exam confirmation thank you screen
- Make sure CUE.01 Linux still shows CUE.01 on thank you page 

## Issue / Card

Fixes [WD-21511](https://warthogs.atlassian.net/browse/WD-21511)

## Help

[WD-21511]: https://warthogs.atlassian.net/browse/WD-21511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/webapp/shop/cred/views.py
+++ b/webapp/shop/cred/views.py
@@ -17,6 +17,10 @@ from webapp.shop.api.datastore import (
     handle_confidentiality_agreement_submission,
     has_filed_confidentiality_agreement,
 )
+from webapp.shop.cred.constants import (
+    TAEXAM_PROC_EXAM_MAPPING,
+    TAEXAM_PROC_STATE,
+)
 from webapp.shop.decorators import (
     credentials_group,
     credentials_admin,
@@ -429,18 +433,11 @@ def check_cred_exam_start_time(
 
 
 def get_taexam_to_procexam_mapping(ta_exam: str) -> int | None:
-    mappings = {
-        "cue-01-linux": 2,
-        "cue-02-desktop": 3,
-    }
-    return mappings.get(ta_exam)
+    return TAEXAM_PROC_EXAM_MAPPING.get(ta_exam)
 
 
 def is_proctoring_enabled(ta_exam: str) -> int | None:
-    mappings = {
-        "cue-01-linux": True,
-    }
-    return mappings.get(ta_exam, False)
+    return TAEXAM_PROC_STATE.get(ta_exam, False)
 
 
 @shop_decorator(area="cred", permission="user", response="html")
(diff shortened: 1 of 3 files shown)


webapp/shop/cred/views.py
```diff
@@ -431,8 +431,16 @@ def check_cred_exam_start_time(
 def get_taexam_to_procexam_mapping(ta_exam: str) -> int | None:
     mappings = {
         "cue-01-linux": 2,
+        "cue-02-desktop": 3,
     }
-    return mappings.get(ta_exam, None)
+    return mappings.get(ta_exam)
+
```
abhigyanghosh30, line 437: Instead of having a duplicate mapping here, can we just not add the desktop exam to proc mapping? This way we don't need to do this extra check. We can enable this once the exam is taken out of beta.

abhigyanghosh30: Or we can even set the value to be -1 for the desktop exam indicating it is not valid and in line 694 the check becomes `get_taexam_to_procexam_mapping(data["ta_exam"])>0`

abhigyanghosh30 (changes requested):

webapp/shop/cred/views.py
```diff
@@ -431,8 +431,16 @@ def check_cred_exam_start_time(
 def get_taexam_to_procexam_mapping(ta_exam: str) -> int | None:
     mappings = {
         "cue-01-linux": 2,
+        "cue-02-desktop": 3,
```
abhigyanghosh30, line 434: Also I'd say to make this a dictionary as a constant at the top of the file or move the constants in a different `settings.py` file. So when we need to make small changes to exams, we don't have to mess with the huge `views.py` file

abhigyanghosh30 (changes requested):

webapp/shop/cred/views.py
usamabinnadeem-10: I separated the TA exam to Proc exam mapping from the fact that some exams are themselves not supposed to have proctoring enabled. There may be a case where we would want to have the actual Proc exam but still not have proctoring enabled for that exam. The -1 approach seems a bit hardcoded.

webapp/shop/cred/views.py
usamabinnadeem-10: I agree.

abhigyanghosh30 (approved):

PR #15276, opened 2025-06-27
feat: remove wordpress json redirect

samhotep:
## Problem
We have a security vulnerability due to an exposed wordpress json endpoint that exposes user information.

## Done

- Removed redirects to admin.insights.ubuntu.com, which is only accessible via the VPN.

## QA

- This branch has been deployed to staging. Open the staging url here https://staging.ubuntu.com/
- Open the staging url and check that blogs pages work without issue https://staging.ubuntu.com/blog.

## Help

### DIFF
--- a/redirects.yaml
+++ b/redirects.yaml
@@ -982,12 +982,6 @@ blog/canonical-announces-ubuntu-22-04-lts-support-for-flexran-reference-software
 blog/intel-compute-stick-now-comes-with-ubuntu: /download/iot/intel-iot
 /blog/cups-remote-code-execution: /blog/cups-remote-code-execution-vulnerability-fix-available
 
-# Copied from https://github.com/canonical-web-and-design/blog.ubuntu.com/blob/master/redirects.yaml
-
-# Wordpress CMS pages & resources
-blog/admin(?P<page>(/.*)?): https://admin.insights.ubuntu.com/admin{page}
-blog/wp-(?P<page>.+): https://admin.insights.ubuntu.com/wp-{page}
-
 # Archive pages
 blog/page/(?P<page>[0-9]+)/?: /blog/archives?page={page}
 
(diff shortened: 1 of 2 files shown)


redirects.yaml
```diff
@@ -982,12 +982,6 @@ blog/canonical-announces-ubuntu-22-04-lts-support-for-flexran-reference-software
 blog/intel-compute-stick-now-comes-with-ubuntu: /download/iot/intel-iot
 /blog/cups-remote-code-execution: /blog/cups-remote-code-execution-vulnerability-fix-available
 
-# Copied from https://github.com/canonical-web-and-design/blog.ubuntu.com/blob/master/redirects.yaml
-
-# Wordpress CMS pages & resources
-blog/admin(?P<page>(/.*)?): https://admin.insights.ubuntu.com/admin{page}
-blog/wp-(?P<page>.+): https://admin.insights.ubuntu.com/wp-{page}
```
Onibenjo: All these return 404 now

Is that the intention?
Because they return 404 now for example https://ubuntu-com-15276.demos.haus/blog/wp-json/wp-ben compared to https://ubuntu.com/blog/wp-json/wp-ben

Onibenjo (commented):
A question

redirects.yaml
akbarkz: indeed, what is the intention of removing these redirects? If all they do is redirecting people to WordPress admin URL.

akbarkz:
https://staging.ubuntu.com/blog/wp-json/wp/v2/ is still returning JSON. Is that due to content-cache?

samhotep:
> https://staging.ubuntu.com/blog/wp-json/wp/v2/ is still returning JSON. Is that due to content-cache?

Yes, it should be redirecting to admin.insights now.

redirects.yaml
samhotep: You're right. The redirect will still need the VPN to work in any case

Onibenjo (approved):
LGTM

PR #15409, opened 2025-08-04
fix size issue on chart

abhigyanghosh30:
## Done

- Fix size issue on Pro CVE chart

## QA

- Visit [/pro/linux-patch-management-with-pro](https://ubuntu-com-15409.demos.haus/pro/linux-patch-management-with-pro)

## Issue / Card

Fixes [#WD-24405](https://warthogs.atlassian.net/browse/WD-24405)

## Screenshots

image

## Help

### DIFF
--- a/templates/pro/linux-patch-management-with-pro/bar-chart.html
+++ b/templates/pro/linux-patch-management-with-pro/bar-chart.html
@@ -17,8 +17,6 @@
     margin-left: 32px;
   }
   #myChart {
-    max-width: 100%;
-    height: 400px;
     margin: inherit;
   }
 </style>


minkyngkm:
@abhigyanghosh30 on medium screens
image

abhigyanghosh30:
> @abhigyanghosh30 on medium screens image


@minkyngkm Can you try reloading the page? it works for me

image

minkyngkm:
Can we add a min-width? It becomes too small and only displays fine after reloading.

abhigyanghosh30:
> Can we add a min-width? It becomes too small and only displays fine after reloading.

Min width doesn't fix the issue. Once you shrink the viewport width and then stretch it again, the chart stays at the shrunk state. Seems like this might be a javascript library issue in chart.js

terameta (dismissed):
LGTM. Thank you for your contribution

abhigyanghosh30:
> image
> After shrinking the width of the page chart disappears. If I again enlarge the screen, it never shows up again. Also, if we are hiding chart on small screens, we shouldn't show related copy of the chart.

image
it is supposed to show on small screens as well

abhigyanghosh30:
@terameta @minkyngkm I think I found a fix. It seems that chart.js doesnot like `margin: 0 auto`; The best way to override the value coming from Vanilla was to just inherit the margin from the parent which hopefully would have a sensible margin and it seems to work

Issue #13106, opened 2023-08-24
"Components certified with this system" table doesn't have legends for `[x]`


nobuto-m:
Forwarded from:
https://answers.launchpad.net/ubuntu-certification/+question/707714

For example in this page:
https://ubuntu.com/certified/201908-27313
There is a table as "Components certified with this system".

However, there is no definition or legend for [x] although the definition of [✓] or [?] exists. A potential question would be "what's the difference between being blank [ ] and [x]? Does [x] mean the system failed in the certification explicitly?".

![image](https://github.com/canonical/ubuntu.com/assets/4356209/12364b8e-6b29-4a9e-b3d3-cf5ac291630b)

---
*Reported from: https://ubuntu.com/certified/201908-27313*

mtruj013:
@eliman11 could you have a look at this please?

eliman11:
> [@eliman11](https://github.com/eliman11) could you have a look at this please?

I think this page doesn't fall under our scope, so tagging @leiaru-c here instead :)

eliman11:
Thanks for the feedback @nobuto-m! @tang-mm and I worked on this new design for the table to display the icon labels inside the table itself and included a footnote to explain the blank cells. We will also be dropping the checked icon with the superscript "3" as the text is too small to be legible and differentiable from the other checked icon. This will be replaced by the checked icon, with a different label to clarify the meaning. 

<img width="747" height="394" alt="Image" src="https://github.com/user-attachments/assets/aa0dd92a-b6f3-4e20-9257-0bd9e3c6a947" />

@Onibenjo @britneywwc Could you please pick up the changes on Figma [here](https://www.figma.com/design/1gDRefkvA5TnZGWI9tdkZB/ubuntu.com-certified?node-id=675-11535)? Please disregard the PR that was opened earlier as the design has changed since. Let me know if you have any questions - thank you!

Onibenjo:
Thank you @eliman11 
Taking a look into this

Onibenjo:
Closing this as a fix has been merged and deployed.
Thanks for raising.

Issue #13410, opened 2023-12-14
"Components certified with this system" do not show Ubuntu Core compatibility


pieqq:
## Summary

Some certified devices have additional certified components listed at the bottom of their page.

The table that lists the compatibility matrix only mentions classic versions (18.04, 22.04, etc.). There is no mention of Ubuntu Core releases.

## Process

Go to a certified device page that shows an additional component, for instance:

https://ubuntu.com/certified/202302-31257

Click on the component to open its page:

https://ubuntu.com/certified/component/992

→ the component is marked as certified for Core 20.

### Current result

In the "Components certified with this system" section of [the page](https://ubuntu.com/certified/202302-31257):

- There is no mention of Ubuntu Core 20
- There is a red cross under 22.04 for no reason

![image](https://github.com/canonical/ubuntu.com/assets/512471/8ba8aa7b-8e43-4265-9be6-727e7c0a0ad7)

### Expected result

- The table should include Ubuntu Core releases (20, 22...)
- The UC20 column should be marked as supported (green checkmark)
- 22.04 should not be marked with a red cross (which probably means "unsupported", but it's not explained, see #13106)

## Browser details

Not relevant, this happens on the backend.

tang-mm:
Closing this issue as a fix has been implemented by https://github.com/canonical/ubuntu.com/pull/15456. Thanks for reporting it!

The new table now shows the expected result: https://ubuntu.com/certified/202302-31257

<img width="1167" height="548" alt="Image" src="https://github.com/user-attachments/assets/69d5ac6d-5b1e-41db-b49a-ab95758ad42b" />

PR #15456, opened 2025-08-12
feat(certified): update components table layout and styling

Onibenjo:
## Done
- Restructure components table to show releases vertically
- Simplify table styling and remove redundant columns
- Add demo data for testing purposes
- Add explicit "Unsupported" status handling for each LTS release: Unsupported status should only be shown when the status is explicitly set in the database. If no status is set, the cell should be blank.

[Figma](https://www.figma.com/design/1gDRefkvA5TnZGWI9tdkZB/ubuntu.com-certified?node-id=675-11535) 
[Demo](https://ubuntu-com-15456.demos.haus/certified/202412-36076)

## QA

- Check out the [Demo](https://ubuntu-com-15456.demos.haus/certified/202412-36076)
- Check that it follows the design on [Figma](https://www.figma.com/design/1gDRefkvA5TnZGWI9tdkZB/ubuntu.com-certified?node-id=675-11535) 
- Check that statuses show on the table correctly

## Issue / Card

Fixes # https://warthogs.atlassian.net/browse/WD-18361
#13106 

## Screenshots

## Help

### DIFF
--- a/templates/certified/model-details/_components-section.html
+++ b/templates/certified/model-details/_components-section.html
@@ -72,13 +72,5 @@
       <hr class="p-rule" />
       <p class="u-text--muted">*Releases with a blank status have not been tested</p>
     </div>
-    <!-- <ul class="p-inline-list u-hide--medium u-hide--large">
-      <li class="p-inline-list__item">
-        <i class="p-icon--success"></i> <small class="p-">Supported</small>
-      </li>
-      <li class="p-inline-list__item">
-        <i class="p-icon--help"></i> <small>In progress</small>
-      </li>
-    </ul> -->
   </div>
 </section>
--- a/webapp/certified/views.py
+++ b/webapp/certified/views.py
@@ -622,16 +622,6 @@ def certified_model_details(canonical_id):
         "form_factor", model_release["category"]
     )
 
-    # TODO remove (this is for demo)
-    component_summaries[0]["lts_releases"] = {
-        '20.04 LTS': [{'release': '20.04 LTS', 'third_party_driver': True, 'status': 'certified'}],
-        '22.04 LTS': [{'release': '22.04 LTS', 'third_party_driver': False, 'status': 'inprogress'}],
-        '18.04 LTS': [{'release': '18.04 LTS', 'third_party_driver': False, 'status': 'certified'}],
-        'Core 18': [{'release': 'Core 18', 'third_party_driver': True, 'status': 'unsupported'}],
-        'Core 24': [{'release': 'Core 24', 'third_party_driver': False, 'status': 'unsupported'}],
-        'Core 20': [{'release': 'Core 20', 'third_party_driver': True, 'status': 'inprogress'}]
-    }
-
     return render_template(
         "certified/model-details.html",
         canonical_id=canonical_id,


Onibenjo:
@eliman11 Can you take a look into this for review.

Some questions:
- On production, on mobile, the table turns to a card but for the demo I made them scrollable. Is that okay?
- For the comments, there's only one comment per component. How should they be displayed? All or on only 1 row
image

eliman11:
Thanks @Onibenjo for getting on this so quickly! I'll tag @tang-mm here for input as well. My two cents - on some screens the comments column doesn't appear at all on mobile, so it's not immediately obvious that the table can be scrolled horizontally. Is it possible to condense the widths of the columns even further (or more specifically the first column) such that the at least part of the comments column will always be visible? 
Screenshot 2025-08-12 at 14 51 54

I think it makes sense to repeat the comments if there is a possibility for us to input several comments for each release version within a component. But if we can only input one comment per component, could you please keep it for the first row and remove the rules between the rows? Will also wait for @tang-mm to weigh in here!
Screenshot 2025-08-12 at 14 59 50

Also some other comments - 
- [x] Are we able to use the checked icon for the "May require third-party driver May require 3rd-party driver" status as well? The icon used is pixelated and the superscript 3 is too small to be legible. It should be alright to keep the icons the same, as the status label next to it will differentiate the two statuses
- [x] Could we also ensure that the "Comments" heading in the table doesn't get truncated on smaller screen sizes?

tang-mm:
Thanks for adding the changes quickly @Onibenjo @eliman11 ! 

In the data model, each component is associated with a single comment rather than each release, so only one comment will exist. It makes sense to keep the comment in the first row (merged row) like the name field. 

And regarding the third-party icon:
I just realized that it's also possible to have that note on other statuses: "_Unsupported - may require third-party driver_" or  "_In progress - may require third-party driver_".  The data fields `third_party_driver` and `status` are independent. In our internal records, we use superscript for both the checked and crossed icons. However on the previous table, this was another unhandled and misleading information. I apologize for not clarifying this earlier
image

tang-mm:
And to address the issue https://github.com/canonical/ubuntu.com/issues/13410 at the same time, could you please also add _Core 18_.._Core 24_ as new rows into the same table in the HTML template?

eliman11:
> I just realized that it's also possible to have that note on other statuses: "_Unsupported - may require third-party driver_" or "_In progress - may require third-party driver_". The data fields `third_party_driver` and `status` are independent. In our internal records, we use superscript for both the checked and crossed icons. However on the previous table, this was another unhandled and misleading information. I apologize for not clarifying this earlier

I assume these labels can be added to the table as well @Onibenjo? Could we just keep 3 icons - the check, cross, and question mark? So "Unsupported - may require third-party driver" should have a cross icon next to it , and "In progress - may..." would have a question mark.

Onibenjo:
@eliman11 Can you take a look again?
I've updated how the statuses are displayed

eliman11:
Thanks for your work on this @Onibenjo! It looks good to me! @tang-mm did you want to have another look in case you had any additional comments?

templates/certified/model-details/_components-section.html
```diff
@@ -1,113 +1,84 @@
...
+          <tr>
+            <th scope="col">Component</th>
+            <th scope="col">Ubuntu release</th>
+            <th scope="col">Status</th>
+            <th scope="col">Comments</th>
+          </tr>
+        </thead>
+
+        <tbody>
+          {% set releases = ["24.04 LTS", "22.04 LTS", "20.04 LTS", "CORE 20", "18.04 LTS", "CORE 18"] %}
```
tang-mm:
```suggestion
          {% set releases = ["24.04 LTS", "Core 24",  "22.04 LTS", "Core 22", "20.04 LTS", "Core 20", "18.04 LTS", "Core 18"] %}
```
Add more Ubuntu Core releases

tang-mm: Another option for the ordering is 
`["24.04 LTS", "22.04 LTS", "20.04 LTS", "18.04 LTS", "Core 24",  "Core 22",  "Core 20", "Core 18"] `

@eliman11 Could you please advise whether it's clearer to list all LTS first before Core releases, or alternatively?

templates/certified/model-details/_components-section.html
```diff
@@ -1,113 +1,84 @@
...
+                {% endif %}
+              </tr>
+            {% endfor %}
+          {% endfor %}
+        </tbody>
+      </table>
+    </div>
+    <div>
+      <hr class="p-rule" />
+      <p class="u-text--muted">*LTS releases with a blank status have not been tested</p>
```
tang-mm:
```suggestion
      <p class="u-text--muted">*Releases with a blank status have not been tested</p>
```
This applies to both LTS and Core releases

tang-mm (commented):
Thank you @Onibenjo ! The preview looks good to me in general, just two minor edit suggestions

templates/certified/model-details/_components-section.html
Onibenjo: @eliman11 What's your take on this?

eliman11: Hi @tang-mm please correct me if I'm wrong as you're more familiar with the product than I am - but I would imagine it's clearer to list LTSs first before Core. I would defer this decision to you and Matias though!

samhotep:
hey @Onibenjo some checks are failing

samhotep (approved):
LGTM!

PR #15487, opened 2025-08-19
Feat: added tests for view functions

immortalcodes:
## Done

- Added tests for functions within views.py
- While identifying the functions to unit test, following criteria was considered:
     - Do they have solid block of logic which would benefit from tests.
     - Leaving out functions that are dependent on too many services and benefit more from integration testing.
- Aim of the task was to increase the coverage by introducing tests for more functions, we still could add more testing to views.

## QA

- Use `dotrun test-python` to run the tests locally.
- You can replace command with `python3 -W ignore -m unittest discover -s tests -v` to get a verbose description of running tests.
- You might need some api keys or secrets. Let me know if something is needed.

## Issue / Card

Fixes # [WD-22316](https://warthogs.atlassian.net/browse/WD-22316)

## Screenshots

## Help

[WD-22316]: https://warthogs.atlassian.net/browse/WD-22316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/test_views.py
+++ b/tests/test_views.py
@@ -381,6 +381,7 @@ class TestBuildingTutorialsQuery(BaseViewTestCase):
       • The `created_at` field is returned unchanged and
         difficulty levels match the input.
       • When no topic parameter is supplied, an empty list is returned.
+      • When tutorials are missing expected keys like `categories`
     """
 
     def _make_tutorials_docs(self, tutorials):
@@ -481,6 +482,51 @@ class TestBuildingTutorialsQuery(BaseViewTestCase):
             resp = view()  # Flask Response
             data = resp.get_json()  # directly parse JSON
         self.assertEqual(len(data), 0)
+
+    def test_building_tutorials_query_missing_keys(self):
+        tutorials = [
+            {
+                "title": "Tutorial 1",
+                "slug": "tutorial-1",
+                "author": "Author 1",  # missing categories
+                "difficulty": 3,
+                "created_at": "2023-01-01T00:00:00Z",
+            },
+            {
+                "title": "Tutorial 2",
+                "slug": "tutorial-2",
+                "categories": ["c2"],
+                "author": "Author 2",  # missing difficulty
+                "created_at": "2023-02-01T00:00:00Z",
+            },
+            {
+                "title": "Tutorial 3",
+                "slug": "tutorial-3",
+                "author": "Author 3",  # missing categories and difficulty
+                "created_at": "2023-03-01T00:00:00Z",
+            },
+            {
+                "title": "Tutorial 4",
+                "slug": "tutorial-4",
+                "categories": ["c2"],
+                "author": "Author 4",
+                "difficulty": 2,
+                "created_at": "2024-02-01T00:00:00Z",
+            },
+        ]
+        docs = self._make_tutorials_docs(tutorials)
+
+        with self.app.test_request_context("/tutorials?topic=c2"):
+            view = build_tutorials_query(tutorials_docs=docs)
+            resp = view()  # Flask Response
+            data = resp.get_json()  # directly parse JSON
+        self.assertEqual(len(data), 2)
+        self.assertEqual(data[0]["title"], "Tutorial 4")
+        self.assertEqual(data[1]["title"], "Tutorial 2")
+
+        # Check that the created_at field is parsed correctly
+        self.assertEqual(data[0]["created_at"], "2024-02-01T00:00:00Z")
+        self.assertEqual(data[1]["created_at"], "2023-02-01T00:00:00Z")
 
 
 class TestMatchTags(TestCase):
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -660,11 +660,13 @@ def build_tutorials_query(tutorials_docs):
         tutorials = [
             doc
             for doc in tutorials_docs.parser.tutorials
-            if topic in doc["categories"]
+            if topic in doc.get("categories", [])
         ]
 
         tutorials = sorted(
-            tutorials, key=lambda k: k["difficulty"], reverse=True
+            tutorials,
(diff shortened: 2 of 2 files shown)


tests/test_views.py
```diff
@@ -48,72 +59,357 @@ def test_shorten_acquisition_url(self):
...
+                "mirror_bandwidth": "1000",
+                "mirror_countrycode": "GB",
+            },
+            {
+                "link": "https://mirror2.example.com/ubuntu/",
+                "mirror_bandwidth": "2000",
+                "mirror_countrycode": "US",
+            },
+        ]
+        mock_parse.return_value = type("F", (), {"entries": mirrors})()
```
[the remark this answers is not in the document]
britneywwc: Seconding this suggestion to improve test readability

tests/test_views.py
```diff
@@ -48,72 +59,357 @@ def test_shorten_acquisition_url(self):
...
+        )
+
+    def test_missing_architecture_aborts(self):
+        from werkzeug.exceptions import HTTPException
+
+        with self.app.test_request_context("/?version=22.04"):
+            with self.assertRaises(HTTPException) as cm:
+                download_thank_you("desktop")
+        self.assertEqual(cm.exception.code, 400)
+
```
britneywwc, line 293: We should also add a test for when architecture is present but no version is passed

tests/test_views.py
```diff
@@ -48,72 +59,357 @@ def test_shorten_acquisition_url(self):
...
+            },
+        )
+
+    def test_account_query_no_user(self):
+        with self.app.test_request_context("/any-url"):
+            resp = account_query()
+        self.assertEqual(resp.status_code, 200)
+        self.assertEqual(resp.get_json(), {"account": None})
+
+class TestBuildingTutorialsQuery(BaseViewTestCase):
```
britneywwc, line 371: Please add a docstring to indicate which methods are being tested in the class

tests/test_views.py
```diff
@@ -48,72 +59,357 @@ def test_shorten_acquisition_url(self):
...
+                "difficulty": 2,
+                "created_at": "2023-02-01T00:00:00Z",
+            },
+            {
+                "title": "Tutorial 3",
+                "slug": "tutorial-3",
+                "categories": ["c1"],
+                "author": "Author 3",
+                "difficulty": 1,
+                "created_at": "2023-03-01T00:00:00Z",
```
britneywwc, line 433: We can also add a test for an unexpected tutorial field and check if the index builds as expected without the invalid entry
```suggestion
                "created_at": "2023-03-01T00:00:00Z",
                "unexpected_field": "test"
```

tests/test_views.py
```diff
@@ -48,72 +59,357 @@ def test_shorten_acquisition_url(self):
...
+        ]
+        docs = self._make_tutorials_docs(tutorials)
+
+        with self.app.test_request_context("/tutorials"):
+            view = build_tutorials_query(tutorials_docs=docs)
+            resp = view()  # Flask Response
+            data = resp.get_json()  # directly parse JSON
+        self.assertEqual(len(data), 0)
+
+class TestMatchTags(TestCase):
```
britneywwc, line 532: Please add docstring here to indicate which methods are being tested

britneywwc (commented):
Thank you for this piece of work!

tests/test_views.py
immortalcodes: The view function is only using categories and difficulty and hence all other fields already act us unexpected fields and it still builds without issues so that automatically gets tested

britneywwc: In that case we should add tests for when `categories` and/or `difficulty` fields are not present

.github/workflows/pr.yaml
```diff
@@ -160,7 +160,7 @@ jobs:
 
       - name: Run tests and generate coverage report
         run: |
-          coverage run --source=. -m unittest $(find tests/ -name "test_*.py" ! -name "test_marketo.py")
+          coverage run --source=webapp -m unittest $(find tests/ -name "test_*.py" ! -name "test_marketo.py")
```
britneywwc, line 163: The [codecov report](https://app.codecov.io/gh/canonical/ubuntu.com/pull/15487) still shows a decrease in coverage, I'm guessing that it should be an increase due to the added tests

immortalcodes: The reason is quite interesting, 
- The current codecov report compares the coverage between base that had all directories and current PR that just has webapp
- Test function automatically cover themselves.
- This means that when we include tests in coverage that means we include  percentage in coverage that is self-tested and hence gives us a false idea of our code coverage.

Coverage without tests on main branch
image
Coverage with tests directory on main branch
image
image

immortalcodes:
https://github.com/canonical/ubuntu.com/pull/15487#discussion_r2339126532

Regarding this comment, keyerrors are not being handled within the function so we would not be actually testing any intended good/bad logic of function. It is just a lack of implementation in the function.
What do you think?

britneywwc:
> [#15487 (comment)](https://github.com/canonical/ubuntu.com/pull/15487#discussion_r2339126532)
> 
> Regarding this comment, keyerrors are not being handled within the function so we would not be actually testing any intended good/bad logic of function. It is just a lack of implementation in the function. What do you think?


I would suggest to add a KeyError handling as a drive-by so that we could catch any existing/future pages that may be failing

.github/workflows/pr.yaml
britneywwc: Ah I see, the tests are giving a false percentage as they are all marked as 100% covered. Since the scope of this work is adding tests to the views function, it seems like the coverage has increased from 42% to 44%. 

Nice work! 🎉

PR #15495, opened 2025-08-20
Update notice sitemap logic

mtruj013:
## Done

- [List of work items including drive-bys - remember to add the why and what of this work.]

## QA

## Issue / Card

Fixes #

## Screenshots

## Help

### DIFF
--- a/webapp/security/api.py
+++ b/webapp/security/api.py
@@ -180,3 +180,22 @@ class SecurityAPI:
             raise SecurityAPIError(error)
 
         return sitemap_response.json()
+
+    def get_sitemap_notices(self, limit, offset):
+
+        parameters = {"limit": int(limit), "offset": int(offset)}
+
+        parameters = {k: v for k, v in parameters.items() if v is not None}
+
+        filtered_parameters = urlencode(parameters, doseq=True)
+
+        try:
+            sitemap_response = self._get(
+                f"sitemap/notices.json?{filtered_parameters}"
+            )
+        except HTTPError as error:
+            if error.response.status_code == 404:
+                return None
+            raise SecurityAPIError(error)
+
+        return sitemap_response.json()
--- a/webapp/security/views.py
+++ b/webapp/security/views.py
@@ -300,42 +300,19 @@ def notices_feed(feed_type):
 # USN API
 # ===
 def single_notices_sitemap(offset):
-    notices = []
-    max_items = 100
-    api_limit = 10
     offset = int(offset)  # Convert offset to int
 
     # max limit is 10, so we need to make multiple requests up to 100
-    for batch_offset in range(offset, offset + max_items, api_limit):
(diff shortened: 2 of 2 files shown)


konf/site.yaml
```diff
@@ -1015,6 +1017,8 @@ staging:
           /security/released/cves/.*\.json,
           /security/sitemap/cves\.json,
           /security/sitemap/cves/.*\.json,
+          /security/sitemap/notices\.json,
+          /security/sitemap/notices/.*\.json,
```
immortalcodes, line 1021: What is the purpose of adding the paths here.

webapp/security/api.py
```diff
@@ -180,3 +180,22 @@ def get_sitemap_cves(self, limit, offset):
             raise SecurityAPIError(error)
 
         return sitemap_response.json()
+
+    def get_sitemap_notices(self, limit, offset):
+
+        parameters = {"limit": int(limit), "offset": int(offset)}
+
+        parameters = {k: v for k, v in parameters.items() if v is not None}
```
immortalcodes:
```suggestion
parameters = {}
if limit is not None:
    parameters["limit"] = int(limit)
if offset is not None:
    parameters["offset"] = int(offset)
```

Since int(None) will raise an error.

webapp/security/views.py
```diff
@@ -300,42 +300,19 @@ def feed_entry(notice, url_root):
...
-        batch_notices = batch_response.get("notices", [])
-
-        if not batch_notices:
-            break
-
-        notices.extend(batch_notices)
-
-        if len(batch_notices) < api_limit:
-            break
+    response = security_api.get_sitemap_notices(limit=100, offset=offset)
```
immortalcodes, line 305: If limit is something that we define and it doesn't come from URL, we shall embed it with in the `get_sitemap_notices` function

immortalcodes: Or we can have a default value.

webapp/security/api.py
immortalcodes: I would also leave a comment or a check that limit has to be below 100

konf/site.yaml
mtruj013: Good question, all sec api endpoints are served under the u.com base path so if we don't add these here the url won't resolve to the intended endpoint

webapp/security/api.py
mtruj013: Nice catch, thanks! About the limit, it actually doesn't need to be below 100 anymore. Being able to increase the limit without facing performance issues was one of the reasons behind having a dedicated endpoint. It only returns 2 fields and neither are heavy so it handles a high limit without issue

webapp/security/views.py
mtruj013: Makes sense, updated!

mtruj013:
Thanks @immortalcodes , could you have another look?

PR #15655, opened 2025-09-23
chore: Update cve chip markup

mtruj013:
## Done

- ~~Updated CVE chips to be a tags instead of buttons as intented in the [original design ](https://www.figma.com/design/nUJEBpOPTW74RqDRuMQXxB/24.04-ubuntu.com---CVEs---Sites?node-id=949-22691&m=dev)~~  Aligned CVE chip functionality with how it's done on USN pages

## QA

- View the site locally in your web browser at: https://ubuntu-com-15655.demos.haus/security/CVE-2024-44975#status
    - Scroll through the table and see that the chips retain the intended style and that they route to the correct page on click

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-14904

### DIFF
--- a/templates/security/cves/cve.html
+++ b/templates/security/cves/cve.html
@@ -207,7 +207,7 @@
                       aria-controls="drawer">Toggle table of contents</button>
             </div>
             <ul>
-              {% if cve.description %}
+              {% if cve.description or cve.ubuntu_description%}
                 <li class="p-side-navigation__item">
                   <a class="highlight-link is-active" href="#description">Description</a>
                 </li>
@@ -244,7 +244,10 @@
 
       <main class="col-9">
         <div class="p-section">
-          <h2 id="description" class="section-heading">Description</h2>
+          {% if cve.description or cve.ubuntu_description %}
+            <h2 id="description" class="section-heading">Description</h2>
+          {% endif %}
+
           {% if cve.description %}<p>{{ cve.description }}</p>{% endif %}
 
           {% if cve.ubuntu_description %}


templates/security/cves/cve.html
```diff
@@ -315,14 +315,14 @@ <h2>Status</h2>
                             <span class="u-text--muted">{{ status.release_codename }}</span>
                             {% if status.pocket_desc and status.pocket_desc.label != "Ubuntu Pro" %}
                               &nbsp;
-                              <button class="p-chip p-tooltip--top-center"
-                                      aria-describedby="{{ status.pocket }}-tooltip"
-                                      style="vertical-align: unset">
+                              <a class="p-chip p-tooltip--top-center u-no-margin--bottom"
+                                 aria-describedby="{{ status.pocket }}-tooltip"
+                                 href="{{ status.pocket_desc.href }}">
```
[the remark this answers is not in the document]
mtruj013: aria-describedby is already handling this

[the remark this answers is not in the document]
mtruj013: Keyboard navigation is working as expected for both hover and click state

britneywwc (dismissed):
Looks good, thank you! Just a tiny question about the tooltip. Now that the entire tooltip is an anchor element, should there be any additional styling to indicate that it is an interact-able element? CC @eliman11 as well

eliman11:
> Looks good, thank you! Just a tiny question about the tooltip. Now that the entire tooltip is an anchor element, should there be any additional styling to indicate that it is an interact-able element? CC @eliman11 as well

Sorry the demo is throwing an error

eliman11:
Hi @mtruj013, @britneywwc sent me screenshots via MM and I was wondering if the removal of the links in the tooltips was intentional? If it wasn't in the CVE tooltips before, I think it might've been added because of the updates we made to the tooltips in the notices page. IIRC the security team asked for those links to be added in the tooltips for the notices, so I just wanted to check if we intentionally didn't want them only on notices pages and not on CVEs?

mtruj013:
@eliman11 according to the [original design](https://www.figma.com/design/nUJEBpOPTW74RqDRuMQXxB/24.04-ubuntu.com---CVEs---Sites?node-id=949-22691&m=dev) the links were never supposed to be in the tooltips, the chips themselves were supposed to be the links. We settled on having them in the tooltips instead because at the time there was a vanilla bug when using the `p-chip` component on an `a` tag. Since the relevant [vanilla ticket](https://github.com/canonical/vanilla-framework/issues/5317) that was blocking this has been addressed, we can update to serve the intended design

@britneywwc re: additional styling to indicate that it is an interact-able element
I might be misunderstanding the question, but on both chrome and ff the cursor updates on hover indicating that it's an interactive element:

https://github.com/user-attachments/assets/8e15bffd-0597-4ae0-9715-3d98e5c8d1a0

I'd even argue that the cursor change on hover for non-interactive chips (ie, how the chips currently behave on prod) is a bug

eliman11:
> @eliman11 according to the [original design](https://www.figma.com/design/nUJEBpOPTW74RqDRuMQXxB/24.04-ubuntu.com---CVEs---Sites?node-id=949-22691&m=dev) the links were never supposed to be in the tooltips, the chips themselves were supposed to be the links. We settled on having them in the tooltips instead because at the time there was a vanilla bug when using the `p-chip` component on an `a` tag. Since the relevant [vanilla ticket](https://github.com/canonical/vanilla-framework/issues/5317) that was blocking this has been addressed, we can update to serve the intended design

Ah right thank you for the additional context! I just remembered we (or it might've been me and Britney, I can't recall) had a similar conversation about how having the chip as an actual link means that the tooltip wouldn't be visible to users on mobile and tablet, as the click would redirect users to another page rather than triggering the tooltip, which is why we opted for the links within the tooltip itself.

eliman11:
Hi @mtruj013 I messaged Juan about this and he agreed we can use the same approach we used for the USN individual pages ([example](https://ubuntu.com/security/notices/USN-7769-2)), where the chips open the tooltip on click and the links are placed inside the tooltip instead of being linked to the chip. I made this change on the notices change because the tooltip can't be triggered on hover on mobile and tablet, and so the tooltips would show for a second before the user is redirected. By having the tooltips trigger on click instead, it allows mobile and tablet users to also view the tooltip content as users on desktop would.

As we have it on the USN page, could we ensure that only 1 tooltip can be open at a time? So if you click into another chip, if a tooltip is already opened, it would close and the new one would open instead. Also if the user clicks out of the tooltip, it will close. Please let me know if this works!

mtruj013:
> As we have it on the USN page, could we ensure that only 1 tooltip can be open at a time? So if you click into another chip, if a tooltip is already opened, it would close and the new one would open instead. Also if the user clicks out of the tooltip, it will close.

Done! Vid here since demos still aren't working:

https://github.com/user-attachments/assets/980e95d0-3148-4e68-811a-25d3bd4ebd07

eliman11:
LGTM, thanks for your work and additional context on this task @mtruj013!

templates/security/cves/cve.html
```diff
@@ -703,6 +703,49 @@ <h3>Other references</h3>
       });
     };
     setUpDynamicSideNav();
+
+    function handleTooltips() {
```
britneywwc: There are existing scripts for `handleTooltips` and `setUpDynamicSideNav` that is also used in the other individual notices pages. We should reuse these scripts if applicable. 

```
<script src="{{ versioned_static('js/dist/dynamic-toc.js') }}" defer></script>
<script src="{{ versioned_static('js/dist/notices.js') }}" defer></script>
```

britneywwc (commented):
I'm noticing that the dynamic TOC does not highlight the section as intended. E.g on http://0.0.0.0:8001/security/CVE-2024-44975#status, the TOC is still highlighted on **Description**.

Perhaps importing the `dynamic-toc.js` might fix this bug

templates/security/cves/cve.html
mtruj013: The highlighting thing is a really weird bug, went down a bit of a rabbit hole with this. It doesn't seem like it's consistent (some pages have it some don't, and it's also not consistent across specific anchors). Using dynamic-toc.js fixes it but only if the relevant ids and `section-heading` classes are added to headings (doesn't work on divs or p tags), which is also quite strange as the two versions of the script are almost identical except for the `rootMargin` and there doesn't seem to be anything in the code that suggests that the anchors/classes _must_ be applied to semantic headings 🤔 

That said, seems like a pretty easy fix to just add a heading for that section so that's what I've done. @eliman11 lmk if you disagree, looks like this vs original design [here](https://www.figma.com/design/nUJEBpOPTW74RqDRuMQXxB/24.04-ubuntu.com---CVEs---Sites?node-id=949-23124&m=dev):

image

britneywwc (approved):
LGTM 🚀  Thanks for making the changes!

templates/security/cves/cve.html
eliman11: That makes sense to me, thanks both for flagging!

PR #15656, opened 2025-09-23
 Fix Error Handling in Certified Platform Details Functions

Onibenjo:
### Problem
The certified_platform_details and certified_platform_details_by_release functions were incorrectly handling API errors, causing all HTTP errors (including 404s) to be returned as 500 Internal Server Errors instead of the appropriate HTTP status codes. This resulted in poor user experience and incorrect error reporting on Sentry.

Before: When the certified platforms API returned a 404 (platform not found), users would see a generic 500 error page instead of the proper 404 page.

## Done
- Refactored Error Handling Logic : Created a centralized handle_api_error helper function in webapp/certified/helpers.py that properly handles different types of API errors:
  - HTTP 404 errors : Now correctly return abort(404) to show the proper "Not Found" page
  - Other HTTP errors : Return abort(500) with Sentry logging for debugging
  - General exceptions : Return abort(500) with Sentry logging
- Improved User Experience : Users now see appropriate error pages:
  - 404 pages when a platform doesn't exist
  - 500 pages only for actual server errors

## QA

- Check out the links below
  - https://ubuntu-com-15656.demos.haus/certified/platforms/12196/22.04%20LTS
    - It should show a 404 page
  - compare that with production http://ubuntu.com/certified/platforms/12196/22.04%20LTS
  - Check that other pages load correctly
    - https://ubuntu-com-15656.demos.haus/certified/platforms/14169/22.04%20LTS

## Issue / Card

Fixes #

## Screenshots

## Help

### DIFF
--- a/webapp/certified/helpers.py
+++ b/webapp/certified/helpers.py
@@ -1,9 +1,7 @@
 import html
 import bleach
 import markdown
-import requests
 from markupsafe import Markup
-from flask import current_app, abort
 
 
 def get_download_url(model_details):
@@ -122,24 +120,3 @@ def _get_category_pathname(form_factor):
         return "socs"
     else:
         return form_factor.lower() + "s"
-
-
-def handle_api_error(api_call):
-    """
-    Centralized error handling for API calls in certified views.
-
-    :param api_call: A callable that makes the API request
-    :return: The result of the API call if successful
-    :raises: Aborts with appropriate HTTP status codes on error
-    """
-    try:
-        return api_call()
-    except requests.exceptions.HTTPError as error:
-        if error.response.status_code == 404:
-            abort(404)
-        else:
-            current_app.extensions["sentry"].captureException()
-            abort(500)
-    except Exception:
-        current_app.extensions["sentry"].captureException()
-        abort(500)
--- a/webapp/certified/views.py
+++ b/webapp/certified/views.py
@@ -1,5 +1,4 @@
 import talisker.requests
-import talisker.sentry
 import requests
 import math
 
@@ -19,7 +18,6 @@ from urllib.parse import urlencode
 from webapp.certified.helpers import (
     _get_category_pathname,
     get_download_url,
-    handle_api_error,
 )
 
 session = Session()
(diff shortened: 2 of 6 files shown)


webapp/certified/views.py
```diff
@@ -159,7 +159,19 @@ def _parse_query_params(all_releases, all_vendors):
...
+
+    except requests.exceptions.HTTPError as error:
+        if error.response.status_code == 404:
+            abort(404)
+        else:
+            current_app.extensions["sentry"].captureException()
+            abort(500)
+    except Exception as error:
+        current_app.extensions["sentry"].captureException()
+        abort(500)
```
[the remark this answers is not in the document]
Onibenjo: Fixed

webapp/certified/helpers.py
```diff
@@ -120,3 +122,24 @@ def _get_category_pathname(form_factor):
         return "socs"
     else:
         return form_factor.lower() + "s"
+
+
+def handle_api_error(api_call):
```
samhotep: @Onibenjo this is great! Would this function be better off as a decorator that can be applied whereever we need this error handling (perhaps whereever there's a possible HTTPError)?

webapp/certified/views.py
```diff
@@ -177,39 +184,52 @@ def certified_platform_details(platform_id):
...
-        if release in certificate["releases"]
-    }
-    return render_template(
-        "certified/platforms/platform-details.html",
-        category_pathname=_get_category_pathname(platform["category"]),
-        platform=platform,
-        releases=releases,
-        selected_release=release,
-    )
+    except Exception:
```
samhotep: If we use a decorator, we won't have to catch this error again, not so?

webapp/certified/views.py
```diff
@@ -177,39 +184,52 @@ def certified_platform_details(platform_id):
...
+                ),
+                platform=platform,
+                releases=releases,
+                selected_release=None,
+            )
+
+        # Filter only certificates for the release specified in the URL
+        platform["certificates"] = {
+            canonical_id: certificate
+            for canonical_id, certificate in platform["certificates"].items()
```
samhotep: We should be careful with accessing the values here, in case `platform["certificates"]` returns `None` this code could break.
It's worth first checking to see that `platform["certificates"]` is set and is a dict

webapp/certified/views.py
```diff
@@ -177,39 +184,52 @@ def certified_platform_details(platform_id):
...
     )
 
-    # If the release specified in the URL is not available for this platform,
-    # render the page for all releases
-    if release not in releases:
+    try:
+        # Get the set of all releases available for this platform
+        releases = set(
+            release
+            for _, certificate in platform["certificates"].items()
```
samhotep: same comment as on line `217`

webapp/certified/views.py
Onibenjo: ah yes.
This was the current implementation. I will update it to do some error checks

webapp/certified/views.py
Onibenjo: moved to decorator now

webapp/certified/views.py
Onibenjo: added checks for this now

Onibenjo:
@samhotep I've addressed comments
You can take another look again
Thanks

samhotep (approved):
LGTM!

Issue #15576, opened 2025-09-03
Issues with screen readers on Ubuntu website.


edr-xix:
### Summary

When using the website, the menu is inaccessible with VoiceOver.
The options presented to VoiceOver, which is a screenreader for the visually impaired are just announced as “menu item” and do not have proper labeling for screenreaders.

### Steps to reproduce the behavior

Prerequisites ensure you have an iOS, iPadOS, or macOS device
One. Enable voiceover, which is a built-in screen reader for people who are visually impaired.
Two. Go to the Ubuntu website navigate the website using the screen reader commands and click on the menu
Three. You will find that the menu is  announced as “menu” which is expected behavior however, every item inside the menu is announced as “menu item” with no way of determining which of those options you are focused on.

### Expected behavior

The expected behavior would be that the screenreader is able to determine what are the items inside the menu.
For example, login or resources however, all the menu items are just marked as “menu  item” to the screenreader making it impossible for a visually impaired or blind user to navigate the website.

### Browser/device details

Browser, Safari.
Device, iPhone 16 Pro.
Operating system, iOS 18.6.2

### Reported from

https://ubuntu.com/

immortalcodes:
Hi @edr-xix ,
Thanks for reaching out, we are committed to make our sites accessible to all.
I just tried to browse navigation on 
- Chrome ( Ubuntu)
- Safari (Iphone)

The screen-reader and voice-over were able to recite the name of menu items and were able to help me in navigating the site.
Could you please provide more details about the issue, a demo video would be really helpful as well.

edr-xix:
Hello, thanks for the reply. I have attached a screen recording of VoiceOver navigating the ubuntu website.
As a quick heads up between when I first reported this issue and now I have updated to iOS 26 the issue still remains.
Notice how voiceover only announces “menu item,” but does not say what each menu item is.
https://github.com/user-attachments/assets/9680cf96-dd1b-4f6d-8373-92e951673a89

akbarkz:
@edr-xix Thanks for reporting the issue! The fix has been pushed. Feel free to re-open the issue if it still reproduces on your side.

PR #15719, opened 2025-10-08
[WD-26500] make top menu items accessible by defining their role correctly

akbarkz:
## Done

- Correctly define the "role" tag on the top menu items for a voiceover, which includes:
   - Products
   - Use Cases
   - Support
   - Community
   - Download Ubuntu
   - All Canonical
   - Sign in/Account

Previously these menu items were pronounced as just "Menu Item" by the voiceover on Apple OS-es (see the linked GitHub issue below).

## QA

- Open the demo link on any browser
- Turn on the voiceover
- Click through the menu items in top navigation
- Check that the voiceover correctly pronounces the menu items
- Check on different screen resolutions, as for tablet and mobile screen navigation looks slightly different
- Preferrably check on MacOS and iOS as well, because that's where this issue was reported

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-26500
Addresses https://github.com/canonical/ubuntu.com/issues/15576

### DIFF
--- a/static/js/src/navigation.js
+++ b/static/js/src/navigation.js
@@ -788,10 +788,10 @@ if (accountContainer) {
     .then((response) => response.json())
     .then((data) => {
       if (data.account === null) {
-        accountContainer.innerHTML = `<a href="/login" class="p-navigation__link" style="padding-right: 1rem;" tabindex="0" onclick="event.stopPropagation()">Sign in</a>`;
+        accountContainer.innerHTML = `<a href="/login" class="p-navigation__link" style="padding-right: 1rem;" tabindex="0" role="button" onclick="event.stopPropagation()">Sign in</a>`;
       } else {
         window.accountJSONRes = data.account;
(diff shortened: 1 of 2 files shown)


akbarkz:
> It reads the items fine now, but there's a small bug on desktop when you tab into the first item in a category. It'll incorrectly read out that it has children the first time you focus on it and then not again, loom [here](https://www.loom.com/share/2b9e684758a34ca690bd1495ea4d5486?sid=fd73829c-e3a4-480a-bb07-4f58bd8dfe7c). On mobile it's the same issue but it's on all list items. Do you get the same?

Thanks for the remark, MariaPaula!
I watched the video and checked on my browser as well, it's the same. The voiceover doesn't say that the menu item has children, what it says is - "list 4 items", meaning there is a list that consists of 4 items. Which is correct, there are 4 menu items positioned vertically. If you focus on the menu item to the right, it will say "list 3 items", which is also correct since there are 3 more menu items positioned vertically. So I believe this is expected.

PR #15782, opened 2025-10-29
feat: /about/release-cycle desktop view redesign

mtruj013:
## Done

- Applied redesign as per [design](https://www.figma.com/design/fOlOv7ENep0DhB0ShY6hgC/25.10-Release-cycle-page---Commercial?node-id=7414-14955&p=f&m=dev) and [copy doc](https://docs.google.com/document/d/1EV5ckGXgPrIHg7joYMxZTqqZhzi51ekwlPjUt_dPbiQ/edit?pli=1&tab=t.0)

## QA

- View the site locally in your web browser at: https://ubuntu-com-15782.demos.haus/about/release-cycle
- Check page against design and design notes 
  - Note: Compatible components are found under Anbox, ie https://ubuntu-com-15782.demos.haus/about/release-cycle?product=anbox&release=anbox+cloud+lxd+image&version=all

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-22797

### DIFF
--- a/konf/site.yaml
+++ b/konf/site.yaml
@@ -55,7 +55,7 @@ env:
     secretKeyRef:
       key: auth-consumer
       name: launchpad-imagebuild
-
+  
   - name: MARKETO_API_URL
     value: https://066-EOV-335.mktorest.com
 
@@ -150,7 +150,7 @@ env:
     secretKeyRef:
       key: app-secret
       name: proctor360
-
+      
   - name: PROCTOR360_APP_ID
     secretKeyRef:
       key: app-id
@@ -160,6 +160,21 @@ env:
     secretKeyRef:
       key: secret
       name: sitemaps
+
+  - name: WORDPRESS_APPLICATION_PASSWORD
+    secretKeyRef:
+      key: wordpress-application-password
+      name: wordpress-api
+
+  - name: WORDPRESS_USERNAME
+    secretKeyRef:
+      key: wordpress-username
+      name: wordpress-api
+
+  - name: GITHUB_TOKEN
+    secretKeyRef:
+      key: GITHUB_TOKEN
+      name: github-products-api
 
 memoryLimit: 512Mi
 
@@ -206,7 +221,7 @@ production:
 
         - name: SENTRY_DSN
           value: https://0293bb7fc3104e56bafd2422e155790c@sentry.is.canonical.com//13
-
+        
         - name: WORDPRESS_APPLICATION_PASSWORD
           secretKeyRef:
             key: wordpress-application-password
@@ -226,7 +241,7 @@ production:
       env:
         - name: HTTPS_PROXY
           value: http://squid.ps6.internal:3128
-
+        
         - name: HTTP_PROXY
           value: http://squid.ps6.internal:3128
 
@@ -317,7 +332,7 @@ production:
           secretKeyRef:
             key: api-key
             name: trueability
-
+        
         - name: PROCTOR360_BASE_URL
           value: https://prod1ext.proctor360.com
 
@@ -325,15 +340,15 @@ production:
           secretKeyRef:
             key: app-secret
             name: proctor360
-
+            
         - name: PROCTOR360_APP_ID
           secretKeyRef:
             key: app-id
             name: proctor360
-
+        
         - name: MARKETO_API_URL
           value: https://066-EOV-335.mktorest.com
-
+        
         - name: MARKETO_API_CLIENT
           secretKeyRef:
             key: api_client
@@ -343,7 +358,7 @@ production:
           secretKeyRef:
             key: api_secret
             name: marketo
-
+      
         - name: CREDLY_URL
           value: https://api.credly.com/v1
 
@@ -442,10 +457,18 @@ production:
         ]
       service_name: ubuntu-com-security-api
 
-    - paths: [/security/notices\.json, /security/notices/.*\.json]
+    - paths:
+        [
+          /security/notices\.json,
+          /security/notices/.*\.json,
+        ]
       service_name: ubuntu-com-security-api-notices
 
-    - paths: [/security/flat/notices\.json, /security/flat/notices/.*\.json]
+    - paths:
+        [
+          /security/flat/notices\.json,
+          /security/flat/notices/.*\.json,
+        ]
       service_name: ubuntu-com-security-api-flat-notices
 
     - paths: [/security/updates/.*]
@@ -742,7 +765,7 @@ staging:
       secretKeyRef:
         key: app-secret
         name: proctor360
-
+        
     - name: PROCTOR360_APP_ID
       secretKeyRef:
         key: app-id
@@ -833,13 +856,13 @@ staging:
       env:
         - name: HTTPS_PROXY
           value: http://squid.ps6.internal:3128
-
+        
         - name: HTTP_PROXY
           value: http://squid.ps6.internal:3128
 
         - name: NO_PROXY
           value: "10.24.0.132,10.24.0.23,.internal,ubuntu.com,.ubuntu.com,snapcraft.io,.snapcraft.io,jujucharms.com,.jujucharms.com,maas.io,.maas.io,conjure-up.io,.conjure-up.io,netplan.io,.netplan.io,canonical.com,.canonical.com,launchpad.net,.launchpad.net,linuxcontainers.org,.linuxcontainers.org,cloud-init.io,.cloud-init.io,vanillaframework.io,.vanillaframework.io,anbox-cloud.io,.anbox-cloud.io,juju.is,.juju.is,dqlite.io,.dqlite.io,charmhub.io,.charmhub.io,canonical.design,.canonical.design"
-
+        
         - name: SEARCH_API_KEY
           secretKeyRef:
             key: google-custom-search-key
@@ -929,7 +952,7 @@ staging:
           secretKeyRef:
             key: app-secret
             name: proctor360
-
+            
         - name: PROCTOR360_APP_ID
           secretKeyRef:
             key: app-id
@@ -1051,11 +1074,19 @@ staging:
           /security/updates/.*,
         ]
       service_name: ubuntu-com-security-api
-
-    - paths: [/security/notices\.json, /security/notices/.*\.json]
+    
+    - paths:
+       [
+          /security/notices\.json,
+          /security/notices/.*\.json,
+       ]
       service_name: ubuntu-com-security-api-notices
 
-    - paths: [/security/flat/notices\.json, /security/flat/notices/.*\.json]
+    - paths:
+        [
+          /security/flat/notices\.json,
+          /security/flat/notices/.*\.json,
+        ]
       service_name: ubuntu-com-security-api-flat-notices
 
     - paths: [/security]
@@ -1238,7 +1269,7 @@ demo:
       secretKeyRef:
         key: auth-consumer
         name: launchpad-imagebuild
-
+    
     - name: MARKETO_API_URL
       value: https://066-EOV-335.mktorest.com
 
@@ -1298,7 +1329,7 @@ demo:
       secretKeyRef:
         key: app-secret
         name: proctor360
-
+        
     - name: PROCTOR360_APP_ID
       secretKeyRef:
         key: app-id
@@ -1328,12 +1359,7 @@ demo:
         name: discourse-api
 
     - name: FLASK_ENV
-      value: "demo"
-
-    - name: GITHUB_TOKEN
-      secretKeyRef:
-        key: GITHUB_TOKEN
-        name: github-products-api
+      value: "demo"    
 
     - name: WORDPRESS_APPLICATION_PASSWORD
       secretKeyRef:
@@ -1347,3 +1373,8 @@ demo:
 
     - name: CANONICAL_CLA_API_URL
       value: https://cla.staging.canonical.com
+    
+    - name: GITHUB_TOKEN
+      secretKeyRef:
+        key: GITHUB_TOKEN
+        name: github-products-api


eliman11:
Thanks @mtruj013! Sorry for the delay. I've recorded a video with questions to Lidia so once we have clarity there I'll come back here with more notes but for now - 

**Input dropdowns:**
- [ ] On page load, the default values for the selections should be: Product = Ubuntu, Release type = Ubuntu, Version = All versions and Ubuntu Pro subscription = I don’t know
- [ ] With all the selectors filled with these default values on page load, if the user selects a different product the other fields should clear and the Release type and Version dropdown should show this error state (unless Lidia's requested otherwise since)
Screenshot 2025-10-30 at 17 23 20

- [ ] Could we remove the asterisk from "Product" as all the fields are required
- [ ] "No" and "I don't know" should be 2 separate options in the Ubuntu Pro subscription dropdown
- [ ] If the text within an input field no longer fits as the screen resizes, it should be truncated with ellipses
Screenshot 2025-10-30 at 13 16 51


**Rows**:
- [ ] If users are subscribed to Pro, we should no longer show the grey dividers and instead only show the notification at the bottom 
Screenshot 2025-10-29 at 17 08 19

- [ ] If there’s nothing to display under “Out of standard support..” this row should be hidden
- [ ] If compatible components are empty, the accordion should not show
- [ ] Could we shorten dates so we take the first 3 letters of the month always (i.e. Sep 2025 instead of September 2025)
- [ ] The accordion seems to be broken for some products - i.e. not expanding/collapsing the "Compatible components" information within it as expected. For example for Product = Kubernetes, Release type = Canonical Kubernetes

mattea-turic:
Thanks for all your work on this @mtruj013 !

Notes:
**Hero:**
- [ ] Could you use the [75/25 split](https://vanillaframework.io/docs/patterns/hero#2575-signpost) here please?

**Find out your current coverage:**
- [ ] Needs a rule above this section

**Canonical Ubuntu lifecycle and release cadence key concepts:**
- [ ] Wrap the basic section w/heading and desc. text in a shallow (using reg. atm I think)

_**Tabbed section [Ubuntu Releases]**_
- [ ] The link in "Ubuntu release cycle" is super long – any way of shortening this @eliman11 ?
- [ ] Messaged you about the chart and if we can use [chip comp](https://vanillaframework.io/docs/patterns/chip) here. Lmk what's easiest for you!

_**Tabbed section [Packaging Formats]**_
- [ ] Under "Snap packages", this copy "...broader access—similar to debs." should use en-dash and spacing, "broader access – similar to debs.

**_Tabbed section [Editions and Flavors]_**
- [ ] As above. Under "Ubuntu Core", the copy, "...and reliable—perfect for large-scale..." should be "...and reliable – perfect for large-scale..."

mattea-turic:
Thank you @mtruj013 !

Notes:

**"Find out your current coverage:"**
- [ ] Could the rectangle here bleed outside the margins of the cell by 24 pixels/1.5 rem?
- [x] For the tooltip, could you please check the annotation [here](https://www.figma.com/design/fOlOv7ENep0DhB0ShY6hgC/25.10-Release-cycle-page---Commercial?node-id=7414-15361&t=zoTDXw0nA0VZAIud-1), as there are a few tweaks that'll need to be made

For the table on mb/tb:
- [x] FWe should add box-shadow-deep for the fixed header to separate it from the content
- [x] I think it could also benefit from some more padding at the top – 24 px should be fine 
- [x] There's truncation outside the chip; can this be avoided?
Screenshot 2025-11-25 at 11 06 02 AM<br>
- [x] Could we use truncation to avoid overlap on very small screens:
Screenshot 2025-11-25 at 11 11 58 AM<br>
- [ ] The tooltip is a little too wide – could we introduce a max. width of 3 columns on mobile? And if there isn't a comfortable amount of space at the right edge (24 px), then can have it below (left-aligned) the "information" icon 
IMG_1889<br>
Screenshot 2025-11-25 at 2 05 09 PM<br>

- [x] Could we add spacing between the break/bug support and legacy add-on so that it mirrors what we've got btwn ESM and break/bug support
Screenshot 2025-11-25 at 11 08 15 AM<br>


**"Canonical Ubuntu lifecycle..." (repo. coverage tab):**
- [x] Could you pls remove the rule under the first item in the legend, so "SSM for pkgs in Main" and "ESM for pkgs in Universe aren't separated)
- [x] Now that we have confirmation for chips, I've updated the design – bg colour should be #000000
- [x] On mobile there is truncation (adding image below), could we avoid this? I would say stacking could be helpful here, but it blurs the connection between what is Ubuntu Pro and what isn't :/ not sure if you have any ideas @eliman11 ?
IMG_1888<br>

eliman11:
Input dropdown:
Thank you so much for your patience on this page! Sorry it took so long for a review

**Input dropdowns:**
- [x] Error state should have the red background color and red underline from the [Vanilla component](https://vanillaframework.io/docs/base/forms#validation)
- [x] Since we default to "All versions" we shouldn't show the error state for this field - i.e. only "Release type" should have an error state when the user selects a new product but hasn't specified the release type
- [x] If the copy inside the dropdowns exceed the container, could we please add ellipses? 
Screenshot 2025-11-25 at 14 13 13

**Table:**
- [x] Could we apply sentence case to all table inputs so it's always consistent? For example, it should read "Until xx" rather than "until xx" 
- [x] Could we please apply a shadow ($box-shadow) to the bottom of the header so there's a visual distinction between the sticky header and the content?
Screenshot 2025-11-25 at 13 40 55

- [x] Similarly, could we have the header sticky on desktop too - apologies I didn't annotate this on Figma!
- [x] On mobile, the table truncates some of the content. Could we maintain the dimensions of the table on medium breakpoints and allow for horizontal scrolling on mobile (annotation [here](https://www.figma.com/design/fOlOv7ENep0DhB0ShY6hgC/25.10-Release-cycle-page---Commercial?node-id=7516-62251&m=dev))
Screenshot 2025-11-25 at 14 25 49

**Tooltip:**
- [x] In addition to Mattea's comments above, would we be able to drop the grey highlight that shows on hover/active state when the tooltip is visible? Since the container is aligned with the icon the grey highlight makes the tooltip look misaligned 
Screenshot 2025-11-25 at 13 39 05

**Chart under "Repository coverage" tab:**

- [x] Please drop the divider between "Standard Security Maintenance for packages in Main" and "Expanded Security Maintenance for packages in Universe"
- [x] Could you please add alt text for the chart (copy [here](https://www.figma.com/design/fOlOv7ENep0DhB0ShY6hgC/25.10-Release-cycle-page---Commercial?node-id=7839-6649&m=dev))

mtruj013:
@mattea-turic @eliman11 thanks both, ready for another look.

Addressed everything except making the border extend beyond the table cells and having the sticky header + horizontal scrolling on mobile. Neither one of these is feasible with css for a semantic table and while I can make it work with some js, the solution would be fragile and unnecessarily complex for what are minor visual details rather than core functionality

mattea-turic:
This is looking really good, thank you @mtruj013 !!

Notes:
**"Find out your current coverage"**
- [x] The shadow works well :) Could thewe just remove this rule (or if it can't be removed, then making it transparent): 
Screenshot 2025-11-28 at 4 10 43 PM<br>
- [x] On mobile, it's not immediately obvious that the table allows for horizontal scrolling. I wonder if we could do a sort of progressive gradient like this (adding image below) or making the scroll indicator dark and visible the whole time @eliman11 – or even just some extra copy?  
Screenshot 2025-11-28 at 4 28 17 PM<br>

- [ ] For the tooltip, could I just verify we've got 16 px both top and bottom, as the bottom looks to have a little too much length

No other comments!

eliman11:
Thanks @mtruj013 the page looks great! Would it be possible to hide the shadow for the table header until the user scrolls through the page and the header becomes sticky? (The page wasn't implemented in the end) but Max linked [this article](https://css-tricks.com/books/greatest-css-tricks/scroll-shadows/) for how scroll shadows could work with CSS when we designed a similar functionality. I find that the drop shadows look a bit awkward when the table is shorter. Acknowledge that this is a late ask though, so if it requires too much effort I think it would be better to drop the sticky headers and shadows altogether IMO, wdyt @mattea-turic?
Screenshot 2025-11-28 at 16 37 41

> * [ ]  On mobile, it's not immediately obvious that the table allows for horizontal scrolling. I wonder if we could do a sort of progressive gradient like this (adding image below) or making the scroll indicator dark and visible the whole time @eliman11 – or even just some extra copy?

Ah nice catch, the scroll bar does show up on my phone but it's white. My preference would be for the gradient as I think it's more obvious especially for longer tables that don't fit on the screen entirely, but again if it requires too much effort to change then making the scroll indicator dark would be ideal.
Screenshot 2025-11-28 at 17 33 14

mtruj013:
@eliman11 @mattea-turic thanks both, final round?

Mattea to answer your q:

> For the tooltip, could I just verify we've got 16 px both top and bottom, as the bottom looks to have a little too much length

16px is indeed what we have 
Screenshot 2025-12-01 170219


What you're likely seeing is a bit of extra bottom margin on the text element:
Screenshot 2025-12-01 170236

That's still needed if there are compatible components (the other possible text elements in the tooltip) so I've removed the bottom margin conditionally depending on if those are available and spacing should now be correct.

mattea-turic:
Thank you @mtruj013 !! This looks pretty final to me :)

 > I've removed the bottom margin conditionally depending on if those are available and spacing should now be correct.

This is perfect, thanks!

The only note I have is about the colour of the top rule in the table – could we maintain the "default" styling as in [here](https://vanillaframework.io/docs/base/tables)

This is super minor, though, so I'm +1ing!

eliman11:
Tysm for your patience with this page :') Just one tiiiny note on mobile - could we have the gradient apply to the table but not the input fields above it? Will also +1 anyway though 
Screenshot 2025-12-01 at 17 53 13

templates/about/partials/_desktop-releases-table.html
```diff
@@ -0,0 +1,124 @@
+<table aria-label="Selected product coverage details"
+             id="js-product-coverage-table" class="release-cycle-table is-desktop u-hide--medium u-hide--small">
```
Onibenjo:
```suggestion
             id="js-product-coverage-table-desktop" class="release-cycle-table is-desktop u-hide--medium u-hide--small">
```

This has the same ID as the mobile [table](https://github.com/canonical/ubuntu.com/pull/15782/files#diff-33c01c69cf525ac113fc46d471cadc52aae97f49ff42e510e82003354c250db5R2).

templates/about/release-cycle.html
```diff
@@ -10,380 +10,541 @@
...
+          <nav class="p-tabs js-tabbed-content">
+            <ul class="p-tabs__list u-no-margin--bottom"
+                data-maintain-hash="true"
+                role="tablist">
+              <li class="p-tabs__item" role="presentation">
+                <a class="p-tabs__link"
+                   id="ubuntu-releases-tab"
+                   role="tab"
+                   aria-controls="ubuntu-releases"
+                   aria-selected="true">Ubuntu releases</a>
```
Onibenjo, line 189: This is missing an href.
This makes the link not tabbable.

https://github.com/user-attachments/assets/049f181b-9849-4bd0-9e5d-a90b5434d6fb

templates/about/release-cycle.html
```diff
@@ -10,380 +10,541 @@
...
+              </li>
+              <li class="p-tabs__item" role="presentation">
+                <a href="#ubuntu-kernel-release-cycle"
```
Onibenjo:
```suggestion
                <a href="#repository-coverage"
```

templates/about/release-cycle.html
```diff
@@ -10,380 +10,541 @@
...
+                   class="p-tabs__link"
+                   id="repository-coverage-tab"
+                   role="tab"
+                   aria-controls="repository-coverage">Repository coverage</a>
+              </li>
+              <li class="p-tabs__item" role="presentation">
+                <a href="#openstack-release-cycle"
```
Onibenjo:
```suggestion
                <a href="#packaging-formats"
```

templates/about/release-cycle.html
```diff
@@ -10,380 +10,541 @@
...
+                   class="p-tabs__link"
+                   id="packaging-formats-tab"
+                   role="tab"
+                   aria-controls="packaging-formats">Packaging formats</a>
+              </li>
+              <li class="p-tabs__item" role="presentation">
+                <a href="#canonical-kubernetes-release-cycle"
```
Onibenjo:
```suggestion
                <a href="#editions-and-flavors"
```

Onibenjo (commented):
Some comments.

Also the shadow seems to be hidden on tablet but shows on desktop.
Tablet
Image
Desktop
Image

templates/about/partials/_desktop-releases-table.html
mtruj013: It's actually not needed anymore, this is a copy/paste from v1 where there was one table

templates/about/release-cycle.html
```diff
@@ -10,380 +10,541 @@
...
+    {{ vf_basic_section(
+      title={"text": "Canonical Ubuntu lifecycle
+            <br />
+            and release cadence key concepts"},
+      items=[
+        {
+          "type": "description",
+          "item": {
+            "type": "text",
+            "content": "Canonical publishes new releases of Ubuntu and other products in our portfolio on a regular cadence, enabling the community, businesses and developers to plan their roadmaps with the certainty of access to newer open source upstream capabilities."
```
Onibenjo, line 171: what is on the copydoc is this

`Canonical publishes new releases of Ubuntu on a regular cadence, enabling the community, businesses and developers to plan their roadmaps with the certainty of access to newer open source upstream capabilities.`
but this makes sense compared to what is on the copydoc as there are other products

templates/about/release-cycle.html
```diff
@@ -10,380 +10,541 @@
...
+          </div>
+          <hr class="p-rule--muted"/>
+          <div class="p-section--shallow">
+            <div class="grid-row">
+              <div class="grid-col-2">
+                <h3 class="p-heading--5">Interim releases</h3>
+              </div>
+              <div class="grid-col-4">
+                <p>
+                  Ubuntu's interim releases are designed for users and teams who move fast and need access to the latest kernels, languages, and toolchains. They provide cutting-edge features and hardware support every six months, but with only 9 months of updates. For long-term stability, production environments should use the LTS version while interim releases suit those prioritizing speed and rapid feature testing.
```
Onibenjo:
```suggestion
                  Ubuntu's interim releases are designed for users and teams who move fast and need access to the latest kernels, languages, and toolchains. They provide cutting-edge features and hardware support every six months, but with only 9 months of updates. For long-term stability, production environments should use the LTS version, while interim releases suit those prioritizing speed and rapid feature testing.
```

templates/about/release-cycle.html
```diff
@@ -10,380 +10,541 @@
...
+          </div>
+          <hr class="p-rule--muted"/>
+          <div class="p-section--shallow">
+            <div class="grid-row">
+              <div class="grid-col-2">
+                <h3 class="p-heading--5">Ubuntu Pro</h3>
+              </div>
+              <div class="grid-col-4">
+                <p>
+                  Ubuntu Pro is Canonical's enhanced security and support subscription for Ubuntu systems. It includes up to 12 years of security coverage (ESM and Legacy Support), Kernel Livepatch, Landscape, FIPS 140-2 certified packages, and CIS hardening, and optional enterprise-grade support.
```
Onibenjo:
```suggestion
                  Ubuntu Pro is Canonical's enhanced security and support subscription for Ubuntu systems. It includes up to 12 years of security coverage (ESM and Legacy add-on), Kernel Livepatch, Landscape, FIPS 140-2 certified packages, and CIS hardening, and optional enterprise-grade support.
```

static/js/src/product-releases.js
```diff
@@ -28,68 +114,160 @@ productSelection.addEventListener("change", () => {
...
 });
 
 // Add selected options as URL parameters on submit
 selectorForm.addEventListener("submit", (event) => {
   event.preventDefault();
 
+  // Lock the button and show spinner
+  const rect = submitButton.getBoundingClientRect();
+  submitButton.style.width = rect.width + "px";
+  submitButton.style.height = rect.height + "px";
```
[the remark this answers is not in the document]
mtruj013: Not sure I agree with this. The button size has to come from getBoundingClientRect() since it depends on the actual rendered copy/layout. Moving these to css variables doesn't seem like it'd have any benefit (js still has to calculate and set the values), it would just add an extra step which I would argue makes things less clear

templates/about/release-cycle.html
mtruj013: Left a comment for the pdm

templates/about/release-cycle.html
```diff
@@ -10,380 +10,541 @@
...
+             role="tabpanel"
+             aria-labelledby="ubuntu-releases-tab">
+          <div class="p-section--shallow">
+            <div class="grid-row">
+              <div class="grid-col-2">
+                <h3 class="p-heading--5">Ubuntu release cycle</h3>
+              </div>
+              <div class="grid-col-4">
+                <p>
+                  Ubuntu releases a new version every six months. Releases of Ubuntu get a development codename ('{{ releases.latest.name }}') and are versioned by the year and month of delivery &ndash; for example, Ubuntu 25.10 was released in October 2025.
```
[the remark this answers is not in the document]
mtruj013: Left a comment for the pdm

mtruj013:
Thanks @Onibenjo! Addressed the majority and left comments for the rest, could you have another look?

templates/about/release-cycle.html
```diff
@@ -10,380 +10,542 @@
...
+              <div class="grid-col-4">
+                <p>
+                  At the end of the 10 year ESM period, the Legacy add-on can be purchased to cover additional 5 years of the lifetime of Ubuntu LTS, giving a total of 15 years security maintenance and support for LTS releases.
+                </p>
+              </div>
+            </div>
+          </div>
+          <hr class="p-rule--muted" />
+          <div class="u-fixed-width">
+            <h3 class="p-heading--5">Security and support coverage for LTS releases</h3>
```
Onibenjo, line 341:
```suggestion
            <h3 class="p-heading--5">Ubuntu security and support coverage overview</h3>
```

templates/about/release-cycle.html
```diff
@@ -10,380 +10,542 @@
...
+          </div>
+          <hr class="p-rule--muted" />
+          <div class="p-section--shallow">
+            <div class="grid-row">
+              <div class="grid-col-2">
+                <h3 class="p-heading--5">Ubuntu Core</h3>
+              </div>
+              <div class="grid-col-4">
+                <p>Ubuntu Core is an all-snap edition designed for appliances and IoT. With a lightweight, strictly confined architecture, it's highly secure and reliable &ndash; perfect for large-scale, embedded deployments.</p>
+              </div>
```
Onibenjo, line 522: There seems to be a missing CTA here

Image

templates/about/release-cycle.html
```diff
@@ -10,380 +10,542 @@
...
+            <div class="grid-row">
+              <div class="grid-col-2">
+                <h3 class="p-heading--5">Ubuntu Flavors</h3>
+              </div>
+              <div class="grid-col-4">
+                <p>Ubuntu flavors are community-maintained variants that offer different desktop environments and user experiences. The release cycle for flavours may differ from those described here. For release details and support policies specific to a flavor, please consult its official website or documentation.</p>
+              </div>
+            </div>
+          </div>
+        </div>
```
Onibenjo, line 550: There is supposed to be a cta link here per the copydoc

Image

templates/about/partials/_pro-chip.html
```diff
@@ -0,0 +1,13 @@
...
+		{{ image(url="https://assets.ubuntu.com/v1/1b64bc23-cof_svg.svg",
+			alt="",
+			width="15",
+			height="15",
+			hi_def=True,
+			loading="lazy"
+			) | safe
+		}}
+	</span>
+	<span class="p-chip__value">Ubuntu Pro</span>
```
Onibenjo, line 12:
```suggestion
	<span class="p-chip__value">
	    <span class="u-hide--small">Ubuntu</span>  Pro
	</span>
```

I think this should be hidden on mobile, this should be

Image

Onibenjo (commented):
Aside the copydocs comments and this, everything else looks good.

Really great work on this. This is a loooot 🚀

templates/about/release-cycle.html
mtruj013: She confirmed that the current version (matching figma) is correct and made the change on the doc

templates/about/partials/_pro-chip.html
mtruj013: This is a bit outdated, I was instructed to drop the LTS chip and to rename the other one to Ubuntu Pro in all instances since that's the official name of the product

templates/about/release-cycle.html
mtruj013: Assuming the figma is correct here, @eliman11 could you confirm in Lidia's absence?

eliman11: I would think so too as Lidia was reviewing the copy on Figma directly

templates/about/release-cycle.html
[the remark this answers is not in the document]
mtruj013: Updated to pull from releases.yaml

templates/about/partials/_pro-chip.html
Onibenjo: Alright 
Noted

templates/about/release-cycle.html
Onibenjo: Thanks for clarifying

Onibenjo (approved):
LGTM. Thanks

PR #15987, opened 2026-01-30
remove redirect for 404 page

akbarkz:
## Done

Removed a redirect of the /404 page to home page, because in cases when a non-existent RTD docs page is requested, for example, ubuntu.com/something/docs, it should show 404 page, instead of failing (currently it fails with 503 error). To make this work the /404 page should be able to open by its URL.

## QA

- Open https://ubuntu-com-15987.demos.haus/404 and check that it opens and does not redirect to the home page.


muhammad-ali-pk (approved):
LGTM!

PR #16000, opened 2026-02-05
Fix: Failing tests : Updating test cassettes

immortalcodes:
## Done

- Updated test cassettes with the latest api calls for the test functions.
- Removed obsolete api calls
- Updated pr workflows to use .env

## QA

- Make sure the tests pass as intended.

## Issue / Card

Fixes # https://warthogs.atlassian.net/browse/WD-33493

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -105,7 +105,7 @@ To update cassettes, run the tests with the `VCR_RECORD_MODE` environment variab
 
 ```bash
 # Record new cassettes
-VCR_RECORD_MODE=all dotrun test-python
+dotrun -e VCR_RECORD_MODE=all test-python
 ```
 
 Alternatively, add `VCR_RECORD_MODE=all` to your `.env.local` file temporarily while re-recording.


tests/cassettes/TestRoutes.test_ceph_docs.yaml
```diff
@@ -1,181 +1,315 @@
...
+      - Thu, 05 Feb 2026 12:09:12 GMT
       Referrer-Policy:
       - strict-origin-when-cross-origin
-      Server:
-      - Apache/2.4.18 (Ubuntu)
+      Strict-Transport-Security:
+      - max-age=15724800; includeSubDomains
       Transfer-Encoding:
       - chunked
       X-Content-Type-Options:
```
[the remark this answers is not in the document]
immortalcodes: Will remove from .gitignore in next commit

tests/test_routes.py
```diff
@@ -20,13 +20,18 @@ def _get_vcr_kwargs(self):
         """
         This removes the authorization header
         from VCR so we don"t record auth parameters
+
+        IMPORTANT: If there are changes to the APIs called in these tests,
```
petesfrench, line 24: Can we move the functionality of refreshing the cassettes into a dedicated script please

immortalcodes: The method to update the cassettes is just toggling record_mode from "none" to "all"/"new-episodes", I feel a readme on updating it would be more apt.
What do you feel?

petesfrench: I think this is a bad idea for a number of reasons:
- This requires someone to come in the the `test_routes.py` file and find this comment to know that they need to do in the first place. The fix is essentially hidden (adding this to the readme or hacking would help though)
- Someone could very easily forget to change the value back
- This doesn't address the issue of the cassetts becoming stale in the first place

immortalcodes: I agree with point-1 and by readme I meant a separate section in readme or even a separate small read-me file.
The stale cassettes aren't as big of an issue because the intended usage of these cassettes is mock network data like how we have mock-test-data for testing db-operations.
Even a script would not help with stale cassettes, we would need a dedicated mechanism in form of a periodic job that updates the cassettes.

PR #16006, opened 2026-02-10
feat: automatically fill out the pr description with demo link and jira link

petesfrench:
## Done

- On PR creation adds 'or visit the demo link' to the QA steps
- If there is a valid Jira ticket number (e.g. 'wd-12345') in the branch name, link the jira ticket
- Here is a [PR](https://github.com/canonical/ubuntu.com/pull/16007) I made from this branch. I did not touch the description

## QA

- Create your own PR off this branch, include a valid jira ticket number in the branch name
- Check the demo link is correctly formed and placed logically
- Check the jira ticket link is correctly formed and placed logically
- Create your own PR off this branch, without a jira ticket number
- See only the demo link is added

### DIFF
--- a/.github/workflows/update-pr-description.yaml
+++ b/.github/workflows/update-pr-description.yaml
@@ -41,8 +41,8 @@ jobs:
             exit 0
           fi
           
-          # Add demo link after "Check out this feature branch"
-          sed "s|Check out this feature branch|Check out this feature branch or visit the [demo link]($demo_link)|g" original_body.txt > temp_body.txt
+          # Replace __DEMO_URL__ placeholder with actual demo link
+          sed "s|__DEMO_URL__|$demo_link|g" original_body.txt > temp_body.txt
           
           if [ -n "$jira_link" ]; then
             sed "s|Fixes #|Fixes $jira_link|g" temp_body.txt > updated_body.txt


.github/workflows/update-pr-description.yaml
immortalcodes, line 1: The job gives an unsuccessful status even when it has succeeded
https://github.com/canonical/canonical.com/actions/runs/21893818776/job/63205444233?pr=2271

.github/workflows/update-pr-description.yaml
```diff
@@ -0,0 +1,44 @@
...
+            echo "Found Jira ticket: $jira_ticket"
+          else
+            jira_link=""
+            echo "No Jira ticket found in branch name"
+          fi
+          
+          gh pr view "$pr_number" --repo "$repo" --json body -q .body > original_body.txt
+          
+          # Add demo link after "Check out this feature branch"
+          sed "s|Check out this feature branch|Check out this feature branch or visit the [demo link]($demo_link)|g" original_body.txt > temp_body.txt
```
immortalcodes: If we could replace a variable like $DEMO-LINK with the actual demo link that would be nice and the reason is:
Say I want to link demo-link.com/openstack/managed page, for this I would have to again edit the PR and insert "openstack/managed" to the link.
If it was some placeholder, I could have done $DEMO-LINK/openstack/managed and the action would automatically put the relevant link there.

petesfrench: That would be nice, how do you propose we get the desired path? We could potentially fix it from the PR name but there is no consistent format people are using (that I know of)

immortalcodes: I think we can have a place holder like --DEMO_LINK-- and we can just replace all the instances of this placeholder with the actual link.
What do you think?

petesfrench: I see, that is an excellent idea. I will have to apply the change upstream for it to take effect on this PR, as the template is pulled from main.

immortalcodes:
Works for me!
Nice
image
Just minor nik, when I do "__Demo_link__/data" , it leads to a double slash

immortalcodes (approved):

PR #16229, opened 2026-04-10
Engage page rebranding/improvements

petesfrench:
## Done

- Applies the rebranding found in this [design](https://www.figma.com/design/GjfNl9xEzE5JSdYpAkO3DD/ubuntu.com-engage---Sites?node-id=1260-21959&p=f&m=dev)
- Implements new component parsers in the Discourse module. [PR here](https://github.com/canonical/canonicalwebteam.discourse/pull/222). These can be reviewed in tandem.

## QA

- Open the [DEMO](https://ubuntu-com-16229.demos.haus/engage/)
- Check the design against the different types of engage page:
[whitepaper](https://ubuntu-com-16229.demos.haus/engage/test-whitepaper)
[blog](https://ubuntu-com-16229.demos.haus/engage/test-blog)
[roadshow](https://ubuntu-com-16229.demos.haus/engage/test-roadshow)
- Check against legacy engage pages

## Issue / Card

Fixes 
https://warthogs.atlassian.net/browse/WD-35872
https://warthogs.atlassian.net/browse/WD-21992


eliman11:
Thanks @petesfrench!! Could you set up example pages with an image in the hero? Also perhaps recreating an events/roadshow and webinar page to see what the new layout looks like for those page types ([designs here btw](https://www.figma.com/design/GjfNl9xEzE5JSdYpAkO3DD/ubuntu.com-engage---Sites?node-id=1260-21959)).

- [x] Sorry for the last min ask, could we add a right aligned, 6-col rule to the top of the content section? To match the way we do it on [case studies](https://canonical.com/case-study/oediv-de), like this - 
Screenshot 2026-04-15 at 09 29 23

- [x] Also could we display the meta_image above the button like in the ss above
- [x]  Resources section should use resources pattern
- [x] There seems to be excessive padding for the headings compared to the headings on [Vanilla](https://vanillaframework.io/docs/base/typography)
Screenshot 2026-04-15 at 09 56 31

- [x] On pages with no resource section, could the content section use deep padding at the bottom? And on pages with a resource section, the content section should have reg padding as the resource section should then have the deep padding

**Responsiveness:**
- [x] If the viewport is shortened, could we make sure the form is scrollable
Screenshot 2026-04-15 at 09 42 27

- [x] On tablet and mobile, could we reorder the content such that the CTA button or form appears before the content ([designs here](https://www.figma.com/design/GjfNl9xEzE5JSdYpAkO3DD/ubuntu.com-engage---Sites?node-id=1764-29228&t=KgHDJ5C4IU7oRxo6-1))
- [x] Also to note - the rule that should be added under the hero section on desktop should appear between the form/button section and the content section

petesfrench:
> * [ ]  Sorry for the last min ask, could we add a right aligned, 6-col rule to the top of the content section? To match the way we do it on [case studies](https://canonical.com/case-study/oediv-de), like this -

I don't understand this one.

> * [ ]  There seems to be excessive padding for the headings compared to the headings on [Vanilla](https://vanillaframework.io/docs/base/typography)

This happened when a heading is below a `p`. It is a base part of Vanilla.

eliman11:
Thanks @petesfrench! Are the BrightTalk embeds in webinars not showing because of the demo? ([example](https://ubuntu-com-16229.demos.haus/engage/pt/microcloud-virtualizacao-custos-previsiveis))

**General**:
- [ ] Is there a way we can remove the additional padding around the bullet points so it matches the styling on [Vanilla](https://vanillaframework.io/docs/patterns/lists#bulleted-with-horizontal-divider) better?

**Forms and primary CTA:** 
- [x] Can we have the images on the right to fill the width of the container so we don't show the container with the background?
Screenshot 2026-04-29 at 23 08 00

- [x] Can we hida the meta image and button for webinar engage pages with the BrightTalk embeds as there's nothing to "register" for

Screenshot 2026-04-29 at 23 14 43

> * [x]  If the viewport is shortened, could we make sure the form is scrollable
- [x] This looks much better now - but it seems to cut off the submit button at the bottom of the form, unless you scroll all the way down the page([example](https://ubuntu-com-16229.demos.haus/engage/canonical-day-paris-2026))

https://github.com/user-attachments/assets/cf8e954c-9402-4069-9d51-f034f82b0406

**Hero:**
- [x] Hide CTAs on the hero - it should only be on the right column 

**Responsiveness**

- [x] On mobile, the form shouldn't have this additional scroll bar 
Screenshot 2026-04-29 at 23 22 15

- [x] Is there a way we can remove some of this padding in the hero on tablet/mobile? It seems a bit too space out atm
Screenshot 2026-04-29 at 23 25 31

eliman11:
Thanks @petesfrench! 

As mentioned on MM I have a wider q on how we're using the primary_cta field for the button on the right in case there are no forms. For example [here](https://ubuntu-com-16229.demos.haus/engage/google-next-2026#register-section), since the primary_cta related fields are currently being used as anchor links in the hero, the button takes you to a #register-section that isn't on the page. I'm not sure what's technically feasible but the idea was to house URLs in the primary_link field ([meeting notes](https://docs.google.com/document/d/1qhc878MPf7uI9or9sM4FH1b30sueLHzoPQpxCnH09jM/edit?tab=t.ti0xflvbr0nv#heading=h.s85gqmv30j7r)). Does it introduce too much complexity to reuse an existing field in a different way? Should we instead set up new fields on Discourse for this button on the right and remove the guidelines around using the primary_cta field so we don't break this functionality for existing pages?

Other than that - 
**For webinars:**
- [x] Can we have the content section above it wrap in a regular section instead of a deep?
- [x] Do we have any flexibility around changing the background of the webinar embed to a paper background? If not. could we have that entire strip in a white background (full width) instead of just having the white container around the embedded section? 
Screenshot 2026-06-01 at 13 57 42

**Forms:**
- [x] (Mentioned on MM) Can we make the form sticky only for content that is longer than the forms
- [x] If possible can we hide the scrollbar so they aren't visible by default. Alternatively if not possible can we truncate the form to the first 3 fields, and "hide" the rest of the fields behind a "view more" link? 

**Related pages section:**
- [x] In the description field instead of displaying "It displays a title, description, and metadata (author and date)." could we pull from the "subtitle" field from the listed posts ([example page with this section](https://ubuntu-com-16229.demos.haus/engage/openstack-commands-cheat-sheet#the-form))
- [x]  Hide the author and date published info as we don't have this for engage pages
- [x] Image thumbnails should still be visible on tablet and mobile
Screenshot 2026-06-01 at 14 21 02

static/js/src/resources.js
```diff
@@ -13,26 +13,40 @@
  * XXX: This is a hack until we have an individual resource component pattern.
  * https://github.com/canonical/vanilla-framework/issues/5771
  */
-const fetchAndRenderResources = (tag = "", resource = "") => {
+const fetchAndRenderResources = (
+  tag = "",
+  resource = "",
+  language = "",
```
Ataku-B: The language param isn't used, are there plans to implement it later? Also with this change, the JSDoc above no longer reflects the method accurately.

templates/engage/base.html
Ataku-B: Is there a valid case where only webinar_date is present without youtube_id and vimeo_id? In that case we wouldn't be handling the deep padding correctly.

static/js/src/resources.js
petesfrench: Good catch, engage pages no longer require the JS implementation of resources.js. I have reverted the changes to this file

templates/engage/base.html
petesfrench: I have updated to pass the deep padding param into the template. I also merged them into a single value `embedded_video` for simplicity, this works as there will only be one video on the page

eliman11:
Thanks @petesfrench! Webinar embeds seem to be missing again ([example 1](https://ubuntu-com-16229.demos.haus/engage/cra-compliance-microcontrollers), [example 2](https://ubuntu-com-16229.demos.haus/engage/managed-kubeflow-microsoft-azure))

Other than that, I still have this outstanding question-
> As mentioned on MM I have a wider q on how we're using the primary_cta field for the button on the right in case there are no forms. For example [here](https://ubuntu-com-16229.demos.haus/engage/google-next-2026#register-section), since the primary_cta related fields are currently being used as anchor links in the hero, the button takes you to a #register-section that isn't on the page. I'm not sure what's technically feasible but the idea was to house URLs in the primary_link field ([meeting notes](https://docs.google.com/document/d/1qhc878MPf7uI9or9sM4FH1b30sueLHzoPQpxCnH09jM/edit?tab=t.ti0xflvbr0nv#heading=h.s85gqmv30j7r)). Does it introduce too much complexity to reuse an existing field in a different way? Should we instead set up new fields on Discourse for this button on the right and remove the guidelines around using the primary_cta field so we don't break this functionality for existing pages?

petesfrench:
@eliman11 

> As mentioned on MM I have a wider q on how we're using the primary_cta field for the button on the right in case there are no forms. For example [here](https://ubuntu-com-16229.demos.haus/engage/google-next-2026#register-section), since the primary_cta related fields are currently being used as anchor links in the hero, the button takes you to a #register-section that isn't on the page. I'm not sure what's technically feasible but the idea was to house URLs in the primary_link field ([meeting notes](https://docs.google.com/document/d/1qhc878MPf7uI9or9sM4FH1b30sueLHzoPQpxCnH09jM/edit?tab=t.ti0xflvbr0nv#heading=h.s85gqmv30j7r)). Does it introduce too much complexity to reuse an existing field in a different way? Should we instead set up new fields on Discourse for this button on the right and remove the guidelines around using the primary_cta field so we don't break this functionality for existing pages?

I think we can make a check to see if the primary_link is not an anchor link or not (ie. starts with 'https'). If it is and anchor, exclude it. This will mean pages without a form, where the primary_cta is something like "#register-now", won't have the button appear. If the link is a valid link, it will still be included. Pages with forms will be unaffected by this change.

eliman11:
Sounds good thanks @petesfrench! Can we also hide the thumbnail in addition to the button we're hiding? As it would look a bit strange to have a sticky thumbnail with no button. And just flagging to make sure we include that check for the regional pages as well. Thanks so much for your patience on this!! :)

petesfrench:
> Sounds good thanks @petesfrench! Can we also hide the thumbnail in addition to the button we're hiding? As it would look a bit strange to have a sticky thumbnail with no button. And just flagging to make sure we include that check for the regional pages as well. Thanks so much for your patience on this!! :)

Thumbnails should not be rendered if there is no button already. Do you have an example where this occures?

eliman11:
> Thumbnails should not be rendered if there is no button already. Do you have an example where this occures?

Not that I've seen, was just noting it down as a condition we need to include but all good if you've taken care of it already! Ty

Ataku-B:
@petesfrench There are currently differences in the background color of the engage pages and the webinar embeds in some bits, were we aware of this?
[Example 1](https://ubuntu-com-16229.demos.haus/engage/building-sovereign-cloud)
image

[Example 2](https://ubuntu-com-16229.demos.haus/engage/managed-kubeflow-microsoft-azure)
image

britneywwc (approved):

PR #16411, opened 2026-06-02
Add CSP nonce attribute to inline scripts cont.

mtruj013:
## Done

- Continuation of https://github.com/canonical/ubuntu.com/pull/16402

## QA

- Pick a demo listed below and run the following: `curl -sI https://ubuntu-com-16411.demos.haus | grep -i content-security-policy | tr ';' '\n' | grep script-src`
- See that nonce-<id> shows up in the response
- Check the pages still load as expected
- Demo list:
  - https://ubuntu-com-16411.demos.haus/pricing/pro
  - https://ubuntu-com-16411.demos.haus/pricing/pro
  - https://ubuntu-com-16411.demos.haus/security/cves
  - https://ubuntu-com-16411.demos.haus/security/notices

## Issue / Card

Partial work towards completing WD-36594

### DIFF
--- a/templates/account/checkout.html
+++ b/templates/account/checkout.html
@@ -37,7 +37,7 @@
       </div>
     </div>
 
-    <script>
+    <script nonce="{{ csp_nonce }}">
       window.stripePublishableKey = "{{ get_stripe_publishable_key }}";
     </script>
     <script src="{{ versioned_static('js/dist/shopCheckout.js') }}" type="module"></script>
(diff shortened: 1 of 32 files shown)


petesfrench (dismissed):
LGTM

petesfrench:
The `nonce` for /security/cves has a double dash `nonce--8Jom-_ZxXIbk-5arUw8N`

mtruj013:
@petesfrench this is a valid return from [secrets.token_urlsafe()](https://docs.python.org/3/library/secrets.html#secrets.token_urlsafe) which can include `-` or `_` even at the beginning of the id. It's also random so if you try it again you'll likely get a value without the dashes

petesfrench (approved):

PR #16414, opened 2026-06-03
feat: add CSP nonce to inline scripts and remove unsafe-inline from script directive 

mtruj013:
## Done

- Continuation of https://github.com/canonical/ubuntu.com/pull/16402, final pr

## QA

- Pick a demo listed below and run the following: `curl -sI https://ubuntu-com-16414.demos.haus/ | grep -i content-security-policy | tr ';' '\n' | grep script-src`
- See that nonce-<id> shows up in the response
- Check the pages still load as expected
- Demo list:
  - https://ubuntu-com-16414.demos.haus/about/release-cycle
  - https://ubuntu-com-16414.demos.haus/wsl
  - https://ubuntu-com-16414.demos.haus/certified
  - https://ubuntu-com-16414.demos.haus/engage
  - https://ubuntu-com-16414.demos.haus/tutorials

## Issue / Card

Partial work towards completing WD-36594

### DIFF
--- a/webapp/constants.py
+++ b/webapp/constants.py
@@ -63,7 +63,6 @@ CSP = {
         "*.g.doubleclick.net",
         "extend.vimeocdn.com",
         "tracking-api.g2.com",
-        "'unsafe-inline'",
     ],
     "font-src": [
         "'self'",
@@ -81,7 +80,6 @@ CSP = {
         "*.livechat-static.com",
         "'unsafe-eval'",
         "'unsafe-hashes'",
-        "'unsafe-inline'",
     ],
     "connect-src": [
         "'self'",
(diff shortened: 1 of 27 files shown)


webapp/constants.py
```diff
@@ -81,7 +80,6 @@
         "*.livechat-static.com",
         "'unsafe-eval'",
```
petesfrench, line 81: I wonder should/could we remove `unsafe-eval`

petesfrench (approved):

webapp/constants.py
mtruj013: We should, I thought this was already a part of the overall spreadsheet epic but looks like it isn't. I'll add a ticket

PR #16439, opened 2026-06-10
Failing links

Skazitron:
## Done
- Replaced fonts that had commas in them with fonts that didn't
- Fixed failing fonts issue on u.com

## QA
- Open the demo
- Open browser console.
- Ensure there are no font related error logs

## Issue / Card
https://warthogs.atlassian.net/browse/WD-36663


petesfrench (approved):

anthonydillon:
I do not think this has fixed the font not loading. What QA was taken here?

Skazitron:
@anthonydillon we've pushed this PR to try and get around the `ERR_RESPONSE_HEADERS_MULTIPLE_CONTENT_DISPOSITION`. We're tracking it [here](https://warthogs.atlassian.net/browse/ISREQ-2759)

We should technically not be getting that response since the updated fonts don't use commas, but we still are. So, the issue probably has something to do with the networking layer.

PR #16443, opened 2026-06-10
WD-36956 - /blog/author/<author-name> rebrand

Ataku-B:
## Done

- Updated /blog/author/<author-name> to match the rebranded design

## QA

- Open the [DEMO](https://ubuntu-com-16443.demos.haus/blog/author/g-aguiar-noury)
- Check that the author blog page matches the [Figma design](https://www.figma.com/design/CB22O7EPgyiUQheHVpOHQ9/26.04-Blog-rebranding---Sites?node-id=1521-7310&m=dev)

## Issue / Card

Fixes [WD-36956](https://warthogs.atlassian.net/browse/WD-36956)

[WD-36956]: https://warthogs.atlassian.net/browse/WD-36956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/blog/author.html
+++ b/templates/blog/author.html
@@ -26,7 +26,7 @@
         "padding": "shallow",
         "item": {
           "type": "html",
-          "content": '<p class="p-text--small-caps">' ~ total_posts ~ ' posts</p>'
+          "content": '<p class="p-text--small-caps">' ~ total_posts ~ ' post' ~ ('s' if total_posts|int != 1 else '') ~ '</p>'
         }
       },
       {
--- a/templates/blog/author.html
+++ b/templates/blog/author.html
@@ -26,7 +26,7 @@
         "padding": "shallow",
         "item": {
           "type": "html",
-          "content": '<p class="p-blog-author-post-count">' ~ total_posts ~ ' posts</p>'
+          "content": '<p class="p-text--small-caps">' ~ total_posts ~ ' posts</p>'
         }
       },
       {
@@ -35,7 +35,7 @@
           "is_highlighted": false,
           "attrs": {
             "src": 'https://assets.ubuntu.com/v1/f16c40d0-Favicon+-+CoF.svg' if author.id == 217 else author.avatar_urls['96'],
-            "alt": "alt-text",
+            "alt": "Canonical logo" if author.id == 217 else author.name,
             "width": "96",
             "height": "96"
           }
--- a/templates/blog/author.html
+++ b/templates/blog/author.html
@@ -35,7 +35,9 @@
           "is_highlighted": false,
           "attrs": {
             "src": 'https://assets.ubuntu.com/v1/f16c40d0-Favicon+-+CoF.svg' if author.id == 217 else author.avatar_urls['96'],
-            "alt": "alt-text"
+            "alt": "alt-text",
+            "width": "96",
+            "height": "96"
           }
         }
       },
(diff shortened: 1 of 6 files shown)


templates/blog/author.html
```diff
@@ -10,77 +12,47 @@
...
+          "content": '<p class="p-blog-author-post-count">' ~ total_posts ~ ' posts</p>'
+        }
+      },
+      {
+        "type": "signpost_image",
+        "item": {
+          "is_highlighted": false,
+          "attrs": {
+            "src": 'https://assets.ubuntu.com/v1/f16c40d0-Favicon+-+CoF.svg' if author.id == 217 else author.avatar_urls['96'],
+            "alt": "alt-text",
```
petesfrench: Please add descriptive alt text or don't include it at all so the screen reader skips it.

templates/blog/author.html
```diff
@@ -10,77 +12,47 @@
...
+  {% call(slot) vf_hero(
+    title_text=author.name,
+    layout='25/75',
+    blocks=[
+      {
+        "type": "description",
+        "padding": "shallow",
+        "item": {
+          "type": "html",
+          "content": '<p class="p-blog-author-post-count">' ~ total_posts ~ ' posts</p>'
```
petesfrench:
```suggestion
          "content": '<p class="p-text--small-caps">' ~ total_posts ~ ' posts</p>'
```
No need for a custom font here.

petesfrench:
Most pages only show 11 posts, leaving an empty space. interestingly, the last page did show the full 12.
image

templates/blog/author.html
```diff
@@ -10,77 +12,47 @@
...
+  {% call(slot) vf_hero(
+    title_text=author.name,
+    layout='25/75',
+    blocks=[
+      {
+        "type": "description",
+        "padding": "shallow",
+        "item": {
+          "type": "html",
+          "content": '<p class="p-text--small-caps">' ~ total_posts ~ ' posts</p>'
```
petesfrench: Bit of an edge case, but if the author has 1 post, it will say '1 posts'

petesfrench (commented):
The spacing for the pagination is uneven on mobile:

Image

Ataku-B:
@petesfrench The way I implemented shrinking the pagination widget was to use u-hide--small to hide the outer elements to reduce the shown page count to 3 on small screens. The margins of the pagination items are handled [here](https://github.com/canonical/vanilla-framework/blob/63878be875af9dfcb98d38402ce5ac1abde4149d/scss/_patterns_pagination.scss#L32-L41) in Vanilla. The reason for the spacing bug is that the hidden pagination item is the 2nd child, which consumes the large margin. I propose that we upstream the solution to Vanilla, otherwise, I feel like solving it here is going to be messy, and we would need to replicate that workaround everywhere we want a responsive-width pagination.

petesfrench:
> @petesfrench The way I implemented shrinking the pagination widget was to use u-hide--small to hide the outer elements to reduce the shown page count to 3 on small screens. The margins of the pagination items are handled [here](https://github.com/canonical/vanilla-framework/blob/63878be875af9dfcb98d38402ce5ac1abde4149d/scss/_patterns_pagination.scss#L32-L41) in Vanilla. The reason for the spacing bug is that the hidden pagination item is the 2nd child, which consumes the large margin. I propose that we upstream the solution to Vanilla, otherwise, I feel like solving it here is going to be messy, and we would need to replicate that workaround everywhere we want a responsive-width pagination.

We could upstream it to Vanilla if it being using in other projects. But bare in mind this will expand the scope of this work a fair bit.

petesfrench (approved):
LGTM

PR #16476, opened 2026-06-18
Csp styles 4

Skazitron:
## Done
- Removed inline style= attributes.
- Added static/sass/_utilities.scss atomic properties

## QA
- Open demo
- Navigate to pages that are modified by `Files changed`
- Ensure there are no breaking CSS changes (eg. buttons work as expected and things don't look strange)
- Open console and ensure there are no errors caused by removing inline styles.


.github/workflows/pr.yaml
britneywwc, line 1: I remember this being added recently, what is the reasoning behind removing it again?

static/sass/_utilities.scss
```diff
@@ -0,0 +1,92 @@
...
+// (cf. Vanilla's numeric grid columns)
+.u-table-col--half {
+  width: 50% !important;
+}
+
+.u-table-col--two-fifths {
+  width: 40% !important;
+}
+
+// Spacing helpers. Scale (rem):
```
britneywwc: Can we use Vanilla's vertical spacing instead of introducing new classes?

static/sass/_utilities.scss
britneywwc, line 1: Can you please explain why a separate utilities class is needed here? We should try and use Vanilla defined utilities where possible to ensure consistency across our sites

static/sass/styles.scss
```diff
@@ -1157,6 +1158,29 @@ hr.p-separator.is-shallow {
     padding-right: 1.5rem;
     width: 100%;
   }
+
+  table {
```
britneywwc: Can we import the table styles from Vanilla here? Default table stylings should match Vanilla styles

britneywwc (changes requested):
Can you please break this PR up into smaller chunks with descriptive QA steps?
It is quite hard to review on GitHub with the big number of files changed.

I would suggest having a PR with just the scss changes, and then breaking up the templates into groups to QA.

.github/workflows/pr.yaml
Skazitron: @britneywwc we'll use revert this deleted block when merging into `csp-styles`, so you can ignore it for now.

static/sass/_utilities.scss
Skazitron: The separate utility classes that I've introduced don't have a corresponding Vanilla defined class. The `u-svx` classes come close but you can't pick the direction for the spacing (which is why people felt the need to use inline styles in the first place).

static/sass/styles.scss
Skazitron: Good catch. On it!

static/sass/_utilities.scss
Skazitron: Can do. Thanks for spotting.

Skazitron:
I'm closing this pull request since it will no longer be necessary after creating the chunked PRs.

PR #16520, opened 2026-06-30
feat: added sitewide json-ld

Skazitron:
## Done
- Added sitewide json-ld

## QA
- Run the branch locally
- Run `curl -s http://0.0.0.0:8001/ | grep -A30 'application/ld+json'`
- Check the `<head>` block for Organization.
- Confirm that this works behind proxied pages like account by running `curl -s http://0.0.0.0:8001/account | grep -c '"@type": "Organization"'`

## Issue / Card
https://warthogs.atlassian.net/browse/WD-36480


templates/templates/base.html
```diff
@@ -79,6 +79,34 @@
...
+      <script type="application/ld+json">
+        {
+          "@context": "https://schema.org",
+          "@type": "Organization",
+          "@id": "https://ubuntu.com/#organization",
+          "name": "Ubuntu",
+          "url": "https://ubuntu.com/",
+          "logo": {
+            "@type": "ImageObject",
+            "url": "https://assets.ubuntu.com/v1/c55060fd-COF%20android-chrome-512x512.png",
```
Skazitron, line 91: These values were LLM generated. Please let me know if there are any changes we need to make before it's ready for prod.

templates/templates/base.html
```diff
@@ -79,6 +79,34 @@
...
+            "width": 512,
+            "height": 512
+          },
+          "sameAs": [
+            "https://x.com/ubuntu",
+            "https://www.linkedin.com/company/234280",
```
rk-8: Non-blocking - recommend replacing this URL with https://www.linkedin.com/company/canonical/. LLMs prefer vanity URLs.

templates/templates/base.html
```diff
@@ -79,6 +79,34 @@
...
+            "https://www.facebook.com/ubuntulinux/",
+            "https://www.instagram.com/ubuntu_os/",
+            "https://www.tiktok.com/@canonical_ubuntu",
+            "https://ubuntu.social/@ubuntu",
+            "https://github.com/canonical",
+            "https://launchpad.net/ubuntu",
+            "https://www.wikidata.org/wiki/Q381"
+          ]
+        }
+      </script>
```
rk-8, line 114: Non-blocking suggestion: since Epic 2 leans on the @id graph model, we could link this entity to the Canonical org entity by adding a parentOrganization property here, so the two graphs stitch together when the canonical.com Organization block lands (that block would carry the mirror subOrganization property):

```json
"parentOrganization": {
  "@type": "Organization",
  "@id": "https://canonical.com/#organization",
  "name": "Canonical",
  "url": "https://canonical.com/"
}
```

rk-8 (commented):
I've looked through the PR and checked it against the spec document. Everything looks to be in order, with base template injection, entity graph modelling, baseline fields and manually curated profiles. A couple of small items: 
- Let's use the vanity URL for LinkedIn 
- Currently we are creating two parallel organization IDs, one for Ubuntu and one for Canonical. If possible, we should use the properties "parentOrganization" and "subOrganization". We can set the parent to https://canonical.com/#organization and the sub to https://ubuntu.com/#organization.

Onibenjo (approved):

rk-8 (commented):
Have looked through at new version. Ready to go.

PR #16525, opened 2026-06-30
WD-36594: Remove 'unsafe-inline' from script CSP using nonces

Ataku-B:
## Done

- Added `nonce="{{ csp_nonce }}"` to all `<script>` tags, both inline and external

## QA

- Open the [DEMO](https://ubuntu-com-16525.demos.haus/)
- check various pages affected by this PR, eg:
- check release chart renders in /about page
- check release cycle tabs + product releases data render in /about/release-cycle page
- check blog filter auto-submit behavior in /blog filters page
- check product card interactions in /blog/product-cards page
- check Typeform embed loads in /cloud/public-cloud-survey page
- check BrightTALK webinar embed loads in engage webinar pages

## Issue / Card

Fixes [WD-36594](https://warthogs.atlassian.net/browse/wd-36594)

[WD-36594]: https://warthogs.atlassian.net/browse/WD-36594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Skazitron:
@Ataku-B Good work on your PR! A few notes:
-  There are certain scripts added by GTM that are being flagged by CSP. An example of this would be jsdelivr. Would there be a way to circumvent this? This appears to be a problem on c.com as well, so maybe we can create a ticket to resolve it.
- Same with jspm, bat.bing, z.clarity (see attached pics)
- After a talk with the marketing team, I think we can get rid of the Usabilla GTM, so we can ignore that for now.
- Aside from that I don't really see any weird issues with UX or design. So good job on that!
image
image
image
image

Ataku-B:
@Skazitron Thank you for the review! I have added them to connect-src for now, but for jspm specifically, I think it was caused by this [runtime CDN import](https://github.com/canonical/ubuntu.com/blob/fec1bf4e957469e06c1b93e38c78cee9bb736e0b/static/js/src/cookie-policy-with-callback.js#L6)

We could add this as a proper dependency instead, what do you think?

Skazitron:
@Ataku-B thanks. I think you should ask @petesfrench  if there's a reason for that before we decide to make it static.

Skazitron:
@Ataku-B After our discussion on MM, I think it's fine if we add this as a proper dependency for now to prevent tech debt.

Ataku-B:
`static/js/src/cookie-policy-with-callback.js` file was importing uuid at runtime from https://jspm.dev/uuid, a CDN
ES module. This required jspm.dev in both script-src-elem and connect-src in the CSP, and introduced an uncontrolled external dependency on every page load.

Changes:
- Add uuid as a dependency
- Change the import from the CDN URL to the npm package specifier
- Add cookie-policy-with-callback as an esbuild entry point so the
  dependency is bundled statically
- Update the template to load the bundled file from js/dist/ and remove
  the now-unnecessary type="module" attribute
- Remove jspm.dev from script-src-elem and connect-src in the CSP

Skazitron:
@Ataku-B thank you for all your work! Approving.

Skazitron (approved):

PR #16533, opened 2026-07-02
feat: block form submissions with missing mandatory fields

britneywwc:
## Done

- Block marketo submissions for forms with missing mandatory fields
- Mandatory fields are decided based on Marketing's requirement and also the structure on `form-fields.html`

## QA

- Go to https://ubuntu-com-16533.demos.haus/ai#get-in-touch or any other form page
- Submit the form without sending mandatory contact fields (set as not required on this PR)
- See that an error message pops up and no submission goes through Marketo
- Fill up a newsletter form, checks that it submits to Marketo as expected

## Issue / Card

Fixes [WD-37603](https://warthogs.atlassian.net/browse/WD-37603)

## Screenshots

## Help

[WD-37603]: https://warthogs.atlassian.net/browse/WD-37603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/ai/form-data.json
+++ b/templates/ai/form-data.json
@@ -79,7 +79,7 @@
               "type": "tel",
               "id": "phone",
               "label": "Phone number",
-              "isRequired": false
+              "isRequired": true
             }
           ]
         }
--- a/templates/ai/form-data.json
+++ b/templates/ai/form-data.json
@@ -79,7 +79,7 @@
               "type": "tel",
               "id": "phone",
               "label": "Phone number",
-              "isRequired": true
+              "isRequired": false
             }
           ]
         }
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -1033,7 +1033,7 @@ def enrich_acquisition_url(acquisition_url, utm_dict, approved_utms):
 # Marketo asset API on every submission. Keyed by form id, each entry stores
 # a (expiry_timestamp, required_fields) tuple.
 _marketo_required_fields_cache = {}
-_MARKETO_REQUIRED_FIELDS_TTL = 60 * 60  # 1 hour
+_MARKETO_REQUIRED_FIELDS_TTL = 6 * 60 * 60  # 6 hours
 
 
 def get_marketo_required_fields(form_id):
--- a/webapp/views.py
+++ b/webapp/views.py
@@ -1029,6 +1029,48 @@ def enrich_acquisition_url(acquisition_url, utm_dict, approved_utms):
     return enriched_url
 
 
+# Cache Marketo form field definitions to avoid fetching them from the
+# Marketo asset API on every submission. Keyed by form id, each entry stores
+# a (expiry_timestamp, required_fields) tuple.
+_marketo_required_fields_cache = {}
+_MARKETO_REQUIRED_FIELDS_TTL = 60 * 60  # 1 hour
+
+
+def get_marketo_required_fields(form_id):
+    """Return a ``{field_id: label}`` dict of a Marketo form's required
+    fields.
+    """
+    cache_key = str(form_id)
+    cached = _marketo_required_fields_cache.get(cache_key)
+    if cached and cached[0] > time.time():
+        return cached[1]
+
+    try:
+        data = marketo_api.get_form_fields(form_id).json()
+    except Exception:
+        marketo_sentry_report(
+            "Failed to fetch Marketo form field definitions",
+            exception=True,
+            form_id=form_id,
+        )
+        return cached[1] if cached else {}
+
+    required_fields = {}
+    for field in data.get("result", []):
+        if field.get("required"):
+            field_id = field.get("id")
+            if field_id:
+                label = field.get("label") or field_id
+                # Marketo labels end with a colon
+                required_fields[field_id] = label.rstrip(": ").strip()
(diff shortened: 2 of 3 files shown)


templates/shared/forms/form-fields.html
britneywwc, line 1: Note: undo changes before merging

templates/ai/form-data.json
britneywwc, line 1: Note: undo changes before merging

templates/shared/forms/form-fields.html
```diff
@@ -31,44 +31,44 @@
             <div class="col">
               <ul class="p-list">
                 {% if fieldset_id == "about-you" %}
-                  <label class="is-required" for="firstName">First name:</label>
-                  <input required
```
anthonydillon, line 35: @britneywwc why did you remove the browser validation for required?

britneywwc: I unchecked it for QA purposes so we can verify that BE will reject the field if FE required fields are bypassed (i.e through scripting). Also added a comment to undo changes on `form-fields.html` and `form-data.json` after QA passes

webapp/views.py
```diff
@@ -1059,6 +1059,32 @@ def marketo_submit():
...
     )
+
+    # Reject submissions missing any mandatory field. Empty values are
+    # dropped from form_fields above, so this catches both missing and
+    # blank fields (e.g. spam that only posts a valid formid).
+    required_fields = {
+        "firstName": "First name",
+        "lastName": "Last name",
+        "email": "Email",
+        "company": "Company",
```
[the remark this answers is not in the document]
britneywwc: This is a valid comment, I'm looking into another solution where we get the required fields from Marketo and compare it against the payload before submitting. This way the checks will be most robust and can also apply to forms with different required fields.

webapp/views.py
```diff
@@ -1012,6 +1012,48 @@ def enrich_acquisition_url(acquisition_url, utm_dict, approved_utms):
     return enriched_url
 
 
+# Cache Marketo form field definitions to avoid fetching them from the
+# Marketo asset API on every submission. Keyed by form id, each entry stores
+# a (expiry_timestamp, required_fields) tuple.
+_marketo_required_fields_cache = {}
```
Skazitron, line 1035: There is a separate in-memory cache that will exist for every pod on u.com. So it will be 8 API calls every hour. Thoughts?

britneywwc: Yeah that's right, 8 calls/hr is well within Marketo's rate limit. But since these forms definitions are static, I've bumped the TTL to 6h so 8 calls/6h

webapp/views.py
```diff
@@ -1059,6 +1101,27 @@ def marketo_submit():
...
+    if form_id:
+        required_fields = get_marketo_required_fields(form_id)
+        missing_required = [
+            label
+            for field, label in required_fields.items()
+            if not form_fields.get(field, "").strip()
+        ]
+        if missing_required:
+            flask.flash(
+                "There was a problem submitting your form. Please complete "
```
Skazitron, line 1138: This looks a little cluttered when there are too many required fields. What do you think?

webapp/views.py
```diff
@@ -1059,6 +1101,27 @@ def marketo_submit():
...
     )
+
+    # Reject submissions missing Marketo required fields.
+    # Calls cached Marketo API to get required fields for the form ID
+    # check if any fields are missing.
+    form_id = form_fields.get("formid")
+    if form_id:
+        required_fields = get_marketo_required_fields(form_id)
+        missing_required = [
+            label
```
Skazitron, line 1132: I believe the logic for this is not working as expected. It displays **all** the required fields when you miss just one required field.

Skazitron (changes requested):
@britneywwc thank you for your PR. Aside from the comments I left, there's a few more thing to note:
- the logic to display the required field error appears to be getting triggered even if you input values for **all** fields
- I don't see entries on marketo, but this probably related to the error subroutine getting triggered.

webapp/views.py
britneywwc: Good catch, I added case-insensitivity check to fix this

webapp/views.py
britneywwc: I think it's okay to leave as is, our aim it to alert users which fields are missing if any. These fields will be set as required on the form so it's more of a guardrail

Skazitron:
@britneywwc thanks for addressing my comments. Taking another look!

webapp/views.py
Skazitron: Also, this cache would refresh each time a commit is pushed to main, which happens very often. Would that be a concern?

Skazitron:
@britneywwc Left one more comment about the caching, but other than that it looks good.

Skazitron (dismissed):

webapp/views.py
britneywwc: That's okay. The main thing here is that we want to reduce API calls on form submission, so pushes to main are okay

Skazitron (approved):

PR #16538, opened 2026-07-03
implement response caching and circuit breaker for Discourse API interactions

Ataku-B:
## Done

- Added `webapp.discourse_cache.py` with `cached_fetch` and `install_topic_cache`, providing a shared cache and circuit breaker for Discourse-backed views.
- Wrapped the main Discourse API's `get_topic` method with the shared cache and breaker using `install_topic_cache`. 
- Refactored Engage page views and related endpoints to use `cached_fetch`, improving performance and resilience to rate-limiting.
- Updated `/takeovers_json` to cache its data for 5 minutes using `cached_fetch`.
- Added a 503 error handler to render a styled error page when the circuit breaker is open, visually it looks the same as the 500 page since the canonicalwebteam.directory_parser library hardcodes the sitemap's error-page exclusions, 400/401/403/404/410/429/500/502.html, and 503 isn't in that list

These changes collectively make the application more robust against Discourse API rate limits, improve user experience during outages, and reduce unnecessary load on upstream services.

## Issue / Card

Fixes discourse being overloaded causing engage pages to return 500

### DIFF
--- a/webapp/discourse_cache.py
+++ b/webapp/discourse_cache.py
@@ -61,8 +61,14 @@ def cached_fetch(cache, key, fetcher, ttl, max_size=512):
     cached = cache.get(key)
     now = time.time()
 
-    if cached and now - cached["ts"] < ttl:
-        return cached["data"]
+    if cached:
+        # "Not found" results expire quickly so newly published
+        # pages appear promptly
+        entry_ttl = ttl if cached["data"] else min(ttl, 60)
+        if now - cached["ts"] < entry_ttl:
+            cache.pop(key, None)
+            cache[key] = cached
+            return cached["data"]
 
     if now < _DISCOURSE_COOLDOWN["until"]:
         if cached:
@@ -116,14 +122,25 @@ def install_topic_cache(api, ttl=600):
     """
     topic_cache = {}
     fetch_topic = api.get_topic
+    check_updates = api.check_for_topic_updates
 
     def get_topic(topic_id):
         return cached_fetch(
             topic_cache,
-            topic_id,
+            # str() so int and str topic ids share one cache entry
+            str(topic_id),
             lambda: fetch_topic(topic_id),
             ttl=ttl,
         )
 
+    def check_for_topic_updates(topic_id, last_updated=None):
+        updated, updated_at = check_updates(topic_id, last_updated)
+        if updated:
+            # Drop the cached topic so the caller's re-parse sees the
+            # new content instead of latching the stale cached copy
+            topic_cache.pop(str(topic_id), None)
+        return updated, updated_at
+
     api.get_topic = get_topic
+    api.check_for_topic_updates = check_for_topic_updates
     return topic_cache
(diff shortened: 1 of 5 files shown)


webapp/discourse_cache.py
```diff
@@ -0,0 +1,129 @@
...
+    def get_topic(topic_id):
+        return cached_fetch(
+            topic_cache,
+            topic_id,
+            lambda: fetch_topic(topic_id),
+            ttl=ttl,
+        )
+
+    api.get_topic = get_topic
+    return topic_cache
```
Onibenjo, line 146: `Category` views do a live `check_for_topic_updates`, then re-parse through the cached `get_topic` and latch `index_last_updated` — serving pre-edit content until the *next* edit

```suggestion
    topic_cache = {}
    fetch_topic = api.get_topic
    check_updates = api.check_for_topic_updates

    def get_topic(topic_id):
        return cached_fetch(
            topic_cache,
            # str() so int and str topic ids share one cache entry
            str(topic_id),
            lambda: fetch_topic(topic_id),
            ttl=ttl,
        )

    def check_for_topic_updates(topic_id, last_updated=None):
        updated, updated_at = check_updates(topic_id, last_updated)
        if updated:
            # Drop the cached topic so the caller's re-parse sees the
            # new content instead of latching the stale cached copy
            topic_cache.pop(str(topic_id), None)
        return updated, updated_at

    api.get_topic = get_topic
    api.check_for_topic_updates = check_for_topic_updates
    return topic_cache
```

webapp/discourse_cache.py
```diff
@@ -0,0 +1,129 @@
...
+    per-view dict living in a closure, so it is scoped to the worker process).
+    ``max_size`` caps the number of entries: once reached, the oldest entry
+    (by insertion order) is evicted to keep per-worker memory bounded even
+    when bots crawl large numbers of distinct query-parameter combinations.
+    """
+    cached = cache.get(key)
+    now = time.time()
+
+    if cached and now - cached["ts"] < ttl:
+        return cached["data"]
```
Onibenjo: Hits never refresh insertion order and `None` results get the full TTL — so a crawler probing unique `/engage/<junk>` paths fills the cache with junk and evicts every real page's stale fallback right before the breaker opens:

```suggestion
    cached = cache.get(key)
    now = time.time()

    if cached:
        # "Not found" results expire quickly so newly published
        # pages appear promptly
        entry_ttl = ttl if cached["data"] else min(ttl, 60)
        if now - cached["ts"] < entry_ttl:
            cache.pop(key, None)
            cache[key] = cached
            return cached["data"]
```

webapp/views.py
```diff
@@ -542,9 +564,21 @@ def engage_page(language, page):
...
+                        try:
+                            page_metadata = cached_fetch(
+                                page_cache,
+                                related_url,
+                                lambda related_url=related_url: (
+                                    engage_pages.get_engage_page(related_url)
+                                ),
+                                ttl=900,
+                            )
+                        except ServiceUnavailable:
```
Onibenjo: `cached_fetch` re-raises non-429 `HTTPError` when nothing's cached, so one flaky related URL still 500s the whole page:

```suggestion
                        except (ServiceUnavailable, HTTPError):
```

Onibenjo (commented):
some comments

Onibenjo (dismissed):

Onibenjo (approved):

PR #16541, opened 2026-07-03
feat: block YisouSpider user agent with 403 response

Onibenjo:
## Done

- Block the YisouSpider crawler with a 403 at the ingress (production `nginxServerSnippet`): across two access-log samples taken 5 hours apart it accounted for ~60% of all requests to Discourse-backed pages (/engage, /community, /takeovers), exhausting the shared Discourse API rate limit and causing the intermittent 500s in production.

## QA

- Not testable locally (production-only ingress config); after deploy, `curl -sI -A "YisouSpider" "https://ubuntu.com/community?cb=1"` should return 403 while a normal browser user agent returns 200, and the "Too Many Requests" count in the ubuntu.com-discourse pod logs should stay near zero.

## Issue / Card

Production incident: Discourse 429s causing 500s on /engage and /community pages (same incident as #16538).


Ataku-B (approved):
LGTM!

PR #16558, opened 2026-07-08
feat: cache Discourse-backed content pages for 30 minutes

immortalcodes:
## Summary

- Discourse-backed content pages currently rely on the `flask_base` default `Cache-Control: max-age=60`. 
- This adds a longer `max-age` (30 min) for the infrequently-changing, publicly-identical pages backed by Discourse, applied centrally in the existing `cache_headers` after_request hook in `webapp/handlers.py`.

## How it works

- Runs **before** `flask_base`'s `set_cache_control_headers` (after_request runs in reverse registration order; `init_handlers` is registered after `FlaskBase.__init__`).
- `flask_base` only stamps its default `max-age=60` when `max-age` isn't already set, so our value is preserved — and `flask_base` still layers on `stale-while-revalidate=86400` and `stale-if-error=300`.
- Net effect: **user-perceived latency is unchanged** (SWR still serves instantly after expiry); only CDN→origin revalidation frequency drops (~30× fewer).

## QA
- Run the app locally and see apt headers are set for cache-control

### DIFF
--- a/webapp/handlers.py
+++ b/webapp/handlers.py
@@ -35,12 +35,48 @@ from webapp.shop.flaskparser import UAContractsValidationError
 from webapp.certified.helpers import convert_markdown_to_html
 from canonicalwebteam.flask_base.env import get_flask_env
 
+# Setting long cache headers for pages that are backed by Discourse and
+# don't change often, to reduce revalidation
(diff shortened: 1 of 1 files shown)


Onibenjo (approved):
LGTM

PR #16581, opened 2026-07-10
feat: block script injections for marketo forms

Skazitron:
## Done
- Added patterns to prevent script injections into Marketo

## QA
- Checkout this branch locally or use the demo
- Ensure all tests pass.
- Go to /contact-us and fill in the fields with realistic data and in comments enter one of the flagged strings: example: <svg onload=alert(1)>
  - This request will fail silently and the user will be taken to a thank-you page.
  - It will also raise an error on sentry

## Issue / Card
https://warthogs.atlassian.net/browse/WD-37742?search_id=db0c4825-ecf1-477e-a964-33f332fb2fb6


webapp/constants.py
```diff
@@ -15,6 +15,30 @@
     }
 }
 
+# Substrings that indicate a Marketo form submission is a script/command
+# injection attempt (path traversal, XSS, SSRF, SQLi probes, etc.) rather
+# than a genuine lead. Matched case-insensitively against every submitted
+# field value in webapp.views.marketo_submit.
+MARKETO_INJECTION_PATTERNS = [
```
britneywwc, line 30: We can look into using a library which includes injection patterns, I think this list is a good start but not comprehensive of all potential script injection patterns. A few more to note stated by Beatrijs in [the ticket](https://warthogs.atlassian.net/browse/WD-37161?focusedCommentId=1126663) should also be covered, such as the static string/domain blocklist. If there is an existing open source library that can cover these it will be most preferred

Skazitron:
Hi @britneywwc, thanks for taking a look. I addressed your comments but I had one caveat:

For our specific usecase, there is just one package called `pymodsecurity` but it relies on a C++ library caled `libmodsecurity3` which needs to be built separately. Since our apps are all containerized, we'll need to add this to the Dockerfile and add other dependencies that this library relies on as well. 

I personally want to use what we have right now, i,e. the hardcoded patterns since it will be easier to maintain overall, and we won't need to refactor how we're handling forms. (this new library will require huge refactors).

What do you think?

britneywwc:
@Skazitron I think we should still have some guardrails on top of the script injection patterns. Having the patterns cover some of the reported script injections is a good start but not robust to all possible attacks. An example would be that an attacker could easily look into our codebase and identify the patterns that are not covered, and attempt another attack.

The form inputs are submitted from the HTML file, processed by the JS `prepare-form-inputs` and lastly submitted through our Flask endpoint `/marketo/submit`. We can consider sanitising/validating the input along the process, depends where you find fit, and then also checking them against the injection patterns. 

I would suggest looking into libraries on pypi/npm to prevent the need of installing packages/modifying the way we containerise our application

webapp/views.py
```diff
@@ -1124,6 +1145,30 @@ def marketo_submit():
...
+        field, pattern = injection_match
+        marketo_sentry_report(
+            "Marketo form submission blocked: injection attempt detected",
+            field=field,
+            pattern=pattern,
+            form_fields=form_fields,
+        )
+        return flask.redirect(
+            f"/thank-you?{urlencode({'referrer': referrer})}"
+        )
```
britneywwc: We shouldn't redirect to the thank you page if the submission is rejected, perhaps a `#contact-form-failure` notification banner will be more appropriate. I also think we should exclude Sentry reports here, as they might bloat our alerts with spam injection attempts.

webapp/views.py
```diff
@@ -1124,6 +1145,23 @@ def marketo_submit():
...
+    # Drop submissions that look like script/command injection probes
+    # instead of forwarding them to Marketo. No Sentry report is raised
+    # here, since a public endpoint like this can be hit by scanners
+    # repeatedly and would otherwise bloat alerts with spam attempts.
+    if find_injection_attempt(form_fields):
+        flask.flash(
+            "There was an issue submitting the form.",
+            "contact-form-fail",
+        )
+        return flask.redirect(f"{referrer}#contact-form-fail")
```
britneywwc: We should use `urlparse`/`urlencode` for `referrer` or simply just redirecting back to `#contact-form-fail`

webapp/views.py
```diff
@@ -1175,11 +1175,6 @@ def marketo_submit():
     visitor_data = {
         "userAgentString": flask.request.headers.get("User-Agent"),
     }
-    referrer = (
-        flask.request.referrer
-        if flask.request.referrer
-        else "https://ubuntu.com"
-    )
```
britneywwc, line 1182: This shouldn't be removed, it is still referenced in other parts of the code

Skazitron: This just got moved up to line 1148, so it should be fine.

britneywwc (approved):
LGTM!

PR #16583, opened 2026-07-10
fix: extend cache TTL for Discourse-backed content pages

immortalcodes:
## Done
- Extended the discourse content-cache max-age to 3 hours
- Added additional discourse routes

## QA
- This can only be QAed after the PR is merged


Skazitron (approved):
LGTM!

PR #16597, opened 2026-07-14
Fix broken Article JSON-LD schema in blog article template

Skazitron:
## Done
- Fix broken json-ld on article pages

## QA
- VIsit the live ubuntu.com/blog/\<article\> pages
- Open console log -> json-ld parsing error
- Visit the demo and go to /blog/\<article\> pages
- No json-ld parsing error

## Issue / Card
https://warthogs.atlassian.net/browse/WD-37776

### DIFF
--- a/templates/blog/article.html
+++ b/templates/blog/article.html
@@ -32,29 +32,29 @@
   {% endif %}
   <script type="application/ld+json">
     {
-      "@context": "http://schema.org",
-      "@id": "https://ubuntu.com/#article",
+      "@context": "https://schema.org",
+      "@id": "{{ self.canonical_url() }}",
       "@type": "Article",
-      "name": "{{ article.title.renderered|safe }}",
+      "name": "{{ article.title.rendered|safe }}",
       "headline": "{{ article.excerpt.raw }}",
       "author": {
         "@type": "Person",
         "name": "{{ article.author.name }}"
       },
       "datePublished": "{{ article.date_gmt }}",
-      {
-        %
-        if article.image and article.image.source_url %
-      }
+      "dateModified": "{{ article.modified_gmt }}",
+      {% if article.image and article.image.source_url %}
       "image": "{{ article.image.source_url }}",
-      {
-        %
-        endif %
-      }
-      "url": "{{ request.url }}",
+      {% endif %}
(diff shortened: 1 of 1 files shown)


templates/blog/article.html
```diff
@@ -32,29 +32,29 @@
   {% endif %}
   <script type="application/ld+json">
     {
-      "@context": "http://schema.org",
-      "@id": "https://ubuntu.com/#article",
+      "@context": "https://schema.org",
+      "@id": "{{ self.canonical_url() }}",
```
anthonydillon: This URL redirect to an entire different blog section of the canonical site.

templates/blog/article.html
```diff
@@ -32,29 +32,29 @@
...
       "@type": "Article",
-      "name": "{{ article.title.renderered|safe }}",
+      "name": "{{ article.title.rendered|safe }}",
```
anthonydillon, line 38: Can you also fix the duplicated block in the head of the page please?

anthonydillon (changes requested):
Thanks @Skazitron few comments in the code. I was testing the path `/blog/ubuntu-server-a-platform-made-for-enterprise-scale`.

Skazitron:
@anthonydillon Thanks for the review. I've addressed your comments. Feel free to take another look.

anthonydillon:
@Skazitron I didn't mean to remove the author block but to update the other schema.org block in the head of the page.

Skazitron:
Hey @anthonydillon sorry about the mixup. The first json+ld block that's inside of  `<head>` is for the entire website. It's defined inside of `templates/base.html`. Given that context, there are two ways we can handle this:
- Refactor the existing json+ld blocks. (We may be able to generate them on the backend instead, but this would go beyond the scope of this ticket).
- Have multiple blocks on the same page. Which is completely fine according to Google's own [metrics](https://developers.google.com/search/docs/appearance/structured-data) (Use the schema markup validator tool)
  - Note: Since the demos are not working, you can try https://canonical.com/blog/maas-installation-bare-metal-provisioning-is-easier-than-ever, which has the schema blocks defined in the exact same way.

What do you think?

anthonydillon (approved):
LGTM :+1:

PR #16625, opened 2026-07-22
fix: surface engage edits faster (shorter engage cache + SWR cap)

Onibenjo:
## Done

Engage edits were invisible for hours: `EngagePages` had no freshness signal, so the shared 24h app cache (and CDN on top) hid changes. The real fix is upstream — `EngagePages` now self-invalidates on edit (canonicalwebteam.discourse#232), keeping engage on the shared cache **and circuit breaker** — so this PR only tightens the CDN layer:

- 30-min CDN `max-age` for engage/takeovers.
- Cap `stale-while-revalidate` at 5 min for engage/takeovers only (flask-base defaults it to 24h). Docs/tutorials/community keep the long default.

Earlier revisions split engage onto its own cache; reverted because it also split the circuit breaker across one shared quota.

Upstream PR: https://github.com/canonical/canonicalwebteam.discourse/pull/232

## QA

- `curl -sI https://ubuntu-com-16625.demos.haus/engage/<page>` → `max-age=1800, stale-while-revalidate=300, stale-if-error=3600`
- `curl -sI 'https://ubuntu-com-16625.demos.haus/engage/<page>?preview=true'` → short default (long cache skipped)
- `curl -sI https://ubuntu-com-16625.demos.haus/ceph/docs/<page>` → long `stale-while-revalidate` (docs unchanged)
- Edit an engage page in Discourse: appears within ~30 min publicly, instantly with `?preview=true`

## Issue / Card

N/A

## Screenshots

N/A

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,9 @@ canonicalwebteam.blog==6.8.4
 canonicalwebteam.search==2.1.2
 canonicalwebteam.templatefinder==1.0.0
 canonicalwebteam.image-template==1.9.0
-canonicalwebteam.discourse==7.8.0
+# TESTING: demo the probe throttle (canonicalwebteam.discourse#232)
+# alongside this PR. Re-pin to ==7.9.0 before merge.
+canonicalwebteam.discourse @ git+https://github.com/canonical/canonicalwebteam.discourse.git@feat/throttle-freshness-probes
 canonicalwebteam.form-generator==2.2.0
 canonicalwebteam.directory-parser==1.2.10
 canonicalwebteam.markdown-response==0.2.0
--- a/webapp/app.py
+++ b/webapp/app.py
@@ -43,7 +43,7 @@ from canonicalwebteam.form_generator import FormGenerator
 from canonicalwebteam.markdown_response import MarkdownResponse
 
 from webapp.certified.views import certified_routes
-from webapp.constants import CACHE_TTL
+from webapp.constants import CACHE_TTL, ENGAGE_CACHE_TTL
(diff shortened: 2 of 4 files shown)


immortalcodes:
Could you help me understand if the upstream layer is self-invalidating the updated pages, why are we tightening the CDN layer with stale while re validate to be 5 minutes?

Onibenjo:
> Could you help me understand if the upstream layer is self-invalidating the updated pages, why are we tightening the CDN layer with stale while re validate to be 5 minutes?

@immortalcodes 

We're reducing the time for the engage pages due to edits happening on discourse taking longer to appear on the website.

There was a case where an engage page was edited but changes did not appear on the website after 3 hrs (which is supposed to be the cache time) so the stale while revalidate kicked in and the content was being served even hours after the cache ttl expired (likely due to 429s on the api)

Some incidents [here](https://chat.canonical.com/canonical/pl/sgf1a8jq3tf3zb7i1g5d8nd4qo) and [here](https://chat.canonical.com/canonical/pl/p31hg97117yfzyc9z7tbufgauh)

immortalcodes:
I mean it looks fine, I am just worries that we are not sending too many requests to discourse and might cause any outage

Onibenjo:
@immortalcodes understandable, but currently, it's the circuit breaker and retry that's holding it up based on logs on the pods

immortalcodes (dismissed):
LGTM, while this is not blocking what I would suggest is to have a better way to manage different cache-times/routes for /community and /engage pages, this might soon become unreadable

PR #16633, opened 2026-07-24
chore: update desktop sitemaps (WSL, authd, ADSys, Steam)

edibotopic:
Sitemaps entries for WSL, authd, and ADSys were previously added as per the [URL migration documentation](https://documentation.ubuntu.com/rtd-proxy/how-to/migrate/#sitemap-index-records).

The use of the `/page/` redirect in the URLs may be causing issues with indexing, so this PR replaces that with the default docs version (`stable` in most cases).

The PR also adds an entry for the Steam snap documentation, which has just been migrated: https://ubuntu.com/docs/steam/


templates/sitemap_index.xml
```diff
@@ -142,6 +142,7 @@
     <loc>https://ubuntu.com/docs/snapcraft/7/sitemap.xml</loc>
   </sitemap>
   <sitemap>
+    <loc>https://ubuntu.com/docs/steam/doc-sitemap.xml</loc>
     <loc>https://ubuntu.com/docs/checkbox/main/doc-sitemap.xml</loc>
```
Skazitron, line 148: I see that there are two entries of `<loc>` under this `<sitemap>`. Can you please address it?

Skazitron (approved):

PR #16648, opened 2026-07-27
csp: trim report-only policy to the directives under test

Onibenjo:
## Done

- `CSP_REPORT_ONLY` was a full copy of the enforced policy. It now carries only the directives under test (`script-src-elem`, `connect-src`), the nonced ones, and `report-uri`.
- Saves **1.6 kb on every response** (report-only: 5kB → 4kB). Enforced policy unchanged.

## Why

`/login` is 502-ing on production for logged-out users. The app returns a valid 302, but the response headers total 16,773 B and the IS-managed edge nginx in front of the cluster buffers 16k. The ingress is already at 32k, so it isn't the constraint.

Directives absent from a CSP policy are unrestricted and report nothing, so the 9 untested ones were duplicated bytes with no added signal.

## QA

- `/login` headers: 16 kB → ~15 kB, under the 16,384 limit.
- Report-only still drops every source in `_CSP_REPORT_ONLY_REMOVALS` and still carries the nonce; `tests/test_csp.py` needs no changes.
- After deploy: `curl -sI https://ubuntu.com/login` returns 302, not 502.

Leaves ~1.2 KB of headroom. If more is ever needed, dropping `connect-src` from the report-only policy frees a further 3.4 KB.

## Issue / Card

Fixes #

## Help

### DIFF
--- a/webapp/constants.py
+++ b/webapp/constants.py
@@ -410,12 +410,32 @@ _CSP_REPORT_ONLY_REMOVALS = {
 }
 
 
+# CSP directives that receive the per-request nonce. With 'strict-dynamic' in
+# script-src-elem, the nonce authorises page scripts (and scripts they load),
+# replacing the removed 'unsafe-inline'.
+NONCED_DIRECTIVES = ("script-src", "script-src-elem")
+
+
 def _build_csp_report_only(csp):
-    stricter = {directive: list(values) for directive, values in csp.items()}
-    for directive, stale_values in _CSP_REPORT_ONLY_REMOVALS.items():
+    """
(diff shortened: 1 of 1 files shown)


srebrenkonp (approved):
This will reduce the headers just for the tiny fraction which should squeeze it in 16k. Limitation is not on "our" end; real solution should be to increase values on IS side to reflect what we have in `konf/site.yaml`.

srebrenkonp:
> Leaves ~1.2 KB of headroom. If more is ever needed, dropping `connect-src` from the report-only policy frees a further 3.4 KB.

I agree, that can be the next step if this doesn't work.

PR #16690, opened 2026-08-19
Auto-Apply Filters on Certified Search

Ataku-B:
## Done

- Removed the "Apply filters" button on Certified search and made filters auto-apply instantly when changed
- Added js to restore page scroll when filters are applied for usability
- Added loading indicator to "show more" link when expanding filter options

## QA

- Open the [DEMO](https://ubuntu-com-16690.demos.haus/certified/laptops)
- Add and remove filters to check that they apply immediately without the need of a secondary action
- Check that the page restores the scroll position after applying a filter

## Issue / Card

Fixes [WD-37896](https://warthogs.atlassian.net/browse/WD-37896)

[WD-37896]: https://warthogs.atlassian.net/browse/WD-37896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/static/js/src/certified-search-results.js
+++ b/static/js/src/certified-search-results.js
@@ -269,8 +269,8 @@ function handleFilterClick(e) {
       urlParams.delete(name);
       // Only category pages
       if (pathCategory === id) {
-        const newURL = `/certified?${urlParams.toString()}`;
-        window.history.pushState({ path: newURL }, "", newURL);
+        window.location.replace(`/certified?${urlParams.toString()}`);
+        return;
       }
       // Append back deleted params
       // If multiple selected
@@ -319,11 +319,7 @@ function handleFilterClick(e) {
   }
 
   const newURL = `${window.location.pathname}?${urlParams.toString()}`;
-  window.history.pushState({ path: newURL }, "", newURL);
-}
-
-function submitFilters() {
-  window.location.replace(window.location.href);
+  window.location.replace(newURL);
 }
 
 /**
@@ -393,7 +389,7 @@ function clearFilters() {
   } else {
     objUrl.search = "";
   }
-  window.history.pushState({ url: objUrl }, "", objUrl);
+  window.location.replace(objUrl.toString());
 }
 
 // function to ensure only the option which has been changed is appended to the URL
@@ -445,11 +441,6 @@ function wireStaticFilterHandlers() {
       }
     },
   );
-
-  const submitFiltersButton = document.querySelector(".js-submit-filters");
-  if (submitFiltersButton) {
-    submitFiltersButton.addEventListener("click", submitFilters);
-  }
 
   const clearFiltersButton = document.querySelector(".js-clear-filters");
   if (clearFiltersButton) {
--- a/static/js/src/certified-search-results.js
+++ b/static/js/src/certified-search-results.js
@@ -356,13 +356,24 @@ function toggleExpandFilters(e, element) {
   const { name, value } = element;
   const { category, vendor, release } = retrieveSelectedFilters();
 
+  setFilterLinkLoading(element, true);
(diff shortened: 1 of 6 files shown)


static/js/src/certified-search-results.js
```diff
@@ -269,8 +269,8 @@ function handleFilterClick(e) {
       urlParams.delete(name);
       // Only category pages
       if (pathCategory === id) {
-        const newURL = `/certified?${urlParams.toString()}`;
-        window.history.pushState({ path: newURL }, "", newURL);
+        window.location.replace(`/certified?${urlParams.toString()}`);
```
Skazitron: Can we replace `window.location.replace` with something that's UTM friendly?

static/js/src/certified-search-results.js
```diff
@@ -393,7 +389,7 @@ function clearFilters() {
   } else {
     objUrl.search = "";
   }
-  window.history.pushState({ url: objUrl }, "", objUrl);
+  window.location.replace(objUrl.toString());
```
Skazitron: Same goes for this one.

Skazitron (changes requested):
Thanks for your work on this @Ataku-B. It looks good for the most part, but I think it would be nicer if we implemented a debouncing mechanism to prevent unexpected behavior when filters are toggled on/off too fast. 

I also think it would be wise if we got UX to review this first. cc. @eliman11

eliman11:
As mentioned on MM - giving a +1 provided we go back to the version where selecting and applying new filters will take users to the top of the page, rather than having the screen flash briefly and preserving the scroll position as this version was more disorienting than the first. Thanks so much for your thorough investigation Atakan!

Skazitron:
Good work getting the UX+1! There is a minor design concern. When toggling on a filter the left part of the box gets cut off. I see the same issue on both Firefox and Chrome.

image

static/js/src/certified-search-results.js
```diff
@@ -36,6 +36,36 @@ tabBtn3.addEventListener("click", (e) => {
...
 let filterLimit = 5;
 
+let filterNavigateTimer = null;
+
+const SCROLL_POSITION_KEY = "certifiedFiltersScrollY";
+
+// Filter changes reload the page, which otherwise resets the scroll position
+// to the top. Stash it before navigating away; an early inline script in
+// base.html restores it as soon as the new page starts loading.
+function saveScrollPosition() {
```
Skazitron, line 46: From Elaine's comment, I thought we were getting rid of the scroll to last position thing. Toggling on a filter changes the list, so it makes sense to scroll to the top again.

Correct me if I'm wrong.

Ataku-B:
> Good work getting the UX+1! There is a minor design concern. When toggling on a filter the left part of the box gets cut off. I see the same issue on both Firefox and Chrome.
> 
> image

@Skazitron Good eye! Elaine and I were aware of this, we decided it's better to tackle this in the sidebar visuals ticket. It is currently blocked by design.

static/js/src/certified-search-results.js
Ataku-B: Initially Elaine requested this scroll retention feature. Last I spoke with Elaine I couldn't figure out a way to make the page reload with the scroll already applied, so we were getting a split second flash of the page scrolling down upon load, and she told me if we can't get it to stay scrolled we should revert, but last night I managed to get it to work like she initially requested. Elaine is OOO starting today, but I felt safe to assume she would approve this iteration as well as it is what she had initially wanted.

Skazitron (approved):
LGTM 👍🏽

PR #16696, opened 2026-08-20
Post release engage page fixes

petesfrench:
## Done

- submit filter options with 'Apply filter' button
- remove JP language options
- don't redirect Chinese engage pages to the regional site 
- minor visual adjustments

## QA

- Open the [DEMO](https://ubuntu-com-16696.demos.haus/engage)
- Check the specifications listed in [the ticket](https://warthogs.atlassian.net/browse/WD-38385) are fullfilled
- Check the filter works as normal
- Check Chinese pages are accessible and are under ubuntu.com

## Issue / Card

Fixes https://warthogs.atlassian.net/browse/WD-38385


eliman11:
Thanks so much Pete! Will there be a separate PR for the changes on the regional sites for the guardrail against the #the-form and #register-section primary_links and the length of the rule under the hero section?

Also checking for "don't redirect Chinese engage pages to the regional site" point in your description - we don't currently redirect to Chinese engage pages right? The option that shows on the English page is for Chinese (Traditional), which does live in the asme category as all ubuntu.com engage pages, whereas the regional page is for Chinese (Simplified) pages.  

- [x] "Clear filters" should only clear + submit if a query had already been submitted. If you're selecting the filters for the first time in the session, "Clear filters" should only clear your selection, without refreshing the page. 
- [x] Could we use the same button variant for the filters as we do for "Apply filters"? ATM the "Apply filters" button is slightly smaller
Screenshot 2026-08-21 at 10 16 36

petesfrench:
> Thanks so much Pete! Will there be a separate PR for the changes on the regional sites for the guardrail against the #the-form and #register-section primary_links and the length of the rule under the hero section?

Yes, these will need a separate request

> Also checking for "don't redirect Chinese engage pages to the regional site" point in your description - we don't currently redirect to Chinese engage pages right? The option that shows on the English page is for Chinese (Traditional), which does live in the asme category as all ubuntu.com engage pages, whereas the regional page is for Chinese (Simplified) pages.

It was a bug in ubuntu.com that this fixes

> * [ x ]  "Clear filters" should only clear + submit if a query had already been submitted. If you're selecting the filters for the first time in the session, "Clear filters" should only clear your selection, without refreshing the page.


> * [ x ]  Could we use the same button variant for the filters as we do for "Apply filters"? ATM the "Apply filters" button is slightly smaller

petesfrench:
@eliman11 @mattea-turic this is ready for you review

eliman11:
> Yes, these will need a separate request
> 
> >

Got it - should I submit the tickets or would you be able to create and delegate them (not sure if these should only be picked up once this PR is merged so we have a reference point rather than duplicating work?)

- [x] Could we hide the "0" badge? i.e. before selecting a filter, all you see is the label and the chevron, and after selecting one, the badge appears
Screenshot 2026-08-25 at 18 23 10

- [ ] Once filters have been applied, these buttons should stay in it's active state. ATM if I "Apply filters" then the selection shows but the button goes back to it's default, base state. 
Screenshot 2026-08-25 at 18 24 19

- [x] Mentioned this on MM but are we able to add the dark background when banner_class = dark or gradient on Discourse (example Discourse post [here](https://discourse.ubuntu.com/t/how-to-integrate-ubuntu-desktop-with-active-directory/17323))? Demoed on Loom [here](https://www.loom.com/share/a70f4731763742c6922c0436efe2d273), but essentially we'd need to apply the dark background so logos/images still show up with the paper background otherwise on the page.

mattea-turic:
Thanks @petesfrench for your work on this page!

Comments:
---
### Filters:
**General:**
- [x] For the drop downs, should the selected item move to the top? This is how it's been designed for the combo box, but @eliman11 feel free to jump in if you think it shouldn't be the case here!
- [x] The container fill being the same colour as the bg isn't ideal – could you add a muted border around each dropdown? Another option could be to follow the shadow as we've got it [here](https://www.figma.com/design/vbAcS0b73fEVGp9T0rMrqK/25.10-Combo-box---Design-system-contributions?node-id=5136-14111&t=IzfUe6j1sYhbZx3u-1), but this is probably out of scope – and adding the border is an easy fix, I hope
- [x] On tablet, the filters run vertically – could we have them run horizontally instead pls?

**"Tag" filter:**
- [x] There is excess whitespace above the first item – could this be removed pls?
- [x] Could the badge pls replace the chevron? Pls see example [here](https://www.figma.com/design/vbAcS0b73fEVGp9T0rMrqK/25.10-Combo-box---Design-system-contributions?node-id=5131-12737&t=IzfUe6j1sYhbZx3u-1) – basically, the badge should only be present if at least one item is selected, and if it is present, it should replace the chevron.
- [ ] I might be wrong here, but the badge doesn't seem to hug the text? In that, when there are two numbers in the badge, the dimensions/padding are the same as when there is just one number – so a bit claustrophobic
- [x] This is more of a UX question/comment, but should the focus operate on the level of the row, rather than just the checkbox? Similar to how you've implemented in the other filters

### Cards:
**Hero cards**
- [x] The cards at the top have top padding that is inconsistent with the other cards – could we drop this pls, so the image sits right at the top?
- [x] Any way to have the cards the same height? At least the two main ones – no worries if not! 

### "No results:"
- [x] The copy should use an "en-dash" as part of "No results — why not try widening your search?", right @eliman11? Right now is em
- [x] If possible, can this content start at 25%? This way, it'll align nicely with the hero, as currently it's jagged and not so great for reading :')

petesfrench:
@eliman11 

> * [ ]  Once filters have been applied, these buttons should stay in it's active state. ATM if I "Apply filters" then the selection shows but the button goes back to it's default, base state.

I am not sure I follow for this one. In the screen the 'tags' selector was highlighted, not base state. When I appied changes myself, they too remain in the highlighted state:
image

petesfrench:
@mattea-turic 

> * [ ]  For the drop downs, should the selected item move to the top? This is how it's been designed for the combo box, but @eliman11 feel free to jump in if you think it shouldn't be the case here!

They move to the top after the search is complete. Making them move dynamically would add considerable complexity and I would suggest we avoid it *if* possible.

> * [ ]  The container fill being the same colour as the bg isn't ideal – could you add a muted border around each dropdown? Another option could be to follow the shadow as we've got it [here](https://www.figma.com/design/vbAcS0b73fEVGp9T0rMrqK/25.10-Combo-box---Design-system-contributions?node-id=5136-14111&t=IzfUe6j1sYhbZx3u-1), but this is probably out of scope – and adding the border is an easy fix, I hope

I have implemented the border, and based the box-shadow of the styling found [here](https://github.com/canonical/pragma/blob/e8d25fecbe6e70edfc4cb300bad65f691dbbd26e/packages/react/ds-global-form/src/lib/subcomponent/ComboboxInput/common/List/styles.css#L4). Let me know if it needs adjustment.

> * [ ]  I might be wrong here, but the badge doesn't seem to hug the text? In that, when there are two numbers in the badge, the dimensions/padding are the same as when there is just one number – so a bit claustrophobic

How should the badge scale? Equally vertically and horizontally retaining the circle shape or just horizontally to become more oval?

> **Hero cards**

There is a fix in Vanilla for this being worked on seperately

eliman11:
> I am not sure I follow for this one. In the screen the 'tags' selector was highlighted, not base state. When I appied changes myself, they too remain in the highlighted state:

Ack yeah I'm unable to reproduce the bug - it seems to be working as expected now! Ty!

Edit: I removed the last comment I left on editing the dark image containers. I sent the demo to Mohamed and he said he prefers the old version without the background, and will ask marketing managers to update their images instead. Sorry could you please revert this change and keep the image containers with no background as it was previously? Thanks so much

mattea-turic:
@petesfrench thanks for the quick updates here!

> They move to the top after the search is complete. Making them move dynamically would add considerable complexity and I would suggest we avoid it if possible.

No worries, let's leave it!

> I have implemented the border, and based the box-shadow of the styling found [here](https://github.com/canonical/pragma/blob/e8d25fecbe6e70edfc4cb300bad65f691dbbd26e/packages/react/ds-global-form/src/lib/subcomponent/ComboboxInput/common/List/styles.css#L4). Let me know if it needs adjustment.

LGTM, ty!

> How should the badge scale? Equally vertically and horizontally retaining the circle shape or just horizontally to become more oval?

Just as the [component itself](https://vanillaframework.io/docs/patterns/badge) – nothing custom. If it was already at default, then don't worry about making any further changes.

static/sass/_pattern_filter-menu.scss
```diff
@@ -0,0 +1,140 @@
...
+      display: flex;
+      margin-bottom: $spv--small;
+      padding: $spv--x-small $spv--small $spv--x-small $spv--large;
+
+      > span {
+        margin-right: $spv--small;
+      }
+
+      &.is-active {
+        background-color: #d9d9d9;
```
mtruj013: Not a blocker but this and a few other color instances can be replaced with vanilla variables

templates/engage/base.html
```diff
@@ -69,134 +63,135 @@ <h1>{{ metadata["topic_name"] }}</h1>
...
-          <h2>Additional resources</h2>
+    {% if show_webinar %}
+      {% include "engage/shared/_webinar-embed.html" %}
+    {% endif %}
+
+    {% if "youtube_id" in metadata %}
+      <!-- Youtube webinars -->
+      <section class="p-strip is-shallow" id="youtube-section">
+        <div class="row" id="webinar">
+          <iframe title="Webinar"
```
mtruj013: Should we use lite-youtube for performance instead here?

mtruj013 (dismissed):
LGTM!

mtruj013 (approved):

PR #16706, opened 2026-08-24
Feat: Charm fixes for ubuntu.com

immortalcodes:
## Done

- Updated charmcraft.yaml with the the additional github secret.
- Fixed dependency issues that were causing the charm build to fail.

## QA

- See the Rocks and charm build successfully. ( While the pack and deploy  job might fail but you will see the rocks and charms are being successfully  built)
- You can also go to the PS-6 environment and check if the application is up and running.
- Changes in yaml files and requirements.txt are relevant for reivew.

## Issue / Card

Fixes #

## Screenshots

## Help


.github/workflows/deploy.yaml
Skazitron, line 130: Can you please comment out this section? Or else it builds to prod.

immortalcodes: The site is still on PS-5 so it's not an issue.

Skazitron (commented):
The libraries being added are not referenced anywhere in charm.py. Is this expected?

charm/charmcraft.yaml
Skazitron, line 4: There's no gunicorn tuning in this file. Ubuntu.com's `konf` has 2 gunicorn workers. I think we could use more gunicorn workers than 2 per replica. What do you think?

requirements.txt
```diff
@@ -1,3 +1,4 @@
+flask==2.2.5
```
Skazitron, line 1: We don't have gevent here, but gevent is necessary for gunicorn workers to be non-blocking. What do you think? It might be prudent to add it so it doesn't cause issues once the charmed build is deployed.

rockcraft.yaml
```diff
@@ -32,6 +32,12 @@ parts:
       mkdir -p "$CRAFT_PART_INSTALL/flask/app"
```
Skazitron, line 32: There's a line in the Dockerfile (line 58) to fetch `etc/ubuntu-mirrors-rss.xml`. I don't see a substitute for that here.

Skazitron (changes requested):
The biggest issue I see with this build is that konf/site.yaml has 9 different deployments (eg. blog, security, etc), each with its own individual number of replicas. How do we implement that with Juju? Thoughts?

immortalcodes:
> The libraries being added are not referenced anywhere in charm.py. Is this expected?

Could you please elaborate on what libraries we are talking here?

Skazitron:
> > The libraries being added are not referenced anywhere in charm.py. Is this expected?
> 
> Could you please elaborate on what libraries we are talking here?

grafana, prometheus, loki, squid

charm/charmcraft.yaml
Skazitron, line 31: I don't see any way we're getting the env variable CSP_REPORT_ONLY_MODE here.

Skazitron, line 66: I see the marketo api credentials here, but we also need `MARKETO_API_URL`

rockcraft.yaml
```diff
@@ -32,6 +32,12 @@ parts:
...
       cp -r static "$CRAFT_PART_INSTALL/flask/app/"
   flask-framework/install-app:
+    # The built JS/CSS assets are provided exclusively by the `build-ui` part
+    # (which runs `yarn run build` inside the container). Exclude them here so
```
Skazitron, line 43: In the dotrun file, we're getting a bunch of proxy related variables. Example (HTTP_PROXY, NO_PROXY), how are we handling this?

charm/charmcraft.yaml
immortalcodes: The OCI entrypoint is by-default set-up in flask-extension with default number of workers and we don't use konf for it, if we need to change it, we can do it via juju config options

requirements.txt
immortalcodes: It's only needed if we have async gunicorn workers which we don't use in this project.
Docs for more info
https://ubuntu.com/containers/rockcraft/docs/latest/reference/extensions/flask-framework/
image

The charm has been already deployed on staging PS-7 and it loads up.

immortalcodes:
> > > The libraries being added are not referenced anywhere in charm.py. Is this expected?
> > 
> > 
> > Could you please elaborate on what libraries we are talking here?
> 
> grafana, prometheus, loki, squid

They are COS extensions that are baked into flask-framework, they would be use to set-up observability, reference docs here 
https://canonical.com/juju/docs/charmcraft/4/reference/extensions/flask-framework-extension/

charm/charmcraft.yaml
immortalcodes: The secret is a dictionary that has url as key in juju config, hence it is covered.

rockcraft.yaml
immortalcodes: We do these settings using juju config inside the model

charm/charmcraft.yaml
immortalcodes: Where is this variable being used in the codebase?

immortalcodes:
> The biggest issue I see with this build is that konf/site.yaml has 9 different deployments (eg. blog, security, etc), each with its own individual number of replicas. How do we implement that with Juju? Thoughts?

The solution for this is that same charm would be deployed multiple times with routes being configured in ingress-configurator for each deployment, this is how it's also done for canonical.com.

charm/charmcraft.yaml
Skazitron: My bad, meant to say `CSP_REPORT_ONLY_ENABLED`.

Skazitron (dismissed):
LGTM

PR #16718, opened 2026-08-27
chore: bump djlint to 1.44.2 for multiline jinja tag support

Onibenjo:
## Done

Bumped djlint from 1.34.1 to 1.44.2 so it stops flattening the contents of multiline jinja macro calls, a formatter bug djlint fixed in 1.41.0. `.djlintrc` gains `ignore_blocks: macro` to keep macro bodies at column 0 and ignores T038/T039, two new rules that misread tags split across lines as unclosed, while `HACKING.md` now explains which djlint your editor actually runs, which differs between Taskfile (the pinned copy in `.venv`) and dotrun (your global copy, since that venv lives in the container).

## QA

Nothing changes on the site. This only affects the formatter and linter.

- Check out this branch and run `task start`. It reinstalls requirements, so `.venv` picks up the new version on its own.
- `.venv/bin/djlint --version` should say 1.44.2.
- `.venv/bin/djlint templates --lint --profile=jinja --extension=html` gives 489 errors, down from 530 on main. No new rule codes appear; H008, H031 and T027 stop firing.
- `.venv/bin/djlint templates/enterprise-store/index.html --check` prints a diff where the dict inside `vf_basic_section(...)` picks up real nesting. On main the same command says "0 files would be updated", because 1.34.1 believes the flat version is already correct.

No templates are reformatted in this PR, so files on disk still look flat until someone runs `--reformat` on them.

One known gap if you do reformat something: a block whose opening `{{` line also contains a closed HTML tag, such as `<br />`, still flattens on 1.44.2. Six lines repo-wide are affected. Hoisting the string into a `{% set %}` fixes it, and that will be reported upstream.

## Issue / Card

[WD-38520](https://warthogs.atlassian.net/browse/WD-38520)

## Screenshots

N/A

## Help

[QA steps](https://discourse.canonical.com/t/qa-steps/152) - [Commit guidelines](https://discourse.canonical.com/t/commit-guidelines/148)

https://claude.ai/code/session_01A4jozykGLxQxJfeorhpDvX

[WD-38520]: https://warthogs.atlassian.net/browse/WD-38520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #16733, opened 2026-09-01
docs: record the ignore_blocks macro trade-off

Onibenjo:
## Done

Documents a trade-off in the `ignore_blocks: macro` setting added in #16718 that I described wrongly at the time: it keeps macro bodies at column 0 as intended, but reformatting a macro file also pulls its parameters back to column 0. Also notes that `.djlintrc` is read from the project directory by whichever djlint runs, so config reaches your editor even when the pin does not.

Docs only, no behaviour change.

## QA

- Read the "Macro formatting" section added to `HACKING.md`.
- To see it: `.venv/bin/djlint templates/macros/_vf_highlighted-cta.jinja --check` wants to move `cta_text` to column 0. Removing `ignore_blocks` from `.djlintrc` and rerunning instead indents `) -%}` and the whole macro body, which is the worse of the two.

The setting is staying as it is. Both options are stable and neither reproduces the hand formatting currently in the macro files, so this is about setting expectations before someone reformats one.

## Issue / Card

WD-38520

## Screenshots

N/A

## Help

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -223,6 +223,28 @@ python -m pip install -U djlint
 A global install is still worth having, since some of our other repos do not pin
 djLint at all. It only needs to stay off the old versions.
 
+Note that `.djlintrc` is read from the project directory by whichever `djlint`
+runs, so config changes reach your editor even when the pin does not. Options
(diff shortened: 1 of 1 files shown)


Ataku-B (approved):
LGTM

PR #16739, opened 2026-09-03
WD-37898: Certified Search Autocomplete

Ataku-B:
## Done

- Add autocomplete to the certified search bar

## QA

- Open the [DEMO](https://ubuntu-com-16739.demos.haus/certified/search)
- Alternatively, check out this feature branch
- Run the site using the command `task start`
- Type various configuration names into the search bar and check the autocomplete suggestions

## Issue / Card

Fixes [WD-37898](https://warthogs.atlassian.net/browse/WD-37898)

[WD-37898]: https://warthogs.atlassian.net/browse/WD-37898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


static/js/src/certified-search-results.js
```diff
@@ -489,3 +489,156 @@ if (filters2Elm.querySelector("input") || filters3Elm.querySelector("input")) {
...
+
+const AUTOCOMPLETE_MIN_CHARS = 3;
+const AUTOCOMPLETE_MAX_SUGGESTIONS = 5;
+const AUTOCOMPLETE_DEBOUNCE_MS = 200;
+
+function initCertifiedAutocomplete() {
+  const searchInput = document.querySelector("#certified-search");
+  const suggestionsList = document.querySelector(
+    "#certified-search-suggestions",
+  );
```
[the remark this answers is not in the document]
Ataku-B: Rejecting this as it is outside the scope of this task.

webapp/certified/views.py
```diff
@@ -729,6 +738,56 @@ def nxp_contact():
     return render_template("certified/202309-32027/contact-us.html")
 
 
+def certified_autocomplete():
+    """
+    Search-box suggestions for /certified/search, scoped to the same
+    category/vendor/release filters currently applied on that page.
```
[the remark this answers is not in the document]
Ataku-B: Rejecting this as it is outside the scope of this task.

immortalcodes:
QA fails for me
image
image
image

webapp/certified/views.py
```diff
@@ -729,6 +740,89 @@ def nxp_contact():
...
+    selected_releases = request.args.getlist("release")
+    category__in = (
+        ",".join(selected_categories) if selected_categories else None
+    )
+    major_release__in = (
+        ",".join(selected_releases) if selected_releases else None
+    )
+    vendor = selected_vendors or None
+
+    def fetch_models(**field_filter):
```
immortalcodes, line 770: Why create functions when we are anyway not using them outside of this block ?

webapp/certified/views.py
```diff
@@ -729,6 +740,89 @@ def nxp_contact():
...
+            if not name or key in seen:
+                continue
+            seen.add(key)
+            suggestions.append(
+                {"model": name, "make": (result.get("make") or "").strip()}
+            )
+            if len(suggestions) == AUTOCOMPLETE_MAX_SUGGESTIONS:
+                break
+
+    add_unique(fetch_models(model__icontains=query))
```
immortalcodes, line 804: Would this not automatically get suggestions equal to `AUTOCOMPLETE_MAX_SUGGESTIONS` ?
Why call this again in next line with if condition?

webapp/certified/views.py
```diff
@@ -729,6 +740,89 @@ def nxp_contact():
...
+            # Suggestions are a non-essential enhancement - never break the
+            # search box over a flaky upstream API call
+            sentry_sdk.capture_exception()
+            return []
+        return response.get("results", [])
+
+    seen = set()
+    suggestions = []
+
+    def add_unique(results):
```
immortalcodes, line 791: Functions should ideally operate as "black boxes." When a function modifies a global(in this case the seen and suggestions) variable without returning a value, its behavior becomes invisible to anyone reading the function call.

webapp/certified/views.py
```diff
@@ -729,6 +740,89 @@ def nxp_contact():
...
+    tokens = query.split()[:AUTOCOMPLETE_MAX_QUERY_TOKENS]
+    for i in range(1, len(tokens)):
+        if len(suggestions) == AUTOCOMPLETE_MAX_SUGGESTIONS:
+            break
+        add_unique(
+            fetch_models(
+                make__icontains=" ".join(tokens[:i]),
+                model__icontains=" ".join(tokens[i:]),
+            )
+        )
```
immortalcodes, line 821: Could you please help me understand what are we doing here?

static/js/src/certified-search-results.js
immortalcodes, line 1: Do you think we can abstract the `autocomplete_suggestion `functionality here such that when we initialise it  on a page with apt `url` to fetch suggestions from and the relevant `Classes` for `input` and `suggestions panel`, it could work for any autocomplete_suggestion on any page?

This is non-blocking and is just nice to have.

immortalcodes (changes requested):
Thanks for this feature 😃 
Left some comments

webapp/certified/views.py
Ataku-B: This is due to a limitation with the Certified API. The two calls hit different API fields, so the first one can't be relied on to fill the slots:

- model__icontains=query searches the configuration/model name (e.g. "16 Aurora")
- make__icontains=query searches the vendor name (e.g. "Alienware")

If the user just searches with the vendor name (e.g. "Alienware") the first call would return nothing so we need to send another call with the make__icontains field. The check is there to avoid making this second network call if the first one was sufficient to fill the autocomplete slots. We are not allowed to request changes to the Certified API at this time, thus the need for this workaround.

I'm adding a comment to clarify the behaviour.

webapp/certified/views.py
Ataku-B: This is also due to a limitation with the Certified API, similar to the above. It can't search "make" and "model" together as one phrase, only as separate fields:

model__icontains searches the configuration/model name (e.g. "XPS 13")
make__icontains searches the vendor name (e.g. "Dell")
If someone searches "dell xps", neither field on its own contains that whole phrase, so both single-field calls above return nothing. Since we can't change the API to search across both fields at once, we split the query at each word boundary and retry it as a combined request (e.g. for "dell xps 13", first try "dell / xps 13", then " dell xps / 13" and so on).  It's capped to the first 4 words so an unusually long query can't trigger unbounded extra requests. Some vendor names have multiple words so just isolating the first word as the vendor name would not work.

I'm updating the comment to indicate that this is a Certified API workaround.

templates/certified/search/_search-results-main-block.html
```diff
@@ -24,5 +24,7 @@
   </div>
 </div>
 
+<script nonce="{{ csp_nonce }}" type="text/javascript"
+        src="/static/js/src/autocomplete.js"></script>
 <script nonce="{{ csp_nonce }}" type="text/javascript"
         src="/static/js/src/certified-search-results.js"></script>
```
immortalcodes, line 28: We can either convert autocomplete in an ES-module and then export and use `initAutocomplete` inside certified-results, this way we need not to remember to import autocomplete everytime and it becomes cleaner.
Just this line
`<script nonce="{{ csp_nonce }}" type="module"
        src="/static/js/src/certified-search-results.js"></script>`

immortalcodes (commented):
Just one comment, rest looks good.
Thanks

immortalcodes (approved):
LGTM! thanks

PR #16748, opened 2026-09-04
feat: Add compliance information to release cycle page

petesfrench:
## Done

- Add compliance information to release cycle page (pulling from https://github.com/canonical/product-architecture/)
- Update to use more modern filter (introduced in this [pr](https://github.com/canonical/ubuntu.com/pull/16696))

## QA

- Open the [DEMO](https://ubuntu-com-16748.demos.haus/about/release-cycle)
- Alternatively, check out this feature branch
- Run the site using the command `task start`

## Issue / Card

Fixes [WD-37769](https://warthogs.atlassian.net/browse/WD-37769)

[WD-37769]: https://warthogs.atlassian.net/browse/WD-37769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


templates/about/partials/_tablet-mobile-releases-table.html
```diff
@@ -1,106 +1,102 @@
...
-            {% include "about/partials/_coverage-chip-pro.html" %}
-          {% endif %}
-        {% elif version.legacy_supported.notes %}
-          {{ version.legacy_supported.notes | capitalize}}
-        {% endif %}
-      </td>
-    </tr>
-  {% endfor %}
-</tbody>
-</table>
```
petesfrench, line 106: Line 1-106 is formatting changes

eliman11:
Thanks Pete sorry for the delay in reviewing this :') 
- [x] The chip for "Covered with legacy add-on" is cropped - could we reduce the width of the first column to allow for more space for the 3  first columns in the orange section?
- [ ] ATM if you have the dropdown open and scroll down it lays on top of the header rather awkwardly. Could we have it such that table heading will be on top of the dropdown instead? Fine if it's too complicated to implement, it just looks a bit buggy this way.
Screenshot 2026-09-08 at 11 03 26

- [x] Remove "Clear" option in the "Compliance standards" dropdown
- [x] If all the options are selected, the "Select all" link should also be disabled
- [x] Remove "Clear filters" link besides the "View coverage" button. It's slightly confusing in this context as there's no way to completely clear all the selections, as otherwise the table would show no values. 
- [x] On medium and small screens, the new "Compliance standards met" info should be added as a row under "Legady add-on" rather than having it's own column on the right
Screenshot 2026-09-08 at 11 10 32

Also tagging @mattea-turic for visual review.

lidia-luna:
Hello @eliman11 and @petesfrench , thanks for the progress on this. A few comments

1. Need to include the error message when not selecting a release. Otherwise when changing the product it looks like the page is broken

Screenshot 2026-09-08 at 12 52 45
 
2. For compliance standards filter, I think when you select one, we need to show them when they are green, (achieved) if it's any other status it counts as None. Also for products with no compliance details, then that filter can be disabled. The name of the category could be "Active compliance standards" for clarity.

3. There are two dots next to chips that don't fit, I don't think they are needed.

Screenshot 2026-09-08 at 12 51 04

4. This one is just my preference, Do we have to keep such a big shadow for the haeding of the table?
Screenshot 2026-09-08 at 12 58 14

eliman11:
> 1. Need to include the error message when not selecting a release. Otherwise when changing the product it looks like the page is broken

Indeed - sorry for not including it earlier. Please see the design [here](https://www.figma.com/design/iVtKFFULTKHyV58RKULoZU/ubuntu.com-about---Sites?node-id=384-18689&t=7HWBbOHFNbHTFCl1-1) for the error state in case a release hasn't been selected

> There are two dots next to chips that don't fit, I don't think they are needed.

@lidia-luna This is due to the full copy not fitting on the chips. I've asked Pete above if we are able to make the columns wider, else we'll need to shorten the copy to "With legacy add-on", unless you have other suggestions. 

> 7\. This one is just my preference, Do we have to keep such a big shadow for the haeding of the table?

We can drop the shadows altogether, would something like this look better to you @lidia-luna?
Screenshot 2026-09-08 at 14 40 29

Additionally, could we increase the width of the dropdown so that "Observability" isn't cropped?
Screenshot 2026-09-08 at 14 31 44

lidia-luna:
> This is due to the full copy not fitting on the chips. I've asked Pete above if we are able to make the columns wider, else we'll need to shorten the copy to "With legacy add-on", unless you have other suggestions.

+1 to both making the columns wider and also dropping the "covered with"

> We can drop the shadows altogether, would something like this look better to you @lidia-luna?

Yes, I think it's less overwhelming without the shadows.

+1 to your other suggestions @eliman11 

cc @petesfrench , thanks!

eliman11:
Thanks @petesfrench a couple of outstanding changes:

- [ ] Can you please drop the shadows from the header per Lidia's request
- [ ] Drop "Covered" from the chips  the copy of the chips to "With legacy add-on" 
- [ ] Change table header from 'Compliance standards met' -> 'Compliance standards status'
- [ ] Sorry Pete seeing it on the demo the gap between the dropdown feels a bit awkward - can we have it directly under the dropdown instead? (last change on these filters I promise)
Screenshot 2026-09-24 at 17 40 55

- [ ] When an input is invalid can we hide the table altogether
Screenshot 2026-09-24 at 17 42 02

